PR #28, opened 2016-02-09
Import task cleanup

alexsoble:
## What's new

* Fix object count diffs in import task report
* Deprecate `--district` flag: only working in 1 district at this point
* Remove `--recent-only` flag: not a real thing yet


kevinrobinson:
:+1:

PR #29, opened 2016-02-09
Add default database.yml for local development

kevinrobinson:
So that after installing Ruby, Postgres, etc., a developer can clone the repo, seed the database and start the app, and everything just works.


kevinrobinson:
@alexsoble From some feedback, we chatted about this the other day, what do you think?

alexsoble:
@kevinrobinson Works for me!

kevinrobinson:
@alexsoble :+1:

Issue #5, opened 2016-02-08
Umbrella - Redesign student profile page


kevinrobinson:
From https://github.com/codeforamerica/somerville-teacher-tool/issues/373

The graphs are the first thing the user sees – but they're not necessarily the most useful!

Mockups v1:
[profile.redesign.mockups (2).pdf](https://github.com/studentinsights/studentinsights/files/122154/profile.redesign.mockups.2.pdf)

Mockups v2:
[profile.redesign.v2 (3).pdf](https://github.com/studentinsights/studentinsights/files/122155/profile.redesign.v2.3.pdf)

Some options for next steps:
+ [x] Build the interventions panel
+ [x] Add tab selection by clicking on panels
+ [x] ELA: Add STAR and MCAS charts
+ [x] Improve layout and styling for demographics section
+ [x] Improve header styling (@BWieber in https://github.com/studentinsights/studentinsights/issues/93)
+ [x] Math: Add STAR and MCAS charts (@erose)
+ [x] Attendance: Add in charts
+ [x] Attendance: Add in text summaries for behavior incidents
+ [x] Attendance: Improve styling for behavior incidents (@bw120 in https://github.com/studentinsights/studentinsights/pull/139)
+ [x] ELA: Add DIBELS numbers (https://github.com/studentinsights/studentinsights/issues/23 and https://github.com/studentinsights/studentinsights/pull/283 and https://github.com/studentinsights/studentinsights/issues/291)
+ [x] Profile: design iterations, add full timeline view (@erose in https://github.com/studentinsights/studentinsights/issues/117)

Notes and interventions flow (@kevinrobinson):
+ [x] Design: structure interventions by what outcomes they're attempting to move?
+ [x] Improve initial rendering for notes
+ [x] Add UI for adding a note, endpoint
+ [x] Improve rendering for new interventions and progress notes
+ [x] Add UI for adding an intervention, endpoint (@BWieber and @kevinrobinson in https://github.com/studentinsights/studentinsights/issues/115)

Data questions (Related: https://github.com/studentinsights/studentinsights/issues/4 and https://github.com/studentinsights/studentinsights/issues/1):

+ [x] Fix bug with assessment dates in demo data
+ [x] Investigate nil STAR assessments (https://github.com/studentinsights/studentinsights/issues/91)

Action items from @jgeiser's feedback on 2/12:
+ [x] Show the type of note (eg., SST meeting)
+ [x] Move 51a from notes to services
[...]

PR #37, opened 2016-02-09
Student profile - add 'none' for read-only notes and interventions

kevinrobinson:
Also removes placeholder "add" buttons to make clear this is read-only at the moment.

Before:
screen shot 2016-02-09 at 5 41 50 pm

After:
screen shot 2016-02-09 at 5 41 31 pm


kevinrobinson:
Working on https://github.com/studentinsights/studentinsights/issues/5

PR #45, opened 2016-02-10
Student profile v2 - Prototyping revisions to notes and interventions flow

kevinrobinson:
This is a prototype of work in https://github.com/studentinsights/studentinsights/issues/5.  The pull request here isn't production-quality code, and is meant for prototyping user experience and design ideas.  This is read-only code, so there's little risk in deploying it for user testing.

Here's the main hypothesis about user needs, and how they'll interact with the product:
- Important meeting or event?  The user can capture that the meeting happened and take free-text notes.  Tracking the event accurately is an important part of building trust in the product's data, especially for new users, so we'll focus on a few specific events to start (eg., SST meeting, important parent communication, 51a filling).  This would intentionally remove options that exist in the v1 profile page, many of which are unused or not accurate enough to make decisions based on them.
- Important conversation or something else to track?  The user can take free-text notes on that too.
- SST and MTSS meetings are often places where intervention decisions are made, information can be captured about services being provided to high-need students (eg., counseling, tutoring).  The prototype here frames these as "services" rather than "interventions," trying to clarity what exactly this means.  The intent is that this captures when a staff member is actively and continually delivering services to a student.  This removes many options from the v1 profile page, since many were unused and others duplicate information tracked in other systems of record in the school (eg., after-school tutoring).  Capturing more services here is out of scope for the current project plan, but high on the priority list for a future iteration (e.g, tracking which students are receiving 1:1 reading interventions).

The profile page has some revisions to the upper right panel, showing which services the student receives, and which staff are involved in service delivery or have taken notes:
screen shot 2016-02-10 at 6 36 02 pm

Notes are shown in the column on the left, and services actively being delivered are shown in the column on the right.  Services addressing reading, math and behavior or attendance outcomes are colored to indicate that.
screen shot 2016-02-10 at 6 14 06 pm

When the user clicks "Take notes":
screen shot 2016-02-10 at 6 14 39 pm

When the user clicks "Record service delivery":
screen shot 2016-02-10 at 6 14 50 pm

On the technical side, this adds https://github.com/JedWatson/react-select and its dependencies.  My thinking was that on the backend, we'll add new tables for this data, and either perform a one-time migration from the existing notes and intervention tables, or more likely, update the read path on the v2 profile to read from the new tables and existing tables.  This isn't a direct translation, since v2 removes some services, and this is something that we'll need to get feedback on and test with this prototype.  But the idea would be that interventions being removed would be interpreted as plain notes in the v2 UI.


kevinrobinson:
This also updates the demo seeding tasks to create interventions and notes, for local UI testing.  They're minimally accurate, mostly just keeping the shape of the data.

PR #54, opened 2016-02-11
Profile v2 - UI read and write paths for event notes

kevinrobinson:
This is part of the student profile v2 page: https://github.com/studentinsights/studentinsights/issues/5.

This PR adds the read and write paths to the UI for the backend work in https://github.com/studentinsights/studentinsights/pull/51.  It's in support of this feature:

image

This moves a bit further towards a Flux-style design, hoisting most state to the `PageContainer`, and centralizing responses to user actions there.  It defines props for components to use: `actions` to perform user actions and `requests` as a bag of state for tracking in-flight requests.  At the top-level component, it uses an `api` object to make actual server calls.

This also adds a few high-level integrations tests in Jasmine for the profile v2 page, and jasmine-jquery for doing assertions like `expect(element).toContainText('foo')`.  The application.js dependencies are made more explicit, as part of factoring out more pieces of code.


kevinrobinson:
This PR also updates the travis.yml to install PhantomJS 2, to work around problems with its support for Function.bind.  Swapping another polyfill and es5-shim didn't work, but this does the trick and removes the need for the polyfill altogether.

PR #58, opened 2016-02-14
Adding math charts to profile page v2

erose:


kevinrobinson:
@erose This looks like a great first step!  I'd love to merge this if you resubmit.  You can probably see there's good places to factor out some shared code between `ELADetails` and `MathDetails`, but I think it's cool to do that in a second pass.

erose:
Sure! (yeah some refactoring is in order here). Actually didn't realize I'd closed this when I deleted the branch. New pull request submitted.

Issue #4, opened 2016-02-08
Investigate data irregularities with attendance


alexsoble:
From https://github.com/codeforamerica/somerville-teacher-tool/issues/96

alexsoble:
## TODO 

+ [ ] Handle past attendance records with invalid dates
+ [ ] Investigate {absences} + {tardies} != 100% of events

alexsoble:
First, investigating attendance events that are neither absences not tardies.

Wrote a script to analyze the attendance data dump from X2: f49bd4972308b6e24210bc5b6a79b5715274653f

Ran this against the first 100,000 lines (out of 361,099) because the CSV parser hits malformed rows after that. So this analysis should be considered ___incomplete___, but it's a start.

Output from the script (aggregates only, of course):

```
COUNTS FOR INDICATOR == "1" (TRUE)

att_tardy_ind  =>  38205 out of 99983
att_tardy_ind_02  =>  0 out of 99983
att_dismissed_ind  =>  10280 out of 99983
att_dismissed_ind_02  =>  0 out of 99983
att_excused_ind  =>  1091 out of 99983
att_excused_ind_02  =>  0 out of 99983
att_absent_ind  =>  49963 out of 99983
att_absent_ind_02  =>  0 out of 99983


COUNTS FOR INDICATOR == "0" (FALSE)

att_tardy_ind  =>  61778 out of 99983
att_tardy_ind_02  =>  99983 out of 99983
att_dismissed_ind  =>  89703 out of 99983
att_dismissed_ind_02  =>  99983 out of 99983
att_excused_ind  =>  98892 out of 99983
att_excused_ind_02  =>  99983 out of 99983
att_absent_ind  =>  50020 out of 99983
att_absent_ind_02  =>  99983 out of 99983
```

From the X2 data dictionary:

ATT_ABSENT_IND => Absent?
ATT_ABSENT_IND_02 => Absent PM?
ATT_DISMISSED_IND => Dismissed?
ATT_DISMISSED_IND_02 => Dismissed PM?
ATT_EXCUSED_IND => Excused?
ATT_EXCUSED_IND_02 => Excused PM?
ATT_TARDY_IND => Tardy?
ATT_TARDY_IND_02 => Tardy PM?

## What this means

* First of all: nothing conclusive, because we still have to analyze the remainder of the data

* PM columns all blank so far

* Small number of excused absences/tardies; my intuition is that these events are likely low priority to import because they don't necessarily indicate a student in need of extra support

[...]

PR #64, opened 2016-02-15
Remove counter_cache from attendance events

kevinrobinson:
This removes the use of `counter_cache` on attendance models to fix a bug in yearly counts for discipline events, absences and tardies that affects the v1 profile page.  This may be related to https://github.com/studentinsights/studentinsights/issues/4 -  I only focused on the immediate problem here though and didn't dig further.

I think the root bug here is from a misunderstanding about how `counter_cache` works.  The work in https://github.com/studentinsights/studentinsights/commit/b7aba340fbec0079e65fac9218de44548645b061 created models for `Tardy`, etc. and added `counter_cache` options to them, along with the the corresponding columns to the `StudentSchoolYear` model.  But the migration to add these columns set the default counters to 0.  Adding `counter_cache` only adds hooks to the ActiveRecord lifecycle events to increment and decrement this column (http://guides.rubyonrails.org/association_basics.html#counter-cache), and on its own it won't do anything smart to set the initial value if there are existing associations.  So as a result the `counter_cache` columns all under-count.  I think in order to do a refactoring like this in the future, we'd need to read the current counts and set them as the default value in the migration code.

Relatedly, I discovered that calling `student.most_recent_school_year.absences.size` on a collection with `counter_cache: true` will lead to reading the value from the counter cache.  So in order to see the inconsistency in the database you have to use `.length` to load all the rows and count them, or `.count` to explicitly issue a count query.  I was using student 422 as a test case.

```
irb(main):049:0> student.most_recent_school_year.absences.size
=> 0
irb(main):050:0> student.most_recent_school_year.absences.count
=> 9
irb(main):051:0> student.most_recent_school_year.absences.length
=> 0
irb(main):053:0> student.most_recent_school_year.absences_count
=> 9
```

From an engineering perspective, this was counter-intuitive to me (and I assume to @alexsoble too).  If we want to precompute values for performance, I think it'd be better to do this in the nightly import process or in a separate precompute task so that the dataflow stays explicitly one-way.

From a product perspective, the UI features on the profile v1 page that rely on this are showing inaccurate data in production now, so this will fix that.  I'm not sure the optimization to precompute the counts here has much product impact - for a single school year this should be scanning at most ~180 rows at the end of the school year.  Right now, the median absences is 3 and median tardies is 1.

Also, the charts on the profile v1 page (that this optimization was intended to serve), is geared towards seeing attendance events over several years, which isn't super important for SST/MTSS meetings.  It's deprecated and we're actively building the v2 profile page, so it makes more sense to fix forward.  The attendance, tardies and discipline data on the v2 profile page is unaffected by this bug and already working in production, so there's no net feature loss.

### DIFF
--- a/app/models/absence.rb
+++ b/app/models/absence.rb
@@ -1,4 +1,4 @@
 class Absence < ActiveRecord::Base
-  belongs_to :student_school_year, counter_cache: true
+  belongs_to :student_school_year
   validates_presence_of :student_school_year, :occurred_at
 end
--- a/app/models/discipline_incident.rb
+++ b/app/models/discipline_incident.rb
@@ -1,4 +1,4 @@
 class DisciplineIncident < ActiveRecord::Base
-  belongs_to :student_school_year, counter_cache: true
+  belongs_to :student_school_year
   validates_presence_of :student_school_year, :occurred_at
 end
--- a/app/models/student.rb
+++ b/app/models/student.rb
@@ -122,18 +122,6 @@ class Student < ActiveRecord::Base
     end
   end
 
-  def absences_count_by_school_year
-    student_school_years.map(&:absences_count)
-  end
-
-  def tardies_count_by_school_year
-    student_school_years.map(&:tardies_count)
-  end
-
-  def discipline_incidents_count_by_school_year
-    student_school_years.map(&:discipline_incidents_count)
-  end
-
   def serialized_student_data
     {
       student: self,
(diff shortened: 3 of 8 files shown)


kevinrobinson:
Also, for absence data, this was affecting 80% of students (presumably any who had absences after the counter columns were added).

```
def absence_errors(student)
  if student.most_recent_school_year.absences_count == student.most_recent_school_year.absences.length
    []
  else
    [{ student: student, error: :absences }]
  end
end

students_with_absence_errors = Student.all.flat_map {|student| absence_errors(student) };nil
(100 * students_with_absence_errors.length / Student.all.length).to_i.to_s + '%'
=> '80%'
```

Removing the columns will mean we don't have to validate this either.

kevinrobinson:
FYI @alexsoble this worked as expected, and performance is fine for the handful I spot-tested (eg., student 316).

PR #81, opened 2016-02-17
The logo is now clickable and takes you back to the home page.

erose:


kevinrobinson:
Awesome, thanks Mohan for spotting this and @erose for making it happen!

PR #83, opened 2016-02-17
Fix STAR "Math" VS. "Mathematics"

alexsoble:
## Notes

+ No more importing information about STAR from X2 Aspen

+ STAR's SFTP server is the only source of truth about STAR assessments
  + Aspen X2 was storing these assessments with the subject "Math"
  + The same information was coming in from STAR's SFTP server with the subject "Mathematics"

+ Use only "Mathematics", validate STAR subjects as either "Reading" or "Mathematics"

### DIFF
--- a/app/demo_data/fake_star_math_result_generator.rb
+++ b/app/demo_data/fake_star_math_result_generator.rb
@@ -6,8 +6,8 @@ class FakeStarMathResultGenerator
     @student = student
   end
 
-  def star_math_assessment_id
-    @assessment_id ||= Assessment.find_by_family_and_subject('STAR', 'Math').id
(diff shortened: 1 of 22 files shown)


kevinrobinson:
@alexsoble This is an awesome improvement!  It looks like there's a bug somewhere, this is production right now:

At [https://somerville.studentinsights.org/schools/hea](https://somerville.studentinsights.org/schools/hea):

screen shot 2016-02-17 at 10 07 26 am

From looking quickly, I think it's a UI or controller bug, since the v1 and v2 student profile pages both show data for STAR Math.  I'll take a look later this morning.

PR #85, opened 2016-02-17
Fix bug with matching first and last name in searchbar

kevinrobinson:
This came in a @codeforboston user testing session yesterday, with help from some awesome Northeastern user interaction students.

Spaces in names weren't being handled correctly.  If you tried to disambiguate a first name with multiple matches by typing the last name, you'll get no matches back.  This fixes that with a slightly smarter match algorithm and adds test cases demonstrating the behavior.


kevinrobinson:
Examples:
screen shot 2016-02-17 at 1 30 16 pm
screen shot 2016-02-17 at 1 30 23 pm
screen shot 2016-02-17 at 1 30 28 pm

PR #86, opened 2016-02-17
Minimal fix for viewing MCAS math data in UI

kevinrobinson:
This is leading to MCAS charts not having data on student profile v1 and v2 pages, so this makes the minimal fix to get it working again.  Will rename the UI code to 'mathematics' to simplify, and add some related tests after confirming this fixes the problem.

### DIFF
--- a/app/charts/student_profile_chart.rb
+++ b/app/charts/student_profile_chart.rb
@@ -42,9 +42,9 @@ class StudentProfileChart < Struct.new :data
     {
       star_series_math_percentile: percentile_ranks_to_highcharts(data[:star_math_results]),
       star_series_reading_percentile: percentile_ranks_to_highcharts(data[:star_reading_results]),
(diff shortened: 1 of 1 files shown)


kevinrobinson:
Actually in looking at this more closely, I don't think there's much value in renaming MCAS to be mathematics in the UI code.  STAR is still called "math," so no need to leak these database key details out to the UI, everything can just be "math" there.

As for testing, I added a minimal test case here (essentially typing it twice), and I think a next step would be to either generate the JS fixtures for specs from Rails, or write an RSpec test that compares the JS fixture for Jasmine tests to the format produced by the Rails serialization code.  I tried a quick pass at this, but it also means updating the student factory to produce a fully valid student with all the related assessment and attendance data.  I think it makes sense to punt on that for now, but if it comes up again we can add that to better test the seam better API and UI.

PR #90, opened 2016-02-17
Refactor serialized_student_data out of model and into StudentProfileChart

kevinrobinson:
There's already instance and class methods called `serialize_data` on the `Student` model, so this one's a bit confusing.  It's only used to grab some fields for use in `StudentProfileChart`, so this moves it there and isolates it, and we can simplify that code on its own later.


spec/charts/student_profile_chart_spec.rb
```diff
@@ -15,8 +15,7 @@
   describe '#chart_data' do
     let(:student) { FactoryGirl.create(:student) }
     it 'has the expected keys' do
-      puts student.serialized_student_data.to_json
```
kevinrobinson: oops, unrelated leftover `puts`

PR #92, opened 2016-02-17
Factor out StudentProfileHeader, in prep for improving it

kevinrobinson:
This is part of https://github.com/studentinsights/studentinsights/issues/5.

No visual changes, this is a pure refactoring in prep for other improvements.

No regressions:
screen shot 2016-02-17 at 5 45 13 pm

### DIFF
--- a/app/assets/javascripts/student_profile_v2/student_profile_header.js
+++ b/app/assets/javascripts/student_profile_v2/student_profile_header.js
@@ -1,0 +1,55 @@
+(function() {
+  window.shared || (window.shared = {});
+  var dom = window.shared.ReactHelpers.dom;
+  var createEl = window.shared.ReactHelpers.createEl;
+  var merge = window.shared.ReactHelpers.merge;
+  var Routes = window.shared.Routes;
+
+  var styles = {
+    titleContainer: {
+      fontSize: 16,
+      padding: 20
+    },
+    nameTitle: {
+      fontSize: 20,
+      fontWeight: 'bold'
+    },
+    titleItem: {
+      padding: 5
+    }
+  };
+
+  /*
+  This pure UI component renders top-line information like the student's name, school and
+  classroom.
+  */
+  var StudentProfileHeader = window.shared.StudentProfileHeader = React.createClass({
+    displayName: 'StudentProfileHeader',
+
+    propTypes: {
+      student: React.PropTypes.object.isRequired
+    },
+
+    render: function() {
+      var student =  this.props.student;
+      return dom.div({ style: styles.titleContainer },
+        dom.a({
+          href: Routes.student(student.id),
+          style: styles.nameTitle
+        }, student.first_name + ' ' + student.last_name),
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/student_profile_v2/student_profile_header.js
```diff
@@ -0,0 +1,55 @@
+(function() {
+  window.shared || (window.shared = {});
+  var dom = window.shared.ReactHelpers.dom;
+  var createEl = window.shared.ReactHelpers.createEl;
+  var merge = window.shared.ReactHelpers.merge;
+  var Routes = window.shared.Routes;
```
kevinrobinson: This is roughly analogous to `require` in Ruby.

app/assets/javascripts/student_profile_v2/student_profile_header.js
```diff
@@ -0,0 +1,55 @@
...
+
+  var styles = {
```
kevinrobinson: These define styles (similar to what you would do in CSS), and you use them in the component (see https://facebook.github.io/react/tips/inline-styles.html).

This approach somewhat controversial and bleeding edge (rather than using CSS), but it's just as simple to use.  If you're curious about the rationale, [these slides](https://speakerdeck.com/vjeux/react-css-in-js) has more context about challenges with using CSS.

app/assets/javascripts/student_profile_v2/student_profile_header.js
```diff
@@ -0,0 +1,55 @@
...
+  var StudentProfileHeader = window.shared.StudentProfileHeader = React.createClass({
+    displayName: 'StudentProfileHeader',
+
+    propTypes: {
+      student: React.PropTypes.object.isRequired
+    },
+
+    render: function() {
+      var student =  this.props.student;
+      return dom.div({ style: styles.titleContainer },
```
kevinrobinson: If you Google for React examples, you'll find JSX that looks like HTML with brackets.  This works the same way but is just using plain JavaScript syntax instead.

spec/javascripts/student_profile_v2/student_profile_header_spec.js
```diff
@@ -0,0 +1,31 @@
...
+  };
+
+  SpecSugar.withTestEl('high-level integration tests', function() {
+    it('renders note-taking area', function() {
+      var el = this.testEl;
+      helpers.renderInto(el);
+
+      expect(el).toContainText('Daisy Poppins');
+      expect(el).toContainText('Arthur D Healey');
+      expect($(el).find('a.homeroom-link')).toContainText('102');
```
kevinrobinson: These are some high-level Jasmine tests.  They render the component on the page with fixture data for a student, and then test whether things like the student's name and school are on the page.  You can run Jasmine tests at [http://localhost:3000/teaspoon/default](http://localhost:3000/teaspoon/default) when Rails is running.

PR #101, opened 2016-02-20
Added bullet points and spacing

BWieber:
Hi Student Insights Team! 

Sorry about the Rails mixup - the header now looks very much like the mock up. In this current version, there are links which make the text _blue_, all except "Grade 5" - it looks a little odd with one being different. 

Also, I am not sure how to incorporate if/else logic into react to display 1st, 2nd, 3rd, etc. Let's get that on Tuesday! 

_Brandon_

### DIFF
--- a/app/assets/javascripts/student_profile_v2/student_profile_header.js
+++ b/app/assets/javascripts/student_profile_v2/student_profile_header.js
@@ -11,10 +11,11 @@
       padding: 20
     },
     nameTitle: {
-      fontSize: 20,
+      fontSize: 30,
       fontWeight: 'bold'
     },
     titleItem: {
+      fontSize: 25,
       padding: 5
     }
   };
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/student_profile_v2/student_profile_header.js
```diff
@@ -40,16 +41,16 @@
         dom.a({
           href: Routes.school(student.school_id),
           style: styles.titleItem
-        }, student.school_name),
+        }, '\n\u2022 ' + student.school_name + '\n\u2022'),
```
kevinrobinson: You can use the • character directly in JavaScript, rather than escaping.

kevinrobinson: By including the bullet in the body of the `A` tag, it means the bullet itself is clickable, which probably isn't what someone would expect.  It'd be better to split it out to its own `dom.span` tag, which will also let you apply any padding if you need to.

kevinrobinson:
@BWieber awesome, this is a great start!  If you can attach a screenshot of what this looks like locally, this is about ready to merge.

If my comments make sense and you want to take those suggestions then go for it and push those commits on top.  Or if you have questions about them or think a different approach would be better that's great too. :)

kevinrobinson:
As for making convert the grade number to a user-facing string, that can be split out to a method (something like `formatGradeNumber: function(grade) { ... }`, happy to work with you on that Tuesday.

BWieber:
Hi Kevin,

I pushed the adjusted branch - each bullet point now has its own dom.span, so they are no longer clickable. Take a look here: 

profile header bullets

kevinrobinson:
@BWieber Nice work, this looks great!  Thanks again for all your help with this, :ship: .

PR #104, opened 2016-02-23
Refactoring profile js

erose:
DRYing up the code related to displaying charts in profile.js. Introduced a new component, chart.js, which wraps the quadsToPairs function and some other settings. In the future it should probably incorporate ProfileChartSettings.js from the old profile folder.

This pull request deals with all of the `TODO: factor out` comments in profile v2 charts code, except for `percentileYAxis`, which doesn't belong in all charts. (More thought should be given to this).

Let me know any more info that's needed. (There's screenshot since this is pure refactoring and has no visual effect.)


app/assets/javascripts/helpers/react_helpers.js
```diff
@@ -6,6 +6,8 @@
     dom: React.DOM,
     createEl: React.createElement.bind(React),
     merge: function() {
+      // Combines several objects into one.
+      // e.g. merge({a: 1, b: 2}, {c: 'hi!'}) = {a: 1, b: 2, c: 'hi!'}
```
kevinrobinson: nice!

app/assets/javascripts/student_profile_v2/chart.js
```diff
@@ -0,0 +1,77 @@
...
+  window.shared || (window.shared = {});
+  var dom = window.shared.ReactHelpers.dom;
+  var createEl = window.shared.ReactHelpers.createEl;
+  var merge = window.shared.ReactHelpers.merge;
+
+  var ProfileChartSettings = window.ProfileChartSettings;
+  var HighchartsWrapper = window.shared.HighchartsWrapper;
+
+  // Component for all charts in the profile page.
+  window.shared.Chart = React.createClass({
```
kevinrobinson: Nice description and comments here too!  The `Chart` name is a bit generic, I'd recommend something a bit more explicit, maybe something like `ProfileHistoricalChart`?

app/assets/javascripts/student_profile_v2/chart.js
```diff
@@ -0,0 +1,77 @@
...
+    displayName: 'Chart',
+
+    propTypes: {
+      series: React.PropTypes.arrayOf( // you can plot multiple series on the same graph
+        React.PropTypes.shape({
+          name: React.PropTypes.string.isRequired, // e.g. 'Scaled score'
+          data: React.PropTypes.array.isRequired // [year, month, date, value] quads
+        })
+      ),
+      title: React.PropTypes.string.isRequired, // e.g. 'MCAS scores, last 4 years'
```
kevinrobinson: `title` is a Highcharts options (where you can pass the text and alignment or other options), but this component only allows modifying the text.  So I'd suggest making that explicit with `titleText` or something like that.

app/assets/javascripts/student_profile_v2/chart.js
```diff
@@ -0,0 +1,77 @@
...
+      return merge(ProfileChartSettings.base_options, {
+        xAxis: merge(ProfileChartSettings.x_axis_datetime, {
+          plotLines: this.x_axis_bands,
+          min: this.props.timestampRange.min,
+          max: this.props.timestampRange.max
+        })
+      });
+    },
+
+    getDefaultProps: function() {
```
kevinrobinson: This is pretty minor, but I'd suggest moving this method up so it's together with `propTypes`.  In general I think it's helpful practice to keep all the React lifecycle methods up front.

app/assets/javascripts/student_profile_v2/chart.js
```diff
@@ -0,0 +1,77 @@
...
+        intervalBack: intervalBack,
+        timestampRange: {
+          min: moment(now).subtract(4, 'years').toDate().getTime(),
+          max: now.getTime(),
+        }
+      };
+    },
+
+    render: function() {
+      var self = this;
```
kevinrobinson: Alternately you can pass the `this` value to `.map` like so: `this.props.series.map(function(obj) { ... }, this)`.  Cool either way.

app/assets/javascripts/student_profile_v2/chart.js
```diff
@@ -0,0 +1,77 @@
...
+      return createEl(HighchartsWrapper, merge(this.baseOptions(), {
+        title: {
+          text: this.props.title,
+          align: 'left'
+        },
+        series: this.props.series.map(function(obj){
+          return {name: obj.name, data: self.quadsToPairs(obj.data || [])}
+        }),
```
kevinrobinson: This is a bit tricky since `series` is an option to Highcharts that has the shape `{name, data}` but this adds a translation step that I don't think you'd expect when using this component.  I'd suggest either remove the translation `quadsToPairs` and require callers to do that translation themselves, or rename `series` to something that makes clear that the value will be translated and not passed through directly to Highcharts.  Happy to talk this out more tonight too!

kevinrobinson:
@erose This looks awesome!  I left a few comments mostly just clarity suggestions, let me know what you think either way and then let's merge this tonight! :tada:

PR #108, opened 2016-02-24
Risk Bubble

BWieber:
I figured I'd do this in pieces -  I made the bubble for the profile header. I'll implement flex box next, and get "Risk Level" to display. 

risk bubble

### DIFF
--- a/app/assets/javascripts/student_profile_v2/risk_bubble.js
+++ b/app/assets/javascripts/student_profile_v2/risk_bubble.js
@@ -1,0 +1,56 @@
+(function() {
+  window.shared || (window.shared = {});
+  var dom = window.shared.ReactHelpers.dom;
+  var createEl = window.shared.ReactHelpers.createEl;
+  var merge = window.shared.ReactHelpers.merge;
+
+  var styles = {
+    riskBubble: {
+      fontSize: 20,
+      width: '35',
+      height: '35',
+      color: 'white',
+      borderRadius: '30',
+      paddingTop: '3',
+      textAlign: 'center',
+      marginTop: '10',
+      marginRight: '25',
+      display: 'inline-block',
+      float: 'right'
+    },
+
+    riskItem: {
+      fontSize: 25,
+      padding: 5,
+      float: 'right',
+      marginTop: '3'
+    }
+};
+
+  var RiskBubble = window.shared.RiskBubble = React.createClass({
+    displayName: 'RiskBubble',
+
+    propTypes: {
+      riskLevel: React.PropTypes.number.isRequired
+    },
+
+    render: function() {
+      return dom.div({},
+        dom.span({
+          style: merge(styles.riskBubble, { backgroundColor: this.bubbleColor() })
+        }, this.props.riskLevel),
+        dom.span({
+          style: styles.riskItem
+        }, "Risk Level")
+      );
+
+    },
+
+    bubbleColor: function() {
+      if (this.props.riskLevel === 1) return 'green';
+      if (this.props.riskLevel === 2) return '#ffc41d';
+      if (this.props.riskLevel === 3) return 'red';
+    }
+  })
+
+})();
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -50,6 +50,7 @@
 
 // student profile page v2:
 // pure ui components:
+  //= require ./student_profile_v2/risk_bubble
   //= require ./student_profile_v2/service_color
   //= require ./student/profile_charts/profile_chart_settings
   //= require ./student_profile_v2/feed_helpers
--- a/app/assets/javascripts/student_profile_v2/student_profile_header.js
+++ b/app/assets/javascripts/student_profile_v2/student_profile_header.js
@@ -4,6 +4,20 @@
   var createEl = window.shared.ReactHelpers.createEl;
   var merge = window.shared.ReactHelpers.merge;
   var Routes = window.shared.Routes;
+  var RiskBubble = window.shared.RiskBubble;
+
+
+  var formatGradeNumber = function(grade) {
+      if(grade === 1){
+        return "1st";
+      } else if (grade === 2) {
+        return "2nd";
+      } else if (grade === 3) {
+        return "3rd";
+      } else {
+        return grade + "th";
+      }
+  }
 
   var styles = {
     titleContainer: {
@@ -50,7 +64,7 @@
         }, '•'),
         dom.span({
           style: styles.titleItem
-        }, student.grade),
+        }, formatGradeNumber(student.grade) + ' Grade'),
         dom.span({
           style: styles.titleItem
         }, '•'),
@@ -58,7 +72,10 @@
           className: 'homeroom-link',
           href: Routes.homeroom(student.homeroom_id),
           style: styles.titleItem
-        }, 'Homeroom ' + student.homeroom_name)
+        }, 'Homeroom ' + student.homeroom_name),
+        createEl(RiskBubble, {
+          riskLevel: student.student_risk_level.level
+        })
       );
     }
   });


kevinrobinson:
@BWieber Nice!  A few pieces of feedback for you here.

This has the commit from https://github.com/studentinsights/studentinsights/pull/101 in it, which is cool, but you pushed an additional commit to that branch that this branch doesn't include.  This is a pretty normal workflow and to be expected, but we'll have to resolve that before merging this pull request, otherwise that additional commit from the other branch will be undone.  This is a tricky one, since git thinks it can resolve things automatically at the moment, but the automatic resolution would undo part of the commit on the other branch.  You can see this looking at the diff in this pull request - the nice UTF8 bullets are gone.

To fix this, you can checkout to `risk_box` and then `git merge origin/master` to merge master in, since it now has your commits from that other branch.  When you do that, you'll see there are merge conflicts.  That's because git detects a conflict between the additional commit https://github.com/BWieber/studentinsights/commit/a3379db6ad5dfa78f276357b73572580371c397e and the risk bubble commit https://github.com/BWieber/studentinsights/commit/8d67e810173d438b229c263fb3b81b01a03b398e, and can't resolve it automatically.

These are the relevant lines:
screen shot 2016-02-24 at 10 17 24 am
screen shot 2016-02-24 at 10 17 33 am

This should be straightforward to resolve, and if you need more info on merge conflicts GitHub's a [good source](https://help.github.com/articles/resolving-a-merge-conflict-from-the-command-line/).  Then you can push up the branch again, and the conflict should be resolved!

kevinrobinson:
Also, this pull request always uses `3` for the risk level, but it needs to show the actual data :)  It looks like this data isn't actually being sent down to the UI at the moment, so I added that in https://github.com/studentinsights/studentinsights/pull/118.  After merging master in (above), you should be able to use this.  The shape of the data looks like this, and `level` is the field you want:

screen shot 2016-02-24 at 9 19 58 am

kevinrobinson:
And finally, checking this out locally, it looks like the risk level bubble looks bigger than the screenshot here.  This is what I see:

screen shot 2016-02-24 at 9 15 59 am

I think the key things is that this needs to be `display: inline-block` in order for sizing rules to have the effect you want.  And then you can tone down the different size params a bit.

I'll let you take a look at this, and am happy to help finish this off later this week too if that's helpful.  I should have allocated more time last night to look at this with you too, so my bad and we can make sure to do that next time. :)

kevinrobinson:
@BWieber Cool, if you can make it tomorrow, let's work together to finish these off and get them merged.  If you can't make it tomorrow that's cool too just let me know and I can help out.

BWieber:
Hi Kevin!

I can definitely make it tomorrow - I have the controller work done too. I
might need a little git help to ensure that I don't keep posting conflicts
- I think I may not be pulling from the right repo!

kevinrobinson:
@BWieber Yeah no worries!  I can help resolve it, but let's try doing it together then so we can talk it out.  Thanks again for your help!

BWieber:
Merged with the upstream to clear up previous conflicts!

BWieber:
Merged `format_student_grade` to ease the branching.

kevinrobinson:
For my own memory, the next steps we talked about at Code for Boston were to factor out RiskLevel to its own React component, and then update the color logic.  Then we can :ship: it!

BWieber:
Hi Kevin! 

The risk bubble is now a part of it's own component, and is currently being successfully rendered. The coloring is also in full effect, 1 (Green), 2 (Yellow), 3 (Red). I believe all of the students currently have a 2 value, but I tested it in development and it ran successfully.  

Here is the current image: 

risk_bubble_pr

As you can see, the current Risk Level and bubble are a bit offset. I discovered what was causing the problem, but I went through a few iterations without success. 

The code in question is this: 

```
    render: function() {
      return dom.div({},
        dom.span({
          style: merge(styles.riskBubble, { backgroundColor: this.bubbleColor() })
        }, this.props.riskLevel),
        dom.span({
          style: styles.riskItem
        }, "Risk Level")
      );

    },

    bubbleColor: function() {
      if (this.props.riskLevel === 1) return 'green';
      if (this.props.riskLevel === 2) return '#ffc41d';
      if (this.props.riskLevel === 3) return 'red';
    }
  })
```

This renders the above image. The reason the element is offset is because of the dom.div call. I have tried to remove it and it works... except the bubbleColor function no longer renders, this no bubble. I tried a various set of attempts, but to no avail - I kept getting the following error:

`Uncaught Error: Invariant Violation: RiskBubble.render(): A valid ReactComponent must be returned. You may have returned undefined, an array or some other invalid object.`

Using a console.log test, RiskBubble was indeed being called - but I couldn't discover what was throwing this error. 

I'll take a further look soon. Let me know if you know the answer! 

_Brandon_

kevinrobinson:
@BWieber nice debugging, and awesome write-up!  :smile:

You're spot on that it's that `div` that's causing the problem - `div` is a "block" element, which means by default it starts a new line and takes up the entire width of the container.  The other elements, like `a` and `span` are "inline" elements, which means they flow like text would, left to right.  This article looks like it has a decent [explanation](http://designshack.net/articles/css/whats-the-deal-with-display-inline-block/).

Here, you can change that `div` to a `span` and that should help move in the right direction.

BWieber:
Hi Kevin, 

Thanks for the resource and the suggestion - the risk bubble is no being properly rendered with the right color and style! Take a look: 

risk_bubble_final

Risk Bubble will properly respond to the integers 0, 1, 2, and 3. Does the risk level go any higher than that? I also changed the coloring options to match the Homeroom page - the bubbles have consistent coloration across the app.  Otherwise, I think this PR will get this closed. Let me know if it needs anything else ! 

_Brandon_

kevinrobinson:
@BWieber Awesome!  :smile:  I will take a look at this tomorrow if I'm able, and if not let's definitely get this fully shipped this week.  Have a great weekend, and thanks again for all your help!

app/assets/javascripts/student_profile_v2/risk_bubble.js
```diff
@@ -0,0 +1,56 @@
...
+  window.shared || (window.shared = {});
+  var dom = window.shared.ReactHelpers.dom;
+  var createEl = window.shared.ReactHelpers.createEl;
+  var merge = window.shared.ReactHelpers.merge;
+
+  var styles = {
+    riskBubble: {
+      fontSize: 20,
+      width: '35px',
+      height: '35px',
```
kevinrobinson: For React styles, you can use plain numbers here (eg., `35` instead of `35px` like you would in CSS).  React will always interpret numbers as being pixels.

db/schema.rb
```diff
@@ -188,11 +188,10 @@
 
   create_table "services", force: true do |t|
```
kevinrobinson: I don't think this should be part of this pull request, right?

kevinrobinson:
@BWieber This looks good!  One minor piece of style feedback, and then a stray migration got checked in here, but if you can take care of those let's ship this! :+1:  

Also, @johnrobbjr had some design suggestions in https://github.com/studentinsights/studentinsights/issues/151#issuecomment-195026223, but let's merge this first and address that separately.

BWieber:
I _believe_ this should have removed the stray migration change. I pulled from master and pushed it back.

kevinrobinson:
@BWieber This looks like something drifted from the screenshot here, looking at the demo site on [this page](https://somerville-teacher-tool-demo.herokuapp.com/students/5/profile).

screen shot 2016-03-16 at 9 13 46 am


I haven't deployed this to production yet, so am going to revert this and then we can investigate without having to do a hot fix.

@BWieber I am a little backed up this week, but let me know if you aren't able to reproduce this locally.

PR #113, opened 2016-02-24
formatGradeNumber function 

BWieber:
Hi Kevin, 

@erose and I worked on this, and got it up and running :) Unfortunately the students in the demo file are only in the 5th grade, but as you can see it is working for that use-case. Take a look! 

formatstudentgrade

### DIFF
--- a/app/assets/javascripts/student_profile_v2/student_profile_header.js
+++ b/app/assets/javascripts/student_profile_v2/student_profile_header.js
@@ -5,6 +5,18 @@
   var merge = window.shared.ReactHelpers.merge;
   var Routes = window.shared.Routes;
 
+  var formatGradeNumber = function(grade) {
+      if(grade === 1){
+        return "1st";
+      } else if (grade === 2) {
+        return "2nd";
+      } else if (grade === 3) {
+        return "3rd";
+      } else {
+        return grade + "th";
+      }
+  }
+
   var styles = {
     titleContainer: {
       fontSize: 16,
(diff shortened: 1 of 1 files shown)


kevinrobinson:
@BWieber :facepunch: This looks good, thanks!

To merge this, you'll have to do a similar merge as I described in https://github.com/studentinsights/studentinsights/pull/108#issuecomment-188304087, since there are additional commits on the other branches.  My recommendation would be to finish off https://github.com/studentinsights/studentinsights/pull/108 first, then move to this one.  And in general, I find that when making changes like this that are so closely related and changing together in the span of a few hours, it's often easier to work on one branch rather than a separate ones.

This does the trick for our limited use case as you mention, so I'd say after merging in the other commits, let's :ship: this.  FWIW, [numeral.js](http://numeraljs.com/) is a library that handles these kinds of string conversions and is usually a better bet when you want something more general.

BWieber:
Hey Kevin! 

I checked out both branches and I can identify any merge conflicts. I switched to both the `risk_box` branch and the `format_student_grade` branch, and both of them were already up to date with master. 

Let me know if there is another step I am missing, otherwise, all is up to date!

kevinrobinson:
@BWieber cool, let's talk it out manana. :)

kevinrobinson:
@BWieber let's close this and move the work to https://github.com/studentinsights/studentinsights/pull/108 like we just talked about!

BWieber:
Closed - Branched was merged with `risk_box`

PR #118, opened 2016-02-24
Profile v2 - include risk level when serializing student data, move serialization to controller

kevinrobinson:
This is part of https://github.com/studentinsights/studentinsights/pull/5.

This adds in data needed for https://github.com/studentinsights/studentinsights/pull/108.  It removes the generic-seeming `serialize_data` method from the `Student` model, since it's only used by `students#profile`.  It moves the method into the controller and adds a spec.  This also revealed an older dead method in the controller that is now removed.


kevinrobinson:
@BWieber going to merge this, since it unlocks finishing off https://github.com/studentinsights/studentinsights/pull/108.

Issue #115, opened 2016-02-24
Profile v2 - Allow saving services to the database and rendering them


kevinrobinson:
This is a part of https://github.com/studentinsights/studentinsights/issues/5.

Rough steps, using https://github.com/studentinsights/studentinsights/pull/51 and https://github.com/studentinsights/studentinsights/pull/54 as models.

In Rails:
- [x] https://github.com/studentinsights/studentinsights/pull/132 - Make migrations and models (for `services` and `service_types`), test in Rails console (@BWieber)
- [x] https://github.com/studentinsights/studentinsights/pull/158 - Read path for services in students_controller#profile
- [x] https://github.com/studentinsights/studentinsights/pull/158 - Add new controller action to enable saving, update specs

In React (@kevinrobinson):
- [x] https://github.com/studentinsights/studentinsights/pull/124 - Factor out `RecordService` component
- [x] https://github.com/studentinsights/studentinsights/pull/125, https://github.com/studentinsights/studentinsights/pull/131 - Update read path to show list of services
- [x] https://github.com/studentinsights/studentinsights/pull/131 - Update read path for "notes" to show older interventions and progress notes
- [x] https://github.com/studentinsights/studentinsights/pull/136 - Add `actions` and `api` to the UI code to enable saving
- [ ] Render custom intervention type in UI
- [ ] Fix UI bug on IE with note button width (see fix note button width: https://github.com/studentinsights/studentinsights/pull/126#issuecomment-188940902)

kevinrobinson:
@BWieber I've got some time today and tomorrow, so am going to grab the next Rails tasks here to try to get this feature fully shipped this week.  Thanks so much for your help starting this off!  :+1:

BWieber:
No problem thanks Kevin! I wish I had more time to dedicate 😼

kevinrobinson:
@BWieber no worries, I'm thankful for all the help you've been able to give!  I'm just taking advantage of time I have while I do :)

alexsoble:
"Allow saving services to the database and rendering them" . . . this we've got in production!

PR #124, opened 2016-02-24
Factor out RecordService component, in prep for enabling recording services in database

kevinrobinson:
This is part of https://github.com/studentinsights/studentinsights/issues/115 and https://github.com/studentinsights/studentinsights/issues/5.

This refactors the UI code for recording a new service into its own component and adds some basic specs for it.

No visual or functional regressions (although "Save" is still intentionally disabled until the backend pieces are ready):
screen shot 2016-02-24 at 12 53 12 pm

This also led to some cleanup in `InterventionsDetails`, which can ultimately be simplified or factored further as part of the additional steps in https://github.com/studentinsights/studentinsights/issues/115.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -54,6 +54,7 @@
   //= require ./student_profile_v2/quad_converter
   //= require ./student_profile_v2/academic_summary
   //= require ./student_profile_v2/take_notes
+  //= require ./student_profile_v2/record_service
   //= require ./student_profile_v2/summary_list
   //= require ./student_profile_v2/profileChart.js
   //= require ./student_profile_v2/student_profile_header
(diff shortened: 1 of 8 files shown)


kevinrobinson:
@BWieber this is the first piece of the UI code for #115.

Also, this uses the following placeholder ids for the `service_types` right now:
```
{
  502: 'Attendance Officer',
  503: 'Attendance Contract',
  504: 'Behavior Contract',
  505: 'Counseling, in-house',
  506: 'Counseling, outside',
  507: 'Reading intervention',
  508: 'Math intervention'
}
```

You can use the same values on the server side and things will just work, or if you end up using different values that's cool and I can update these ids later to match.  :+1: either way!

PR #131, opened 2016-02-26
Profile v2 - UI read path for rendering services and notes for deprecated tables

kevinrobinson:
This updates the schools#profile controller code and UI code to render the new `services` table that will be written as part of the backend work in https://github.com/studentinsights/studentinsights/issues/115.  It also adds updates the fixture data in the controller to simulate this in development mode, and adds placeholder fixtures for the `service_types` constants table.

The notes UI shows data from `event_notes`, but also merges in data from the deprecated `interventions`, `notes` tables.  Doesn't yet show progress notes.
screen shot 2016-02-26 at 5 21 03 pm

The "services" list now shows services data (previously it actually read from the old interventions table):
screen shot 2016-02-26 at 5 16 56 pm

"Services" in the summary panel, which used to show the deprecated intervention data, now shows the new `services` data as well:
screen shot 2016-02-26 at 5 18 33 pm

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -46,6 +46,7 @@
 
 // student profile page v2:
 // components:
+  //= require ./student_profile_v2/service_color
   //= require ./student/profile_charts/profile_chart_settings
   //= require ./student_profile_v2/educator
   //= require ./student_profile_v2/highcharts_wrapper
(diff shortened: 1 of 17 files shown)


kevinrobinson:
This is not 100% done yet - tests haven't been updated, but sharing it while still in progress work since it overlaps with other work in https://github.com/studentinsights/studentinsights/issues/115.  cc @BWieber

kevinrobinson:
There's a series of commits on top, factoring out `NotesList` and `ServicesList` into their own UI components, updating the intervention demo data to include progress notes, and rendering progress notes in the UI.  It also fixes a bug with the timestamps and sort order of notes.

This is how deprecated interventions and progress notes look:
screen shot 2016-02-27 at 3 17 32 pm

PR #132, opened 2016-02-28
Added services and service_type models

BWieber:
Hi Kevin! 

I added the models we needed to get this piece rolled out. I borrowed heavily from the `event_notes` model and migration. 

I also noticed that while `event_notes` `belongs_to` educator, educator does not have a `has_many` association. I added one for services, but perhaps it was intended that an educator would not have many event_notes (though it does seem like an educator could have quite a few!) 

Let me know how this looks ! 

_Brandon_

### DIFF
--- a/app/models/educator.rb
+++ b/app/models/educator.rb
@@ -13,6 +13,7 @@ class Educator < ActiveRecord::Base
   has_many    :interventions
   has_many    :progress_notes, through: :interventions
   has_many    :student_notes
+  has_many    :services
 
   validates :email, presence: true, uniqueness: true
   validates :local_id, presence: true, uniqueness: true
--- a/app/models/educator.rb
+++ b/app/models/educator.rb
@@ -13,8 +13,7 @@ class Educator < ActiveRecord::Base
   has_many    :interventions
   has_many    :progress_notes, through: :interventions
   has_many    :student_notes
-  has_many    :services
-
+  
   validates :email, presence: true, uniqueness: true
   validates :local_id, presence: true, uniqueness: true
 
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -3,6 +3,6 @@ class Service < ActiveRecord::Base
   belongs_to :educator
   belongs_to :service_type
 
-  validates_presence_of :educator_id, :student_id, :event_note_type_id, :recorded_at
+  validates_presence_of :educator_id, :student_id, :service_type_id, :recorded_at, :date_started
 
 end
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -1,8 +1,9 @@
 class Service < ActiveRecord::Base
   belongs_to :student
-  belongs_to :educator
+  belongs_to :provided_by_educator, class_name: 'Educator'
+  belongs_to :recorded_by_educator, class_name: 'Educator'
   belongs_to :service_type
 
-  validates_presence_of :educator_id, :student_id, :service_type_id, :recorded_at, :date_started
+  validates_presence_of :recorded_by_educator_id, :provided_by_educator_id, :student_id, :service_type_id, :recorded_at, :date_started
 
 end
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -1,0 +1,8 @@
+class Service < ActiveRecord::Base
+  belongs_to :student
+  belongs_to :educator
+  belongs_to :service_type
+
+  validates_presence_of :educator_id, :student_id, :event_note_type_id, :recorded_at
+
+end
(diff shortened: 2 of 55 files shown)


BWieber:
Hmm, I just noticed that my branch has merge conflicts for some reason - I am sure that I pulled before I created a new branch! I'll get the work done on the controller, but I'll wait till Tuesday to commit unless you can guide me a bit here - I don't want to keep creating conflicts!

app/models/educator.rb
```diff
@@ -13,6 +13,7 @@ class Educator < ActiveRecord::Base
   has_many    :interventions
   has_many    :progress_notes, through: :interventions
   has_many    :student_notes
+  has_many    :services
```
kevinrobinson: This will have to be updated to use a different column if we revise the `services` table to store both which educator _recorded_ the service and which educator is _providing_ the service.  Since we don't have a need to query in either way at the moment, you could also just remove the association since they're easy to add when we do need them. :)

app/models/service.rb
```diff
@@ -0,0 +1,8 @@
+class Service < ActiveRecord::Base
+  belongs_to :student
+  belongs_to :educator
+  belongs_to :service_type
+
+  validates_presence_of :educator_id, :student_id, :event_note_type_id, :recorded_at
```
kevinrobinson: `event_note_type_id` looks like a leftover copy-pasta, this should be `:service_type_id`.

db/migrate/20160228205806_create_services.rb
```diff
@@ -0,0 +1,13 @@
+class CreateServices < ActiveRecord::Migration
+  def change
+    create_table :services do |t|
+      t.integer :student_id
+      t.integer :educator_id
```
kevinrobinson: The `educator_id` might mean who created this service in the product, or which teacher is providing the service.  I was thinking it would make sense to capture both explicitly (eg., `recorded_by_educator_id` and `assigned_by_educator_id`) and is how the [fixture data](https://github.com/studentinsights/studentinsights/blob/master/app/controllers/students_controller.rb#L167) works now.  Would that work here?

db/migrate/20160228205806_create_services.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.integer :service_type_id
+      t.text :text
+      t.datetime :recorded_at
```
kevinrobinson: This might be a useful thing to add, but we also need the date that the educator started providing the service as well.  In the UI, it's the "When did they start?" date:
screen shot 2016-02-29 at 12 32 25 pm

This is different than the date the user recorded this information, so we can add a `date_started` field here to capture this.

spec/factories/services.rb
```diff
@@ -0,0 +1,6 @@
+FactoryGirl.define do
+  factory :service do
+    educator ""
```
kevinrobinson: Unless this is being used, let's remove it.

spec/models/service_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Service, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
kevinrobinson: Same comment here, let's either add tests or remove this stub file.

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160217010342) do
+ActiveRecord::Schema.define(version: 20160228210723) do
```
kevinrobinson: This is the source of the merge conflict.  You can see this if you try to merge master into your branch, or try to rebase your branch onto master:

screen shot 2016-02-29 at 12 24 24 pm

This is happening because other migrations were merged into master after the point you cut this branch.  So `master` has one version number for the schema.  And your code created its own version number when you ran it locally.  So this is saying "hey, maybe there are upstream changes that your migration hasn't considered."  This is usually not an actual conflict, but the warning here is a helpful reminder and easy to fix.

We can do this tomorrow in person, but you can typically resolve this by looking at other recent migrations, verifying there's nothing conflicting, and then updating the `schema.rb` file to take the more recent version number.  More background is [here](http://stackoverflow.com/questions/2979059/is-the-version-number-in-rails-schema-rb-used-for-anything/4047884#4047884).

kevinrobinson:
@BWieber this is a great start, nice work!  I left a few comments on some fields that need some small revisions, and info about the merge conflict.  Happy to work on this with you tomorrow to get it sorted out and merged. :)

BWieber:
Looks like there are no more conflicts, Kevin!  Thanks for the help :)

Issue #8, opened 2016-02-08
Umbrella - Create SPED referral packet


kevinrobinson:
From https://github.com/codeforamerica/somerville-teacher-tool/issues/100

Teachers and administrators want to be able to create a SPED referral packet from all the the information in Student Insights.  This is a laborious manual process now, across several computer and paper data systems that we can help automate.

This will likely be a PDF, and will require working closely with school partners to iterate on the exact requirements.  https://github.com/codeforamerica/somerville-teacher-tool/issues/100 has more background on the types of information that might be important.
#### Rough first steps
- [x] https://github.com/studentinsights/studentinsights/pull/137 - "Hello world" PDF (@eudaimonious)
- [x] Verify that download UX is good on Windows IE (@kevinrobinson)
- [x] Add raw unstyled data from text mock (https://github.com/studentinsights/studentinsights/pull/165)
- [ ] Add info in the 'Additional Notes' section https://github.com/studentinsights/studentinsights/issues/497 (@bwieber)
#### Research and design steps
- [x] Get feedback from school folks on text mock in https://github.com/studentinsights/studentinsights/pull/165
#### Work based on feedback from our end user
- [x] Add birth date https://github.com/studentinsights/studentinsights/issues/498 (@eudaimonious)
- [ ] Academic Assessments should be grouped by test/subject with ACCESS at the top. https://github.com/studentinsights/studentinsights/issues/499 (@eudaimonious)
- [ ] The DIBELS test should show scores not percentile rankings https://github.com/studentinsights/studentinsights/issues/500 (@iqnivek)
- [ ] Add ELL Status under Language https://github.com/studentinsights/studentinsights/issues/501
#### Add a button so folks can access it!
- [x] Add a button to the profile to access the PDF. Be very explicit about the beta nature of this feature. https://github.com/studentinsights/studentinsights/issues/502 (@eudaimonious)
##### Open questions

Do we need this additional information?  Is it new, or is it possibly available in district systems like Aspen already?
- Prior school enrollments?
- Retention?
- Grades?
- Report cards?
- In-school and out-of-school suspensions?
- Health, medical, nursing?
- Specific evidence of RTI?

After that, we can learn more about what else we need here, whether it involves adding a separate UI up front to ask for a few details, whether we want to record all packets that are generated for the case history, or whatever else comes up.  It'll be easier to do that working with school folks as referrals come up, so iterations here might require a little more communication.

kevinrobinson:
[...]

PR #137, opened 2016-03-02
Create sped referral form

eudaimonious:
Switched out the pdfkit gem for the wicked_pdf gem because wicked_pdf is  better documented and pdfkit was never fully implemented in this application. Created a route at students/:id/sped_referral.pdf and render a pdf with the student's name. One small step toward https://github.com/studentinsights/studentinsights/issues/8

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,7 +16,6 @@ gem 'jquery-rails'
 gem 'jquery-ui-rails', '~> 5.0.3'
 gem 'net-sftp'
 gem 'net-ssh'
-gem 'pdfkit'
 gem 'probability'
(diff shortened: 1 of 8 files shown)


kevinrobinson:
This is amazing, thank you @eudaimonious! :)

Confirmed this works as expected on OSX for local development too.  I'll check that the UX for the download is as expected on IE as well through browserstack after deploying to the demo site.

Here's what an awesome hello world PDF looks like in Chrome:

screen shot 2016-03-02 at 8 47 14 am

kevinrobinson:
Also, I sent an invite to the studentinsights team, which is the step I forgot last night that enables tagging you as working on an issue. :)

alexsoble:
Thanks for the pull request @eudaimonious! :bow:

PR #139, opened 2016-03-02
improved styling on behavior incidents

bw120:
screen shot 2016-03-01 at 9 46 21 pm


app/assets/javascripts/student_profile_v2/attendance_details.js
```diff
@@ -7,6 +7,31 @@
   var ProfileChart = window.shared.ProfileChart;
   var Scales = window.shared.Scales;
 
+  var styles = {
+    box: {
+      border: '1px solid #eee',
+      padding:'15px',
+      marginTop: '10px',
+      marginBottom: '10px',
```
kevinrobinson: In React styles, it will interpret numbers for properties like `margin`, etc. as pixels so you can use plain numbers here like `10`.

app/assets/javascripts/student_profile_v2/attendance_details.js
```diff
@@ -7,6 +7,31 @@
...
+      width: '50%'
+    },
+    item: {
+      width: '33%',
```
kevinrobinson: For layout styling like this, where we want three items to have equal width and take up the full width of the container, could we use [flexbox](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Using_CSS_flexible_boxes)?

app/assets/javascripts/student_profile_v2/attendance_details.js
```diff
@@ -7,6 +7,31 @@
...
+      display: 'inline-block',
+      paddingBottom: '2%'
+    },
+    itemHead: {
+      fontWeight: 'bold',
+    },
+    header: {
+      padding: '5%',
```
kevinrobinson: For padding and margins, let's use actual pixel numbers since these don't need to be fully responsive.  In my experience making small styling changes like look good in a fully responsive is a lot of work with not much benefit to the end user. :)

app/assets/javascripts/student_profile_v2/attendance_details.js
```diff
@@ -60,12 +85,14 @@
...
-          dom.div({}, 'Code:' + incident.incident_code),
-          dom.div({}, 'Location: ' + incident.incident_location),
-          dom.div({}, 'Description: ' + incident.incident_description)
-        );
+        return dom.div({ style: styles.box, key: incident.occurred_at },
+          dom.div({ sytle: styles.header },
+            dom.div({ style: styles.item }, dom.span({ style: styles.itemHead }, 'Date: '), dom.span({}, moment.utc(incident.occurred_at).format('MMM D, YYYY'))),
+            dom.div({ style: styles.item }, dom.span({ style: styles.itemHead }, 'Code: '), dom.span({}, incident.incident_code)),
+            dom.div({ style: styles.item }, dom.span({ style: styles.itemHead }, 'Location: '), dom.span({}, incident.incident_location))
+          ),
```
kevinrobinson: If it's helpful, you can also factor out the code for making an item into method, something like:

```
renderItem: function(label, content) {
  return dom.div({ style: styles.item }, dom.span({ style: styles.itemHead }, label), dom.span({}, content));
}
```

This isn't super necessary, just pointing out how you can refactor React UI code like plain JS code. :)

kevinrobinson:
@bw120 This looks awesome and is a great improvement, thanks! :)

I left a few small comments, let me know what you think.  I can also help finish this off and merge it if you're busy this week.  And if you have time or ideas for other improvements next week that's awesome too!  Whatever you prefer. :)

kevinrobinson:
@bw120 Actually sorry to be confusing. :)  I'm going to merge this to get it out since it's a great improvement already, and then we can make other improvements separately in a new pull request.  Thanks!

PR #164, opened 2016-03-03
Remove dead link to Dropbox design history

kevinrobinson:
We should figure this out, but it's broken at the moment.


kevinrobinson:
cc @AmirHadjihabib in case you have an updated link or can help fix!

PR #165, opened 2016-03-03
SPED referral - Adding minimal text example for iterating on design

kevinrobinson:
This is part of https://github.com/studentinsights/studentinsights/issues/8.

This adds a minimal text example of what a SPED referral packet might look like, to help facilitate design iterations.  This is to help iterate on the content and process for this packet's UX, more than the final layout or look.  It uses https://github.com/codeforamerica/somerville-teacher-tool/issues/100#issue-80810855 as the starting point for what should be in this packet.

We can merge this when we're confident enough that these are worth moving forward with building.  Alternately, we'll close this if this isn't even in the right direction and we need to approach this dirrectly.  There are other major open design questions for this feature, those should be tracked in https://github.com/studentinsights/studentinsights/issues/8 and not here.


kevinrobinson:
@alexsoble Here's what I was thinking on privacy and collaboration.

I'll reach out to our Somerville amigos for feedback on this over email and docs, so we're free to discuss existing packets as models without worrying about privacy on GitHub.  Then separately we can move that discussion here for more public contributions.  Let's discuss more over email and then update here when we've sorted that out, seem cool?

alexsoble:
@kevinrobinson That seems like a good flow!

eudaimonious:
Spoke with Jill about the mock-up today!

1. Need birth date, which appears on the mock-up but is not data available in the db.
2. Don't need attendance percentage, which appears on the mock-up.
3. Academic Assessments should be grouped by test/subject with ACCESS at the top.
4. The DIBELS test should show scores not percentile rankings.
5. Add a button to the profile to access the PDF. Be very explicit about the beta nature of this feature.
6. Future goal: Instead of a list, each test/subject should be presented as a line chart.
7. Future goal: Won't want to include all notes in the additional notes section, user should pick which. Preferred way to do this is probably exporting to a Microsoft Word document instead of a PDF. For now though it's easy enough to copy and paste from the PDF.

@kevinrobinson Any more notes on your end?

eudaimonious:
Forgot something:
A. We should add ELL Status under Language.

kevinrobinson:
@eudaimonious @jsgeiser these are amazing notes!  The only other items I had were:
 - loading speed on this page is not important, since it's used less frequently, and outside of meetings
 - while developing, it's better to throw errors loudly and explicitly if any data is missing, rather than failing gracefully and generating a packet that might be missing important data
 - the first few packets will need to be reviewed for accuracy by school amigos, especially at first

Issue #159, opened 2016-03-02
Fix bug with randomness in Student when seeding local database


kevinrobinson:
This happens intermittently.  This example is from https://github.com/studentinsights/studentinsights/commit/5cad8c783fab9dd4957fcb09fba48dc5a8730140.

Re-running the seed task leads to the same error in other places in the stack, or a successful run.  I suspect there's some randomness in generating the base `Student` model.

```
Creating demo schools, homerooms, interventions...
Seeding constants...
Creating demo educators...
Creating students for homeroom #1...
rake aborted!
ActiveRecord::RecordInvalid: Validation failed: Student can't be blank
/Users/krobinson/Documents/github/studentinsights/app/demo_data/fake_student.rb:185:in `block in add_deprecated_notes'
/Users/krobinson/Documents/github/studentinsights/app/demo_data/fake_student.rb:185:in `times'
/Users/krobinson/Documents/github/studentinsights/app/demo_data/fake_student.rb:185:in `add_deprecated_notes'
/Users/krobinson/Documents/github/studentinsights/app/demo_data/fake_student.rb:8:in `initialize'
/Users/krobinson/Documents/github/studentinsights/db/seeds/demo/demo.seeds.rb:42:in `new'
/Users/krobinson/Documents/github/studentinsights/db/seeds/demo/demo.seeds.rb:42:in `block (2 levels) in define_seed_task'
/Users/krobinson/Documents/github/studentinsights/db/seeds/demo/demo.seeds.rb:41:in `times'
/Users/krobinson/Documents/github/studentinsights/db/seeds/demo/demo.seeds.rb:41:in `block in define_seed_task'
Tasks: TOP => db:seed:demo => db:seed:demo:demo
```

kevinrobinson:
Think this is related, but didn't investigate: https://travis-ci.org/studentinsights/studentinsights/builds/113422806

eudaimonious:
Sorry for the commit spam. I was rebasing stuff around and didn't realize github would link based on issue number. I've got a PR for you here: https://github.com/studentinsights/studentinsights/pull/167

kevinrobinson:
[...]

PR #167, opened 2016-03-06
Seed data bug

eudaimonious:
The Student model has a uniqueness validator on the local_id attribute. That attribute was being generated with rand(1000) in the seed data file so from time to time you'd get a non-unique value. That's fixed here. https://github.com/studentinsights/studentinsights/issues/159

I also added pry because I needed to set a breakpoint to find the issue, and I think pry is quite popular so others might be wanting it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,6 +39,8 @@ group :development, :test do
   gem 'faker'
   gem 'launchy'
   gem 'phantomjs'
+  gem 'pry'                   # Set a breakpoint in your ruby code by adding `binding.pry`
+                              # See https://github.com/pry/pry
   gem 'rack-test'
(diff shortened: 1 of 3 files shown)


kevinrobinson:
@eudaimonious nice fix, thanks! :)
image

This looks good, and adding pry is great too.  The Travis build is failing on adding a `Service` model here, similar to the second comment in https://github.com/studentinsights/studentinsights/issues/159.  I must have mucked this up when updating the demo data generators.

Either way we should merge this, since it fixed the root problem in the `Student`, but I'll see if I can take a look at the `Service` bug tomorrow before Tuesday.

Thanks again, this is awesomely helpful! :)

PR #169, opened 2016-03-09
Added header for the incident history

bw120:
I added incident history header. Adjusted styling based on comments from the previous pull request. 

screen shot 2016-03-08 at 8 43 00 pm


kevinrobinson:
Awesome, thanks @bw120!

image

PR #170, opened 2016-03-09
Add note about rebuilding Docker container when Gemfile changes

kevinrobinson:
This came up with @erose tonight at Code for Boston.


erose:
Sounds good to me!

PR #172, opened 2016-03-09
Fix bug with seeding test db, factor out constant seeding into DatabaseConstants

kevinrobinson:
This fixes the flaky test failures we've been seeing like https://travis-ci.org/studentinsights/studentinsights/builds/113422806, and closes https://github.com/studentinsights/studentinsights/issues/159.

### DIFF
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -18,6 +18,7 @@ class Assessment < ActiveRecord::Base
   end
 
   def self.seed_somerville_assessments
+    Assessment.destroy_all
     Assessment.create!([
       { family: "MCAS", subject: "Mathematics" },
       { family: "MCAS", subject: "ELA" },
--- a/db/seeds/database_constants.rb
+++ b/db/seeds/database_constants.rb
@@ -1,0 +1,15 @@
+class DatabaseConstants
+  # Used for initializing the constants in a new database.  This
+  # is factored out of the main seed file so it can be used for
+  # RSpec tests as well.
+  #
+  # These are all specific to the initial Somerville deployment, but
+  # are used for local development as well and relied on in tests.
+  def seed!
+    Assessment.seed_somerville_assessments
+    InterventionType.seed_somerville_intervention_types
+    EventNoteType.seed_somerville_event_note_types
+    ServiceType.seed_somerville_service_types
+    nil
+  end
+end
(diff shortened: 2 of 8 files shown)


kevinrobinson:
Ah, this is more complicated and interleaves with the import step.  Looking more.

kevinrobinson:
@alexsoble This is what I mentioned in chat just now.

#### The problem with my attempted fix:
My fix here introduced a bug since it consolidated the seeding logic and moved creating `Assessment` types into the seeding process for both local development and test.  This led to [failures](https://travis-ci.org/studentinsights/studentinsights/builds/114765664) in the importer specs for that use `McasRow`.  That code reads in the data dumped from X2 about assessments, reads in rows one at a time and then creates new `StudentAssessment` records.  But in the process, it looks for `Assessment` records that define the particular type of assessment (eg., MCAS math).  If it doesn't find the assessment type it expects, `McasRow` creates a new `Assessment` record as a side effect in addition to adding the `StudentAssessment` record.  These specs were written with the assumption that the `Assessment` table would always be empty, and so my change broke that.

#### The immediate fix in this pull request:
So my additional commits here clarify the specs to be explicit that they expect no `Assessments`.  And since I moved the `Assessment` table to be created as part of seeding now, I added `before/after` blocks to these importer specs so that their behavior is still tested from a clean database.  This effectively updates the specs to test what they were testing before this PR, so we can merge it and make the RSpec tests reliable again, and also can move `Assessment` to be created in the seed process.  I also added one spec to test the case when there are `Assessment` records already in the database (since I think this is the regular production path).  So unless this seems crazy I'll merge this now to fix that and keep things moving.

#### Stepping back:
I think the common import flow is to import data into a database that already has `Assessment` records.  I added one spec to cover this path as well.  But I think it would simplify to move the creation of `Assessment` records fully into the seeding process, and remove it altogether from the import process, so that `McasRow` can always assume those `Assessment` records are present, and doesn't have to do that seeding itself.  I can do that separately if that seems reasonable, but wanted to get your input.

alexsoble:
@kevinrobinson Great summary of the issue. Essentially, this work is moving `Assessments` from the universe of things we learn about from X2 to the universe of things we pre-define. That makes sense because we already have pre-defined `Assessments` baked into the rest of the code base, including the front-end.

kevinrobinson:
@alexsoble Awesome, thanks for the help! :)

PR #168, opened 2016-03-09
added hover styling to table rows

lilybarrett:

PR #174, opened 2016-03-09
Factor out SlicePanel components and styles, revise hover styling

kevinrobinson:
This builds on the awesome work @lilybarrett did in https://github.com/studentinsights/studentinsights/pull/168.  It breaks up the pieces in the `SlicePanel` component into separate files and stylesheets to make them easier to discover in the future.

@lilybarrett I noticed this morning that the text on the left inside each row becomes a subtle dark blue on hover, so I fixed that up to stay the same.
screen shot 2016-03-09 at 8 21 06 am

I also realized in using this a bit this morning that the underline styling looks a lot like a link, which I didn't realize when we were talking about it last night.  This might be confusing to folks, since links usually take you away to a new page when you click on them, but these rows act more like buttons, where you click on them and the page changes in response.  So here's another iteration, trying to make them feel "clickable" and communicate _clicking here will change something on the page_ more than _this is a link and clicking will take you to another page_

screen shot 2016-03-09 at 8 24 58 am

@lilybarrett Let me know what you think, and if this seems cool I can merge it later today or tomorrow AM.  And thanks again for your help!

### DIFF
--- a/app/assets/stylesheets/components/fixed_table.scss
+++ b/app/assets/stylesheets/components/fixed_table.scss
@@ -1,0 +1,22 @@
+.FixedTable {
+  table {
+    table-layout: fixed;
+    border-collapse: collapse;
+
+    td {
+      padding: 0;
+    }
+
+    tr.clickable-row:hover {
+      outline: 1px solid lighten($gray, 40%);
+
+      a:hover {
+        color: $gray;
+      }
+    }
+  }
+
+  .bar {
+    background-color: rgba(49, 119, 201, 0.75);
+  }
+}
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/components/fixed_table.scss
```diff
@@ -0,0 +1,22 @@
+.FixedTable {
+  table {
+    table-layout: fixed;
+    border-collapse: collapse;
+
+    td {
+      padding: 0;
+    }
+
+    tr.clickable-row:hover {
```
kevinrobinson: This is the actual diff for the styling - it's my mistake mixing the refactoring work together with the style change, which makes it difficult to see what actually changed.

lilybarrett:
Makes a lot of sense! Thanks.

Issue #147, opened 2016-03-02
Fence in data import with integrity checks


alexsoble:
Via slack conversation with @kevinrobinson:

* Write queries to fence in what we expect the database to look like
* Run as a task after data import
* If it detects something is wrong in our data universe, it calls us

alexsoble:
Ran against production data.

```
Uh oh! Data check failed:
Validation failed: Student can't be blank
```

Will investigate! And keep running until all the data is valid. :blush:

alexsoble:
All data integrity checks now pass. Best of all, they helped identify and fix another data problem (#185).

Good suggestion @kevinrobinson!

PR #183, opened 2016-03-12
Data sanity check

alexsoble:
## Notes

+ Add more validation to `Educator`, `StudentAssessment`
+ Add `IntegrityCheck` model to check for valid data
+ #147


alexsoble:
Travis build isn't starting. Ruby specs pass locally:

ruby specs pass

PR #184, opened 2016-03-12
Stylish newsfeed

erose:
I added a title to the page:
screen shot 2016-03-12 at 5 36 58 pm

I made 'Absence' and 'Tardy' events smaller (because there are a lot of them):
screen shot 2016-03-12 at 5 34 04 pm

Outlined events and added badges (color-coded the same way as the columns) to every event:
screen shot 2016-03-12 at 5 34 50 pm

Let me know what you think!

This is a quick & dirty implementation that I would like to clean up once I know if this is a good direction for the page or not. A filthy nest of ternary operators. Also I'm wondering where one puts common styles for the page (I copy+pasted the "Notes" and "Services" title style) and what styles are in CSS vs. in code.


alexsoble:
@erose A very pretty first pass. Making absences / tardies smaller looks like a smart move! :blush:

app/assets/javascripts/student_profile_v2/profile_details.js
```diff
@@ -3,14 +3,46 @@
...
   var merge = window.shared.ReactHelpers.merge;
-
+  var FeedHelpers = window.shared.FeedHelpers;
   var QuadConverter = window.shared.QuadConverter;
 
+  var Colors = {
+    math: '#e8e9fc',
+    reading: '#ffe7d6',
+    procedural: '#e8fce8'
+  };
```
kevinrobinson: These would be awesome to pull out to [themes.js](https://github.com/studentinsights/studentinsights/blob/master/app/assets/javascripts/helpers/theme.js#L4) but that could be a separate piece of work after, whatever you prefer.

app/assets/javascripts/student_profile_v2/profile_details.js
```diff
@@ -3,14 +3,46 @@
...
+  var styles = {
+    feedCard: {
+      marginBottom: '10px',
```
kevinrobinson: In React styles, you can use a plain number and it will always interpret that as a pixel value.  And for font sizes, let's use pixels instead of points.  I'll add a doc about this, since it's come up a few times in the last few weeks.

app/assets/javascripts/student_profile_v2/profile_details.js
```diff
@@ -3,14 +3,46 @@
...
+      borderRadius: '10px',
+      fontSize: '10pt'
+    },
+    feedCardHeader: {
+      fontSize: '120%',
```
kevinrobinson: Can you try sticking with absolute pixel values for these sizes?  The rest of the app isn't responsive for font size (and doesn't need to be since we aren't optimizing this for mobile use), so expressing this as a percent doesn't really add more over pixels.

app/assets/javascripts/student_profile_v2/profile_details.js
```diff
@@ -22,55 +54,113 @@
...
+          type: 'STAR Math',
+          message: name + ' scored in the ' + score + 'th percentile on the Math section of STAR.',
+          date: QuadConverter.toDate(quad),
+          color: Colors.math
+        });
+      });
+      _.each(this.props.feed.deprecated.interventions, function(obj){
+        events.push({
+          type: 'Note',
+          message: _.template("<%=name%> (Goal: <%=goal%>)")(obj),
```
kevinrobinson: Let's stick with using React for rendering HTML rather than Underscore's templating.  This looks like just string concatenation, so you can do:

```
message: obj.name + '(Goal: ' + obj.goal + ')'
```

app/assets/javascripts/student_profile_v2/profile_details.js
```diff
@@ -22,55 +54,113 @@
...
-              moment(obj.date).format("MMMM Do, YYYY") + ": " + obj.message
-            );
-          })
+        dom.h4({style: styles.title}, 'Full Case History'),
+        this.renderCardList()
+      )
+    },
+
+    renderCardList: function(){
+      return dom.div({style: {marginTop: '10px'}},
```
kevinrobinson: `10` for pixels, you can drop the "px" in React styles.

app/assets/javascripts/student_profile_v2/profile_details.js
```diff
@@ -22,55 +54,113 @@
...
+        this.getEvents().map(this.renderCard)
+      );
+    },
+
+    renderCard: function(event){
```
kevinrobinson: For the branching here, it looks like the styling for most pieces change so it might be simpler to do two separate code paths that render the data differently.  But the ternaries seem fine to me too, up to you.

app/assets/javascripts/student_profile_v2/profile_details.js
```diff
@@ -22,55 +54,113 @@
...
+      var key = [event.date, event.message].join();
+      return dom.div(
+        {
+          key: key,
+          style: merge(styles.feedCard,
+            event.compressed ? {fontSize: '80%'} : {border: '1px solid #eee'}
+          )
+        },
+        dom.div(event.compressed ? {style: {paddingLeft: '10px'}} : {style: {padding: '10px'}},
```
kevinrobinson: You can remove the "px" here too.

kevinrobinson:
@erose This is a great improvement!  I left a few comments for you about mostly low-level code style things, looks great overall.  Let me know what you think and it'd be great to ship this and get some input to hear what folks think! :+1:

app/assets/javascripts/student_profile_v2/profile_details.js
erose: Cool, good to know where styles go. I'll do that as a separate pass, since I'm real tired.

app/assets/javascripts/student_profile_v2/profile_details.js
erose: Since absences/tardies have `fontSize: 80%` on the whole div, if I didn't use percentages here I would have to make separate styles for `feedCardHeaderCompressed`, `feedCardHeader`, `feedCardBodyCompressed`, etc.

app/assets/javascripts/student_profile_v2/profile_details.js
erose: Bodacious!

app/assets/javascripts/student_profile_v2/profile_details.js
erose: The ternaries were pretty bad IMO, I switched to branching.

app/assets/javascripts/student_profile_v2/profile_details.js
erose: Bodacious!

app/assets/javascripts/student_profile_v2/profile_details.js
erose: Sure!

app/assets/javascripts/student_profile_v2/profile_details.js
erose: Bodacious!

app/assets/javascripts/student_profile_v2/profile_details.js
kevinrobinson: Ahh I see, I missed the `80%` inline.  Yeah I think my general suggestion would be: if there are font size changes, let's see if we can express them in pixels (with percentages we lose isolation, in that one change can cascade and in general it'd be better to avoid that).  And if there are places where there are many font size changes, then it's probably worth looking at simplifying to a fewer number of font sizes.

So here, would removing the sizing on the container simplify?  Then each element is sized individually and with pixels, and they're each isolated and not interdependent?  Let me know if I'm being confusing here. :)

app/assets/javascripts/student_profile_v2/profile_details.js
kevinrobinson: :+1:

app/assets/javascripts/student_profile_v2/profile_details.js
erose: I removed the percentage -- should be ready to merge now!

erose:
Last revisions complete.

kevinrobinson:
Awesome improvement! :)

I captured two other thoughts after trying this out in https://github.com/studentinsights/studentinsights/issues/217 and https://github.com/studentinsights/studentinsights/issues/218, and would love to hear your thoughts on those too, especially if you have other ideas for how to improve!

PR #203, opened 2016-03-18
Cut over to v2 profile page, fix minor authorization bug that was not exploited

kevinrobinson:
This adds in a redirect so that `students#show` redirects to `students#profile`.  It moves the previous show action the `deprecated_v1_profile` action, so that the old page is still accessible, just not navigable from the product.  Changing from `show` to `deprecated_v1_profile` involved touching a few places like the JS code that runs for that page, and various specs.

Relatedly, the `before_filter` call in `students_controller` needed updating, which revealed what looks like an authorization bug.  I quickly tried testing whether there was a security bug here, and confirmed nothing exposes any student data publicly.  I then worked on fixing forward and didn't dig further into it yet, but my understanding is that there were two minor authorization bugs that:

  1. exposed the full list of enrolled students (along with their grade and classroom, but nothing else) to other Healey teachers in the search bar autocomplete (or by curling that route)
  2. exposed the v2 profile page, and the v2 write path for notes and services to teachers at the Healey who may not have authorization to those students.

I verified in server logs and Mixpanel that no authorized data was actually exposed, and testing this fully in production is difficult since it involves interaction with the production LDAP service to confirm the bug for various levels of access.  We should investigate this further (and how we can test this more easily) as part of expanding to other schools  cc @alexsoble.

I'm reasonably confident that the RSpec changes here address these bugs, so am going to ship this improvement now, but @alexsoble if you can give this a post-review when you're able that would be super helpful.


app/controllers/students_controller.rb
```diff
@@ -110,7 +114,10 @@ def service
 
   def names
     q = params[:q]
-    sorted_students = search_and_score(q, Student.with_school)
+    authorized_students = Student.with_school.select do |student|
```
kevinrobinson: This runs each student through the authorization check.

app/csv_exporters/student_profile_csv_exporter.rb
```diff
@@ -32,7 +32,7 @@ def mcas_mathematics_results
   end
 
   def mcas_ela_results
-    student.mcas_ela_results.map do |results|
+    student.mcas_ela_results.map do |result|
```
kevinrobinson: This came up in fixing the export button - these were latent syntax bugs that just hadn't been exercised in the RSpecs.

spec/controllers/students_controller_spec.rb
```diff
@@ -1,16 +1,30 @@
 require 'rails_helper'
 
 describe StudentsController, :type => :controller do
-  let(:educator) { FactoryGirl.create(:educator_with_homeroom) }
```
kevinrobinson: Removing this level of nesting to express this closer to the actual specs.

spec/features/export_profile_feature_spec.rb
```diff
@@ -1,5 +1,6 @@
 require 'rails_helper'
 require 'capybara/rspec'
+require 'csv'
```
kevinrobinson: This came up in testing this file in isolation, must have been `require` pollution where it was pulled in elsewhere.

kevinrobinson:
Confirmed manually in the demo site and in production that this looks good!  I also tested the authorization changes didn't regress any functionality on the production site, but as mentioned above, we don't have a good way to thoroughly test for all roles in production at the moment.  I opened https://github.com/studentinsights/studentinsights/issues/205 as a first step.

PR #206, opened 2016-03-18
Fix bug with new event_note_types in UI not being consistent with server, add tests

kevinrobinson:
There was older UI code that had `event_note_type_id` values that didn't match the server models.  This led to two problems:

  1. on the read path for existing notes, the "badge" showing which type of note was null
  2. on the write path, the write would succeed but we'd be writing invalid `event_note_type_id` keys

UI bug on the read path:
screen shot 2016-03-17 at 8 41 14 pm

The first bug makes sense, but I'm confused by the second one, since there's a [`validates_presence_of` validation](https://github.com/studentinsights/studentinsights/blob/master/app/models/event_note.rb#L6) on the `EventNote` model.  I would have thought this would lead to the controller throwing a server error when manually testing adding event notes, but it does not and the POST completes so the feature works as expected (except for the read-side badge bug after).  Here's the local Rails log showing that it works locally:

```
  Parameters: {"event_note"=>{"event_note_type_id"=>1, "text"=>"wat", "student_id"=>6}, "id"=>"6", "student"=>{}}
  Educator Load (0.4ms)  SELECT  "educators".* FROM "educators"  WHERE "educators"."id" = 1  ORDER BY "educators"."id" ASC LIMIT 1
  Student Load (0.2ms)  SELECT  "students".* FROM "students"  WHERE "students"."id" = $1 LIMIT 1  [["id", 6]]
   (0.2ms)  BEGIN
  SQL (0.3ms)  INSERT INTO "event_notes" ("created_at", "educator_id", "event_note_type_id", "recorded_at", "student_id", "text", "updated_at") VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING "id"  [["created_at", "2016-03-18 02:32:43.530066"], ["educator_id", 1], ["event_note_type_id", 1], ["recorded_at", "2016-03-18 02:32:43.529329"], ["student_id", 6], ["text", "wat"], ["updated_at", "2016-03-18 02:32:43.530066"]]
   (1.6ms)  COMMIT
```

And there's an RSpec spec that [tests this explicitly](https://github.com/studentinsights/studentinsights/blob/master/spec/controllers/students_controller_spec.rb#L263).

So I'm a little confused.  Going to ship this fix and then investigate further separately as to why this wasn't caught in the spec or manual testing.


kevinrobinson:
I also manually ran a migration on the console to convert `EventNotes` that had `event_note_type_id == 1` to be `301`, for SST Meeting.

kevinrobinson:
Ahh yeah, `validates_presence_of` just validates the presence of a value, not that the value is a valid foreign key and the foreign record exists.  https://github.com/studentinsights/studentinsights/issues/209 could help.

PR #220, opened 2016-03-21
Limit Rollbar error reporting to production only

kevinrobinson:
Right now the development environment is set to use the `production` env and production API key, so local development errors alert as if they were production errors.  We could set this up to work for multiple environments, but for now this just aborts loading Rollbar unless we're in the production environment.

cc @alexsoble probably want to turn this off for staging too, or look at supporting multiple environments in Rollbar.


kevinrobinson:
I also resolved all those Rollbar issues.

PR #222, opened 2016-03-22
Update config for different deployments to declare SHOULD_REPORT_ANALYTICS

kevinrobinson:
MixPanel data includes the staging site, when it should only include the production deployment.  There's no way to distinguish right now, so I updated the Heroku config for each deployment, and then updated Rollbar and Mixpanel reporting to use this instead of the environment check.

screen shot 2016-03-22 at 7 24 47 am

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ So far, Student Insights can import CSV and JSON and can fetch data from AWS and
 
 ### LDAP
 
-The project is configured to use LDAP as its authentication strategy in production. To use database authentication (in a production demo site, for example) set a `DEMO_SITE` environment variable. Authentication strategies are defined in `educator.rb`.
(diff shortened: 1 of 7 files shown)


kevinrobinson:
Added two other changes here.  Using `ENV['DEMO_SITE']` checks for any value in the environment variable, so `DEMO_SITE=false` will enable the flag.  This adds an `EnvironmentVariable.is_true` method to check the actual string value and strictly require it to be `true` or `TRUE`.

I also swapped the authorization branching to use `SHOULD_USE_LDAP` instead of the environment itself, and added `DEPLOYMENT_KEY` for logging purposes, which isn't used for any functional branching.

screen shot 2016-03-22 at 8 15 56 am

kevinrobinson:
Verified this manually on demo, staging and production deploys.

Issue #133, opened 2016-02-29
Improve color mapping for lines on charts


kevinrobinson:
Across sparklines and details charts, particularly on the profile v2 page and the STAR charts, it'd be simpler for users if the colors for line charts had a consistent meaning.  The STAR charts encode the change across the visible range with color (eg., red is regression, blue is growth), so this might be a good starting point for all charts.

kevinrobinson:
@sentientmachine started taking a look at this last night!

sentientmachine:
Plan: Of the charts that show in the v2 student page, the color of the charts should be a smooth gradient between red and blue to indicate an overall assessment of how good this rise is relative to the rise of the other students.  That way if you see an increasing diagonal line that's green, you can know that this is the best of the group, if a decreasing diagonal line is green then that decline is the "smallest loser", otherwise if red it's the "worst of all users, who are also declining".

sentientmachine:
The rest of the plan is simple, fly to the moon, shrink the moon, grab the moon....
![img_20160308_211318](https://cloud.githubusercontent.com/assets/330247/13623292/7decafdc-e573-11e5-9fbd-db36f9412c41.jpg)

kevinrobinson:
:moon: :full_moon_with_face: :smile:

sentientmachine:
I've got the date of the leftmost position.  I was able to get the indexes of the points to its left and right, and I was able to get the values of the points left and right.  Next is to get the percentage the date is between the two dates, and then apply that same ratio to the y values to get the final leftmost y value.  Then compare that to the rightmost value and apply a gradient between green and red for how much it goes up.

kevinrobinson:
@sentientmachine nice!  :+1:

sentientmachine:
I'm done with all the changes, I'm trying to do a git push to push my branch, but I get an authentication error:
fatal: unable to access 'https://github.com/studentinsights/studentinsights.git/
We'll get it sorted out tuesday.

kevinrobinson:
@sentientmachine nice!  Yeah I think this is from trying to push to the main Student Insights repo.

So I think what you'll need to do is:
 1. Fork this repo, using the button in the upper right of the GitHub UI.
 2. Add your fork (eg., https://github.com/sentientmachine/studentinsights) as a new remote repository to  your local git repository, using the command line git commands on your machine.
 3. Push the work on the branch in your local repository to your GitHub fork.
[...]

PR #230, opened 2016-03-22
Color the chart lines green or red based on the visible line.

sentientmachine:
Before all the charts had grey lines like this:
student_insight_colors_original

So I changed it so that when the viewable window has lines that go up, it is colored blue:
student_insight_colors_updated2

And like this:

student_insight_colors_updated3

What I do is get the date of the this.props leftmost position on the viewable window on the SVG. I was able to get the indexes of the points to its left and right, and I was able to get the y values at those points. I got the percentage the date is between the two dates, and then apply that same ratio to the y values to get the final leftmost y value. Then I compare that to the rightmost value and apply a color between green and red for how much it went up or down.

### DIFF
--- a/app/assets/javascripts/student_profile_v2/sparkline.js
+++ b/app/assets/javascripts/student_profile_v2/sparkline.js
@@ -59,24 +59,24 @@
       }
 
       var color = function(val) {
-          //val represents a number of how much the line on the chart increased
-          //or decreased.
+          //val represents a number of how much the line on the 
+          //chart increased or decreased.
 
-          //If the line in the viewable area of the chart increases, color it green
-          //otherwise color it red.
+          //If the line in the viewable area of the chart 
+          //increases, color it green, otherwise color it red.
 
-          //if (d >= 0){ return 'green'; }
-          //else       { return 'red'; }
-          val = val * 2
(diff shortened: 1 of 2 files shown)


sentientmachine:
This addresses: https://github.com/studentinsights/studentinsights/issues/133

alexsoble:
@sentientmachine Thanks for the PR Eric!

When I pulled down this branch to my local machine, I noticed that some of the sparklines got colors but others didn't, even when there seemed to be a trend:

olaf

rapunzel

Could you take a look to see if you can reproduce this for the seed demo on your own machine? Thanks!

sentientmachine:
I was able to reproduce your observation.  To address your concern I doubled the sensitivity of the colorization of the line, that way small moves will show as partially red, and large moves will be fully red.

kevinrobinson:
From talking with @sentientmachine @alexsoble @erose in person at Code for Boston, some other feedback that's come up from user testing this week impacts this work and has become higher priority.  So we decided to leave this be for now, and revisit it after those other design improvements land.

@sentientmachine thanks for all your help with this, and for embracing the beautiful and messy process that is new product development! 😄

PR #231, opened 2016-03-23
Risk Bubble Styling

BWieber:
Hey Kevin, 

Let's call this step COMPLETE! 

Here's the image: 

riskbox final

I found out that I my code in development didn't make it into the request. 

The code in question:

```
    render: function() {
      return dom.span({},
        dom.span({
          style: merge(styles.riskBubble, { backgroundColor: this.bubbleColor() })
        }, this.props.riskLevel),
        dom.span({
          style: styles.riskItem
        }, "Risk Level")
      );

    },
```

Where you see `return dom.span` was `return dom.div` in the last PR, causing the element separate itself onto another line. It looks like because the other PR was reverted, the `formatGradeNumber` function is also being added here, to ensure that the proper wording is associated with the grade levels. 

Hope it looks good! 

_Brandon_

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -50,6 +50,7 @@
 
 // student profile page v2:
 // pure ui components:
+  //= require ./student_profile_v2/risk_bubble
   //= require ./student_profile_v2/service_color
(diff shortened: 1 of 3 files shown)


kevinrobinson:
:ship:

Issue #209, opened 2016-03-18
Investigate adding foreign key constraints or validations


kevinrobinson:
There was a minor database-level data integrity bug I introduced in https://github.com/studentinsights/studentinsights/pull/206, and is now fixed.  I was mixed up about `validates_presence_of`, and it'd be great to tighten these all to foreign key constraints (or the Rails validation equivalent).

eudaimonious:
working on this now

erose:
Hey @eudaimonious, if you're still working on this, can you assign yourself? (If not, no pressure!) Helps people navigate the issues list to see what they can work on.

eudaimonious:
Thanks Eli! Looks like I have access to do that sort of stuff with gh issues now. Yay! Well, I merged the change but I know that caused some failures due to production data. I'm closing this but @kevinrobinson might want to re-open or start a new issue if this did indeed get rolled back.

erose:
@eudaimonious: Yup, I added you as a "collaborator" on Github (not sure if you get notified about that), so you now have full write access here.

Issue #211, opened 2016-03-18
Remove v1 profile page and related code


kevinrobinson:
This is part of the cleanup for https://github.com/studentinsights/studentinsights/issues/5.

We should wait a week or two until after v2 is fully rolled out and stable.  Especially with adding new bits to the v2 profile page like DIBELS and ACCESS, it's useful to have the v1 page still around for quick validation.

There's also some CSV export logic and really good authorization tests that are written for the v1 profile page, so should move those over to the new profile page too.

Ultimately, this means we remove the redirect that's there now, and the `profile` action becomes the `show` action.

alexsoble:
@kevinrobinson PR #281 covers the controller and view part of this issue.

It looks like there's also a data model / data migration aspect here too.

The `StudentNote` and `ProgressNote` models are used in V1 but not V2. We should delete those models (and their corresponding specs, helpers, generators, etc.) as part of this work.

There are 0 StudentNotes and 49 ProgressNotes in the production database. These were added by Jill at MTSS / SST meetings. In order to not lose this data, we'll need a migration that turns `ProgressNote` data into `EventNote` data.

kevinrobinson:
Awesome!

Yeah and to clarify I think those models are used in v2, read and then
rendered as 'older note' or something like (although maybe not).

I had been thinking we'd just keep the old models around but I think doing
a full migration sounds like a better way to simplify. :thumbsup:

alexsoble:
Oh yeah, you are correct! I see the "Old progress note" / "Old intervention" bit now, didn't notice before. 

In that case I think a full migration is a "nice to do" refactor but not critical. The `StudentNote` and `ProgressNote` models are still being exercised by Profile V2 so they are not just cruft, unlike the view / controller code.

alexsoble:
Almost done here. The only code left in `app/assets/javascripts/student` is the `profile_charts` directory. Some of the V1 profile chart code is being used by V2, some is not.

So remaining steps are:

1. [ ] Figure out which `profile_charts` code is unused, delete that.
2. [ ] Move `profile_charts` code in use over to the V2 folder.
3. [ ] Get rid of the "V2" part of files, components, and folders that have it in their name. Would be a little confusing for a new contributor to see V2 everywhere when there's no more V1.

Issue #9, opened 2016-02-08
Handle students who become inactive


alexsoble:
From https://github.com/codeforamerica/somerville-teacher-tool/issues/447.

## Notes

+ Only active students come through the X2 export
+ When they become inactive, we want to destroy their records or give them an inactive status
+ One way to do this would be to detect which students are no longer in the export file
+ But better would be to get the actual Active/Inactive data from X2

alexsoble:
As a first step, dove a bit into the contents of the `STD_ENROLLMENT_STATUS` column. 

Here are the different column values present in X2:

### Enrollment statuses

"Active"
"Graduate"
"Withdrawal-24"
"SummerWithdrawal-20"
"Withdrawal-20"
"Withdrawal-23"
"Withdrawal"
"Not Attending"
"Withdrawal-22"
"DropOut"
"DropOut-36"
"DropOut-35"
"SummerDropOut-36"
"SummerWithdrawal-22"
"Withdrawal-21"
"SummerWithdrawal-21"
"SummerGraduate"
"SummerWithdrawal"
"SummerDropOut-34"
"Excluded"
"SummerWithdrawal-24"
"DropOut-34"
"Attainment"
"DropOut-30"
"SummerDropOut-35"
"PreRegNew"
"DropOut-31"
"SummerDropOut-30"
"SummerAchievement"
"Achievement"
"SummerWithdrawal-23"
"Services-41"
"SummerMaxAge"
"MaxAge"
"DropOut-32"
""
"DropOut-33"
"SummerAttainment"
"SummerDeceased"
"Deceased"
"SummerExcluded"
"SummerDropOut-32"
"Certificate"
"SummerDropOut-33"
"SummerDropOut"

alexsoble:
Based on the categories above, looks like we want to:

+ keep and update the record in Student Insights if student is "Active" 
+ destroy the record in Student Insights if student belongs to one of the other categories

alexsoble:
On second thought, that may not be the behavior we want...

If a student graduates or transfers, for example, we may want to keep the records of services and notes that have been entered into the system at MTSS or SST meetings. 

Best thing to do is consult with Uri and Jill at Somerville Public Schools about their preferences for how to handle this.

kevinrobinson:
Good point!

Here's a thought from the train: What if the import always imported all
students, but took the 'active' bit too? Then we could show only active
students in the Student Insights product. That moves towards having less
logic during import. A cost is that it may increase query times if there
[...]

PR #241, opened 2016-03-30
Foreign key constraints

eudaimonious:
We didn't have any foreign key constraints in our database. I read through schema.rb to uncover columns that were foreign keys. I used the add_foreign_key method to add them to postgres. This method isn't available until Rails 4.2 so I added the Foreigner gem to support it.

https://github.com/studentinsights/studentinsights/issues/209


eudaimonious:
closing and re-opening to re-trigger travis

alexsoble:
Thanks for this @eudaimonious!

Helped me learn something new, your PR made me go and read up on foreign keys: https://robots.thoughtbot.com/referential-integrity-with-foreign-keys

My only suggestion here would be, why don't we bump up the Rails version? We're very close to 4.2, and that way we have one less dependency, which is nice. 

Anyway, that's all I've got. Excellent work!

kevinrobinson:
@eudaimonious This is awesomely helpful, thank you! :+1:

db/migrate/20160330000130_add_foreign_key_constraints.rb
```diff
@@ -0,0 +1,33 @@
+class AddForeignKeyConstraints < ActiveRecord::Migration
+  def change
+    add_foreign_key :absences, :student_school_years
```
kevinrobinson: From looking at https://robots.thoughtbot.com/referential-integrity-with-foreign-keys, it seems we can also move `dependent: destroy` expressions here too.  I don't think we're super rigorous about that at the moment, but would be something good to try.

Also, there's a recommendation about verifying first that the current state of the production DB can satisfy these contraints. :)  We can try that out on the staging site, which should be a quick way to validate before deploying this for real.  Added https://github.com/studentinsights/studentinsights/issues/246 for future us.

kevinrobinson:
I tried deploying this to staging and then running the migration.  The migration was successful, but then I ran into errors that I think are completely unrelated but didn't investigate further yet.

```
2016-03-31T02:57:19.110486+00:00 app[web.1]: Started GET "/homerooms/7" for 98.216.106.247 at 2016-03-31 02:57:19 +0000
2016-03-31T02:57:19.112216+00:00 app[web.1]: Processing by HomeroomsController#show as HTML
2016-03-31T02:57:19.112234+00:00 app[web.1]:   Parameters: {"id"=>"7"}
2016-03-31T02:57:19.137306+00:00 app[web.1]: Completed 500 Internal Server Error in 25ms (ActiveRecord: 10.5ms)
2016-03-31T02:57:19.138299+00:00 app[web.1]:
2016-03-31T02:57:19.138307+00:00 app[web.1]: NoMethodError (undefined method `id' for nil:NilClass):
2016-03-31T02:57:19.138308+00:00 app[web.1]:   app/models/intervention.rb:53:in `most_recent_atp'
2016-03-31T02:57:19.138309+00:00 app[web.1]:   app/models/student.rb:204:in `most_recent_atp_hours'
2016-03-31T02:57:19.138310+00:00 app[web.1]:   app/controllers/homerooms_controller.rb:49:in `fat_student_hash'
2016-03-31T02:57:19.138310+00:00 app[web.1]:   app/controllers/homerooms_controller.rb:10:in `block in show'
2016-03-31T02:57:19.138311+00:00 app[web.1]:   app/controllers/homerooms_controller.rb:10:in `show'
```

I suspect this is from the staging site not being in a great state somehow, and not actually related to this migration, but since it takes a few hours to rebuild the staging site I can't try now.  I think the better path is to just ship this and allocate time after to validate and fix forward if anything does come up.

db/migrate/20160330000130_add_foreign_key_constraints.rb
eudaimonious: Agreed. This wouldn't be an across the board change you'd want to make for all instances of a foreign key so as someone who doesn't know all of the business logic yet I guess I'd just base the addition on whether I see a belongs_to relationship or a presence validation.....

eudaimonious:
@alexsoble Upgrading to 4.2 and removing the gem could be a good code quality next step. Or hey, maybe we want to go somewhere past 4.2.....

@kevinrobinson I'm not sure if it's related (if it is we should fix before merging) but the InterventionType class that the failing method refers to seems ultra suspect. The first line in the file is `#Deprecated`

eudaimonious:
Thank you for such a thoughtful review!

db/migrate/20160330000130_add_foreign_key_constraints.rb
kevinrobinson: :+1: yeah I think it has to be a manual walk through the semantics of each relationship, I don't think anyone knows all the relationships off the top of their head (@alexsoble might still!).  Not sure I have any better recommendation but am happy to be another set of eyes, so if that's helpful ever feel free to ping me!

kevinrobinson:
@eudaimonious Yep, that sounds like a great plan!

FWIW, `InterventionType` is only deprecated in the last month (in that the UI for the write path has been removed).  It's still exercised on some older code on the read path (that can be removed in another week or two for https://github.com/studentinsights/studentinsights/issues/211), so I don't think there's any other reason to suspect that it's a code bug there causing issues.  There is also known drift at the moment between staging and production datasets (https://github.com/studentinsights/studentinsights/issues/9), so I think the thing to do is rebuild the staging site and then see if we can still reproduce.  I'm guessing things will work smoothly there, or else digging into it will reveal some other drift, but only one way to find out. :)

Sorry to hold this up, but since it requires validating production data, I'll try to kick it off tonight and then see where we're at tomorrow or in the next few days.

kevinrobinson:
I rebuilt the staging site, deployed this and ran the migration, and things look good! :+1: :ship:

kevinrobinson:
This failed on production on `StudentSchoolYear` pointing to a Student record that didn't exist.  Going to revert this, https://github.com/studentinsights/studentinsights/issues/264 has the next steps for investigation.

Issue #208, opened 2016-03-18
3/22 Code for Boston agenda


kevinrobinson:
Last week: https://github.com/studentinsights/studentinsights/issues/175#issuecomment-198176809.

#### Review priorities (umbrella issues):
- current: [finish Student profile improvements](https://github.com/studentinsights/studentinsights/issues/5)
    - [Show DIBELS data](https://github.com/studentinsights/studentinsights/issues/23)
    - [Show grade-level equivalence for STAR](https://github.com/studentinsights/studentinsights/issues/229)
    - Add ACCESS data
    - Speed up loading of the overview page
    - Finish final bits in https://github.com/studentinsights/studentinsights/issues/5
- next: [Productionize visualizations of STAR academic progress](https://github.com/studentinsights/studentinsights/issues/2)
- after: [SPED referral form](https://github.com/studentinsights/studentinsights/issues/8)

#### Help finish what's in progress:
- Finish any work-in-progress you own
- Help anyone else with work-in-progress if you can (keeping in mind that sometimes piling on more people doesn't help)
- Check out work that's left over from last week.
- Check open pull requests
- Umbrella issues - see if there's a clear piece you can get done on one of the open issues

#### Start some new work:
- Fix an [Open bug](https://github.com/studentinsights/studentinsights/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22+label%3Abug)
- Start on a [help wanted issue](https://github.com/studentinsights/studentinsights/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22)
- Start a [brand new feature](https://github.com/studentinsights/studentinsights/labels/new%20feature)
- Talk it out with Kevin if you have questions or if you're stuck

#### Help new folks pitch in:
- Coordinating tonight: @erose
- Talk about any questions from the project overview
- Help them set up development environment
- Show them the agenda issue and our workflow

PR #248, opened 2016-03-31
Update CONTRIBUTING to link to agenda issues, and emphasize shipping …

kevinrobinson:
…more than pizza

This adds two small improvements on top of @dazzaji's awesome work.  One is that it links to the agenda issues we've been using for each Code for Boston meeting.  The other is that it tweaks the copy about the Tuesday Code for Boston sessions to clarify that while there may be free pizza and drinks at the event, we're primarily there to work together to ship improvements to the product. :)

Usually I'm cool with just merging, but since this is about working at Code for Boston, I'll leave this open for a week or so for any other interested folks to chime in.  cc @alexsoble


kevinrobinson:
Also if folks have input on iterations to try on the agenda issue format (eg., https://github.com/studentinsights/studentinsights/issues/208), this seems like a good place to chime in.

alexsoble:
:+1:

PR #256, opened 2016-04-02
Simplify routes.rb and remove unused routes

kevinrobinson:
This came up in adding a new route for discontinuing a service.  The `resources` function was generating many routes that we don't actually have controller actions for.

Before:
```
$ bundle exec rake routes | wc -l
      82
```

After:
```
$ bundle exec rake routes | wc -l
      35
```


alexsoble:
@kevinrobinson Nice improvement!

PR #257, opened 2016-04-03
Allow discontinuing services

kevinrobinson:
This allows users to discontinue services.  After discontinuing the service, it's still shown in the student profile UI (but faded out), and isn't shown in other views like the overview or STAR pages; those only consider services a student is actively receiving.

#### Server-side changes:
 1. Migration for `DiscontinueService` model
 2. `DELETE /services/:id` endpoint
 3. Added scopes for `Service` to filter by active/discontinued, and updated controllers and specs to provide both types in the student feed and only active services when slicing students by various dimensions.
 4. Add serialization methods for `services` and `event_notes`, merging the `discontinued_service` fields in.

The main idea is to capture the user actions, and full change history.  Since the number of mutations will be relatively low, aggregating over a handful of records on the read path isn't a performance concern.  On the overview pages this could be more of a problem, but the next piece of work is to precompute that. :)

```
irb(main):006:0> Student.find(3).services.size
=> 4
irb(main):007:0> Student.find(3).services.active.size
=> 0
````

#### UI changes:
This adds a confirmation step, which is a new UX idiom.

Discontinue button:
screen shot 2016-04-02 at 9 02 25 pm

On hover:
screen shot 2016-04-02 at 9 02 30 pm

Confirmation:
screen shot 2016-04-02 at 9 35 04 pm

Canceling:
screen shot 2016-04-02 at 9 35 07 pm

After confirming:
screen shot 2016-04-02 at 9 38 25 pm

After deleting:
screen shot 2016-04-03 at 11 06 23 am


kevinrobinson:
Deployed to staging and manually tested the migration and end-to-end functionality on all pages.

kevinrobinson:
This is a way larger PR that it should have been, I didn't do a great job of breaking this into pieces and so there are a lot of changes here. :\

Issue #3, opened 2016-02-08
Precompute data for school overview page


kevinrobinson:
From https://github.com/codeforamerica/somerville-teacher-tool/issues/412

Speed things up!
Precompute assessment results, attendance, discipline incidents — anything that is updated in the database daily or less frequently, and merge in interventions data.

kevinrobinson:
Starting point: https://github.com/studentinsights/studentinsights/compare/feature/more-students-and-homerooms?expand=1

alexsoble:
This is happening now! It speeds things up tremendously, i.e. around 20x in the West Somerville case:

https://github.com/studentinsights/studentinsights/issues/332#issuecomment-209670375

Awesome work @kevinrobinson, thanks for building this and making the page way faster!

PR #270, opened 2016-04-03
Add job for precomputing the part of the school_overview query that is imported

kevinrobinson:
This is to address https://github.com/studentinsights/studentinsights/issues/3, and is part of #5.

This adds a job to precompute the `student_hashes` part of the `schools#show` endpoint, in an effort to speed that up, since that data only updates each day.  This doesn't yet schedule the job to run in Heroku.

#### locally
This is the effect on load times locally, with 466 students in the DB:
Before:
screen shot 2016-04-03 at 2 35 51 pm

After:
screen shot 2016-04-03 at 2 35 23 pm

#### staging a/b
Based on my low budget manual testing, there's a ton of variance in response times here, both in staging and in production.  All my tests were for `ldap_user`, and all trying to hit as warm a path as possible.

My initial tests in staging showed this change (with some responses beforehand up to 2x slower):
Before:
screen shot 2016-04-03 at 2 53 55 pm

After:
screen shot 2016-04-03 at 3 10 52 pm

#### staging and production
But testing a while later led to lower response times overall on both domain.  This is from a few manual tests comparing production and staging (which of course have different profiles, so this is just rough).
Production:
screen shot 2016-04-03 at 3 23 05 pm
screen shot 2016-04-03 at 3 23 30 pm

Staging:
screen shot 2016-04-03 at 3 10 52 pm
screen shot 2016-04-03 at 3 25 42 pm

Overall, these all are in the right direction like we'd expect though.

### DIFF
--- a/app/models/educator.rb
+++ b/app/models/educator.rb
@@ -50,7 +50,7 @@ class Educator < ActiveRecord::Base
             .where(grade: grade_level_access)
             .includes(eager_loads)
     else
-      logger.warn("fell through #students_for_school_overview for educator_id: #{self.id}")
+      logger.warn("Fell through to empty array in #students_for_school_overview for educator_id: #{self.id}")
       []
     end
   end
--- a/app/models/educator.rb
+++ b/app/models/educator.rb
@@ -49,6 +49,9 @@ class Educator < ActiveRecord::Base
             .active
             .where(grade: grade_level_access)
             .includes(eager_loads)
+    else
+      logger.warn("fell through #students_for_school_overview for educator_id: #{self.id}")
+      []
     end
   end
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ branches:
 #   - export PATH=$PATH:$HOME/.local/bin
 
 before_install:
-  - psql --version
+  - echo psql --version
   - gem install -v 1.10.6 bundler --no-rdoc --no-ri
  # workarounds for phantomjs 1, see: https://github.com/travis-ci/travis-ci/issues/3225#issuecomment-177592725
   - mkdir travis-phantomjs
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ branches:
 #   - export PATH=$PATH:$HOME/.local/bin
 
 before_install:
+  - psql --version
   - gem install -v 1.10.6 bundler --no-rdoc --no-ri
  # workarounds for phantomjs 1, see: https://github.com/travis-ci/travis-ci/issues/3225#issuecomment-177592725
   - mkdir travis-phantomjs
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -53,12 +53,6 @@ class SchoolsController < ApplicationController
     end
   end
 
-  def precomputed_student_hashes_key(time_now, authorized_student_ids)
-    timestamp = Time.now.beginning_of_day.to_i
-    authorized_students_key = authorized_student_ids.sort.join(',')
-    [timestamp, authorized_students_key].join('_')
-  end
-
   def serialized_data_for_star
     authorized_students = current_educator.students_for_school_overview(:student_assessments)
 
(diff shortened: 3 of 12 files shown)


kevinrobinson:
There are also some significant differences in the educators and authorizations between staging and production, which I think is from the workarounds that @alexsoble has awesomely figured out.

Concretely, there are 3 educators in staging (1 of which is the `ldap_user` I manually added) that have authorization to see any student data.  In production there are 26 educators.  It might be a good idea to investigate that difference, and figure out if there's another way to externalize that (maybe as a seed or import task).

This is how I ran the task in staging:
`heroku run:detached bundle exec rake precompute:school_overview --app student-insights-staging`

Also, I tried reading from a flat file locally (versus the JSON column in the Postgres record), which was faster and is how the `Rails.cache` would work out of the box, but from poking around online that doesn't fit too well with Heroku's deployment model.  And I think we want the "precompute" semantics more than the "read-through cache that expires each day" semantics.

app/models/educator.rb
```diff
@@ -47,6 +47,9 @@ def students_for_school_overview(*additional_includes)
       school.students
             .where(grade: grade_level_access)
             .includes(eager_loads)
+    else
+      logger.warn("fell through #students_for_school_overview for educator_id: #{self.id}")
```
kevinrobinson: @alexsoble This looks like a bug to me.  This would fall through to a `nil` before, which came up in some of my manual testing.  I was curious how prevalent this was, and found some differences here between staging and production, so this could use some more eyes, maybe we can chat more Tuesday.

kevinrobinson:
I'm not understanding the build failure yet.  Looks like Postgres JSON support isn't installed (which is fine, we could change this to `bytea`).  But this should be included out of the box in Postgres 9.4.1, and I can verify this works fine locally on OSX, and the Travis docs say that's the same version running by default, so I'm not yet sure what the problem is here.

alexsoble:
@kevinrobinson Thoughts on precomputing all of the data including interventions, and then re-computing the value in a background job whenever an educator adds/deletes/edits an intervention?

app/models/educator.rb
alexsoble: @kevinrobinson Yeah, let's talk more at Code for Boston or on Tuesday morning!

alexsoble:
> There are also some significant differences in the educators and authorizations between staging and production, which I think is from the workarounds that @alexsoble has awesomely figured out.

> Concretely, there are 3 educators in staging (1 of which is the ldap_user I manually added) that have authorization to see any student data. In production there are 26 educators. It might be a good idea to investigate that difference, and figure out if there's another way to externalize that (maybe as a seed or import task).

@kevinrobinson I can explain the difference!  :0)

All educator authorization information came from Jill via email. I entered it directly as a big console script into the production environment, didn't save it as a migration or a rake task or anywhere else in the codebase. Would be good for me to do that so that we can keep parity between production / staging. 

Would also be useful because it would help externalize Jill's and other principals' intuitions about who should have access to what student information.

kevinrobinson:
@alexsoble Awesome, that's what I figured but good to verify. :)  And externalizing to a task or code sounds good.

Related to fully recomputing those on edits, we could do that but I don't think we have infrastructure at the moment to trigger a background computation that ultimately leads to a write in the database.  Doing it synchronously in the controller action after flushing a response to the client seems more complex.  Is your thought mostly that this would be simpler, or a way to improve page load times on the overview page?

FWIW, my own thinking has been "every night new data comes in from external systems, and as part of that we index that data for fast serving to users whenever they come."

kevinrobinson:
Got this working!  the problem was the migration code that used the `json` column type failed in Travis (still not sure why it didn't work), so adding a new migration on top to change the column type didn't work, since the earlier migration would still fail. :panda_face:

kevinrobinson:
Verifying this manually in production now, see if you can spot the before and after change. :)

screen shot 2016-04-07 at 1 27 05 am

kevinrobinson:
@jochakovsky FYI this was the issue I mentioned, and other solutions to address would be great too!

Issue #264, opened 2016-04-03
Investigate student_school_years records referring to nonexisting student_ids


kevinrobinson:
This came up trying to run the migration to add foreign key constraints in https://github.com/studentinsights/studentinsights/pull/241.

```
PG::ForeignKeyViolation: ERROR:  insert or update on table "student_school_years" violates foreign key constraint "student_school_years_student_id_fk"
DETAIL:  Key (student_id)=(512) is not present in table "students".
```

In production, there is a gap in student primary keys (from 512-610):
```
irb(main):008:0> Student.all.map(&:id).sort.size
=> 518
irb(main):009:0> Student.all.map(&:id).sort.last
=> 617
```

The next check I ran was:
```
StudentSchoolYear.where(:student_id => (512..617)).size
=> 111
```

I had been hoping this was just a stray record or two and I could remove it manually, but then I stopped since this clearly needs a closer look.  I'm not sure there are actual any data quality issues, but we'll need to figure out what's up here and fixup in order to unblock https://github.com/studentinsights/studentinsights/pull/241.

cc @alexsoble

kevinrobinson:
Just ran into https://github.com/studentinsights/studentinsights/issues/185, possibly related.

alexsoble:
Before running migration to get rid of junk student school years:

`StudentSchoolYear.count # => 2824`

alexsoble:
After running migration: 

`StudentSchoolYear.count # => 2725`

alexsoble:
Data `IntegrityCheck` now runs without throwing any errors.    

👍

kevinrobinson:
:tada: :smile:

PR #273, opened 2016-04-04
Enforce valid associations for StudentSchoolYear

alexsoble:
## Notes

+  #264


kevinrobinson:
This might be a nice time to add in the gem in https://github.com/studentinsights/studentinsights/pull/241, and then the constraints just for these columns.  That way we can start ticking off the remaining constraints rather than trying to do them all in one batch.

PR #274, opened 2016-04-04
Add student school years to data quality check

alexsoble:
## Notes

+ Also make `IntegrityCheck` throw real errors when it hits a problem instead of swallowing them


alexsoble:
Should help prevent future issues like #274.

kevinrobinson:
awesome!

PR #279, opened 2016-04-04
First pass at an onboarding checklist for new schools

alexsoble:
__Work in progress pull request!__

Feel free to add comments, suggestions, ideas.  

🌈  

#278

### DIFF
--- a/docs/somerville_school_onboarding.md
+++ b/docs/somerville_school_onboarding.md
@@ -5,12 +5,17 @@
 _These are based on the [user personas](https://github.com/studentinsights/studentinsights#user-personas-who-were-serving) we are serving. The primary focus at this stage is Principals._
 
 + [ ] Meet with the principal:
-  + [ ] Ask about the school's strengths
-  + [ ] Ask about what kind of data queries s/he asks for or produces most often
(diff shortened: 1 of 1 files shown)


kevinrobinson:
Love it!  :+1: 

I think it's super important we do the full discovery process and learn about the problem context of new schools, but I think we should also start with a description of what the Healey has done already to set the scope of the conversation.

For example if we want to "onboard" them to the Healey way of doing things that means focusing on training and adoption for:

- principals or interventionists want to identify students who are at-risk in particular areas
- principals, interventionists, or teachers want a summary view of the student's profile across academics and behavior
- principals and interventionists want to capture discussions, actions and specific services being provided to students as part of an existing structured intervention process (eg., SST and MTSS meetings)

This isn't meant to close out other conversations - there's probably a whole class of user experiences that would be really impactful to build!  But we'll also want to set a realistic expectation for scoping new feature development, which will change a bit as different schools have slightly different contexts and needs.

alexsoble:
Merging this in now, we can always revise later since it's a documentation PR.

kevinrobinson:
:+1: this is awesome!

PR #281, opened 2016-04-05
Remove Profile V1 code

alexsoble:
## Notes

+ Remove controller and view code for V1
+ Remove specs that apply to V1 and not V2
+ Modify controller authorization specs for V1 to cover V2
+ #211


alexsoble:
Travis not running for some reason.

PR #287, opened 2016-04-06
Implement custom Tablesort function for "disability" column

### LINKED ISSUE #155: Fixing column sorting on overview table and roster table
Sorting by columns that have a logical order doesn't work right.

For example, "Disability" sorts alphabetically, leading to sort order like this:
screen shot 2016-03-02 at 1 21 13 pm

This probably affects columns like FLEP as well.

dsbader:
*Add Tablesort extension function "disability" which sorts in order ('Low < 2','Moderate','High')
*Add disability data sort method as option to students_table
*Allow renderHeader to accept options other than className (e.g., data-sort-method)

Resolves #155


alexsoble:
This looks awesome! 

Good catch. 

🌈

Issue #290, opened 2016-04-06
Make Absences / Tardies / Discipline graphs accurate for multiple years


alexsoble:
Right now they're only pulling in events from the most recent school year, they should show events from all school years.

kevinrobinson:
Resolved in https://github.com/studentinsights/studentinsights/pull/324.

PR #324, opened 2016-04-10
Include attendance data for all previous years in profile page, flex scales for outliers

kevinrobinson:
Addresses https://github.com/studentinsights/studentinsights/issues/290.

This updates the student profile page to include attendance events from previous years in:
 1. the summary section up top
 2. the attendance, tardy and absence detail charts
 3. in the full case history details

It also:
  1. updates the scale functions for projecting values onto the y-axis, so that the full range of data is always shown and the y-scale is compressed if necessary
  2. splits the absences and tardy charts on the details page so they can use independent y-scales

No major changes in query performance, response is still in the 100s of ms:
screen shot 2016-04-10 at 3 31 20 pm

#### Showing more data from previous years
Before, with no data for previous years:
screen shot 2016-04-10 at 3 38 27 pm
screen shot 2016-04-10 at 3 38 52 pm

After, with context about previous years and with absences and tardies split:
screen shot 2016-04-10 at 3 38 45 pm
screen shot 2016-04-10 at 3 38 40 pm
(same result on full case history, not shown here for simplicity since it contains student data that needs to be scrubbed)

#### Showing effects of flexible y-axis scales:
Before, the count of absences overflowed the y scale:
screen shot 2016-04-10 at 3 33 15 pm

After, the scale flexes and pushes the threshold down to show the full range:
screen shot 2016-04-10 at 3 33 09 pm

#### More about y-scale functions
For students that are outliers and have more events than a typical student, the full range of values will be shown and the y-scale will be compressed.  In practice, this means any student who takes up the full visual range of the chart should stick out to users, although students who are extreme outliers will appear visually similar to students who have large values and are at risk.  Concretely, the sparkline for "discipline incidents" will look similar visually for a student with 10 incidents and a student with 50 incidents.  The dotted line for the warning threshold, and the actual number still cue the users as to the actual magnitude.

### DIFF
--- a/app/assets/javascripts/student_profile_v2/attendance_details.js
+++ b/app/assets/javascripts/student_profile_v2/attendance_details.js
@@ -58,7 +58,8 @@
     render: function() {
       return dom.div({ className: 'AttendanceDetails' },
         this.renderDisciplineIncidents(),
-        this.renderAbsencesAndTardies(),
+        this.renderAbsences(),
+        this.renderTardies(),
         this.renderIncidentHistory()
       );
     },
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/student_profile_v2/attendance_details.js
```diff
@@ -78,30 +80,46 @@
...
+    },
+
     renderIncidentHistory: function() {
       return dom.div({ style: styles.container },
         dom.h4({ style: styles.title }, 'Incident History'),
         (this.props.disciplineIncidents.length === 0) ? dom.div({}, 'None') : this.props.disciplineIncidents.map(function(incident) {
-        return dom.div({ style: styles.box, key: incident.occurred_at },
+        return dom.div({
+          style: styles.box,
+          key: [incident.occurred_at, incident.incident_description].join()
```
kevinrobinson: Resolving duplicate key warning that came up in manual testing.

alexsoble:
@kevinrobinson Fantastic! 🌴  🏇  🌴

Issue #291, opened 2016-04-06
Add DIBELS to ELA summary for K-3


alexsoble:
Students in grades K-3 don't have MCAS ELA growth but do have DIBELS scores. In the summary boxes at the top of Profile V2 let's make that substitution for students grades K-3.

kevinrobinson:
And also show the graphs for both below regardless (for students that are say in 4th grade now but still have DIBELS data from last year).  @alexsoble I'm figuring the charts below are straightforward if we have the yearly data required for the summary section, but we should split that bit out if it's more involved, definitely lower priority like you suggested the other day.

PR #327, opened 2016-04-10
Don't send nil DIBELS assessments to UI, fix validation on StudentAssessment to guard, update specs

kevinrobinson:
This starts updating the controller and UI code to filter out DIBELS assessments that have `nil` performance_level values, and updates the demo data generator to not create records like this.

There is a validation on the `StudentAssessment` model that looks like it's trying to enforce this, but a bug prevents this from actually being enforced.  This PR fixes the bug, but I got stuck on updating the associated RSpecs and need to take a break.  This just pushes the current progress.

Related to https://github.com/studentinsights/studentinsights/issues/291.

### DIFF
--- a/app/controllers/students_controller.rb
+++ b/app/controllers/students_controller.rb
@@ -20,7 +20,7 @@ class StudentsController < ApplicationController
       student: serialize_student_for_profile(student),
       feed: student_feed(student),
       chart_data: chart_data,
-      dibels: student.student_assessments.by_family('DIBELS'),
+      dibels: student.student_assessments.by_family('DIBELS').select {|dibels| !dibels.performance_level.nil? },
       intervention_types_index: intervention_types_index,
       service_types_index: service_types_index,
       event_note_types_index: event_note_types_index,
--- a/app/importers/file_importers/star_math_importer.rb
+++ b/app/importers/file_importers/star_math_importer.rb
@@ -22,13 +22,24 @@ class StarMathImporter < Struct.new :school_scope, :client
 
     return if student.nil?
 
-    star_assessment = StudentAssessment.where({
+    existing_star_assessment = StudentAssessment.where({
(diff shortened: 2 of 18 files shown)


kevinrobinson:
Related to https://github.com/studentinsights/studentinsights/pull/303.

alexsoble:
My mistake here @kevinrobinson — I noticed this validation issue last week, but failed to document it in a timely way or push my fix-it-up branch to GitHub. 

😕 

Just pushed my own WIP branch up here, here's the commit that fixes the validation issue: b81b524ee3766a3687757d6b6f018d6918a53013.

It breaks a bunch of specs because of the assessment factories don't have the correct attributes, as I'm sure you noticed!

kevinrobinson:
@alexsoble This has a bunch of commits on top updating specs and related code.  It could use some additional eyes, and almost certainly we'll need to investigate a little bit further to understand the state of the production DB in order to actually ship this as it is.

Main points:
  - The importer code right now would raise on trying to import rows without all valid data, now that the validations are being enforced.  I updated the importer code to catch this for each row, log it and carry on.
  - I'm not sure if we can expect all MCAS scores to have growth percentiles (maybe only recent years have it?), so we might need to relax that and a quick check of the DB should help us learn more.
  - There's some simplification we could do to the factory classes, but not important to do now.

If this seems good to you, next steps would be to poke around the production DB and see how many records would not pass this new validation, or re-run the import task in staging to see what happens.

kevinrobinson:
Going to comment out actual validation for now, so we can get this code into production safely, and investigate how many validations errors there would be and fix them up before flipping them on.  Thanks for the helpful rubber-ducking @eudaimonious!

kevinrobinson:
Dropping this since I let it get stale and this isn't something I'm going to do in the short term.

Issue #145, opened 2016-03-02
Improve generator for behavior incidents in local development


kevinrobinson:
These aren't super realistic, which makes it tough to test product improvements like https://github.com/studentinsights/studentinsights/pull/139/files locally.

#### 1. Generate more realistic event distributions ([code](https://github.com/studentinsights/studentinsights/blob/master/app/demo_data/fake_student.rb#L140))
Current example:
<img width="174" alt="screen shot 2016-03-02 at 9 58 26 am" src="https://cloud.githubusercontent.com/assets/1056957/13464076/566d4e64-e05d-11e5-840b-ccb60d3f3e8a.png">

This is the real distribution on 3/2/16:
<img width="174" alt="screen shot 2016-03-02 at 9 53 05 am" src="https://cloud.githubusercontent.com/assets/1056957/13463935/989e00a4-e05c-11e5-9fe6-d08755057199.png">

We should update to match more closely.  There are also often an unrealistically high number of incidents for a few students (eg., in the 50s), which affects UX work on local development and the demo site.

#### 2. Generate realistic text instead of the placeholders there now ([code](https://github.com/studentinsights/studentinsights/blob/master/app/demo_data/fake_discipline_incident.rb#L3))
<img width="626" alt="screen shot 2016-03-02 at 9 59 49 am" src="https://cloud.githubusercontent.com/assets/1056957/13464107/8522ca0e-e05d-11e5-93d8-534dfe6efb0d.png">

These should be full sentences or paragraphs.  Looking at real production data, these are typically 2 - 4 sentences long.

kevinrobinson:
This came up from @bw120's awesome work on https://github.com/studentinsights/studentinsights/pull/139 at Code for Boston last night.

erose:
I'm interested in working on this. I've noticed some other data issues, let me know if they actually are issues:

 * Right now, absent/tardy events are all within the last six months. But other data extends back to 2010.

erose:
What I've found out so far regarding discipline event distributions: the current thing we do is based on https://github.com/codeforamerica/somerville-teacher-tool/issues/94.

In that issue, the mean is given as `5.2` and the standard deviation as `8.3`. In the code, we use those numbers, but we also assume a normal distribution. That would mean that, of students with discipline events, the majority are clustered around `5.2`.

[...]

Issue #227, opened 2016-03-22
Distribute demo data attendance events evenly across school years


erose:
Right now absences & tardies only occur with in the last 100 days:

    attendance_event.occurred_at = Time.new - (rand * 100).to_i.days

This makes stuff a little difficult to test in the UI. We should evenly distribute them over all school years.

erose:
@kevinrobinson @alexsoble Is it easy to get the real distribution on absences/tardies? Is the current:

![screen shot 2016-04-12 at 4 43 32 pm](https://cloud.githubusercontent.com/assets/1514487/14474994/be715be6-00cd-11e6-80ae-2c14e72384ce.png)

realistic? (Thought I might fix that if I'm fixing attendance events anyway).

kevinrobinson:
@erose Here are the real distributions as of today.  Keep in mind this is quite dependent on the time of the year since it's cumulative from the beginning of the year:

<img width="202" alt="screen shot 2016-05-01 at 10 10 51 pm" src="https://cloud.githubusercontent.com/assets/1056957/14946001/9698735e-0fe9-11e6-855c-99281ac4111e.png">

Also, I'm all for improving this but fully modeling this can be quite complex, so my suggestion would be to focus these improvements on places where it significantly interferes.

erose:
Ah, I didn't realize these were cumulative from the beginning of the year. Okay, that does seem complicated to model. New PR generates them both according to the Absences distribution you gave but makes no attempt to model the change-over-the-school-year.

jemagee:
Assuming that each school year has an ID can’t you just create a belongs to / has many between the school year and the absence?

kevinrobinson:
@erose 👍

kevinrobinson:
@jemagee hi!  I'm not sure I follow what you mean, but the question here is more about what the distributions should be than the relational modeling.

PR #341, opened 2016-04-12
Make demo data more realistic

erose:
Made incident description text much longer and made the distribution closer to what it is in reality, addressing #145.

The `sample_from_distribution` function is a general method to do this, if we ever want to improve generation for other stats. It lets you specify a discrete distribution in (hopefully) a friendlier way that RubyStats.


erose:
Pictures!

Longer incident sentences. (This represents the output of one random seed generation. (Perhaps should be longer?):

screen shot 2016-04-12 at 4 50 13 pm

More realistic distribution of discipline incidents:
screen shot 2016-04-12 at 4 52 54 pm

erose:
Added another commit distributing attendance events evenly across all school years. This addresses #227.

PR #199, opened 2016-03-16
UI changes to Profile V2 page / behavior incident page

bw120:
I added tabs to each column to help the user know what is in each section.
screen shot 2016-03-15 at 9 21 36 pm

I added links to jump down to each section and a link to get back to top of page
screen shot 2016-03-15 at 9 21 29 pm

Adjusted styling so each section looks more visually like a separate section.
screen shot 2016-03-15 at 9 21 21 pm

PR #347, opened 2016-04-13
Profile page: remove time subtext from services summary

kevinrobinson:
This simplifies the services summary panel in the profile page, which also helps it stay clear when resized to smaller screen widths:

Before:
screen shot 2016-04-12 at 8 05 46 pm

After:
screen shot 2016-04-12 at 8 08 36 pm


kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/199 cc @bw120

Issue #228, opened 2016-03-22
Profile page - Add ability to edit and delete notes


kevinrobinson:
Split out as an enhancement to https://github.com/studentinsights/studentinsights/issues/5.

Design should reflect that these are not the primary user interactions, and should require some kind of user confirmation to delete anything.  On the server-side, we could use an append-only log of delete and update actions (in contrast to updating the note records in-place), and merge them at read time since the volume of these changes is expected to be so low.  This is probably most beneficial for deleting, since it keep the audit trail.  As a general user experience and data provenance goal, using an audit trail with recovery can minimize the "confirmation checks" we need to require of users, since users can recover easily and quickly.

For now, we can restrict delete and edit capabilities to admin users, but as part of productionizing this we should test that with school partners.

kevinrobinson:
@kbweaver Thanks for chatting this out in person, and for the great questions!

To capture what we talked about:
- The main scenario for editing is that a user records a note in an SST meeting, and then notices a typo mistake and wants to fix it.  Or they are reviewing a student profile and then notice a typo someone else made and they want to fix it.
- Users in schools (principals in particular) care about maintaining the integrity of the data, and guarding against any mistakes that someone might make that could lead to losing data.  One approach to handling this is to create authorization rules, that restrict which users are able to edit and delete data.  Another approach is to record more history about each change, so we (or end users) can recover if something gets deleted that shouldn't be deleted.  I think for this project we should default towards recording everything and keeping access rules open for everyone in the school.
- For approaches to recording the full history of changes, I'd suggest bias towards a simpler solution over one that's more efficient or generalizable to other cases.  The specifics of what this looks like in the database are a little different for the edit and delete cases, and treating them separately is cool.

To tackle this, you could start with what the database tables and columns would look like to support this, or what the user experience would look like, whatever you prefer.  If it's helpful to write up quick bullet points of what your plan is, I'm happy to review that too.

Thanks for your help! :smile:

kbweaver:
Thanks @kevinrobinson !

We considered a few options for tracking edits.
- Bring in a third-party library like [Paper Trail](https://github.com/airblade/paper_trail) or [Audited](https://github.com/collectiveidea/audited). This would be very generalized and powerful at the cost of maintaining another dependency. @kevinrobinson and @alexsoble didn't anticipate a need for this type of auditing beyond event notes and services, so the business case doesn't seem compelling enough to justify the cost.
[...]

PR #349, opened 2016-04-13
Back-end support for editing event notes

kbweaver:
Part of #228, this change adds tested back-end support for editing event notes. On edit, the previous version is stashed in event_note_revisions. There are no UI changes in this pull request.

To test this change, run `rspec` and test the negative case in the UI by navigating to a student profile and adding an event note.


alexsoble:
Hi @kbweaver! This looks like a strong data model and I love the merge-early-and-often philosophy. `StudentsController` is getting long (> 200 lines), what do you think about breaking out the `create` and `update` actions for `EventNote` into `EventNotesController`?

kbweaver:
Great call on the refactor, @alexsoble . Not only does this improve code hygiene by reducing module size and function complexity, it more clearly and safely maintains backwards compatibility with the existing `/students/:id/event_note` route. I marked that route as deprecated, and we can remove it once we deploy the UI changes and are confident that all clients have the new static assets.

kbweaver:
While testing the UI changes (WIP [here](https://github.com/kbweaver/studentinsights/compare/feature/event-notes-edit...kbweaver:feature/event-notes-edit-ui) if you're interested), I noticed that I'm updating the event note's `recorded_at` attribute to `Time.now` each time you edit a note. I forget whether we discussed this explicitly or not, but it seems unexpected to me. What would you expect? Happy to add another commit to this PR to change this behavior and make `recorded_at` immutable.

alexsoble:
Sorry for the slow response here @kbweaver!

Both `EventNote` and `EventNoteRevision` have a `recorded_at` attribute. I would agree with you that `recorded_at` should be immutable. When an `EventNote` gets revised its `recorded_at` stays the same, and the revision gets its own, newer `recorded at`.

app/controllers/event_notes_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+  before_action :authorize!, except: [:names]
+
+  def authorize!
+    student = Student.find(params[:student_id])
+    educator = current_educator
+    raise Exceptions::EducatorNotAuthorized unless educator.is_authorized_for_student(student)
+  end
+
+  def create
+    clean_params = params.require(:event_note).permit(*[
```
eudaimonious: This works and is easy to read but it would be a bit more idiomatic rails to enapsulate your permissible parameters in a private method. http://edgeguides.rubyonrails.org/action_controller_overview.html#strong-parameters

app/controllers/event_notes_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+      :student_id,
+      :event_note_type_id,
+      :text
+    ])
+
+    event_note = EventNote.new(clean_params.merge({
+      educator_id: current_educator.id,
+      recorded_at: Time.now
```
eudaimonious: It looks like on first creation, the even't note's `recorded_at` will be the same as the `created_at` and `updated_at` values. If an event note revision is created, you set a `recorded_at value` there to match event note's value and then updated the event_note's value for `recorded_at`. I think that this behavior could be simplified by removing the `recorded_at` columns and referring to the models' `updated_at` columns wherever you need that information. You can replicate the behavior of updating the event note's timestamp when an event note revision is created by calling `event_note.touch` http://apidock.com/rails/ActiveRecord/Timestamp/touch

eudaimonious: Ah sorry, didn't notice @alexsoble's earlier comments.

app/controllers/event_notes_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    event_note_id = clean_params[:id]
+    event_note = EventNote.find(event_note_id)
+
+    # Save the state of the existing event note.
+    previous_event_note_revision = EventNoteRevision.where(
+      event_note_id: event_note_id
+    ).order(
+      :version
+    ).take
```
eudaimonious: I don't think `take` will respect your ordering and you'll end up with something slightly random depending on postgres (but most likely the smallest version number rather than the highest). Instead, I think the intention is to call `last`.

app/controllers/event_notes_controller.rb
kbweaver: Factored out the params so I could share them between create and update. Thanks for the tip!

app/controllers/event_notes_controller.rb
kbweaver: Changed the `take` to `last`. Great catch!

erose:
Kevin, you did so much work on this. Let's try to merge this tomorrow!

Issue #359, opened 2016-04-13
Investigate and resolve error with precomputing student hashes job


alexsoble:
Via  https://github.com/studentinsights/studentinsights/issues/332#issuecomment-209645791:

```
ActiveRecord::RecordNotUnique: PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_precomputed_query_docs_on_key"
DETAIL:  Key (key)=(precomputed_student_hashes_1460505600_) already exists.
```

alexsoble:
Syntax issue:

```
irb(main):009:0> PrecomputedQueryDoc.find(this_key)

ActiveRecord::RecordNotFound: Couldn't find PrecomputedQueryDoc with 'id'=precomputed_student_hashes_1460246400_

irb(main):010:0> PrecomputedQueryDoc.find_by_key(this_key)
=> #<PrecomputedQueryDoc id: 4, key: "precomputed_student_hashes_1460246400_", ...>

irb(main):011:0> PrecomputedQueryDoc.exists?(this_key)
=> false

irb(main):012:0> PrecomputedQueryDoc.find_by_key(this_key)
=> #<PrecomputedQueryDoc id: 4, key: "precomputed_student_hashes_1460246400_", ...>

```

alexsoble:
Precompute job runs error-free now.

PR #360, opened 2016-04-13
Fix bug with precomputed doc upsert

alexsoble:
## Notes

+ Finder method wasn't looking up docs correctly because it was searching for a primary key
+ The column named "key" on PrecomputedQueryDoc is not the primary key according to Rails
+ https://github.com/studentinsights/studentinsights/issues/359#issuecomment-209662975

### DIFF
--- a/app/jobs/precompute_student_hashes_job.rb
+++ b/app/jobs/precompute_student_hashes_job.rb
@@ -34,7 +34,8 @@ class PrecomputeStudentHashesJob
     key = precomputed_student_hashes_key(precomputed_time, authorized_student_ids)
 
     # This is a non-atomic upsert
-    PrecomputedQueryDoc.find(key).destroy! if PrecomputedQueryDoc.exists?(key)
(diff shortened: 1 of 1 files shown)


app/jobs/precompute_student_hashes_job.rb
```diff
@@ -34,7 +34,8 @@ def precompute_and_write_student_hashes!(date_timestamp, authorized_student_ids)
     key = precomputed_student_hashes_key(precomputed_time, authorized_student_ids)
 
     # This is a non-atomic upsert
-    PrecomputedQueryDoc.find(key).destroy! if PrecomputedQueryDoc.exists?(key)
+    pre_existing_doc = PrecomputedQueryDoc.find_by_key(key)
```
kevinrobinson: 👍  thanks for fixing up my mistake here!

alexsoble: @kevinrobinson Of course! Thank you for writing all the precompute logic, it is awesome and speeds up the overview soooo much as I shall document shortly! 

😊

Issue #368, opened 2016-04-15
Make experimental status of STAR Reading / Math pages explicit


alexsoble:

kevinrobinson:
Closed by https://github.com/studentinsights/studentinsights/pull/369

alexsoble:
@kevinrobinson You did that absurdly quickly.

kevinrobinson:
just typing, it's the only thing I can really help with at this time of night :)

PR #369, opened 2016-04-15
Add prototype message on STAR pages

kevinrobinson:
We've talked with school partners about this, that this is a prototype that early adopters can experiment with and share feedback.  As we expand to more users, this puts the message front and center so that additional users can see that without in-person conversations.

screen shot 2016-04-14 at 9 30 35 pm


kevinrobinson:
Closes https://github.com/studentinsights/studentinsights/issues/368

PR #391, opened 2016-04-17
WIP – Update services dialog to allow freetext names

### LINKED ISSUE #307: Update services dialog to allow free-text names for service providers
This supports a common SST meeting use case of specific staff that work closely with the school and students but are not district employees (eg., outside counseling providers), or for cases of tutoring programs where the individual providing services is different from session to session.

This path seemed like a good balance between keeping the UI simple and not getting into the challenges of customizing for every different use case, while still allowing some ability to capture provider information for these important edge cases.  We also trade off the ability to perform more precise queries for which staff are providing services, but this capability isn't being used now and isn't an immediate priority.

Separately, but relatedly we also briefly considered updating this UI to support capturing finer-grained interventions like the difference between 1:1 Wilson, regular small group reading intervention, and occasional after school groups.  We didn't move forward with this since it increases challenges with data consistency and training across users and we wanted to focus on successful adoption at the current level of granularity first.

alexsoble:
Fix #307.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -60,6 +60,7 @@
 // pure ui components:
   //= require ./student_profile/risk_bubble
   //= require ./student_profile/service_color
+  //= require ./student_profile/provided_by_educator_dropdown
   //= require ./student_profile/profile_chart_settings
   //= require ./student_profile/educator
   //= require ./student_profile/datepicker
--- a/app/assets/javascripts/student_profile/api.js
+++ b/app/assets/javascripts/student_profile/api.js
@@ -21,7 +21,7 @@
         service: {
           service_type_id: serviceParams.serviceTypeId,
           date_started: serviceParams.dateStartedText,
-          provided_by_educator_id: serviceParams.providedByEducatorId,
+          provided_by_educator_name: serviceParams.providedByEducatorName,
           student_id: studentId
         }
       };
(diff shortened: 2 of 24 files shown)


alexsoble:
TODO: Fixup references to `provided_by_educator_id` in JS fixtures / spec.

alexsoble:
A couple of TODOs left on the UI end...

+ [ ] Add a little button / caret that opens the autocomplete menu for a combobox feel
+ [ ] Ask users to enter in new staff members "LastName, FirstName" — otherwise the ordering on the autocomplete will get messed up and confusing for users

alexsoble:
Cool, now for a little `onProvidedByEducatorChanged` debugging and a JS spec-writing...

alexsoble:
Debugging complete, thanks @bw120!

Classes to write tests for: 

+ [ ] `ProvidedByEducatorDropdown.js`
+ [ ] `StudentsController#get_educator_names_for_dropdown`

alexsoble:
QA steps:

+ [ ] Do a little manual QA in staging environment

+ [ ] In production console: `Service.all.map { |s| s.provided_by_educator.full_name }`

+ [ ] Merge, push to master, migrate

+ [ ] In production console: `Service.all.map { |s| s.provided_by_educator_name }`, should be same as above

alexsoble:
Checks out. 👍

Issue #374, opened 2016-04-15
Simplify or cut "Group quartiles each period" graph


alexsoble:
Feedback from Kathleen and Uri was that this graph is difficult to understand — not clear how to read it and get information out of it.

kevinrobinson:
Addressed in https://github.com/studentinsights/studentinsights/pull/395

Issue #373, opened 2016-04-15
Make "Group progress, since last assessment" graph more prominent


alexsoble:
This is the graph that everyone looks to and loves. Uri and Jill wanted to know about Healey 3rd grade reading progress — they found it here. Kathleen wanted to know about West Somerville 6th grade progress. Everyone loves this graph because it gives them a clear, powerful summary number about progress that they can't get their hands on quickly and easily right now.

kevinrobinson:
Addressed in https://github.com/studentinsights/studentinsights/pull/395

Issue #2, opened 2016-02-08
Umbrella - Productionize STAR change-over-time graphs


kevinrobinson:
From https://github.com/codeforamerica/somerville-teacher-tool/issues/413
- [x] Add STAR math
- [x] Make it navigable
- [ ] Some test coverage
- [ ] Improve STAR reading demo data so we can test locally
- [ ] Add option to clear filters
- [ ] Fix line interpolation issues
- [ ] Investigate reading data points that are close in time, but very different scores (eg., https://somerville.studentinsights.org/students/197?column=profile)
- [ ] update charts to show homeroom names
- [ ] precompute as much data as possible, to speed up server-side load times
- [ ] add axis labels, possibly legends (see https://github.com/studentinsights/studentinsights/issues/79#issuecomment-184983093)

kevinrobinson:
There's some great ideas from a feedback session with Mohan from at Code for Boston here: https://github.com/studentinsights/studentinsights/issues/79#issuecomment-184983093

kevinrobinson:
Capturing some conversations with @jgeiser this morning:

Jill's been gathering and entering a bunch of STAR data manually into a spreadsheet, to see how students have progressed over the last few assessments.  She's looking at every student, from 2nd-8th grade and records their grade-level equivalent.  She did this last fall, and is working on the winter scores now.  She wants to look at this by classroom.

She's looking for where there's a decline, whether it's in one class, whether a particular group of students went down, so she can talk to the teacher and learn more.  Sometimes there are issues with administering the test.  She wants to do this for both reading and math, but starts with reading since it is administered first.

We talked about the prototype charts that use percentiles - Jill said she always want to look at this as grade-level equivalent but that's just her and other principals might not.  She does something similar with DIBELs, starting at performance and then maybe digging deeper.

alexsoble:
Don't think we're actively pushing forward on this @kevinrobinson, mind if I close it?

kevinrobinson:
@alexsoble closing!

PR #395, opened 2016-04-18
Simplify STAR pages, to emphasize recent change and cut aggregates an…

kevinrobinson:
…d longer-term charts

This is part of https://github.com/studentinsights/studentinsights/issues/2.

This is addressing awesome feedback in https://github.com/studentinsights/studentinsights/issues/373 and https://github.com/studentinsights/studentinsights/issues/374, removing the exploratory prototypes and taking the views that are most directly useful.  This page is still in prototype phase.

Changes:
 - focus on recent changes, remove charts showing longer-term trends
 - remove aggregate charts, too complicated and not directly useful
 - add suffix to percentile numbers

Screenshot:
screen shot 2016-04-18 at 10 08 07 am 1

### DIFF
--- a/app/assets/javascripts/components/star_charts_page.js
+++ b/app/assets/javascripts/components/star_charts_page.js
@@ -157,16 +157,12 @@ $(function() {
         ),
         dom.div({ style: { position: 'relative' } },
           dom.div({ style: { flex: 1 } },
-            dom.div({ style: { display: 'flex', padding: 20 } },
-              this.renderScatterplot(merge(sizing, { width: 1200, height: 300 }))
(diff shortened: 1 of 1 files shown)


kevinrobinson:
@alexsoble @jgeiser I think this is a good simplification for now, and then we can revisit more in depth later.  One other thought here is whether we want to encourage looking only at the most recent change, without a fuller context of the student's performance, but I think this can still serve the need to "triage," and then clicking into the student profile page can give that fuller context.

alexsoble:
@kevinrobinson Nice first cut at this!

Issue #372, opened 2016-04-15
Fix missing Services / Notes in school overview


alexsoble:
Uri Harel spotted this one. In the School Overview page, it says that zero percent of Healey students have services or notes assigned to them.

We know that's not the case: they're in the database.

```
irb(main):003:0> Service.count
=> 12
irb(main):004:0> EventNote.count
=> 96
```

kevinrobinson:
Addressed by https://github.com/studentinsights/studentinsights/pull/398.

PR #398, opened 2016-04-18
Symbolize keys after reading precomputed docs

kevinrobinson:
Addresses https://github.com/studentinsights/studentinsights/issues/372, which I think regressed in swapping from a JSON column to a plain string column that we parse manually.


alexsoble:
@kevinrobinson 👍

PR #399, opened 2016-04-18
Added school year headers to events-per-year in newsfeed

### LINKED ISSUE #218: Full case history - section breaks across school years
This is with demo data, so it's not the best example, but with all the absence and tardy data it's hard to see milestone dates like the beginning of a new school year.

 screen shot 2016-03-21 at 8 59 19 am

One idea might just be adding a heading for the start of each school year, or perhaps using the box styling for sections that @bw120 is awesomely trying out in https://github.com/studentinsights/studentinsights/pull/199.

erose:
Fixes #218.
screen shot 2016-04-18 at 12 41 57 pm


kevinrobinson:
Awesome!  🎉 
@erose when merging can you also @ either @alexsoble or me so that we can deploy this to production as well?  Otherwise we can get drift between the demo site and production site, which in general is no bueno. :)

erose:
Sure!

So the demo site is deployed automatically via Travis, but the production site is manually deployed? Makes sense.

kevinrobinson:
yep, exactly!

alexsoble:
Nice @erose!  🎉 🎉

PR #400, opened 2016-04-18
Tweak stylish newsfeed -- aligning event-type badges for easier scanning

### LINKED ISSUE #219: Full case history - improving visual clarify and scanning with horizontal alignment
For certain students in the production deployment, it's difficult to scan the list with the color patches not horizontally align.  There's also a feeling of clutter over a straight table or list from this.

screen shot 2016-03-21 at 9 20 56 am

One improvement might be to fix the label size for the date text, so that there's only two horizontal slots that the colors can appear in - one for the small size items and another for the large size items.

erose:
screen shot 2016-04-18 at 1 21 12 pm

Fixes #219. I hope `display: 'inline-block'` is well-behaved on all browsers, I've heard it might not be

So we should test this on IE. (I don't know how to use BrowserStack right now, but it would be cool to learn!)

### DIFF
--- a/app/assets/javascripts/student_profile/profile_details.js
+++ b/app/assets/javascripts/student_profile/profile_details.js
@@ -262,6 +262,7 @@
         var text = event.message;
       }
 
+      var dateStyle = {display: 'inline-block', width: 180};
       var type_to_color = {
         "Absence": '#e8fce8',
         "Tardy": '#e8fce8',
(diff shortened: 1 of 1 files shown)


kevinrobinson:
@erose nice, awesome improvement! 👍 

There's no problem with using `display: inline-block` in IE, but it's great to think of this and check layout-type changes on IE!  I can help check this tomorrow, but if you want to give browserstack a spin, just visit https://www.browserstack.com/, sign up and then click on IE11 on the latest version of Windows.  Then you can visit the demo site in IE within the browser, and sign in with the demo account.

I think at the moment only @alexsoble and I can deploy branches to the demo site without merging to master, but we could talk through that and change it tomorrow too.  Or since we're pretty confidence this change will be good, you could merge optimistically then verify on browserstack.

erose:
Okay, cool, looks pretty easy. I'll merge optimistically and check.

PR #401, opened 2016-04-19
Add 'this school year' to Attendance & Behavior summary boxes.

### LINKED ISSUE #330: Improve labeling on Student Profile V2 Summary boxes
Attended today's Student Support Team meeting at the Healey school to observe how they used Student Insights. One thing I noticed:

A UI issue caused some confusion during the meeting. One student had an "Absences count" of 3 noted in her summary box at the top of the profile. The staff wanted to double-check that number and so they looked at the student's activity feed. There were many more than 3 absences in the activity feed. It wasn't clear to the staff that the "Absences count" listed at the top referred only to absences in the current school year, not overall number of absences.

Clearly labeling which values refer to current school year only would be helpful here. May also be useful to segment the activity feed events by school year, with headers for each school year. Finally, we may also want some kind of label for the spark lines to make it clear that they are all-time trends as opposed to the current-school-year-only figures directly above them.

erose:
It looks kind of noisy to me. Not sure how to fix that.

screen shot 2016-04-19 at 5 49 22 pm

Fixes #330.


alexsoble:
Here is a design concept that @erose and I made at Code for Boston! 

It gets rid of the sparklines, since they don't correspond to the graphs below anymore.

This is what the tab would look like: 

eli and alex design

Feedback @kevinrobinson?

kevinrobinson:
@alexsoble @erose Nice, seems like a good improvement over the current setup but loses the context about change.  Any thoughts on how to help users understand what's happening over time here at a glance?

awalker4:
What about colored arrows next to the numbers? Absences that are trending up can have a red up arrow, trending down a green down one.

erose:
@awalker4 I like it, sounds cool.

awalker4:
I suppose even just making the numbers red/green would get the point across too.

erose:
Here's the updated version.

screen shot 2016-04-29 at 5 31 25 pm

It wasn't obvious to me what the past point for comparison should be -- how have things changed since three months ago? a month ago? a year ago? So I went with "a year ago" for now, but would love any feedback.

Also, UI feedback would be great.

erose:
It would also be nice for testing if demo data had any attendance events in the last school year -- that should be a quick change to the `.rb` scripts, I can make it.

erose:
Pushed the wrong branch -- now the code I was talking about is actually here.

awalker4:
@erose Comparison from last year seems to make sense if those numbers are for the current year. I like the clean look! I can see how the arrows wouldn't be intuitive at first - maybe a tooltip that, on hover, says  "Student had N tardies last year".

kevinrobinson:
@erose I don't quite have the context from the earlier issue, but I think year-over-year trends might be less immediately useful and actionable as compared to week-over-week changes.  The change to absences-per-week bar charts improved that a lot in the details section, so it might be worth trying to adapt some part of that here.  Up to you, just a thought :)

erose:
@kevinrobinson Yep, makes sense! I pushed some changes -- the arrow changes are now comparing total last week at this time to the total so far this week.

kevinrobinson:
@erose nice!  mind pasting what this looks like?

erose:
Nothing visually has changed since the previous screenshot, except the tooltips, which you reminded me I should include.

screencapture

I'm not sure about the discoverability of these tooltips. Brainstorming solutions -- maybe the arrow should highlight when the cursor is over it? (The whole column already has the "hand" cursor).

erose:
One thing I'm noticing is that, with the demo data, there are rarely any of these events in the past week or in this week. So the trend arrow is nearly always just a purple line.

If we think that total number of these events per student is roughly the same in the demo data as it is in the real data, then maybe we should choose a bigger window to compare. (Last month vs this month?)

Another thought is that, when the window of time we're looking at is so small, we're only really going to be accurate in detecting changes from "no absences" to "some absences" or the other way round. Trying to do it more fine-grained might lead to false positives: if Ben has had a 25% chance of being absent every day since the start of the semester, at the end of any given week we're likely to be displaying an up arrow or a down arrow, even though nothing has really changed.

So maybe we should just try to detect changes from "zero" to "some" or "some" to "zero", instead of the actual numbers.

erose:
Another thought: display total for the last month/school year/big chunk of time, and display a red up-arrow if there's been any incidents so far this week.

There would be no use for the green arrow.

erose:
Talked with @kevinrobinson, redesign is in order.

erose:
Closing this since we're going to go down a different route: bar chart sparklines.

PR #404, opened 2016-04-20
Sped data

eudaimonious:
@kevinrobinson Not sure if we want to merge this yet since I haven't gotten around to an integration test yet. Up to you.

Folks should feel free to do some code review on this in the meantime. See Issue #8 and Issue #165 for context.


app/views/students/sped_referral.pdf.erb
```diff
@@ -1 +1,56 @@
...
+      <li><%= "#{incident.occurred_at.strftime("%-m/%-d/%Y")} - #{incident.incident_code} - #{incident.incident_location}"%></li>
+      <li><%= incident.incident_description %></li>
+    </ul>
+  <% end %>
+</p>
+<p>
+  <h4>Academic assessments:</h4>
+  <% @student_assessments.each do |test| %>
+    <ul style="list-style-type:none">
+      <li><%= "#{test.date_taken.strftime("%-m/%-d/%Y")} - #{test.assessment.family} #{test.assessment.subject} - #{test.percentile_rank.try(:ordinalize) || 'Unknown'} percentile" %></li>
```
kevinrobinson: Not all assessments use `percentile_rank`, you can see which types use which fields here: https://github.com/studentinsights/studentinsights/blob/master/app/models/student_assessment.rb#L14 (and if they have multiple, just grab all the ones that have to be `.present?`)

kevinrobinson:
@eudaimonious This looks great!  👍 

I'd say let's merge this as-is to get in on master since it's not exposed, and then when we add the "Prototype SPED referral" button to the profile page so @jsgeiser can start trying out we can try adding a smoke test (or when verifying different assessment types work or whatever else comes up).

PR #405, opened 2016-04-20
Add milestones to README in place of project priorities

kevinrobinson:
These are stale, update them to use new milestones.


alexsoble:
Good move @kevinrobinson! 👍

PR #409, opened 2016-04-21
Nil risk assessments

brisandbeans:
Included:
- A minor tweak to `fake_user.rb` to occasionally create demo users with no MCAS or STAR assessments and therefore with a nil risk level.
- A change to the sorting code for homeroom rosters, so that nil risk levels are treated as higher than any numeric value.


alexsoble:
Looks great @brisandbeans! Thank you for the PR!

PR #431, opened 2016-04-27
when array is empty, display "No staff" under "Staff Providing Service"

### LINKED ISSUE #425: "Staff providing services" is empty when there are no services
screen shot 2016-04-25 at 3 46 26 pm

Other fields have "None" or "No serices" -- maybe we should use one of those.

ianaroot:
-add ternary to assignment of var educatorNames so that instead of assigning an empty array it assigns ["No staff"]
-fixes #425


kevinrobinson:
@ianaroot thanks!

image

PR #432, opened 2016-04-27
Make sign-in textareas wider, isolate nav CSS more and use flexbox for layout

kevinrobinson:
Before:
screen shot 2016-04-26 at 10 08 56 pm
screen shot 2016-04-26 at 10 09 00 pm

After:
screen shot 2016-04-26 at 10 05 35 pm
screen shot 2016-04-26 at 10 05 38 pm


alexsoble:
@kevinrobinson Excellent! This addresses an issue that Kathleen Seward mentioned was an obstacle (long Somerville emails didn't fit inside small input box). Was going to document, but I suppose there's no need to document when we already have a fix.  😊

PR #434, opened 2016-04-28
Update project description on login page

alexsoble:
It occurred to me as our schools partner Uri was talking the other day about rolling the app out to all elementary schools in Somerville . . . 

It might be a good idea to make the project description on the login page less inaccurate! 

## New look

__Upper part:__
screen shot 2016-04-28 at 3 01 51 pm

__Lower part:__
screen shot 2016-04-28 at 3 07 42 pm

## Copy

+ The list of people working on the app (both on the code side and the schools side) has changed over time. And will continue to change over time as the project evolves. So it doesn't make sense to have this as the first thing someone reads when s/he goes to log on. We should move this to a separate project history or contributors page.
+ Added some (hopefully) pithy sentences about what problems the project aims to fix. Those are more important from the point of view of the folks using the app than who exactly is building it.
+ The "What does it do" section is still inaccurate, because it doesn't mention the school overview page.

## CSS

+ Move CSS classes that were only being used by the about page into `about.scss.erb`.
+ Trimmed CSS, fewer IDs, more classes

This look good @kevinrobinson? Thoughts?

### DIFF
--- a/app/assets/stylesheets/about.scss.erb
+++ b/app/assets/stylesheets/about.scss.erb
@@ -2,71 +2,29 @@
 
 .about-page {
 	text-align: center;
-	margin-left: auto;
-	margin-right: auto;
-	h4 {
-		margin-bottom: 10px;
+
+	.about-big, h4 {
+	  font-size: 24px;
+	  font-weight: 300;
+	  color: $blue;
 	}
 
-	h5 {
-		margin-bottom: 20px;
-	}
-}
-
-.what, .how, .logos {
-	margin-top: 70px;
-}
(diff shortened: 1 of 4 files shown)


kevinrobinson:
@alexsoble This looks like an amazing improvement, I'd say 🚢 this pronto!  Also, I was going to take a pass at updating some of the README product overview Tuesday in advance of the demo night too and now may copy some of your awesome work here :)

alexsoble:
@kevinrobinson, Okay, 🚢 ing! 

:shipit:

Issue #435, opened 2016-04-28
Fix school-based authorization for homerooms


alexsoble:
Feedback from Kathleen Seward, principal at the West Somerville Neighborhood School:

When Kathleen showed the app to her staff, they noticed that the "Homerooms" dropdown in the roster page included homerooms at other schools, and that their access to rosters extended across both Healey and West Somerville.

This is because there's no scoping of roster access by school. Time to fix that.

This is high-priority but not an urgent concern because the information surfaced on the roster page is very general: name, demographic information, risk levels, and recent scores. In-depth notes and behavior information are not surfaced on the roster page.

Issue #338, opened 2016-04-12
Build out views for district-wide administrators


alexsoble:
Right now, the widest view any user has into the system is a school-level view. No way to zoom out and see data views at the district-wide level.

alexsoble:
Thinking about it more closely, this isn't really part of the "second school" milestone. 

We can build out views and logic to have multiple schools on the app, with each staff member seeing the appropriate students for her/his school. 

Views and logic for district-wide staff are separate chunks of work that we can punt on for right now.

PR #440, opened 2016-04-29
Require educators be assigned to school

alexsoble:
This [yak-shaving](http://catb.org/jargon/html/Y/yak-shaving.html) for #435. When scoping educator UI access by school, specs blew up because our test educators weren't all assigned to schools.

Since we punted on building views for district wide educators (#338, explicitly de-prioritized in conversation with Uri), all educators should be assigned a school. 

If you're assigned to zero schools, we can't show you any data. We're not building out the use case for a district-wide admin assigned to all schools. And we haven't run across a use case of an educator who works with multiple schools.

Even if it's not really super essential for #435, this clarifies our data model, adds good validation, and makes our specs simpler.

### DIFF
--- a/app/models/educator.rb
+++ b/app/models/educator.rb
@@ -14,6 +14,8 @@ class Educator < ActiveRecord::Base
 
   validates :email, presence: true, uniqueness: true
   validates :local_id, presence: true, uniqueness: true
+  validates :school_id, presence: true
+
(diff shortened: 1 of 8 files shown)


kevinrobinson:
drive-by comment: 👍 awesome clarification and improvement!

PR #442, opened 2016-04-29
Fix broken links in attendance and behavior

### LINKED ISSUE #438: Fix broken links in Student Profile – Attendance & Behavior Tab
Bug spotted by Leslie Lartey, Counselor Educator at West Somerville:

missing links

erose:
Fixes #438

I also added the first integration test for ProfileBarChart. I copied the style of PageContainerSpec.js, with one exception: in that spec, the function `renderInto`

    renderInto: function(el, props) {
      var mergedProps = merge(props || {}, {
        nowMomentFn: function() { return Fixtures.nowMoment; },
        serializedData: Fixtures.studentProfile,
        queryParams: {}
      });
      return ReactDOM.render(createEl(PageContainer, mergedProps), el);
    },

is calling merge(provided props, default ones), so the default values for a key overwrite the provided values. Is this intentional?

I flipped the definition in my spec because I found it more natural to specify different values for the props in different tests.


kevinrobinson:
@erose nice fix!  This was probably a bug and not intentional, the semantics you have here are much more what I'd expect :)

PR #444, opened 2016-04-30
Attendance events can now occur in the current school year.

### LINKED ISSUE #227: Distribute demo data attendance events evenly across school years

erose:
Fixes #227.

I changed the way we do this to a more discipline-incident-y way (generate overall, not per year). I suspect the resulting distribution isn't accurate.

screen shot 2016-04-30 at 11 42 49 am

Does this match the real distribution?

### DIFF
--- a/app/demo_data/fake_student.rb
+++ b/app/demo_data/fake_student.rb
@@ -147,28 +147,30 @@ class FakeStudent
   end
 
   def add_attendance_events
-    # Probabilities: https://github.com/codeforamerica/somerville-teacher-tool/issues/94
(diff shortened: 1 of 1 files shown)


erose:
For reference: the absences & tardies distribution on the demo site right now.

screen shot 2016-05-01 at 1 28 50 pm

Am I right to assume that this isn't realistic?

erose:
Yo @alexsoble can you get the real distribution from production for me?

Issue #452, opened 2016-05-02
Update list of services in profile page


kevinrobinson:
We're not yet clear on the path forward here, current plan is:

> Remove reading intervention and math intervention, no academic interventions this round.

And discussing over email.  Tracking here as a loose end for this milestone.

kevinrobinson:
Paraphrased from email with Uri:

Reading is being used in MTSS meetings, let's keep Reading Teacher with the definition from the user manual: `Reading Teacher: Student works with a reading specialist at least 4x/week for 30-40 minutes.`

kevinrobinson:
Other context:

Uri currently tracks reading interventions in a spreadsheet, and then twice a year these are bulk uploaded into Aspent in a manual handoff to John.

alexsoble:
Stole this assignment from you @kevinrobinson!  😊

kevinrobinson:
@alexsoble 👍

alexsoble:
We decided in last week's group meeting to go forward with Uri's latest proposal: keep the Reading Intervention option, get rid of Math Intervention, leaving us with a list of 6 services in profile page.

```
Counseling, in-house
Counseling, outside
Reading intervention
Attendance Officer
Attendance Contract
Behavior Contract
```

alexsoble:
Did a preliminary database check to see if we have any Math interventions recorded so far.

Indeed we do: two recorded by the principal at West Somerville, and one recorded by a teacher at the Healey School.

alexsoble:
Since users have already added a couple of Math intervention data points, I'd prefer not to delete data, delete the Math Intervention `ServiceType`, or remove any existing Math interventions from profile pages.

I'm thinking that a better route is simply to remove Math intervention from the array of Service Types passed to the student profile page front-end. So existing interventions stay, but users can't add new ones.

We should also write quick notes to the users who added Math interventions to explain why this button is going away. The reason is to provide fewer options with clearer definitions, so that we can have very high confidence in the data being entered.

alexsoble:
This is complete!

Issue #451, opened 2016-05-02
Provide user manual directly from product


kevinrobinson:
This is subject to change, but we should make the awesome help in the user manual available to users right in the product at the point they're entering data (particularly the notes and services instructions).

erose:
Wow, I never knew this document existed! It's really cool to see all the different things that Notes/Services get used for and to see what the features we're working on look like from the teacher's perspective.

It is cool if I share this with new folks? (Looks like it might have a bit of real student data in it).

kevinrobinson:
@erose It's brand new!  :) Good eyes on that, thanks!  I thought it was fully scrubbed so I removed the document and will re-scrub so we can share with everyone.

kevinrobinson:
This will need to be scrubbed.

alexsoble:
Steps!

+ [ ] Scrub user manual, add to `docs` as Markdown  
+ [ ] Design: "someone sitting in a meeting, not sure what they should" 
+ [ ] Implement

erose:
Design mockup. Here's one thought: (clicking on a question-mark bubble would open a help tooltip).

<img width="1251" alt="screen shot 2016-05-14 at 11 03 19 am" src="https://cloud.githubusercontent.com/assets/1514487/15269123/8179e96a-19c3-11e6-8da5-eb28dd7dcbc3.png">

And another take: adding some friendlier text to the "Which service?". Clicking on it would open the same help tooltip.
<img width="1221" alt="screen shot 2016-05-14 at 11 05 55 am" src="https://cloud.githubusercontent.com/assets/1514487/15269134/e0e11950-19c3-11e6-808c-a0cdd4a3c223.png">

@kbweaver Any design feedback on this?

erose:
Note to self: a similar thing exists on the roster page.
<img width="1079" alt="screen shot 2016-05-15 at 8 01 08 pm" src="https://cloud.githubusercontent.com/assets/1514487/15277618/cec3555c-1ad7-11e6-80f7-975122366d7b.png">

alexsoble:
@erose Nice! I like how the question marks are prominent and purple, draws the eyes. And I like the "help me decide" text a lot too. You can ignore a floaty question mark but you can't ignore "help me decide."

Issue #450, opened 2016-05-02
Research: Clarify "restricted" note concept


kevinrobinson:
Uri:
> There are some notes that are uber restricted (especially 51a and medical info) and get us into student privacy laws/HIPPA.  Possibly a quick fix:  if the word RESTRICTED begins a note, then when teachers see it, it appears to them as "Restricted Note - Please see your administrator for more information."  A little more complex fix - there's a checkbox called Restricted near the note, which embeds a  restricted code into the note field

Alex:
> A) Could you say a little bit about the use case for a Restricted note, as opposed to just some guidelines that we add to the app saying, "Don't write anything in these notes that has to do with XYZ?" In what situation is it important to enter a note but make sure that classroom teachers can't see it, i.e. who will end up looking at the note down the line? 
> B) Could you clarify what the Restricted button would do? Who can see a Restricted note? If Kathleen adds a note, can her AP see it? Can district wide administrators like yourself and Vince see it? Kenya?
> C) Could you add a section to the User Manual clarifying for teachers and principals what kinds of notes should be Restricted? What are the specific criteria? 
> In other words: (A) will help us understand if we should build this or maybe just add guidance about what to enter / not enter, (B) will let me know how to code the permissions if we build the Restricted note feature, and (C) will be important help text to show users about when to use it.

kevinrobinson:
This might involve not doing any software changes, but just noting it as something in the air for this milestone. @alexsoble

alexsoble:
@kevinrobinson Nice, thanks for documenting this! I'll be curious from Uri whether or not building this out is higher priority over expanding to > 2 schools.

kevinrobinson:
Yep!  Just trying to capture all the things out of email, maybe prematurely :)

alexsoble:
No such thing as prematurely! This is great.

alexsoble:
Had a great discussion with Uri and @erose today where we concretized this feature. 

My notes (all quotations from Uri Harel):

## What are restricted notes?

* "Usually comes from the principal"
* "Data point that's sensitive"
* "Certain health data can not be seen by a teacher"
* "51A"
* "Immigration status changing"
* "Some of that data cannot be put in the system"

## What is the use case / rationale?

"Think of the case of a turnover principal. Jill files a 51A or there's a drug situation. You don't want the teacher necessarily to see that. If Jill doesn't put it in, next year a new admin comes in, a new guidance counselor comes in, and they don't have the information."

## Who can see restricted notes?

[...]

PR #453, opened 2016-05-02
Update project priorities

kevinrobinson:
@alexsoble From our conversation last week with Uri.  I also updated the milestones too, but those don't sort so not sure it's super useful.  What do you think makes sense here?  Just trying to visualize the higher-level threads and be able to pick out the issues that matter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,13 @@ This is a web product, integrated with the Aspen SIS, district LDAP authorizatio
 
 
 ## Project priorities
-These are tracked in GitHub with [milestones](https://github.com/studentinsights/studentinsights/milestones).
+Immediate priorites are tracked in GitHub with [milestones](https://github.com/studentinsights/studentinsights/milestones).
+ 1. Whole district classroom teacher rosters and profile in September
+ 2. X2 roll-over to next school year (in July or August)
+ 3. Healey SST/MTSS
+ 4. West SST/MTSS
+ 5. East SST/MTSS
+ 6. PDF Data dump
 
 ## Contributing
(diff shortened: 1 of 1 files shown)


alexsoble:
@kevinrobinson My thoughts:

+ Nice, thank you for adding this. Good to keep the list up to date.

+ On the X2 roll-over: This will end up being an issue if the data input from X2 changes in an unexpected way. If our import scripts prove really robust, there could be no work here at all. It's kind of a "potential issue", something to keep an eye on more than defined work that needs to be done.

+ I wonder which issues we need to resolve at Healey and West before expanding to East. #450, #451, #452? If we agree with Uri that we should resolve these first, then solidifying Healey / West is one chunk of work and expanding to East and beyond is another.

alexsoble:
@kevinrobinson Another way to organize these:

__MAY / EARLY JUNE__

+ Solidify SST/MTSS at Healey and West
+ Run data import for East and see if we run into issues
+ _Last day of current school year: June 12_

__LATE JUNE / JULY / AUGUST__

+ Import data for all elementary schools
+ Build out whole district classroom teacher rosters and profile for September
+ Keep an eye on X2 roll-over to next school year
+ _August 29: 1st day for Staff Prof. Development_
+ _August 30: Staff Prof. Devel./Paras report to work_
+ _August 31: 1st day for students 1-12 (1/2 day)_

__SEPTEMBER / OCTOBER__

+ Debug, handle feedback, input, and scaling issues from going from 2 schools to 7
+ Concierge PDF data dump of student info through SPED referral process

Source for dates: [2016-17 Calendar](http://www.somerville.k12.ma.us/education/page/download.php?fileinfo=QXBwcm92ZWRfMjAxNi0yMDE3X0NBTEVOREFSXzAyMDQxNi5wZGY6Ojovd3d3L3NjaG9vbHMvc2MvcmVtb3RlL2ltYWdlcy9kb2NtZ3IvMjY4M2ZpbGUxNTE3MC5wZGY=&sectiondetailid=19179)

erose:
Thanks for posting this, it helps me a lot to know where the big-picture stuff is headed. Let's definitely share these at the next meeting.

alexsoble:
Merging in!

PR #454, opened 2016-05-02
Added getServiceMessage() function.

### LINKED ISSUE #449: Fix issue where object ID shows up in Service in Profile Feed
Like this (guessing 34 is the Rails ID of this particular Service):

screen shot 2016-05-01 at 7 34 27 pm

cc @erose — might be a quick fix for you since you've been doing work in this area.

erose:
Fixes #449 

Display both the code and the description. 

screen shot 2016-05-01 at 11 05 12 pm

### DIFF
--- a/app/assets/javascripts/student_profile/profile_details.js
+++ b/app/assets/javascripts/student_profile/profile_details.js
@@ -75,6 +75,19 @@
       dibels: React.PropTypes.array,
       chartData: React.PropTypes.object,
       attendanceData: React.PropTypes.object,
+      serviceTypesIndex: React.PropTypes.object
+    },
+
+    getMessageForServiceType: function(service_type_id){
+      // Given a service_type_id, returns a message suitable for human consumption describing the service.
+      var lookup = this.props.serviceTypesIndex;
+      if (lookup.hasOwnProperty(service_type_id)){
+        var text = lookup[service_type_id].name;
+      } else {
+        var text = "";
+      }
+
+      return service_type_id + ": " + text;
     },
 
     getEvents: function(){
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/student_profile/profile_details.js
```diff
@@ -75,6 +75,19 @@
       dibels: React.PropTypes.array,
       chartData: React.PropTypes.object,
       attendanceData: React.PropTypes.object,
+      serviceTypesIndex: React.PropTypes.object
+    },
+
+    getMessageForServiceType: function(service_type_id){
+      // Given a service_type_id, returns a message suitable for human consumption describing the service.
```
alexsoble: Lol @ "suitable for human consumption" — cute turn of phrase!

kevinrobinson:
Nice! will humans in schools know what these database primary key numbers are and want to see them?

erose:
@kevinrobinson: There's `id` and there's `service_type_id`; previously I was displaying just `id`, now I display `service_type_id` along with the short description (from `serviceTypesIndex`). So "503" above isn't the database primary key, it's `service_type_id`.

(I don't know what `service_type_id` is, whether it comes from us or from the school, or whether users want to see it or not. Do you guys know?)

alexsoble:
@erose — it's an internal Rails ID, users won't want to see it.

alexsoble:
@erose — Any `integer` attribute that ends with `_id` is a Rails foreign key used for associations, doesn't have any user facing meaning.

Here's a good place to read more on foreign keys in Rails: http://guides.rubyonrails.org/association_basics.html#options-for-has-many-foreign-key

erose:
Gotcha! Thanks. Fixed here #456.

alexsoble:
@erose No problem! Thanks for the fix!

kevinrobinson:
@erose @alexsoble 👍 sweet!

Issue #439, opened 2016-04-28
Clarify ability to add free-text in educator services dropdown


alexsoble:
Feedback from Kathleen Seward, principal at the West Somerville Neighborhood School:

When adding a Service for a student, Kathleen wanted to put the name of an outside service provider who works with students at West Somerville, but isn't actually on staff at the school.

She didn't see that person's name in the autocomplete, so assumed she would not be able to add that staff member and put her own name instead. 

The autocomplete can accept free-text input. But since that's not immediately clear to our user, we need to brainstorm and implement design changes that will make it obvious and intuitive.

awalker4:
We could have the first row of the autocomplete just mirror what you're typing. If no other names are a match, you'll only have the one option. Quick example:

![example](https://cloud.githubusercontent.com/assets/3846783/14942294/53901854-0f86-11e6-8311-64f95295b15f.png)

alexsoble:
@awalker4 This looks like a really good idea to me!

alexsoble:
@awalker4 Could be a fun fix to work on during Code for Boston.

erose:
Hey @awalker4! Mirror thing looks pretty cool. If you're working on this, assign yourself.

awalker4:
@erose Awesome, I'll look into it! I don't see any option to assign myself though.

erose:
It's a *drumroll.........* permissions problem! You should see the option now.

PR #464, opened 2016-05-06
Add simple name validation for services

### LINKED ISSUE #414: Add validation to services dropdown of educators
Now that we've implemented #307, educators can enter any string they like for the providing educator name when recording a new service.

Educator names come in from the Student Information System as `"LastName, FirstName"` and are stored that way in our database. 

Would be nice to validate that that outside service provider names are entered the same way. This will make the data easier to work with if we need to surface or query it down the line. 

This would include client-side validation that `provided_by_educator_name` includes the string `", "`. It would also add some helpful hints to the user if they've entered a name that's doesn't match the format.

awalker4:
Fixes #414 - Educator names should be entered in the form Last Name, First Name. We should at the very least check that there's a comma in between two words.

### DIFF
--- a/app/assets/javascripts/student_profile/page_container.js
+++ b/app/assets/javascripts/student_profile/page_container.js
@@ -94,10 +94,15 @@
     },
 
     onClickSaveService: function(serviceParams) {
-      this.setState({ requests: merge(this.state.requests, { saveService: 'pending' }) });
-      this.api.saveService(this.state.student.id, serviceParams)
-        .done(this.onSaveServiceDone)
-        .fail(this.onSaveServiceFail);
+	// Very quick name validation, just check for a comma between two words
+	if ((/(\w+, \w|^$)/.test(serviceParams.providedByEducatorName))) {
+	    this.setState({ requests: merge(this.state.requests, { saveService: 'pending' }) });
+	    this.api.saveService(this.state.student.id, serviceParams)
+		.done(this.onSaveServiceDone)
+		.fail(this.onSaveServiceFail);
+	} else {
+	    this.setState({ requests: merge(this.state.requests, { saveService: 'Please use the form Last Name, First Name' }) });
+	}
     },
 
     onSaveServiceDone: function(response) {
--- a/app/assets/javascripts/student_profile/record_service.js
+++ b/app/assets/javascripts/student_profile/record_service.js
@@ -66,7 +66,7 @@
     getInitialState: function() {
       return {
         serviceTypeId: null,
-        providedByEducatorName: null,
+        providedByEducatorName: "",
         momentStarted: moment.utc() // TODO should thread through
       }
     },
@@ -196,8 +196,7 @@
           style: styles.cancelRecordServiceButton,
           onClick: this.onClickCancel
         }, 'Cancel'),
-        (this.props.requestState === 'pending') ? dom.span({}, 'Saving...') : null,
-        (this.props.requestState === 'error') ? dom.span({}, 'Try again!') : null
+        (this.props.requestState === 'pending') ? dom.span({}, 'Saving...') : this.props.requestState
       );
     }
   });
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -5,8 +5,6 @@ class Service < ActiveRecord::Base
   has_many :discontinued_services
 
   validates_presence_of :recorded_by_educator_id, :student_id, :service_type_id, :recorded_at, :date_started
-  validates :provided_by_educator_name, allow_nil: true,
-            format: { with: /(\w+, \w+|^$)/, message: "Please use the form Last Name, First Name" }
 
   def discontinued?
     discontinued_services.size > 0
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -5,6 +5,8 @@ class Service < ActiveRecord::Base
   has_many :discontinued_services
 
   validates_presence_of :recorded_by_educator_id, :student_id, :service_type_id, :recorded_at, :date_started
+  validates :provided_by_educator_name, allow_nil: true,
+            format: { with: /(\w+, \w+|^$)/, message: "Please use the form Last Name, First Name" }
 
   def discontinued?
     discontinued_services.size > 0
--- a/app/assets/javascripts/student_profile/services_list.js
+++ b/app/assets/javascripts/student_profile/services_list.js
@@ -155,8 +155,8 @@
     },
 
     renderEducatorName: function (educatorName) {
-      if (educatorName === null) {
-        return dom.div({}, 'No staff recorded');
+      if (educatorName === "" || educatorName === null) {
+	return dom.div({}, 'No staff recorded');
       } else {
         return dom.div({}, 'With ' + educatorName);
       };
--- a/app/assets/javascripts/student_profile/student_profile_page.js
+++ b/app/assets/javascripts/student_profile/student_profile_page.js
@@ -316,8 +316,8 @@
       var activeServices = this.props.feed.services.active;
       var educatorNamesFromServices = _.pluck(activeServices, 'provided_by_educator_name');
       var uniqueNames = _.unique(educatorNamesFromServices);
-      var nonNullNames = _.filter(uniqueNames, function(id) { return id !== null; });
-      var educatorNames = _.isEmpty( nonNullNames ) ? ["No staff"] : nonNullNames;
+      var nonEmptyNames = _.filter(uniqueNames, function(id) { return id !== "" && id !== null; });
+      var educatorNames = _.isEmpty( nonEmptyNames ) ? ["No staff"] : nonEmptyNames;
 
       var limit = 3;
 
--- a/spec/javascripts/student_profile/page_container_spec.js
+++ b/spec/javascripts/student_profile/page_container_spec.js
@@ -115,6 +115,12 @@ describe('PageContainer', function() {
 	    onProvidedByEducatorName: 'Teacher, Test'
 	});
 	expect(el).toContainText('Saving...');
+
+	// Name can also be blank
+	component.onClickSaveService({
+	    onProvidedByEducatorName: ''
+	});
+	expect(el).toContainText('Saving...');
     });
 
     // TODO(kr) the spec helper here was reaching into the react-select internals,
--- a/spec/javascripts/student_profile/page_container_spec.js
+++ b/spec/javascripts/student_profile/page_container_spec.js
@@ -102,6 +102,21 @@ describe('PageContainer', function() {
       });
     });
 
+    it('verifies that the educator name is in the correct format', function() {
+	var el = this.testEl;
+	var component = helpers.renderInto(el, {});
+
+	component.onClickSaveService({
+	    onProvidedByEducatorName: 'badinput'
+	});
+	expect(el).toContainText('Please use the form Last Name, First Name');
+
+	component.onClickSaveService({
+	    onProvidedByEducatorName: 'Teacher, Test'
+	});
+	expect(el).toContainText('Saving...');
+    });
+
     // TODO(kr) the spec helper here was reaching into the react-select internals,
     // which changed in 1.0.0, this needs to be updated.
     // it('can save an Attendance Contract service, mocking the action handlers', function() {


awalker4:
I've been trying to bubble up this error message to record_service.js but I'm very new to this stack and could use some help. If validation fails, we'll get a 422 response that contains the string "Educator name is invalid". How can I access this message? My guess was in page_container.js - onSaveServiceFail(). That doesn't seem to get the response body though...

erose:
Woop woop thanks for the PR!

I too have never touched this part of the stack, so this may be a half-truth, but looks like the `request` argument to the function `onSaveServiceFail()` is in fact some ~~sort of JQuery thing~~ a [jqXHR object](http://api.jquery.com/jQuery.ajax/#jqXHR) that contains `responseText` and `status == 422`. So you can get the data from there. I do not know why it is not named `response` -- we should dig and try to find that out, then rename it if that is actually what it is.

(I found the above out by throwing a `console.log(request)` into `OnSaveServiceFail`, then trying something that didn't validate.)

I think the error handling is accomplished with JQuery callbacks. `api.js` is involved.

Let's experiment with ways to display the error message! We should make it friendlier than "Educator name is invalid." Maybe we could try putting it in place of the "Try again" message that currently gets displayed on any error.

awalker4:
@erose Just what I needed, thanks! I see we use 'request' in a few places but 'response' elsewhere in page_container. Probably makes sense to name them all 'response'. How does this sound for an error message?

Please use the from 'Last Name, First Name'

In the future maybe I can look into cleaning up error handling elsewhere, since it looks like we mostly just come back with 'Try again'.

awalker4:
Oops, didn't mean to pull in your other changes! Git is fun!

app/assets/javascripts/student_profile/record_service.js
```diff
@@ -197,7 +197,7 @@
           onClick: this.onClickCancel
         }, 'Cancel'),
         (this.props.requestState === 'pending') ? dom.span({}, 'Saving...') : null,
-        (this.props.requestState === 'error') ? dom.span({}, 'Try again!') : null
+        dom.span({}, this.props.requestState)
```
erose: What will this do when `this.props.requestState === 'pending'`? Off-the-cuff, looks like it will display 'Saving...' on top and 'pending' on the bottom, which isn't what we want.

app/models/service.rb
```diff
@@ -5,6 +5,8 @@ class Service < ActiveRecord::Base
   has_many :discontinued_services
 
   validates_presence_of :recorded_by_educator_id, :student_id, :service_type_id, :recorded_at, :date_started
+  validates :provided_by_educator_name, allow_nil: true,
+            format: { with: /\w+. \w+/, message: "^Please use the form Last Name, First Name" }
```
erose: What is the `^` here -- like pointing up?

erose:
I like "Please use the form..." as an error message. Added some code comments. Thanks for this work! Let's get this merged tonight!

Minor stuff: if we can get away without a new dependency, that's slightly preferable. @eudaimonious @ianaroot @BWieber anyone else who knows Rails, do any of you know of a good way to do this (custom error messages in Rails) without this gem? Are any of the suggestions [here](http://stackoverflow.com/questions/808547/fully-custom-validation-error-message-with-rails) workable?

That said, specifying `message: "..."` is really sleek, so if other solutions are clunkier, we should probably go with this one.

erose:
Also, can we see a screenshot of the new error message? (Screenshots are really nice for PRs that change anything remotely visual).

ianaroot:
here is an example of a custom error message from my work codebase without
any gems
```class EventsStaffUsersRole < ActiveRecord::Base

belongs_to :event
  belongs_to :users_role

  validate :organizations_match?
  private
    def organizaitons_match?
      if event.organization_id != users_role.organization_id
        errors.add(:organization_id, "organization id's on event and
staff_users_role do not match.")
      end
    end
  #end
end```
i think this is what you're looking for, right?

On Tue, May 10, 2016 at 12:20 PM, Eli Rose <notifications@github.com> wrote:

> Also, can we see a screenshot of the new error message? (Screenshots are
> really nice for PRs that change anything remotely visual).
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/studentinsights/studentinsights/pull/464#issuecomment-218211269>
>

File: app/models/service.rb (hunk 2 above)
Comment (awalker4) [inline app/models/service.rb]: The gem I found will use that to parse a custom error message. Screenshot incoming!

File: app/assets/javascripts/student_profile/record_service.js (hunk 1 above)
Comment (awalker4) [inline app/assets/javascripts/student_profile/record_service.js]: Oh yeah, good point. Since the string isn't 'error' anymore, I guess I need a smarter way of knowing when to display it. It's not like I can just say requestState != 'pending'.

Comment (awalker4) [inline app/assets/javascripts/student_profile/record_service.js]: Okay, given that there are only two states, I can just update the else part of the 'pending' line.

**Comment by awalker4** [comment]
@ianaroot Thanks for the snippet! Personally, I'd prefer the gem, mostly because it feels like we're just repeating code at that point. Up to you guys, though. Here's what it looks like:

error

**Comment by kevinrobinson** [comment]
@awalker4 Nice!  I have one thought for you, and up to you how you want to proceed.

This is adding a validation at the database level, so for the user to get feedback they need to make the request, try to do the database write, fail the response, and then flow that back through to the UI.  Would it be simpler and give more immediate user feedback if we added the validation on the client-side in the UI code?

File: app/assets/javascripts/student_profile/page_container.js [hunk 3]
```diff
@@ -114,8 +114,8 @@
       });
     },
 
-    onSaveServiceFail: function(request, status, message) {
-      this.setState({ requests: merge(this.state.requests, { saveService: 'error' }) });
+    onSaveServiceFail: function(response, status, message) {
+      this.setState({ requests: merge(this.state.requests, { saveService: response.responseJSON.errors[0]}) });
```
kevinrobinson: @awalker4 One thought here, which is not clear at all from the code, is that these objects right now all hold string keys representing the state of the request.  That's what the UI code is branching on later on, as you awesomely discovered and updated.

I think your approach makes sense here, but any other thoughts you have on how to clarify this in the future are awesome too :)

awalker4:
@kevinrobinson That's a good point. I do like how Rails makes validating simple for you, but I can see the appeal of having a client-side check. Let me see if I can add something!

app/assets/javascripts/student_profile/page_container.js
awalker4: I was a bit hesitant on that bit. I'm not a fan of magic numbers, and also it's not really intuitive at all. I think we can get a lot more mileage out of api.js to have some more consistent return values.

kevinrobinson: Cool, that sounds like a good approach too!  Any thoughts you have here that we can capture in the code are great to help future us and future others 👍

awalker4:
I suppose putting a message on the front end means we don't really need it in Rails, so I can take out that dependency. And actually, @kevinrobinson if I'm not sending a message back up I might as well revert the change to onSaveServiceFail. Then later on we can revisit that response object.

awalker4:
Oops, forgot that educator could be null. I'll just fix that, and while I'm at it, add a fix for #470.

app/models/service.rb
```diff
@@ -5,6 +5,8 @@ class Service < ActiveRecord::Base
   has_many :discontinued_services
 
   validates_presence_of :recorded_by_educator_id, :student_id, :service_type_id, :recorded_at, :date_started
+  validates :provided_by_educator_name, allow_nil: true,
+            format: { with: /\w+. \w+/ }
```
erose: This should be a comma rather than a period, I think?

awalker4: An excellent point! It's weird that once I took the message away, Rails never complained that everything had the wrong form. Guess it's a necessary part of the validate statement.

erose: An excellent... point. I SEE WHAT YOU DID THERE.

I think Rails didn't complain cuz `.` is a wildcard in a regex.

awalker4: Oh man, I didn't even catch that! I dot I was good with puns...

awalker4:
@kevinrobinson Thanks for the note about validating on the front end, I think it makes things look a bit cleaner. What do you guys think, merge time?

app/assets/javascripts/student_profile/page_container.js
```diff
@@ -94,10 +94,17 @@
     },
 
     onClickSaveService: function(serviceParams) {
-      this.setState({ requests: merge(this.state.requests, { saveService: 'pending' }) });
-      this.api.saveService(this.state.student.id, serviceParams)
-        .done(this.onSaveServiceDone)
-        .fail(this.onSaveServiceFail);
+        // Very quick name validation, just check for a comma
+        if ((/\w+, \w+/.test(serviceParams.providedByEducatorName)) ||
```
erose: You're checking against null to ensure that it's not empty, right? Would it be simpler to store whatever the user types in `onProvidedByEducatorTyping` and just check against empty here?

I think then you could even drop the second check entirely, and let the regex take care of checking for empty.

app/assets/javascripts/student_profile/record_service.js
```diff
@@ -78,7 +78,8 @@
     },
 
     onProvidedByEducatorTyping: function(event) {
-      this.setState({ providedByEducatorName: event.target.value });
+      var updatedName = (event.target.value != "") ? event.target.value : null;
```
erose: (Cont. from above). I'd say use an empty string to mean empty, instead of using `null` to mean empty.

app/assets/javascripts/student_profile/record_service.js
```diff
@@ -196,8 +197,7 @@
           style: styles.cancelRecordServiceButton,
           onClick: this.onClickCancel
         }, 'Cancel'),
-        (this.props.requestState === 'pending') ? dom.span({}, 'Saving...') : null,
-        (this.props.requestState === 'error') ? dom.span({}, 'Try again!') : null
+        (this.props.requestState === 'pending') ? dom.span({}, 'Saving...') : this.props.requestState
```
erose: What are the values that we expect `this.props.requestState` to have here? Do we want to show them to the user like this? (Legit asking).

erose:
Thanks for being adaptable, Austin, that's really cool. I added some minor code comments. Screenshot looks good; one passing thought is would that text stand out to the user better if it was colored?

The only big thing missing IMO is a spec (i.e. test). That will let us define the behavior of the feature so we're not afraid of changing code later. The test should validate that we display the right error message when a user enters a name incorrectly, and that we allow the user to enter a name correctly. (Your work on #439 should also get a spec when it's finished).

There's currently a spec for this component at `spec/javascripts/student_profile/record_service_spec.js`, so this test should be added to that file. Up to you -- you can work on it now and/or do this pair-programming style w/me or someone else on Tuesday.

Thanks again for your work, you're really ramping up!

app/assets/javascripts/student_profile/record_service.js
awalker4: When looking at #470 I noticed that we only initialise providedByEducatorName when the user starts typing. Thus a null value meant nothing was entered and we could say "No staff recorded". You're right that an empty string would be the better choice. Fixes the original issue without me having to set the thing back to null if it's blank again. Guess I was working around the issue instead of fixing it!

app/assets/javascripts/student_profile/record_service.js
awalker4: As far as I can tell, we've got pending and error, which in this case I replaced with that message. Does seem kind of overloaded, though, doesn't it? An error message isn't an error state. I'll keep playing around with that. We're in the pending state for less than a second, at least on my machine. I kinda like that I hit the button and it flashes "Saving..." before displaying the new service. Makes the UI feel a bit more dynamic.

awalker4:
I learn best when I dive in headfirst, so thanks for the helpful comments! I like having a bit more time to flesh out the PR, make sure I really know the code. I'll see if I can knock out a spec for this one, but I'd love to pair perhaps on #439! I've never actually tried pair programming before and it seems intriguing. :)

awalker4:
Oops, looks like I can clean up some of those commits.

awalker4:
Okay, there we go - nice and clean pull request. Thanks @erose for the help on the spec! Everything look good here?

erose:
Looks good -- I have just one more question (then we can merge!) -- this allows blank educator names, right? do we want that to be a valid input?

alexsoble:
@erose @awalker4 Yes, blank educator names are valid input (if the user doesn't know who's providing the service). But if we do know who's providing the service and we have a name, we want it to be valid.

alexsoble:
@erose @awalker4 Also want to note that we definitely have at least a couple invalid (i.e. non "Last Name, First Name") entries in the database already.

alexsoble:
@erose @awalker4 This PR doesn't accept blank names as valid input, throws an error. (And surfaces it to the user! Which is nice. Love the user feedback on the client side.)

awalker4:
@alexsoble Any idea what the server side validation will do when it tries to load those two? If that will be an issue I can take the line out of service.rb. The client side stuff will make sure nothing else invalid gets in there.

erose:
@awalker4: Yeah I'd say take it out, data is only coming from one place so we can rely on the client side validation. Also, let's do a test for blank educator name.

awalker4:
@erose All set!

erose:
Awesome. You have write access, so you should be able to merge -- I'll let you do the honors.

Congrats on your first pull request!

alexsoble:
Thanks @awalker4 and @erose! Great PR, and thank you for writing the spec!

PR #466, opened 2016-05-06
Centralize student profile utility methods

### LINKED ISSUE #465: Centralize utility methods (primarily for dealing with quads, school years, etc.) in profile.
In the bar charts and elsewhere, we now have to do stuff with school years a lot. The big one is:

* Given a date, what school year does it fall into?

This is being solved at least twice. QuadConverter.js contains one of these methods. We should have a general place to put these and other utility methods.

erose:
Fixes #465. Probably `toSchoolYear()` shouldn't be in `quad_converter.js` since it has nothing to do with quads. But I ran into snakes and high water fixing that, so leaving it this way for now.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -59,6 +59,7 @@
 
 // student profile page:
 // pure ui components:
+  //= require ./student_profile/quad_converter
   //= require ./student_profile/risk_bubble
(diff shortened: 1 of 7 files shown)


alexsoble:
Thanks for adding those useful code comments @erose!

PR #469, opened 2016-05-07
Added a sublime-project file to version control, so we can share info…

erose:
… about what folders to ignore, etc.

Set this up to ignore files that are in the .gitignore.


alexsoble:
This is great. I'll add whitespace, tabs, and end-of-file preferences to this. Syncing up on whitespace and the like makes PRs a little easier to review since there's no whitespace noise in them.

PR #473, opened 2016-05-10
Only student with enrollment_status == 'Active' are displayed in the …

### LINKED ISSUE #342: Filter out inactive students from roster view
Visited the 6th grade teacher team at the Healey school this morning to show them the latest iteration of Student Insights and get their feedback.

They noticed that the roster view included students who had become inactive. Need to filter out those students to make the roster view accurate.

erose:
…roster. Fixes #342.

Before: 
screen shot 2016-05-09 at 11 43 56 pm

After:
screen shot 2016-05-09 at 11 44 15 pm

Seeya Olaf Mouse!

### DIFF
--- a/app/views/homerooms/show.html.erb
+++ b/app/views/homerooms/show.html.erb
@@ -74,27 +74,30 @@
             <th class="risk sort-default" data-sort-method="reverse_number">
               <span class="table-header">Risk</span>
             </th>
-            <% 
-            #The data-sort-method attribute forces Tablesort to use a specific sorting method, instead of the default sort it uses. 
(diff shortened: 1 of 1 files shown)


alexsoble:
Nice fix, thanks @erose!

alexsoble:
In a future PR, we may want to move this logic out of the view and into the controller. 

A Rails dev's first instinct about where to look for this kind of data-selection logic will be the controller, MVC-style: https://en.wikibooks.org/wiki/Ruby_on_Rails/Getting_Started/Model-View-Controller

But this PR gets the job done (thank you for that!), and in any case it's not adding a very worrisome level of complexity to the view.

alexsoble:
^ @erose

PR #476, opened 2016-05-10
Changed definition of .flash and removed 'field' class from notificat…

### LINKED ISSUE #455: Fix flash messages overlapping with other content
screen shot 2016-05-01 at 7 43 03 pm

screen shot 2016-05-02 at 7 14 43 am

erose:
…ions.

Fixes #455.

Before:
screen shot 2016-05-10 at 4 18 29 pm

After:
screen shot 2016-05-10 at 4 14 26 pm

Not confident in my understanding of the CSS structure so I'd like this to be reviewed before merging.

### DIFF
--- a/app/assets/stylesheets/nav.scss
+++ b/app/assets/stylesheets/nav.scss
@@ -81,7 +81,6 @@
   }
 
   .flash {
-    position: absolute;
     padding: 10px;
   }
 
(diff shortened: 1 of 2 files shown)


alexsoble:
This looks great, quick fix! 

In terms of CSS structure, the "field" class looks like cruft to me. Can't find it in active use anywhere else. 

There's a "field" CSS class in some Devise-generated views for password reset etc., but those views are not used because we're using LDAP and not passwords in this app. I'll remove the unused CSS and views in another PR.

In terms of the flash concept, some reading from the Rails guide in case you aren't already familiar with this @erose: http://api.rubyonrails.org/classes/ActionDispatch/Flash.html

Issue #365, opened 2016-04-14
Refactor StudentsController#show


alexsoble:
We are building up a very complex `@serialized_data` variable in StudentsController#show. The data from that variable gets passed through many components as it builds the StudentProfile page.

Forgetting to pass through the right prop name or calling the wrong part of `@serialized_data` leads to bugs, like this mistake I made: 
#363, fixed via #364

Goal would be to refactor `@serialized_data` toward two outcomes:
1. Fewer keys / values, simpler API for the React components to consume
2. Performance — we are probably duplicating queries

Not urgent (adding a "future" tag) but figured I'd document this soon after fixing a bug here.

kevinrobinson:
@alexsoble Yeah, I think it'd be a great improvement to simplify threading these variables through the UI code as much as they are now.  Another approach is to use more `propTypes` to define the shape of the data the UI expects (eg., [here](https://github.com/studentinsights/studentinsights/blob/master/app/assets/javascripts/helpers/prop_types.js#L30)), like you add in this PR so that these kinds of bugs can result in Jasmine test failures.

Jngai:
Hey @alexsoble I will be working on this.

alexsoble:
Talked about this with @Jngai in our video call — this area of the code isn't really breaking or surfacing bugs, so let's close this and focus our energy on areas where we've seen bugs or issues, like the data import process and the Services area.

PR #483, opened 2016-05-14
Removed attendance/behavior data from chartData because it was duplic…

erose:
…ated in attendanceData and we're only consuming that right now.

Related to https://github.com/studentinsights/studentinsights/issues/365.


erose:
@alexsoble: Will this break anything?

alexsoble:
Ooof, sorry for not answering your question in a timely way @erose!

I trust your judgement / knowledge . . . and if anything breaks, no sweat, we can easily revert!

PR #503, opened 2016-05-18
Some basic tests for students#sped_referral Issue #8

eudaimonious:
not the best tests in the world but it's something...

### DIFF
--- a/spec/controllers/students_controller_spec.rb
+++ b/spec/controllers/students_controller_spec.rb
@@ -541,4 +541,50 @@ describe StudentsController, :type => :controller do
     end
   end
 
+  describe '#sped_referral' do
+    let(:educator) { FactoryGirl.create(:educator, :admin, school: school) }
+    let(:school) { FactoryGirl.create(:school) }
+    let(:service) { FactoryGirl.create(:service, student: student) }
+    let(:student) { FactoryGirl.create(:student, :with_risk_level, school: school) }
+    let(:student_school_year) { FactoryGirl.create(:student_school_year, student: student) }
+
+    def make_request(options = { student_id: nil, format: :pdf })
+      request.env['HTTPS'] = 'on'
+      get :sped_referral, id: options[:student_id], format: options[:format]
+    end
+
+    context 'when educator is not logged in' do
+      it 'does not render a SPED referral' do
+        make_request({ student_id: student.id, format: :pdf })
+        expect(response.status).to eq 401
+      end
+    end
+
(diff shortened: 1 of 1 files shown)


alexsoble:
@eudaimonious Basic tests are a wonderful start! And covering authorization first is a good idea.

spec/controllers/students_controller_spec.rb
```diff
@@ -541,4 +541,50 @@ def make_request(query)
...
+        make_request({ student_id: student.id, format: :pdf })
+      end
+
+      context 'educator has schoolwide access' do
+
+        it 'is successful' do
+          expect(response).to be_success
+        end
+
+        it 'assigns the student correctly' do
```
erose: What is `assigns`?

spec/controllers/students_controller_spec.rb
```diff
@@ -541,4 +541,50 @@ def make_request(query)
...
   end
 
+  describe '#sped_referral' do
+    let(:educator) { FactoryGirl.create(:educator, :admin, school: school) }
+    let(:school) { FactoryGirl.create(:school) }
+    let(:service) { FactoryGirl.create(:service, student: student) }
+    let(:student) { FactoryGirl.create(:student, :with_risk_level, school: school) }
+    let(:student_school_year) { FactoryGirl.create(:student_school_year, student: student) }
+
+    def make_request(options = { student_id: nil, format: :pdf })
```
erose: So in all these tests, we're making requests to the endpoint, then checking that the response contains something, right?

erose:
I have some code questions that are purely for my own edification. You should feel free to merge this at any time.

erose:
Ok I'm merging it cuz it rules.

spec/controllers/students_controller_spec.rb
eudaimonious: I'm just using this to mean that the data stored in the instance variable matches what I expect based on the data in the factories. Not a great way to test honestly but meh.

spec/controllers/students_controller_spec.rb
eudaimonious: I'm not familiar with RSpec. I copy and pasted my way into this. I'm not confident that RSpec actually makes real HTTP  requests. It's probably just a simulation. To my understanding, these tests will fake a request and check the values of the instance variables passed to the template.

eudaimonious:
Sorry I was slow in responding. I'm all for edification!

PR #504, opened 2016-05-18
include react-modal with compatibility shim

kbweaver:
The `react-modal` component that @erose wants to use expects `window.react` and `window.react-dom` to be defined, but our version of React seems to define them as `window.React` and `window.ReactDOM` instead. This is one solution.

Is there a better way to do this or a better place to put any of these files? If so, let us know!

To test: Load up a page, open the JavaScript console, check that `window.ReactModal` is defined, and verify that there are no JavaScript errors. Some light workflow testing can't hurt either.


jochakovsky:
LGTM

PR #513, opened 2016-05-18
Provide help for notes and services

### LINKED ISSUE #451: Provide user manual directly from product

erose:
Resolves #451

This is a new component, the question-mark bubble, which opens a modal dialog when you click on it. The new component appears in two places: next to Notes and next to Services.

screen shot 2016-05-18 at 4 47 19 pm

Here are pictures of the modal dialog. It's a third-party library from https://github.com/reactjs/react-modal. The content is unstyled HTML, `<div>`, `<li>`, `<p>` etc. tags. There are three ways to close a modal dialog: click on one of the close buttons (bottom left or top right), press Escape, or click outside of its bounds.

screen shot 2016-05-18 at 4 47 12 pm
screen shot 2016-05-18 at 4 47 04 pm

Implementation-wise, a big problem I was trying to solve was where to store the content of the dialogs. The most natural place was to pass it in as a prop, but it's a pain to create & maintain a long multiline string containing HTML in JS. So I considered two options:

0. Set up a Rails path to serve the content as a static page, and grab it with JQuery when the user clicks the button.
0. Store the content in the DOM, and grab it on render.

(Both use React's `dangerouslySetInnerHTML` method. We're not rendering user-provided strings, but the method name is scary for a reason and I'l like some independent eyes to make sure we're not vulnerable to XSS.)

I ended up going with 2. because it had fewer moving parts. (But I still have 1. in a branch). I put the help content at the top of `layouts/application.html.erb`, but that seems a little hacky. Maybe we could use Rails's templating system to have the content be in a separate file but still include it in the DOM.

Before merging this, I'd like to have someone verify that it doesn't leave us vulnerable to XSS.

@alexsoble: I know you liked that `(help me decide)` link. I liked it too! And agree that it is less ignorable. But I took it out because the relevant information was in the question-bubble already, and I couldn't think of what help to provide there. (If you're going to enter a service, you probably know which one it is, do you agree/disagree?)


alexsoble:
Awesome, digging in now!

erose:
You rock.

alexsoble:
Couple of thoughts @erose:

— Love the modal approach. The way it fills the screen suits the text-heavy user manual it delivers.

— On `dangerouslySetInnerHTML`: why don't we just store the content we need to serve right in `HelpBubble.js`? Add functions called `textForNotesHelp` / `textForServicesHelp`, drop in the content.

— I see what you're saying about "help me decide". . . What do you think about replacing the purple question marks next to Notes and Services with prompt text or adding text next to them? The prompt text could be "What should I record?" or just "Help." Reason I'm suggesting this is because I've heard users in Somerville schools ask questions out loud but not know to hover or click on question marks for help.

erose:
Replacing the question marks with text sounds good -- maybe "How should I use this?". I also like "Help".

The reason I had not to store in `HelpBubble.js` was because JS has no support for multiline strings, you have to do:

```
var s = "Once a very lazy brown fox..." +
"and then the story continues..."
```

or

```
var k = "Once a very lazy brown fox... \"
"and then the story continues..."
```

Either way, you have to add something at the end of every line. I figured it would be easier to write + change later if we isolated it into an actual HTML document.

I might be being too focused on tech-y elegance at the expense of other stuff here, though. (I mean, putting it in the DOM is kind of a hack, so maybe I'm not even doing so hot on elegance). Let me know what you think. You could also try this out and see if it's a pain to edit.

Another reason was that I thought it would be nicer not to have to add a separate function for every possible help text, but I was wrong; we don't have to do that -- we could just store them as constants and look up by `content_id`.

alexsoble:
@erose Ah right. I made a little prototype with ES6 this week, maybe my brain was stuck on ES6 mode. I wonder if / when it might make sense for us to use a compiler so we can have multiline strings etc.?

Anyway — back to main thread — agreed that `dangerouslySetInnerHTML` is not dangerous in this case. If we serve it from the DOM, why don't we serve it from `app/views/students/show.html.erb` instead of `application.html.erb`, since it's only used on the student profile page?

erose:
Help text implemented! Feedback appreciated.

screen shot 2016-05-19 at 5 19 24 pm

I actually think I was making too big a deal about the multiline thing yesterday. Sublime makes editing w/the slashes pretty easy, and better to have the have the code be readable & local. So now we store these in JS, in `intervention_details.js`, where they are passed to the HelpBubble component.

alexsoble:
Looks great! Just noticed how easy and natural it is to close the modal too, with those three ways of triggering a close . . . And I think the help text is very clear . . . Merging in!

PR #514, opened 2016-05-19
Show user input in services educator dropdown

awalker4:
Fix for #439. We can remind users that this is a free-text input by displaying whatever they've already typed as an option, even if no other names match.

Spec for this incoming. In the meantime, gifs!

teacher test

teacher austin

Notice that we also auto select the first option when they start typing. I thought it made the process feel a bit smoother. Any thoughts?

### DIFF
--- a/app/assets/javascripts/student_profile/provided_by_educator_dropdown.js
+++ b/app/assets/javascripts/student_profile/provided_by_educator_dropdown.js
@@ -59,7 +59,7 @@
         source: '/educators/services_dropdown/' + this.props.studentId,
         delay: 0,
         minLength: 0,
-        autofocus: true,
+        autoFocus: true,
 
         select: function(event, ui) {
           self.props.onUserDropdownSelect(ui.item.value);
--- a/app/assets/javascripts/student_profile/provided_by_educator_dropdown.js
+++ b/app/assets/javascripts/student_profile/provided_by_educator_dropdown.js
@@ -54,13 +54,32 @@
     componentDidMount: function() {
       var self = this;
 
+      // TODO: We should write a spec for this!
       $(this.refs.ProvidedByEducatorDropdown).autocomplete({
         source: '/educators/services_dropdown/' + this.props.studentId,
         delay: 0,
         minLength: 0,
+        autofocus: true,
+
         select: function(event, ui) {
           self.props.onUserDropdownSelect(ui.item.value);
         },
+
+        // Display what the user is typing first
+        response: function(event, ui) {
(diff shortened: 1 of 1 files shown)


erose:
Woohoo, right from closing your first PR to making your second! AND you closed two issues. Rock on!

Will check this out when I get a chance.

erose:
Checked it out. (Big fan of the gifs, by the way).

0. Tests! Here I would test user actions and then verify that they cause things in the DOM.
0. When I click on the dropdown arrow (without any text already in the box), there's a weird visual effect. If there's text already in the box, no weird visual effect.
0. If a valid educator's name is already in the box, and I click on the dropdown button, I see their name displayed twice (once because it's in the box, once because it's in the DB). Struck me as confusing, but open to debate.

awalker4:
Is the weird visual effect the extra lines here? I have a good idea of how to fix that.

screenshot from 2016-05-21 12-50-24

erose:
Yep, that's the one. (I appear to have forgotten, but you have remembered, that a screenshot is worth a thousand words!)

awalker4:
I agree with your 3rd point. Here's what I'm working with now.

testing
When you've typed something that's already in the dropdown, it will only show once. (Have to demo this with "invalid" data until #518 is fixed).

testing w dropdown
If you finish typing a name but then hit the dropdown anyway, your name will appear once but no guarantees that it will be shown first. Seems like it would be better to move that name to the top but I haven't found a clean way to do that yet.

awalker4:
Awesome, figured it out! Now whatever the user has entered will be at the top of the list with no duplicates

good dropdown

erose:
We just enabled https://devcenter.heroku.com/articles/github-integration-review-apps to spawn a deployment of the app just for this pull request. Then a lot of people can check it out without having to add your remote, pull from your remote, etc.

If you merge master into this branch, you'll have the necessary config files, and I can push a button that will spin up the review app. (Later it might be automatic).

awalker4:
Awesome! Will do.

awalker4:
Sorry I've been silent on this for a few weeks! I can't for the life of me figure out how to get a spec on that dropdown. I know we have some tests that will simulate user clicks and whatnot, but I'm in unfamiliar territory here. Since this is more a visual change than anything, I'm thinking it might make sense to just get this piece merged and then get a test together at some point for the sake of completion. Thoughts?

erose:
@awalker4: Sounds good to me, let's get this merged. Just make an "add tests" issue as well. Merge when ready.

PR #525, opened 2016-05-20
Refactor import code

alexsoble:
The `Importer` class is confusing and adds an extra layer of variable pass-through, this PR aims to transfer its responsibilities to the File Importer classes.


alexsoble:
Refactor complete, tests pass, want to manually QA against staging.

PR #528, opened 2016-05-21
Add ability to edit event notes

kbweaver:
The next part of #228 and building off of the API added in #349, this adds the ability to edit event notes on the student profile page.

In this iteration, the notes appear read-only until you hover over them, at which point a border appears and the cursor changes to afford clicking and selecting text. The user can begin modifying the text inline immediately, without any additional clicks to enter a separate mode. The note is saved on blur.

In addition to running the automated tests, the following test cases should be covered manually, particularly in IE 11:
- Enter all sorts of nonsense into the note. HTML-meaningful characters, high-bit characters, etc. In particular, try copying and pasting content from this site or other websites into the note, and verify that unnecessary HTML is stripped.
- Reload the page and verify that the note remains unchanged visually (the HTML itself may be normalized).
- Inject some HTML into the `event_note` table directly and verify that it is escaped when you load the note.

A summary of the code changes in this PR:
- Hooked up the student profile API to the endpoints added in #349.
- Added the `rails-sanitize-js` gem to provide HTML sanitization.
- Factored the `NoteCard` React class into it's own file, mainly to facilitate unit testing.
- Replaced the `NoteCard`'s text with a `contenteditable` div. Added listeners and styles to support inline editing. Added unit tests for `NoteCard` as well as an additional integration test to the `PageContainer` spec to cover these changes.
- Changed `PageContainer`'s `state.requests.saveNotes` from a string to an object, so that it can track requests by event note ID (or `undefined` if there is no ID yet).


kbweaver:
If anyone has Windows and can test this in IE 11, let me know. I've been testing in a VM and noticed that some of the editing actions are a bit slow, so I'm concerned that the HTML sanitization and conversion to/from text may be too expensive. Seems fine to me in Chrome, and things seem slow all around in my VM, so it may not be a problem, but would love if someone could confirm.

erose:
We just enabled https://devcenter.heroku.com/articles/github-integration-review-apps to spawn a deployment of the app just for this pull request. Then a lot of people at CFB can try it out easily.

If you merge master into this branch, you'll have the necessary config files, and I can push a button that will spin up the review app. (Later it might be automatic).

kbweaver:
Sweet! Thanks @erose ! This branch should now be up to date with master. I was hoping to do some light usability testing on Tuesday; this will make it much easier.

kbweaver:
This seemed to work fine in IE 11 on @sentientmachine 's PC. Some usability feedback was that it's unclear whether your changes have been saved successfully.

kevinrobinson:
@kbweaver Awesome, there's lots of great stuff here! Your thoroughness in describing your work is awesome, and would be a great model to share for how to write a pull request. 👍 

I was a little surprised to see HTML sanitation code in here and accepting HTML user input, which seem like security issues that it would be great to avoid.  If I follow, this is a consequence of using `contenteditable` which accepts HTML, and then I see there's some HTML sanitation code to deal with that.  I know I missed talking about this with you last night, sorry about that, but if you could teach me why this is a better approach than using a `textarea` that'd be awesome.  And maybe just chatting next week is easier too, whatever works :)

And as always, thanks so much for your work on this!

PR #536, opened 2016-05-25
Add ability to delete event notes

kbweaver:
The next step for #228, this adds the ability to delete event notes. The delete is a two-step confirmation as opposed to one-step with the ability to undo, since we anticipate deleting notes will be a rare action. The deleted note is stored as an `EventNoteRevision`.

A summary of the code changes in this PR:
- Added a new delete route to the `EventNotesController` with accompanying tests.
- Hooked up the client API to the new route.
- Added a delete button and new state for delete confirmation to the `NoteCard` React class, along with an integration test.


app/controllers/event_notes_controller.rb
```diff
@@ -30,32 +30,32 @@ def update
...
+
+    unless event_note_revision.save
+      render json: { errors: event_note_revision.errors.full_messages }, status: 422
+    end
+
+    if event_note.update!(event_note_params)
+      render json: serialize_event_note(event_note)
     else
-      version = 1
+      render json: { errors: event_note.errors.full_messages }, status: 422
```
eudaimonious: We'll never reach this else clause because you're using `update!` which raises an exception if the update fails. Instead, you could use `update`.

kbweaver:
Will need to coordinate with @erose on #546 to make sure these play nice with each other (`is_restricted` should probably be stashed in `EventNoteRevision` so that we can restore its state if needed).

app/controllers/event_notes_controller.rb
```diff
@@ -30,32 +30,32 @@ def update
...
-    ).order(
-      :version
-    ).last
-
-    if previous_event_note_revision
-      version = previous_event_note_revision.version + 1
+    event_note_revision = build_revision(event_note)
+
+    unless event_note_revision.save
+      render json: { errors: event_note_revision.errors.full_messages }, status: 422
```
eudaimonious: I think a doublerendererror could happen if the save fails - could do `return render json ...` to avoid

app/controllers/event_notes_controller.rb
```diff
@@ -30,32 +30,32 @@ def update
...
+    end
+
+    if event_note.update!(event_note_params)
```
eudaimonious: may want to wrap the `event_note_revision.save` and `event_note.update` calls in an `ActiveRecord::Base.transaction do` block, if it's problematic to have one record save without the other.

alexsoble:
This is falling deep behind the rest of the codebase and hasn't been updated in more than a year; going to close.

PR #542, opened 2016-05-25
Break out project priorities by category

alexsoble:
Categories: 

+ Expansion
+ New feature 
+ Processes we are supporting

Does this do a good job of capturing what we discussed @erose @kevinrobinson?


kevinrobinson:
@alexsoble looks great to me 👍

PR #544, opened 2016-05-25
Update School Deployment grid

alexsoble:
+ Add emoji to reflect the fact that Uri Harel is in the proess of training / onboarding ESCS principal and admins onto the tool


erose:
😆 at use of the phrase "add emoji to reflect the fact that ...".

Issue #532, opened 2016-05-22
Allow school administrators to designate some notes as 'Restricted'.


erose:
We've concretized the feature in https://github.com/studentinsights/studentinsights/issues/450, so now we should move on to implementing it, I think? Here's Alex's summary of the new feature.

## What are restricted notes?

* "Usually comes from the principal"
* "Data point that's sensitive"
* "Certain health data can not be seen by a teacher"
* "51A"
* "Immigration status changing"
* "Some of that data cannot be put in the system"

## What is the use case / rationale?

"Think of the case of a turnover principal. Jill files a 51A or there's a drug situation. You don't want the teacher necessarily to see that. If Jill doesn't put it in, next year a new admin comes in, a new guidance counselor comes in, and they don't have the information."

## Who can see restricted notes?

* PRINCIPAL
* AP
* GUIDANCE COUNSELORS
* DISTRICT-LEVEL ADMIN

## Can classroom teachers enter restricted notes?

* "No, mostly an administrative thing."
* Uri is thinking about in the future developing PD modules for teachers on how to use Student Insights, would include guidance about what to enter / not enter in the Notes section

## How do we explain when to use a Restricted Note?

* "No hard and fast rules . . ."
* Uri can provide some examples in the User Manual, just like he did for regular non-Restricted Notes

johnrobbjr:
Some sketches for a potential v2 for later reference

![img_20160531_202441](https://cloud.githubusercontent.com/assets/9868315/15694703/17fd2bd8-276e-11e6-9748-099d6f2064a1.jpg)

PR #546, opened 2016-05-30
Add restricted notes

erose:
Fixes https://github.com/studentinsights/studentinsights/issues/532.

screen shot 2016-05-30 at 2 37 04 am

screen shot 2016-05-30 at 2 42 22 am

This is a minimum viable product for the restricted notes feature. It's very rough, particularly code-wise -- there are two new components that are almost entirely copy+pastes from old components.

This adds a `is_restricted` column to EventNote, and a `can_view_restricted_notes` column to Educator. Both default to `false`.

Before I merge this, I want to make sure that the authorization is secure. Please give me any feedback on that. I'd like to write another test to ensure that restricted notes don't appear in the regular Notes feed. It may be best to create a new table entirely for restricted notes.


app/controllers/students_controller.rb
```diff
@@ -39,12 +39,33 @@ def sped_referral
...
+  end
+
   # post
   def event_note # DEPRECATED. Use EventNotesController#create instead.
     clean_params = params.require(:event_note).permit(*[
       :student_id,
       :event_note_type_id,
-      :text
+      :text,
+      :is_restricted
```
kbweaver: I don't think `event_note` is called anymore, and this change would need to be made in `EventNotesController#event_note_params` for saving the value of `is_restricted` to work?

app/assets/javascripts/student_profile/interventions_details.js
```diff
@@ -143,6 +150,22 @@
...
 
+    renderRestrictedNotesButtonIfAppropriate: function(){
+      var self = this;
+
+      if (this.props.currentEducator.can_view_restricted_notes){
+        return dom.button({
+          className: 'btn',
+          style: styles.restrictedNotesButton,
+          onClick: function(){
+            location.href = '/students/' + self.props.student.id + '/restricted_notes';
```
kbweaver: I think you can avoid navigating to another page, and save yourself a lot of repetitive code, by making whether or not you are viewing restricted notes part of the state of the `PageContainer` instead.

kbweaver: Alternatively, you could include restricted notes in the feed for authorized educators, and then filter on the client in `NotesList`. Happy to talk about this more tomorrow!

erose: Commented on this below.

erose:
Thanks umpteenly for the feedback @kbweaver!

So we do want to make the restricted notes section feel like it's a separate part of the UI. The original idea was to not even use the same React components as the regular notes, (but then I would have to reinvent the wheel for taking notes, editing them etc. so I did this).

I agree that it would be much easier to make this part of the state somewhere. I like the idea of including it in the feed (for authorized users) and then the button just filters on it or not. I think originally I was driven by worrying about security to make it a totally separate request, but I think that might be a bit paranoid (if the user goes and inspects the DOM, they'll find it there).

What were we trying to accomplish by making it a separate part of the UI? If we know those things, maybe we can keep them even while doing it this way. These are some things I feel like we were trying to accomplish:
  0. Make restricted notes feel more private to the user by creating a lot of space between restricted notes and regular notes in the user's mental model.
  0. Make it sort of scary to use restricted notes.
  0. Ensure that a principal doesn't accidentally reveal restricted notes while sharing their screen.

It seems to me, like you were saying Kevin, that we don't need literally another page + another Ruby action to accomplish these.

Maybe just dump everything in the same feed, have a button that switches between them, and have it pop up a modal that says "Restricted Notes, scare scare scare"?

@alexsoble: What do you think?

app/controllers/students_controller.rb
erose: Yep, I'm not sure why I thought this worked. Mystery for another day -- changed.

kbweaver:
Thanks for the clarification @erose ! Yeah, I think we can accomplish these things without creating an entirely new page. It sounds like we're trying to prevent mode errors by making the mode really obvious and scary, and we don't necessarily need a new page to do that. We could even do a better job of accomplishing #3 using the same page, e.g., by hiding the restricted notes after some period of inactivity, or preventing the admin from bookmarking the page in restricted mode.

app/controllers/event_notes_controller.rb
```diff
@@ -3,7 +3,7 @@ class EventNotesController < ApplicationController
 
   rescue_from Exceptions::EducatorNotAuthorized, with: :redirect_unauthorized!
 
-  before_action :authorize!, except: [:names]
+  before_action :authorize!, except: [:names] # TODO: Why is this here? :names is an action called on students only, right?
```
alexsoble: You are correct @erose, the `except` clause doesn't belong here. There is no `:names` action for `EventNotesController`, so `except: [:names]` does nothing in this context.

app/assets/javascripts/restricted_notes/restricted-notes-details.js
```diff
@@ -0,0 +1,110 @@
...
+            eventNoteTypesIndex: this.props.eventNoteTypesIndex
+          })
+        )
+      );
+    },
+
+    renderTakeNotesSection: function() {
+      if (this.state.isTakingNotes || this.props.requests.saveNotes !== null) {
+        return createEl(TakeNotes, {
+          nowMoment: moment.utc(), // TODO(kr) thread through
```
alexsoble: We probably don't need this KR TODO here? @erose

kevinrobinson: Yeah the idea was to thread through the current time throughout the app, rather than re-generating in separate components.  It's a minor cleanup thing, probably wasn't worth me adding the TODO in the first place :)

alexsoble:
This looks good to me!

__Authorization:__

I looked at the code and clicked through all the main scenarios in the UI, looks solid @erose. Since `can_view_restricted_notes` defaults to `false`, we can merge this very optimistically. No users will see the red Restricted Notes button and Restricted Notes page until we flip that database setting for authorized users. 

__Separate page:__

I like the separate page approach. In the list of reasons that @erose wrote out, I think a separate page approach serves (1) and (3) very well. For (2), in my mental model of this feature the goal isn't so much making Restricted Notes scary as underlining the difference between them and regular notes. 

I also want to add that the Profile page has quite a lot of information being displayed and recorded already. Moving Restricted Notes to a separate page staunches clutter.

erose:
Feedback from John:

  * red tint on the background in Restricted Notes page.

Issue #498, opened 2016-05-18
Add student's birth date to SPED Referrral


eudaimonious:
This is part of https://github.com/studentinsights/studentinsights/issues/8.
The user has confirmed that we should add the student's birth date to the SPED Referral PDFs. Check out its placement in the mockup at https://github.com/studentinsights/studentinsights/pull/165/files

alexsoble:
There is a dependency here! The dependency / blocker is me working with John Breslin to export that data from Aspen / X2 and import it into Student Insights. I will work to unblock that for you. Hopefully by next CFB! @eudaimonious

PR #551, opened 2016-06-01
Add birth date to sped referral packet and dry up date formatting. Is…

eudaimonious:
…sue #498


app/helpers/application_helper.rb
```diff
@@ -1,5 +1,9 @@
 module ApplicationHelper
 
+  def format_date(date)
+    date.strftime("%m/%d/%Y")
```
eudaimonious: hmmm maybe i should protect against nil dates.

alexsoble:
@eudaimonious I don't think we have any nil birthdates but this is always a good idea!

eudaimonious:
Yea, I'm using that function for some other data as well too. And it's pretty generic so it will be good to handle the main error case.

Issue #484, opened 2016-05-14
Educator select dropdown does not close when focus is lost.


erose:
When I open the dropdown

<img width="555" alt="screen shot 2016-05-14 at 10 39 03 am" src="https://cloud.githubusercontent.com/assets/1514487/15269022/1f036e3a-19c0-11e6-8234-7c9d0c98418f.png">

and click elsewhere (without selecting anyone), I expect it to close to allow me to see what's below it. But it doesn't.

PR #552, opened 2016-06-01
Close edu dropdown

QuinnLee:
#484

out

`autocomplete` is a jquery plugin.  First pass for this ticket, an even on the body is binded when the component is mounted, and unbinded when the component is unmounted.

### DIFF
--- a/app/assets/javascripts/student_profile/provided_by_educator_dropdown.js
+++ b/app/assets/javascripts/student_profile/provided_by_educator_dropdown.js
@@ -20,6 +20,7 @@
 
     renderInput: function () {
       return dom.input({
+        ref: 'ProvidedByEducatorDropdown',
         className: 'ProvidedByEducatorDropdown',
         onChange: this.props.onUserTyping,
         placeholder: 'Last Name, First Name...',
@@ -45,25 +46,31 @@
     },
 
     toggleOpenMenu: function () {
-      $('.ProvidedByEducatorDropdown').autocomplete("search", "");
+      $(this.refs.ProvidedByEducatorDropdown).autocomplete("search", "");
     },
-
+    closeMenu: function () {
+      $(this.refs.ProvidedByEducatorDropdown).autocomplete('close');
+    },
     componentDidMount: function() {
       var self = this;
-      $('.ProvidedByEducatorDropdown').autocomplete({
+
+      $(this.refs.ProvidedByEducatorDropdown).autocomplete({
         source: '/educators/services_dropdown/' + this.props.studentId,
         delay: 0,
         minLength: 0,
         select: function(event, ui) {
           self.props.onUserDropdownSelect(ui.item.value);
+        },
+        open: function() {
+          $('body').bind('click.closeProvidedByEducatorDropdownMenu', self.closeMenu);
+        },
+        close: function() {
+          $('body').unbind('click.closeProvidedByEducatorDropdownMenu', self.closeMenu);
         }
       });
     },
-
     componentWillUnmount: function() {
-      $('.ProvidedByEducatorDropdown').autocomplete('destroy');
+      $(this.refs.ProvidedByEducatorDropdown).autocomplete('destroy');
     }
-
   });
-
 })();


app/assets/javascripts/student_profile/provided_by_educator_dropdown.js
```diff
@@ -47,9 +47,12 @@
     toggleOpenMenu: function () {
       $('.ProvidedByEducatorDropdown').autocomplete("search", "");
     },
-
+    closeMenu: function () {
+      $('.ProvidedByEducatorDropdown').autocomplete('close');
```
kevinrobinson: This is a small improvement, but right now this does a query through the whole DOM with `$` (same for the existing code too).  We can explicitly scope this to the component's DOM node like this: https://facebook.github.io/react/docs/top-level-api.html#reactdom.finddomnode, which makes it impossible to possibly affect other sibling components that match this selector.

app/assets/javascripts/student_profile/provided_by_educator_dropdown.js
```diff
@@ -58,12 +61,13 @@
           self.props.onUserDropdownSelect(ui.item.value);
         }
       });
+
+      $('body').bind('click.closeMenu', this.closeMenu);
```
kevinrobinson: Similarly, this affects the whole page which is okay but not ideal.  One option might be to scope the "close to click" to just this component, or another might be too keep the event on `body` like this, but change the event name to be something more unique and specific to this autocomplete widget, so there's no possibility of collision.

This is probably not a real concern, but in general it'd be great to keep components as componentized as we can. :)

kevinrobinson:
@QuinnLee Nice improvement, and nice to meet you last night!  These comments are all minor, curious to hear what you think, and then let's merge this in!

app/assets/javascripts/student_profile/provided_by_educator_dropdown.js
QuinnLee: I can use the react find DOM api, I wasn't sure how much I could refactor and decided to go for consistency. I'll push some refactors.

app/assets/javascripts/student_profile/provided_by_educator_dropdown.js
QuinnLee: I agree, I thought about implementing something similar to [this]( https://github.com/facebook/react/issues/579), but for simplicity I thought this was ideal.

I could do
``` javascript
  componentDidMount () {
    window.__myapp_container.addEventListener('click', this.closeMenu)
  }

  componentWillUnmount () {
    window.__myapp_container.removeEventListener('click', this.closeMenu)
  }
```

to be more idiomatic react.

QuinnLee:
@kevinrobinson Thanks!  My only concern is refactoring vs. consistent style.  I'm not too familiar with the code base as a whole and I didn't want to step on toes.  But I can refactor the existing code + bug fix.

app/assets/javascripts/student_profile/provided_by_educator_dropdown.js
kevinrobinson: Cool! yeah it's not a major thing at all.

app/assets/javascripts/student_profile/provided_by_educator_dropdown.js
kevinrobinson: Cool, thanks for the link!  I'm figuring you're talking about [this](https://github.com/facebook/react/issues/579#issuecomment-68906902)?  You're originally solution seems perfectly idiomatic to me, maybe you could clarify what you mean and what `__myapp_container` is?  I think you mean that is a reference to the top-level DOM node of the React app, but not quite sure.

If this helps clarify on my end, my suggestions were:
  1. Change from binding on the body to `$(React.findDOMNode(this)).bind('click', this.closeMenu)` which scopes the event, but only handles clicks within this component
  2. To make the event namespace more explicit, maybe something like `$('body').bind('click.closeProvidedByEducatorDropdownMenu', this.closeMenu);`, since with the more verbose event namespace, it's very unlikely to ever collide.

No strong feelings either way!

kevinrobinson:
@QuinnLee No problem!  Yeah that's a great thing to be thinking of :)  Any of those kind of small changes seem totally reasonable, but isolated changes is a good instinct. 👍 

Going forward, I probably should bias more towards just merging and then we can make those kinds of refactoring improvements afterward too.

app/assets/javascripts/student_profile/provided_by_educator_dropdown.js
QuinnLee: Yeah, it basically listens for the  click event outside of the component itself to close.  

I'll implement `('body').bind('click.closeProvidedByEducatorDropdownMenu', this.closeMenu);`

QuinnLee:
I ended up putting a ref on the input dom
and reference it instead of querying the dom

`ref: 'ProvidedByEducatorDropdown',` in the input field

`$(this.refs.ProvidedByEducatorDropdown).autocomplete("search", "");`

kevinrobinson:
@QuinnLee Sorry for the delay here, looks great and thanks for this nice improvement! 👍

PR #554, opened 2016-06-01
Add comment describing PageContainer requests state

kevinrobinson:
From the great conversation with @eudamonius and Kevin in person at Code for Boston.

### DIFF
--- a/app/assets/javascripts/student_profile/page_container.js
+++ b/app/assets/javascripts/student_profile/page_container.js
@@ -54,6 +54,10 @@
 
         // ui
         selectedColumnKey: queryParams.column || 'interventions',
+        
+        // This holds the state of network requests, so that UI components can show that to the user.
+        // These might be null (no requests), 'pending' (request currently in-flight), and 'error' or another
+        // value if the request failed.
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/student_profile/page_container.js
```diff
@@ -54,6 +54,10 @@
 
         // ui
         selectedColumnKey: queryParams.column || 'interventions',
+        
+        // This holds the state of network requests, so that UI components can show that to the user.
+        // These might be null (no requests), 'pending' (request currently in-flight), and 'error' or another
+        // value if the request failed.
         requests: {
```
eudaimonious: This is good. The value of saveNotes and discontinueService will be JSON objects where the key is the note or service id and the value is null or a string?

kevinrobinson: Ahh yeah, thanks!  Added that in too.

alexsoble:
Is this mergeable @kevinrobinson?

kevinrobinson:
@alexsoble 👍

Issue #555, opened 2016-06-01
Admin dashboard for assigning educator roles


alexsoble:
Educators have different roles and levels of permissions that govern which student data they can see. Some of this information is imported from X2, such as educators' school assignments and which are Administrators. 

But there is another level of granularity that isn't captured by X2: which staff are guidance counselors, which are assigned to certain grade levels, which work with English Language Learner students only. That information exists in a CSV and a custom database managed by Somerville's data analyst and IT lead, but we've been told that these are difficult to work with and are being phased out.

Currently I'm hand-rolling these permissions to the database using the rails console based on educator roles that Uri sends by email. Creating an admin dashboard that principals can use to assign educator roles cuts me out as the middleman. Will gives principals the power to keep user roles in their schools up-to-date, especially as teachers enter and leave the District and change roles over time.

I'm interested in finding out if [administrate](https://github.com/thoughtbot/administrate) can help us get this done quickly and elegantly.

alexsoble:
Closed via #558 with the help of administrate!

PR #556, opened 2016-06-01
Upgrade Rails to 4.2

alexsoble:
## Notes

+ This is yak-shaving so that we can use thoughtbot/administrate for #555: http://sethgodin.typepad.com/seths_blog/2005/03/dont_shave_that.html

+ Fix up miscellaneous issues in specs, some of them Rails upgrade-related


alexsoble:
Tests pass, looks good on staging too!

PR #558, opened 2016-06-02
Admin dashboard

alexsoble:
Basic admin dashboard for updating staff permissions, as described in #555.

This uses the [thoughtbot/administrate](https://github.com/thoughtbot/administrate) gem. 

__PROS:__

+ No DSLS make it a joy instead of a pain to customize 
+ Pretty and simple interface, expecting it to be easy for principals to use

__CON:__

+ As noted in the repo, "Administrate is still pre-1.0, and there may be occasional breaking changes to the API." So may incur some upkeep down the line if APIs that we use break on the way to 1.0.

Deployed to staging for some manual QA.


alexsoble:
This is in good shape as far as I can tell.

Stylistically it's not very harmonized with the rest of the app, but that can wait — functionality first.

Would love additional 👀  before merging if possible! 

+ Pull branch to your local machine
+ Log into the demo site
+ Visit `/admin`

If you have staging access, you can check this out on staging as well.

Please try to break things!

Goal is to merge sometime Sunday night if possible so we can demo for district partner on Monday.

kevinrobinson:
I just got to check this out and wanted to say it looks great!

alexsoble:
Aw thanks @kevinrobinson!

PR #559, opened 2016-06-02
Draft helptext for Restricted Notes

alexsoble:
note helptext ars


awalker4:
Your first restricted example is also in the first list. Not sure if that was intentional!

alexsoble:
@awalker4 — Good catch, thank you! I think I fixed in the rebase. Will double check!

alexsoble:
screen shot 2016-06-05 at 2 33 21 am

alexsoble:
Fixed during rebase, thanks for catching though @awalker4!

Issue #560, opened 2016-06-06
Use staff data X2 to determine which users get to view Restricted Notes (as much as possible)


alexsoble:
. . . whatever isn't in X2 can be assigned manually by principals and APs using the admin dashboard.

alexsoble:
At the schools for which we've imported data so far (WSNS, HEA, ESCS), here's what we have in the `staff_type` column from X2:

```
Educator.group(:staff_type).distinct.count

=> {""=>1, "Secretary"=>1, "Teacher"=>5, "Administrator"=>6, nil=>146}
```

alexsoble:
Based on what we know from Uri, documented in https://github.com/studentinsights/studentinsights/issues/532#issue-156153578, staff types that should have Restricted Notes access are:

+ PRINCIPAL
+ AP
+ GUIDANCE COUNSELORS
+ DISTRICT-LEVEL ADMIN 

Principal, AP, and, District-Level Admin are captured by `staff_type == "Administrator"` and normalized in our databases as `admin == true`. 

So we can give those users restricted notes access with a migration (current admins) and a sprinkling of import logic (future admins).

Those users will need to use the admin dashboard to give guidance counselors restricted notes access.

alexsoble:
QA'd this migration against staging with 

```
Educator.group(:can_view_restricted_notes).distinct.count
```

Before migration => `{nil=>147, false=>3}`

After migration => `{nil=>1, true=>6, false=>143}`

The `true` values are all the Principals / Assistant Principals, which is what we want.

The `nil` value was an invalid educator record.

alexsoble:
Closed by #562 and #563. Emailed Uri:

> All the administrators with Student Insights access can view Restricted Notes now.

> X2 doesn't have any data about who the guidance counselors are at HEA / WSNS / ESCS. So those principals will need to give their guidance counselors access once we show them how the admin dashboard works.

> Let me know if you have any questions or if I'm missing anything!

PR #562, opened 2016-06-06
Set restricted notes access for users with a nil value

alexsoble:
## What's up with this migration?

+ what we want: admins ==> `true`, non-admins ==> `false` for `can_view_restricted_notes`
+ original migration sets `default` ==> `false`, but that doesn't effect existing users in database
+ #560

## What's up with force: :cascade in schema.rb?

+ from our friends on stackoverflow: "[What is :cascade in rails schema.rb and where did it come from?](http://stackoverflow.com/questions/27697119/what-is-cascade-in-rails-schema-rb-and-where-did-it-come-from)"
+ tl;dr -- this isn't relevant for our production database right now, because we don't plan on dropping any tables. but if we run into situations (i.e. keeping staging/production parity?) where we need to, this will prevent missing foreign key issues


alexsoble:
cc @erose since this affects the excellent `can_view_restricted_notes` feature you built

PR #565, opened 2016-06-08
Redirect to canonical domain name if Heroku domain is accessed directly

kevinrobinson:
I noticed in Mixpanel that several users were accessing through `somerville-teacher-tool.herokuapp.com`, so seems useful to redirect to `somerville.studentinsights.org` so that we converge to the new domain name over time.


kevinrobinson:
Added this environment variable in the production environment.

alexsoble:
Ooooooooh great catch Kevin. I think those users are Uri and maybe Jill.

😊 

Good fix!

Issue #515, opened 2016-05-19
Call MixPanel with isDemoSite = True instead of isDemoSite = undefined


erose:
Right now the `isDemoSite` parameter we pass to Mixpanel can be one of two values: `false` and `undefined`.

<img width="1000" alt="screen shot 2016-05-19 at 2 53 15 pm" src="https://cloud.githubusercontent.com/assets/1514487/15405794/71ba9ad0-1dd1-11e6-82ea-99248c4444db.png">

I assume `undefined` means "true", but obv. it should actually be `true`.

alexsoble:
`undefined` === Staging site I believe

erose:
Okay, so `DEMO_SITE` should be `false` for staging. I can confirm that the config var `DEMO_SITE` is not set on staging right now, so it makes sense that would be undefined.

I assume that `false` means production right now. Maybe let's report `staging`, `production`, `demo` instead of true or false?

I'm wondering if `undefined` really is only staging. [Mixpanel][1] reports 44 unique `undefined`s in the past few days -- where would we be getting all those unique visitors to staging?

[1]: https://mixpanel.com/report/867931/segmentation/#action:segment,arb_event:PAGE_VISIT,bool_op:and,chart_type:pie,from_date:-29,ms_checked:(false:!t,undefined:!t),ms_values:!(false,undefined),segfilter:!((dropdown_tab_index:0,property:(name:isDemoSite,source:properties,type:boolean),selected_property_type:boolean,type:boolean)),segment_type:boolean,to_date:0,type:unique,unit:day

alexsoble:
> I assume that false means production right now. Maybe let's report staging, production, demo instead of true or false?

👍  👍  👍 

> I'm wondering if undefined really is only staging. Mixpanel reports 44 unique undefineds in the past few days -- where would we be getting all those unique visitors to staging?

Hm. I wonder how Mixpanel defines a unique visitor. I also wonder if some of the undefineds could possibly coming from folks running Student Insights on local machines?

erose:
Closed by https://github.com/studentinsights/studentinsights/commit/fd39d8319028f0990fca61dd86d704c96ab13423.

PR #566, opened 2016-06-08
Mixpanel report script on weekly usage

kevinrobinson:
This queries the MixPanel API to make a report of how many total page visits we have for each school, and how many unique users per week.  This filters out the ldap user, and only considers events from deploys that send analytics events and where the DEPLOYMENT_KEY environment variable is set to `production`.  Might be better to just add this explanation to the output.

The idea here is that weekly active users is a good topline engagement measure, and that the breakdown across schools helps track the rollout and next steps if engagement drops.  This can be something that runs each week and emails out, and can be shared with the Code for Boston group as well, since there's no individual teacher or student data here.

Example output:

```
Healey
---------------------------------------------------
Date			Visits		Users
2016-05-09		13		2
2016-05-16		24		4
2016-05-23		4		2
2016-05-30		13		2
2016-06-06		9		3

West
---------------------------------------------------
Date			Visits		Users
2016-05-09		1		1
2016-05-16		4		1
2016-05-23		31		4
2016-05-30		2		1
2016-06-06		0		0

East
---------------------------------------------------
Date			Visits		Users
2016-05-09		0		0
2016-05-16		22		1
2016-05-23		4		1
2016-05-30		0		0
2016-06-06		19		5

TOTAL
---------------------------------------------------
Date			Visits		Users
2016-05-09		14		3
2016-05-16		50		6
2016-05-23		39		7
2016-05-30		15		3
2016-06-06		28		8
```

Verified manually with web UI data on mixpanel.com.

### DIFF
--- a/app/assets/javascripts/helpers/mixpanel_utils.js
+++ b/app/assets/javascripts/helpers/mixpanel_utils.js
@@ -10,7 +10,6 @@
 
       try {
         window.mixpanel.register({
-          'isDemoSite': Env.isDemoSite, // deprecated
           'deployment_key': Env.deploymentKey,
           'educator_id': currentEducator.id,
           'educator_is_admin': currentEducator.admin,
--- a/app/assets/javascripts/helpers/mixpanel_utils.js
+++ b/app/assets/javascripts/helpers/mixpanel_utils.js
@@ -4,9 +4,11 @@
   var Env = window.shared.Env;
 
   window.shared.MixpanelUtils = {
+    isMixpanelEnabled: function() {
+      return (window.mixpanel && Env.shouldReportAnalytics);
+    },
     registerUser: function(currentEducator) {
-      if (!window.mixpanel) return;
-      if (!Env.shouldReportAnalytics) return;
+      if (!Mixpanel.isMixpanelEnabled()) return;
 
       try {
         window.mixpanel.register({
@@ -21,7 +23,7 @@
       }
     },
     track: function(key, attrs) {
-      if (!window.mixpanel) return;
+      if (!Mixpanel.isMixpanelEnabled()) return;
       try {
         return window.mixpanel.track(key, attrs);
       }
--- a/scripts/mixpanel_report.rb
+++ b/scripts/mixpanel_report.rb
@@ -1,0 +1,122 @@
+require 'date'
+require 'json'
+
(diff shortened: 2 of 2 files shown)


kevinrobinson:
@erose @alexsoble Does this help with what we were talking about yesterday?  Also there was a question of whether data in MixPanel is accurate - I spot checked everything I found and it all looked good but if there's things that we're suspicious about I can look into those in more detail too.

kevinrobinson:
related docs: https://mixpanel.com/docs/api-documentation/data-export-api#events-default

alexsoble:
Yo this is fantastic @kevinrobinson!

alexsoble:
No big concerns but Eli and I noticed some weirdness with the `isDemoSite` variable — shows up in Mixpanel as either `false` or `undefined` instead of `false` and `true`.

erose:
I like this Kevin!

Yeah, we should make sure `isDemoSite` is always set. Would be nice to have `isStagingSite` and `isProductionSite` as well.

kevinrobinson:
Yeah, this is from the evolution of our different environment but I think all is working as it should be now, but confirming that would be good. :)

@alexsoble The value `isDemoSite` is from an older iteration and is [deprecated](https://github.com/studentinsights/studentinsights/blob/c76558b09a95ce4d698df04c381700c878dc4f52/app/assets/javascripts/helpers/mixpanel_utils.js#L13).  I pushed a commit to remove it.

@erose As for the different sites, the other important bit is `SHOULD_REPORT_ANALYTICS` which is only set on the production deployment.  We could add other flags but they wouldn't do anything since those envs don't send data.  Or are you thinking there's value in setting up analytics on the demo and staging sites?

Also I just ran into https://github.com/studentinsights/studentinsights/pull/523 and `app.json` too, so while I think the current setup is doing what we want, these are great questions so thanks for your help verifying!

kevinrobinson:
@erose Tracked down what's going on here.  There was a bug that reported events but without the environment or user data [here](https://github.com/studentinsights/studentinsights/pull/566/files#diff-072548747cd7fb3a296af72268e78700R26).  This didn't impact any of the queries, since those filtered by environments, but led to the `undefined` values like you saw here: https://github.com/studentinsights/studentinsights/issues/515

PR #507, opened 2016-05-18
Add a red button for accessing the sped referral packets from the stu…

eudaimonious:
…dent profile. Issue #502 

sped_button

PR #569, opened 2016-06-15
Clarify the unofficial nature of the sped referral packet per Alex's …

eudaimonious:
…comment on PR #507 @alexsoble 
screen shot 2016-06-14 at 8 11 27 pm


kevinrobinson:
👍 It might even be helpful to make it extra huge, so there's no way someone can miss it.

PR #576, opened 2016-06-15
Not all dates are of Date class. At least one is ActiveSupport::TimeW…

eudaimonious:
…ithZone. Check that whatever the date's class is, that it can respond to the strftime method.

@erose whoops. a bug snuck in with that last PR merge. could we merge the fix please?


erose:
Ya! btw, no need to get approval before merging small stuff like this.

PR #577, opened 2016-06-15
Add rake task for emailing Mixpanel usage report out

kevinrobinson:
This adds a rake task that we can run weekly to query Mixpanel for aggregate usage data and then email it out using Mailgun.  It requires environment variables with those keys in order to work.


erose:
Kevin, anything blocking merging this?

kevinrobinson:
@erose nope, mergin'!

Issue #578, opened 2016-06-15
Genericize `restricted-notes-details.js`.


erose:
Let's merge the `intervention-details.js` and `restricted-notes-details.js` components into one.

PR #584, opened 2016-06-21
Simplify restricted notes

erose:
Fixes https://github.com/studentinsights/studentinsights/issues/578.

Refactoring to simplify the backend and parameterize the `notes_details.js` component, so it can be used for both restricted and regular notes.


erose:
This still needs tests.

erose:
Added JS tests!

spec/javascripts/student_profile/restricted_notes_page_container_spec.js
```diff
@@ -0,0 +1,59 @@
...
+      var mergedProps = merge(props || {}, {
+        nowMomentFn: moment.utc,
+        serializedData: Fixtures.studentProfile,
+      });
+      return ReactDOM.render(createEl(RestrictedNotesPageContainer, mergedProps), el);
+    },
+
+    createMockApi: function(){
+      var mockApi = jasmine.createSpyObj('api', ['saveNotes']);
+      // We need to mock a promise object -- this is a poor man's solution.
```
kevinrobinson: This is a start of a good solution, and these are great tests! :)

Here you have to know the order of calls (`.done` chained with `.fail`) since it's just a stub object.  You can use real Promises here with fake values though, something like:

```
var mockResponse = {id: 781, text: 'hi', is_restricted: true, event_note_type_id: 301, student_id: 23};
mockApi.saveNotes.and.returnValue($.Deferred().resolve(mockResponse);
```

That way code like https://github.com/erose/studentinsights/blob/simplify-restricted-notes/app/assets/javascripts/restricted_notes/restricted_notes_page_container.js#L76 will also get exercised all the way through with real values.

app/assets/javascripts/restricted_notes/restricted_notes_page_container.js
```diff
@@ -0,0 +1,129 @@
...
+      Examples include: \
+      <ul> \
+      <li>"Medicine change for Uri on 4/10. So far slight increase in focus."</li> \
+      <li>"51a filed on 3/21. Waiting determination and follow-up from DCF."</li> \
+      </ul>'
+    },
+
+    render: function() {
+      return dom.div({ className: 'RestrictedNotesPageContainer' },
+        dom.div({ className: 'InterventionsDetails', style: {display: 'flex'} },
```
kevinrobinson: Does this className still make sense here?

app/controllers/students_controller.rb
```diff
@@ -7,18 +7,17 @@ class StudentsController < ApplicationController
 
   def authorize!
     student = Student.find(params[:id])
-    educator = current_educator
-    raise Exceptions::EducatorNotAuthorized unless educator.is_authorized_for_student(student)
+    raise Exceptions::EducatorNotAuthorized unless current_educator.is_authorized_for_student(student)
   end
 
-  def show
+  def show(restricted_notes: false)
```
kevinrobinson: I don't think this will work for Rails controller actions like it would for a normal Ruby method - Rails owns the `params` object and injects it into the method scope.  You can use `params.fetch(:restricted_notes, false)` to do this though.

app/controllers/students_controller.rb
```diff
@@ -30,6 +29,11 @@ def show
     }
   end
 
+  def restricted_notes
+    raise Exceptions::EducatorNotAuthorized unless current_educator.can_view_restricted_notes
```
kevinrobinson: To be safe this should probably also do `raise Exceptions::EducatorNotAuthorized unless current_educator.is_authorized_for_student(student)` to be safe, or get added to `authorize!` although I'm sure that this method is sufficient.

app/controllers/students_controller.rb
```diff
@@ -30,6 +29,11 @@ def show
...
+    show(restricted_notes: true)
```
kevinrobinson: It's up to you but my initial read is that the code in the previous `restricted_notes` method is more direct.  Calling the same queries as the `show` action pulls in a bunch of data that isn't used on the restricted notes page.  This is fine from a performance perspective, but is probably unexpected in terms of "why is there chart data on the restricted notes page?"  And with your great change to add the `restricted_notes` param to `student_feed`, we can use that get what we need in the `restricted_notes` action and that's only a few lines long.

kevinrobinson:
@erose This looks good, awesome work on following all the pieces to make this change!  I hadn't realized that the notes and services pieces were still together in `InterventionDetails`, and splitting them out here is a great improvement.

app/assets/javascripts/restricted_notes/restricted_notes_page_container.js
```diff
@@ -0,0 +1,129 @@
...
+
+    onSaveNotesFail: function(request, status, message) {
+      var saveNotes = {}
+      saveNotes[eventNoteParams.id] = 'pending';
+      this.setState({
+        requests: merge(this.state.requests, {saveNotes: saveNotes})
+      });
+    },
+
+    getNotesHelpContent: function(){
```
kevinrobinson: I just noticed this - if we need HTML from JavaScript we should be using React to create it rather than writing it in strings.  As the comment notes, this creates opportunities for XSS that we can avoid altogether.  I'll open a separate issue to refactor this to React elements and update `HelpBubble`, since that seems like a great starter issue for someone! :)

kevinrobinson: created https://github.com/studentinsights/studentinsights/issues/587 as a dive-in!

app/controllers/students_controller.rb
erose: Does `authorize!` not fire before every action here by default, except `names`? line 6, `before_action :authorize!, except: [:names]`.

app/controllers/students_controller.rb
erose: Yeah, agreed. I think having `restricted_notes` as a parameter on `student_feed` is sufficient without having to thread it all the way through.

app/controllers/students_controller.rb
erose: Hmm. Anyway, moot point now, I removed it.

app/controllers/students_controller.rb
erose: I tested by changing `raise Exceptions::EducatorNotAuthorized unless current_educator.is_authorized_for_student(student)` to `raise Exceptions::EducatorNotAuthorized`. When I tried to visit `http://docker:3000/students/25/restricted_notes`, I was redirected to the NotAuthorized page.

erose:
Added another `rspec` test that ensures an educator can't access a student's restricted notes when they don't have access to the student.

app/controllers/students_controller.rb
kevinrobinson: ah yeah I'm totally mixed up, my bad :)

app/controllers/students_controller.rb
kevinrobinson: sweet!

alexsoble:
🎉  🎉

alexsoble:
Nice specs too @erose!

Issue #585, opened 2016-06-22
Design - Create User Interview Guide


johnrobbjr:
Create a document to guide user interviews to answer any questions we may have.

Questions should aim to extract user behaviors while using the application, and eventually drive prioritization of issues and design decisions down the road.

iamctodd:
I'm happy to give this a review once you've got a first version

alexsoble:
Was this issue closed by #586 @johnrobbjr?

johnrobbjr:
@alexsoble not yet, that pull request was for adding the document, but the full interview guide has not been written yet. I hope to get to it soon once life gets a bit less hectic!

alexsoble:
This issue seems dormant, closing.

PR #586, opened 2016-06-22
Adding markdown user interview guide document to Docs folder

johnrobbjr:
Working on research questions and user interview questions outlined in issue #585. The document is in-progress, feedback is encouraged!

@kevinrobinson These might come in handy when the fall comes back around, it's something I'm going to work on over the coming weeks. Might be good to get your opinion on what research questions are important.

### DIFF
--- a/docs/project/user_interview_guide.md
+++ b/docs/project/user_interview_guide.md
@@ -1,0 +1,38 @@
+# Guide for User Interviews
+
+## Description
+Questions to use during a face-to-face user interview to extract user behaviors and attitudes about SST meetings and studentinsights.
+
+## Team Motivations
+* What are the things we should focus on developing?
+* What are goals we can look to accomplish in the long-term?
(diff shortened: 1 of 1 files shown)


kevinrobinson:
@johnrobbjr This is awesome, thanks for sharing this!

These sound like great starting points, and maybe a way we could do this next is for @alexsoble @erose and I to answer these as best we can to capture past observations and conversations.  It might be useful for "learning by doing" and seeing how you're thinking about conducting the interview too.  Maybe that could be for 30min at Code for Boston?  That way we're ready to roll for the fall too. :)

erose:
Merge whenever you're ready, @johnrobbjr!

johnrobbjr:
@kevinrobinson I'm up for that! I'd like to frame some of these research questions as looking forward for partnering with new schools, so having a discussion with you guys would be great to frame those questions as well.

Issue #587, opened 2016-06-22
Refactor HelpBubble and getNotesHelpContent to use React elements instead of strings


kevinrobinson:
The HelpBubble React component takes a string of HTML as an argument, which we should avoid since it introduces potential for security exploits like passing in a `script` tag.  In the current codebase there's no vulnerability, since we're writing the HTML in code ourselves but this would be good to improve.

HelpBubble: https://github.com/studentinsights/studentinsights/blob/master/app/assets/javascripts/student_profile/help_bubble.js#L10

This issue also involved updating call sites to create React elements rather than strings of HTML.

Issue #583, opened 2016-06-21
Long help text screws up placement of `close` button on bottom left.


erose:
<img width="418" alt="screen shot 2016-06-21 at 3 20 11 pm" src="https://cloud.githubusercontent.com/assets/1514487/16243105/b405310a-37c3-11e6-85fc-248808e4da37.png">

PR #595, opened 2016-06-28
Fix help

### LINKED ISSUE #594: "What is this?" for Notes is blank

erose:
Fixes #594, #587, #583.


erose:
I don't know CSS, so would love someone to look at my CSS!

Jrafalak:
I'm not an expert in CSS, but it looks good to me. [Here](https://css-tricks.com/currentcolor/) is the link we were looking at about currentColor at CFB.

app/assets/javascripts/student_profile/help_bubble.js
```diff
@@ -7,11 +7,13 @@
   var HelpBubble = window.shared.HelpBubble = React.createClass({
     propTypes: {
       title: React.PropTypes.string.isRequired, // e.g. 'What is a Note?'
-      content: React.PropTypes.string.isRequired, // HTML string which will be rendered in the modal's box.
+      content: React.PropTypes.object.isRequired, // React DOM objects which will be displayed in the modal text box.
```
kevinrobinson: FYI, you can use `React.PropTypes.element` to indicate it has to be a React element like your good comment says :)

kevinrobinson:
👍 this css looks like it's doing the trick, so I think you know more than you realize :)

PR #597, opened 2016-06-28
Remove intervention types index

erose:
Please don't merge this, I want to see if we should actually remove this.


kevinrobinson:
@erose Nice cleanup, looks great!  👍  I also don't think this is used anywhere anymore, but probably good to sanity-check things after deploying.

Issue #593, opened 2016-06-27
Implement bar chart sparklines for Attendance & Behavior column


erose:
In the Attendance & Behavior column, we currently use line-graph sparklines:
<img width="244" alt="screen shot 2016-06-27 at 3 00 16 pm" src="https://cloud.githubusercontent.com/assets/1514487/16392032/dd337358-3c77-11e6-817c-5b448014454a.png">

But line graphs don't really make sense for the data, which is cumulative per year. We already changed the main graphs to bar charts for this reason, but now we need to change the sparklines in the columns to bar charts as well.

Seems like they should look like the other sparklines, which are implemented in D3. But we could think about other options too.

alexsoble:
This sounds great @erose!

One thought here: for the sparklines, I think some sense of scale for the x-axis will be important. Do the bar chart trends represent the last 6 months, or school year, or the student's entire career ... ? Without some labeling or context on the x-axis it would be hard for the user to know.

erose:
@alexsoble: I agree. Same situation as for the existing line chart sparklines, right? So maybe we should look at adding context to those as well.

<img width="224" alt="screen shot 2016-06-28 at 3 01 10 pm" src="https://cloud.githubusercontent.com/assets/1514487/16428721/268e2880-3d41-11e6-9dba-70b422541552.png">

PR #602, opened 2016-06-29
Bar chart sparklines

erose:
Resolves https://github.com/studentinsights/studentinsights/issues/593.

First pass. Design feedback wanted!

screen shot 2016-06-29 at 4 52 45 pm
screen shot 2016-06-29 at 4 52 51 pm


awalker4:
My eyes keep expecting those horizontal lines to be y=0 and then I get pretty confused when it looks like everything is negative. How would it look with a solid line on the bottom so those rectangles are a bit more grounded?

erose:
@awalker4: I agree, here's how it looks now.
screen shot 2016-06-30 at 1 45 43 pm

alexsoble:
@erose What are the dividing units on the X-axis?

erose:
@alexsoble: The two big vertical lines? Those are starts of school years. (August 15th 2015, August 15th 2014).

I don't think that's clear. Should we try to indicate it better, or just lose them? (since this is just a sparkline not a full graph).

kevinrobinson:
@erose Nice!  This seems like a great improvement to me, and the suggestions of clarifying the time sound 👍 too.

alexsoble:
@erose I think some way to indicate that is important. Maybe a mini-legend at the bottom?

alexsoble:
Hey @erose. I think this is excellent work, I apologize if my suggestions/comments about the X-axis have slowed us down from merging it. Why don't we merge it in now (solid improvement from the line graphs) and we can think more about the X-axis labeling as a next step?

erose:
@alexsoble: Dude, that's not on you! I agree, (merged), I'll think about it more + iterate more when I see it in production.

alexsoble:
Pushed to production! These look nice!

Issue #424, opened 2016-04-25
Share Tablesort code between School Overview and Homeroom Roster


alexsoble:
Several excellent contributors have written code to make the table sorting work better in Student Insights: @dsbader, @brisandbeans, and @ronny2205 in particular! 

In #419 I organized the Tablesort extensions into a single folder under `javascripts`. Some of the extensions are being used only on the School Overview page, others on the Homeroom Roster page. Both pages have tables of students that are very similar and should share this code.

With #422, we now have some duplication: same table sorting feature, two different extensions, being used on the two different pages.

Would be much better to have one extension per table column, and have all extensions shared across both pages.

If I tagged you above (i.e. @dsbader @brisandbeans @ronny2205), this might be a great issue for you to work on! 

Code for Boston tomorrow (Tuesday) night as usual.  :0)

kevinrobinson:
Close https://github.com/studentinsights/studentinsights/pull/604

PR #604, opened 2016-07-06
Share tablesort code

bdougsand:
Addresses issue #424: share more column sort functions between the homeroom list and the school roster views.


erose:
Hi Brian, I spun up a [review app](https://devcenter.heroku.com/articles/github-integration-review-apps) for this PR. ("View deployment"). I'll do some manual testing when I get a chance, but you can test now too.

brisandbeans:
Thanks so much for doing that!

The code appears to work, in the sense that it does what my comments say it should. My remaining concerns are that 1) the sorts conform to user expectations, and that 2) I've actually caught all the places where sort methods can be shared between the homeroom and roster views.

I did eventually "fix" my Docker issue by reverting to 1.12 RC 2. It looks like I [wasn't the only one](https://github.com/docker/compose/issues/3685) to be hit by the issue in RC3.

PR #607, opened 2016-07-28
Turn on data import for all schools

alexsoble:


alexsoble:
There are four homerooms with no name coming through in the import process: one at ESCS, one at KDY, one at AFAS, one at WHCS.

Issue #611, opened 2016-08-11
Remove the SchoolYear and StudentSchoolYear models


alexsoble:
These models add a lot of complexity to the data structure and the spec. Pre-computing which school year each event falls into does save us some work on the client side . . . but not _THAT_ much work. In the very beginning of the project almost all the data visualized was organized by school year, so precomputing and storing these values made a lot of sense. As the project has evolved to deliver more and more sophisticated visualizations to users, fewer of the use cases involve sorting events by school year. So these models are doing less for us. Eventually I see removing them from the codebase and computing school years on the client side as good steps to take.

alexsoble:
Tackling this now because new tooling is showing that the import process is failing sometimes. When it fails, here's the error message:

#### `execution expired (Delayed::Worker.max_run_time is only 14400 seconds)`

Don't have data to back this up, but looking at the attendance import code, each row triggers more than one ActiveRecord query: one query to look up the student in the row, and then more queries to locate the appropriate student school year to associate the record with.

I want to find out of removing the student school year lookups will increase the speed of the import process.

Rough plan:

1. Add association between absences <> students, tardies <> students, discipline incidents <> students so that there's a second path that doesn't involve the student school year. Run the import process in prod so that all event-type objects are associated with both student and student school year.
2. Change feature side code so that student school year isn't called anymore. 
3. Remove code that associates event objects to student school year from import process. Test to see if import processes run faster on average than they did before. (Right now some import processes time out and some succeed so there's some non-determinism involved.)
4. Drop student school year and school year tables; remove models.

Issue #610, opened 2016-08-11
All of the specs are very broken


alexsoble:
. . . but the badge in the repo says green! Very deceptive! 

1. Figure out why the badge is showing green and make it show red.
2. Fix the specs.

alexsoble:
Hmmmm, can't find a way to trigger a Travis build against a specific branch or commit so re-triggering a build against the most recent merged-in PR.

alexsoble:
OK cool! Badge is red now. Time to green-ify it. Looks like something is going wrong with our factories for `SchoolYear` and `StudentSchoolYear`.

alexsoble:
Closing in favor of the more specific #613.

PR #612, opened 2016-08-11
Fix the specs related to StudentSchoolYear factory

alexsoble:
# Notes

+ Part of fixing up #610.
+ You can see why this is making me think about #611.  ;0)
+ Seems like FactoryGirl-created `StudentSchoolYear` objects were trying to create `SchoolYear` objects that had already been created earlier in the FactoryGirl model-minting process, triggering validation errors in the specs.
+ Not quite sure why this FactoryGirl behavior has changed . . . maybe a gem upgrade?!? Will try to investigate later when if I have time; that said, not crucial because this seems to be an issue with the spec code only, not the feature code.
+ This PR removes the StudentSchoolYear factory.
+ Still some failing specs left in #610.

### DIFF
--- a/spec/controllers/students_controller_spec.rb
+++ b/spec/controllers/students_controller_spec.rb
@@ -92,7 +92,7 @@ describe StudentsController, :type => :controller do
           let!(:more_recent_incident) {
             FactoryGirl.create(
               :discipline_incident,
-              student_school_year: most_recent_school_year,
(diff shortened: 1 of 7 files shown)


alexsoble:
Took a quick look at the key models involved (`Student`, `SchoolYear`, `StudentSchoolYear`), no recent changes to any relevant methods or attributes.

PR #617, opened 2016-08-17
Fix symbol that's supposed to be a boolean

alexsoble:
# Note

+ This mistake is a little embarrassing and is causing the progress bar to run in production even though it should be off in production


alexsoble:
Throwback to when I made this mistake: 13b18c23c5044eeef26e26ae5d2561f1507fdd7f

PR #619, opened 2016-08-19
Districtwide access panel

### LINKED ISSUE #338: Build out views for district-wide administrators

alexsoble:
+ Resolve #338


kevinrobinson:
@alexsoble awesome! 🎉

Issue #622, opened 2016-08-21
Fix Student Assessment results that appear off by 1 day in the UI


alexsoble:
```
# Find most recently taken DIBELS student assessment:
dibels = Assessment.find_by_family('DIBELS').student_assessments.order(:date_taken).last

# Check date taken
dibels.date_taken
=> Fri, 01 Jan 2016

# Look up student
dibels.student

# 2016 DIBELS result not appearing on student profile! # 
```

alexsoble:
The record is in the database, just not making it to the UI. Going to throw it in a bunch of debugging statements and see why this isn't showing up.

alexsoble:
```
puts student.dibels.order(:date_taken).pluck(:date_taken).reverse

=>
2016-01-01 00:00:00 UTC
2015-10-01 00:00:00 UTC
2015-05-01 00:00:00 UTC
2015-01-01 00:00:00 UTC
2014-10-01 00:00:00 UTC
```

alexsoble:
Dates are showing up in the UI off by 1 day. 

For example:

```
# db
2016-01-01 00:00:00 UTC

# UI
December 31st, 2015
```

alexsoble:
# The Problem

```
# js

moment("2016-01-01T00:00:00.000Z").toDate()
=> Thu Dec 31 2015 18:00:00 GMT-0600 (CST)

moment("2016-01-01").toDate()
=> Fri Jan 01 2016 00:00:00 GMT-0600 (CST)
```

alexsoble:
This commit fixes the issue for DIBELS: 3e9442b88f25af2115158558a0fd782730fcdff4

Next steps are:

(A) See if this issue is affecting other types of assessments 

and

(B) Plan out a more generalized solution if needed. 

Not sure right now if it's best to handle this client-side or server-side.

alexsoble:
This issue exists for STAR assessments in the student Full Case History (but interestingly not the line graphs).

alexsoble:
This issue exists for MCAS assessments in the student Full Case History (but also not in the line graphs.)

alexsoble:
Checking to see how JS date is computed for the line graphs as opposed to the Full Case History on the Student Profile . . .

alexsoble:
The dates that look right are coming from serialized_data/chart_data.

alexsoble:
Profile dates STAR, MCAS, etc are being computed using the `QuadConverter`; this is improperly returning dates that are one day early. 

```
// js 

[...]

PR #635, opened 2016-08-22
Fix incorrect dates in profile_details.js

alexsoble:
# Notes

+ Start by making the spec describe the incorrect dates we're seeing in #622

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ branches:
 #   - export PATH=$PATH:$HOME/.local/bin
 
 before_install:
+  - export TZ=America/New_York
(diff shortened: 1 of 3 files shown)


alexsoble:
Note to self: 8a661bfb9562fcd9e5131f1741b44aa7acf1bd47 fails on my local machine, succeeds in the cloud. Makes sense that there could be machine/cloud discrepancies since this is a timezone issue.

alexsoble:
Travis is now showing RED, just like we want it to.

alexsoble:
Travis now showing GREEN because we're forcing UTC in the date display.

PR #637, opened 2016-08-23
Adds eslint container (not in the build), for linting JS code

kevinrobinson:
This adds [eslint](http://eslint.org/) with some initial setup so we can use it for linting JS code and helping developers on the team with catching common errors like unused variables, missing propTypes in a React component, etc.  We could also use it to enforce other conventions like spacing, etc. if so desired.  This adds a new container so that this can be run locally and folks can try it out, and demonstrates cleaning up the linting errors in a few files as an example.

Usage:
`$ docker-compose build tools`
`$ docker-compose run tools`
screen shot 2016-08-23 at 8 09 51 am

There are about 1000 mechanical errors on various files, and an interested or enterprising pal from Code for Boston could learn about build tooling and help improve code quality if they were interested in working through a handful of files in a night.  All told, this could take one volunteer a few nights to burn through the whole codebase.  At that point we could add it to the build process and roll it out to everyone.

If anyone is interested, feel free to ping me with questions!
cc @alexsoble @erose FYI or if you have thoughts or feels

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,40 @@
+{
+  "env": {
+    "browser": true
+  },
+  "plugins": [
+    "react"
+  ],
+  "extends": [
+    "eslint:recommended",
+    "plugin:react/recommended"
+  ],
+  "rules": {
+    "indent": ["error", 2],
+    "semi": ["error", "always"],
+    "comma-dangle": ["error", "only-multiline"],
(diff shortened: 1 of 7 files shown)


alexsoble:
Oooooh thanks @kevinrobinson! Excited to check this out!

Issue #627, opened 2016-08-21
Slow Healey overview page load


alexsoble:
Bug report from @erose!

kevinrobinson:
I can't reproduce this now for any school page (using the sweet new district admin links).  It doesn't appear we retain metrics or logs long enough back to find out more information here.  This could definitely happen when requesting before the precompute job finishes, but since that happens in the wee hours of the morning it's unlikely to affect school users.

Logentries has some data showing a lot of variability in response times, with some responses taking >30seconds and timing out, and others completing in hundreds of ms.  It's just a guess, but it seems like something else is going on here.  I think in order to investigate further it'd help to set up more consistent traffic, maybe something like background jobs that query various pages every minute, so we can get a more consistent read on what response times are.  I took a stab at this here: https://github.com/studentinsights/studentinsights/pull/638

@alexsoble @erose Que piensas amigos?

alexsoble:
@erose @kevinrobinson Kevin, it's possible you can't reproduce this anymore because I added the eager loads to the district admin links after Eli's QA: aa51005cb3816bb08667d79d99a52471446395ce

kevinrobinson:
Here's an update.

I ran the pinger script in a few different combinations this morning, and moved it to https://github.com/studentinsights/studentinsights-pinger and deployed it https://dashboard.heroku.com/apps/studentinsights-pinger (private).  There's notes in chat.

I also investigated the memory errors we've been seeing.  These seem particularly relevant for understanding and improving response latency since these error message mean the Heroku runtime is spilling from RAM to disk (to avoid running out of memory and crashing the process).  More info here: https://devcenter.heroku.com/articles/ruby-memory-use

First I tried out https://github.com/schneems/derailed_benchmarks on the recommendation of the Heroku docs.  I just ran the static analysis which analyzes how much memory it takes to just require the project's gem dependencies.  There was nothing unexpected there (eg., Rails takes the most memory, it takes about 43MB to require all the gems).  Output is [here](https://gist.github.com/kevinrobinson/3b43139a14bb30be33a8e87d3f56172b) and [here](https://gist.github.com/kevinrobinson/20e5997d9299e36dd1eee0a55bd66de0).  There are other tools in derailed, but they require running the app in production mode so I didn't try them yet.

[...]

PR #638, opened 2016-08-23
Add pinger script

kevinrobinson:
This is on top of https://github.com/studentinsights/studentinsights/pull/637, which should be merged first.

This uses [CasperJS](http://casperjs.org/), a wrapper of utilities around PhantomJS to sign in and request various pages from Student Insights.  The idea is we could run this every 5min or so to get more data about response time variability related to: https://github.com/studentinsights/studentinsights/issues/627

@alexsoble @erose I'm not sure this is the best idea, but seems like it might help.  Any ideas or thoughts?


alexsoble:
So just to re-phrase in my own words . . . This is a tool that will hit the production app with real requests so that we get a regular, consistent stream of data about page load times.

That sounds really useful, my reaction is 👍 .

kevinrobinson:
@alexsoble Yep, the downside is it makes the server do more work, and maybe drowns out other real activity in the logs.  But I'm not sure how else to investigate what looked like really large variance in response times so it seems like it could give a useful data point.

alexsoble:
@kevinrobinson We could set it to run once per day instead of every 5 minutes to reduce the server-work and the log-drowning.

alexsoble:
@kevinrobinson Also, any reason not to merge this?

kevinrobinson:
@alexsoble Yep, once a day would reduce the load but won't generate more useful data points about performance variability.  I think we'd need to run it regularly for it to be useful for that purpose, but hopefully the site can take a few requests every five minutes just fine :)

I'll merge, but this isn't running anywhere yet.

PR #645, opened 2016-08-25
Remove pinger tool, split out studentinsights-pinger

kevinrobinson:
Removing the work from https://github.com/studentinsights/studentinsights/pull/638, since it's now moved to https://github.com/studentinsights/studentinsights-pinger and a separate Heroku app.


kevinrobinson:
FYI @alexsoble

Issue #647, opened 2016-08-25
Admin Dashboard: display true/false as Yes/No


alexsoble:
Feedback from Uri – displaying as Yes/No will make using the Admin Dashboard easier for many users.

PR #649, opened 2016-08-25
Clean up admin dashboard

### LINKED ISSUE #646: Remove New Educator button from Admin Dashboard
The sole source of Educator data is X2!

erose:
Fixes #646, #647.

@alexsoble I added another custom field, since I couldn't find an easier way to do this in Administrate. (is there full-on API documentation for this? all I found was http://administrate-prototype.herokuapp.com/)

Also changed the way we refer to custom fields, since `Field::CustomFieldName` doesn't work if you refer to it [more than once](https://github.com/thoughtbot/administrate/issues/584). The correct way is just `CustomFieldName`. No idea why it even works in the first place.


alexsoble:
This is looking great @erose! Thanks for tackling this! And yeah the Heroku documentation site for Administrate is the only one I've seen.

PR #651, opened 2016-08-25
Trying scout gem

kevinrobinson:
From recommendation here: https://devcenter.heroku.com/articles/ruby-memory-use#too-much-memory-used-at-runtime


kevinrobinson:
selfie

PR #652, opened 2016-08-26
Experiment with OJ serialization effect on memory usage

kevinrobinson:
Based on this from Scout:

screen shot 2016-08-25 at 8 22 23 pm


kevinrobinson:
selfie, going to try and will revert immediately on any problems

PR #654, opened 2016-08-26
Allow toggling Scout from env variable

kevinrobinson:
From http://help.apm.scoutapp.com/#scout-apm-help-docs


kevinrobinson:
selfie

Issue #656, opened 2016-08-26
All absences from students in production are being grouped into the current day.


erose:
This is in the client, not the data. I think the format of `created_at` changed? Investigating...

kevinrobinson:
@erose This looks good to me, following the links in the Rollbar dashboard: https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/  Can you confirm when you get a chance?

erose:
Looks good!

PR #657, opened 2016-08-26
Using as_json for now just in students/show.html.erb.

erose:
Fixes https://github.com/studentinsights/studentinsights/issues/656.

  * I noticed that student absence + tardy graphs were broken in production. All the data from the past was being rendered as though it happened today.
  * I noticed also in production that clicking on the Overview card didn't work and caused a JS type error. (https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/20/)
  * I checked the DOM and found that the date keys in `serialized_data` (e.g. `created_at`, `updated_at`) had serialized Ruby objects as their values, instead of date strings.
  * I flipped a shit.
  * I replicated this locally, and then experimented with serializing to JSON at a few different points. Eventually I put it in the view layer because the existing tests were expecting the controller layer to return Ruby objects.
  * This fixed both issues and left the specs working.

I'm wondering:
  * how was serialized_data getting serialized before? Is there some default Ruby serialization that the view template does?
  * what changed that caused this to be an issue recently?
  * if calling `as_json` in the view layer is the thing to do, shouldn't we do it everywhere else we serialize data as well? (In `students/restricted_notes.html.erb` and `shared/serialized_data.html.erb`).

### DIFF
--- a/app/views/students/show.html.erb
+++ b/app/views/students/show.html.erb
@@ -1,2 +1,2 @@
 <div id="main"></div>
-<%= tag("div", id: "serialized-data", data: @serialized_data) %>
+<%= tag("div", id: "serialized-data", data: @serialized_data.as_json) %>


kevinrobinson:
@erose This is my mistake, from https://github.com/studentinsights/studentinsights/pull/652 last night, which swapped in a different serialization library.  Thanks for jumping right on this, and if calling `as_json` fixes the immediate problem I'd 🚢 this now.

I'll look more at this in about 30 minutes since it likely affects other pages too.  And apologies for the table-flipping moments here, this is my bad.

erose:
@kevinrobinson: Oh, makes sense! I doubt this was really a big problem, I just freaked out at the time. (my lack of understanding about how serialization happens didn't help).

So there's probably a more nuanced way to do this with `OJ`, I guess this method is bypassing it entirely?

alexsoble:
Thank you for catching and fixing @erose!

kevinrobinson:
@erose Yeah, I'm going to revert https://github.com/studentinsights/studentinsights/pull/652 now instead.  If that doesn't work in the next 10 minutes then I'll merge this to resolve and fix forward from this PR.  I'll be online and in chat for the next hour or so.

kevinrobinson:
nice @alexsoble :)  Also I should have said this up front, but awesome debugging to track that down and write it out so clearly too @erose! 👍

PR #660, opened 2016-08-26
Update Mixpanel report to include email job

kevinrobinson:
Also updates the way Mailgun env variables are read to match the Heroku add-on versus my initial local testing.


kevinrobinson:
Usage is like this:

```
# to send emails:
bundle exec rake report:email_mixpanel_usage

# to print report to stdout:
bundle exec rake report:mixpanel_usage
```

PR #661, opened 2016-08-26
Add MixPanel page view tracking for the student profile page

kevinrobinson:
This wasn't instrumented it seems, probably my mistake.


kevinrobinson:
selfie

PR #664, opened 2016-08-28
Make link behavior more sensible on roster page

alexsoble:
# Notes 

+ Take user to student profile when they click anywhere on a student row in the roster page, not just the student name.
+ The roster page UI doesn't indicate to the user in any way that clicking on the student name will create any sort of special outcome. The hover effect applies to the entire row. So make the behavior (opening student profile) correspond to the entire row too.


alexsoble:
Tested this out in staging, way more intuitive.

PR #672, opened 2016-08-29
Add UTC parsing to cumulativeByMonthFromEvents

kevinrobinson:
Addressing https://travis-ci.org/studentinsights/studentinsights/builds/155916314, originally from https://github.com/studentinsights/studentinsights/pull/602.


kevinrobinson:
@erose trying to help fix the test failure, which looks like a real edge case failure.

PR #679, opened 2016-08-29
Switched out gems and config files, modified Procfile.

erose:
@alexsoble @kevinrobinson: Puma changes. I took out configuration stuff related to workers because we're not using any, correct?


kevinrobinson:
@erose Awesome!  Two thoughts:

1. The [Heroku docs about Puma](https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server) are good and describe how the workers and threads work.  There's always worker processes and I'd suggest starting with two workers, set using the WEB_CONCURRENCY env variable (so you can tune it without having to make code changes).
2. I'd suggest doing a deploy to the staging environment first :)

PR #705, opened 2016-09-23
Remove leftover reference to RosterChart

kevinrobinson:
Addressing https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/26/

This is affecting users now but with a minimal impact now (empty table cells aren't replaced with em dashes).


alexsoble:
Oooooh thank you for catching this @kevinrobinson!

PR #713, opened 2016-09-30
Workaround for CSV download in IE

kevinrobinson:
Related to https://github.com/studentinsights/studentinsights/pull/712/files and https://github.com/studentinsights/studentinsights/issues/711.

@alexsoble From your great links in the issue, trying this on the demo site.

PR #715, opened 2016-09-30
Thread through CSV filename for IE instead of picking it off DOM element

kevinrobinson:
https://github.com/studentinsights/studentinsights/pull/713 looks good, but the filename doesn't work when trying to read it off the A element.

screen shot 2016-09-29 at 11 01 30 pm


kevinrobinson:
👍

Issue #701, opened 2016-09-20
Include descriptions about what the DIBELS performance levels mean


alexsoble:
Include descriptions about what the DIBELS performance levels mean. (i.e. what does "STRATEGIC" mean?) Include raw integer scores for DIBELS along with performance levels. That could let educators see DIBELS change over time.

pauledwardmitchell:
It looks like how the raw integer scores for DIBELS correspond to the three intervention categories changes as students progress through the grades... (page 4 [here](https://dibels.uoregon.edu/docs/DIBELSNextFormerBenchmarkGoals.pdf)) ... not sure how to reflect this in this feature @alexsoble

kevinrobinson:
this is stale now, but for individual assessments this is visible first to project leads as thresholds, and part of work on the reader profile

PR #745, opened 2016-10-19
Dibels info window

pauledwardmitchell:
@alexsoble - I worked on [this card (701)](https://github.com/studentinsights/studentinsights/issues/701) from the Healey teacher feedback.


alexsoble:
Hi @pauledwardmitchell! Thanks for tackling this issue!

I'm going to add a few screenshots of your work right here, using the fake demo data:

screen shot 2016-10-19 at 6 12 34 pm

screen shot 2016-10-19 at 6 12 41 pm

Content-wise, I'm going to email Uri about this, cc you, and see about filling in those blanks on the "one of the teachers should probably write this part!" section.

Style-wise, I think the Help Bubble should be adjacent to the DIBELS score instead of below it, do you agree?

pauledwardmitchell:
Hey @alexsoble - 

I'll work with Uri's feedback - and getting the bubble to show adjacent to the score.  

Would you all prefer something like the help box (seen in pic below) instead of the big help bubble?

screen shot 2016-10-20 at 6 55 58 pm

I'll also circle back on why Travis is failing...

PR #754, opened 2016-10-21
WIP – Serve school overview JSON from endpoint

alexsoble:
... instead of printing it all out on the page.

We'll see if this fixes our memory errors!

### DIFF
--- a/app/assets/javascripts/school_overview/main.js
+++ b/app/assets/javascripts/school_overview/main.js
@@ -8,12 +8,14 @@ $(function() {
     var createEl = window.shared.ReactHelpers.createEl;
 
     function main() {
+      var schoolId = $('#school-id').data('id');
+
       $.ajax({
         url: window.location.href + '/serialized_data',
         method: 'GET',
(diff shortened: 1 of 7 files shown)


alexsoble:
Pushed this work to the staging site. Pages appear to load *much* more slowly with the JSON endpoint route as opposed to the printing-JSON-on-page route. Possibly caused by a lack of precomputing on the staging site?

alexsoble:
Get this error when running the precompute task in staging:

```
> PrecomputeStudentHashesJob.new(STDOUT).precompute_all!(Time.now)
Querying for jobs...
. . . 
Found 6 jobs.
Starting precomputing...
ActiveRecord::StatementInvalid: PG::ProgramLimitExceeded: ERROR:  index row size 3568 exceeds maximum 2712 for index "index_precomputed_query_docs_on_key"
HINT:  Values larger than 1/3 of a buffer page cannot be indexed.
```

alexsoble:
Hmm, the same task is working just fine on staging.

PR #427, opened 2016-04-25
Display grade level on x-axis of charts.

erose:
Fixes #345.

I don't think this is ready to be merged yet; I'm putting it here to make public the stuff I've done on it so far.

screen shot 2016-04-25 at 5 19 14 pm

This is only implemented in line charts right now. Issues I want to resolve:

  - [x] Correctness. Right now we assume that every student started at grade 1 and advanced one grade each year. Seems like a bad assumption. Where would we get the data to do this right, i.e. which grade they've been in each year?
  - [x] Code quality. `getPositionsForYearStarts` in `profile_chart.js` is very similar to `getPositionsForYearStarts` in `profile_bar_chart.js`.
  - [ ] Design. I was hemming and hawing on how to represent a school year. Here I did the easiest thing, which was to call Jan 1. the middle of the school year, and put the grade info below it. But it might be confusing. Some different options I can think of are:
    * creating a visual interval with tick marks at the beginning and end.
    * creating a tick mark at the start of each school year.

PR #760, opened 2016-10-27
Display student grade level on profile charts: rebased

alexsoble:
This is the rebased PR for #427.

Pushed it to staging; looks good!

### DIFF
--- a/app/assets/javascripts/student_profile/attendance_details.js
+++ b/app/assets/javascripts/student_profile/attendance_details.js
@@ -75,7 +75,8 @@
     propTypes: {
       absences: React.PropTypes.array.isRequired,
       tardies: React.PropTypes.array.isRequired,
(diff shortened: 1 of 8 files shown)


alexsoble:
Maybe think about handling "grade 0":

screen shot 2016-10-27 at 11 06 31 am

Probably not much of an issue with the real data, because students won't have Math / ELA test score results that early. But it does crop up in the demo data.

alexsoble:
Okay yeah I do see "Grade NaN" on staging for younger-grade students.

screen shot 2016-10-27 at 11 36 47 am

alexsoble:
Fixed up the NaN issue by not showing these tick marks for Kindergarteners (944a67ea8e310d1360be9dfde4edd64453756bf9), since they have no previous grade levels to show.

alexsoble:
Merging in this excellent work from @erose — better late then never! Thanks for working on this @erose!

Issue #778, opened 2016-12-02
Add "Summer" panel to School Overview page


alexsoble:

alexsoble:
Okay, new thinking about this after a stimulating conversation with @kevinrobinson on Friday ...

The main reason I'm thinking about this is because some services, like Summer Session and Summer Session for English Language Learners, are never going to be "active" during the school year and so they won't ever show up on the School Overview page filters during the school year.

Maybe the answer to this is to break out a separate Summer panel on the School Overview page.

PR #788, opened 2016-12-16
Change services filter on school overview

alexsoble:
# What's the change?

Change the school overview filter from "active services" (i.e. services being administered at this moment") to "past year services" (i.e. services administered within the past year).

# Why do we need it?

As we're bulk-uploading services, some of them are services that have "ended" already — for example, Summer School which ends when the school year begins. But we still want these services to be showing up on the school overview page.

# Issue

 #778


alexsoble:
Closing this PR ...

@kevinrobinson got me thinking about this in a different way!

PR #802, opened 2016-12-23
Mock use of history in PageContainer during test

kevinrobinson:
So that Jasmine tests don't `replaceState`.


kevinrobinson:
@alexsoble approved over the interwebs

PR #803, opened 2016-12-23
Add analytics for clicking columns on student profile, selecting a st…

kevinrobinson:
…udent through search, using overview filters

Also fix a bug with double click handling on column selection.


kevinrobinson:
@alexsoble says 🚢 it's :friday

PR #806, opened 2016-12-23
Try out String#freeze

kevinrobinson:
No change, just trying out `# frozen_string_literal: true` to see how it affects allocations for this endpoint.


kevinrobinson:
with @alexsoble

PR #807, opened 2016-12-23
Remove attachments from event_note serialization in StudentQueryHelper

kevinrobinson:
is this how i skylight

screen shot 2016-12-23 at 2 47 05 pm


kevinrobinson:
with @alexsoble 🎉

PR #815, opened 2016-12-30
Sketch out districtwide admin page

alexsoble:
# What does this look like?

screen shot 2016-12-30 at 3 24 01 pm

# What's it all about?

+ Create a logical homepage for Uri to access all the administrator functions we're building out (like student services bulk upload) 
+ Reduce unnecessary load on the app by avoiding extra trips to the School Overview Page which costs lots of memory 
+ Clean up the navbar


alexsoble:
Here's what the view looks like for educators with districtwide access but no ability to edit educator permissions: 

screen shot 2017-01-03 at 7 53 22 pm

PR #816, opened 2017-01-05
Add PowerBIExporter for flat export of multiple tables

kevinrobinson:
In contrast to `StudentsSpreadsheet`, which exports a single spreadsheet capturing this moment in time, this class exports multiple flat files for showing trends over time.


kevinrobinson:
selfie, cc @alexsoble FYI

PR #826, opened 2017-01-20
Add an school overview page that merges mutable/precomputed data on c…

alexsoble:
…lient side

+ Separate page so we can test and then make default if it's good


alexsoble:
Tests pass locally!

PR #827, opened 2017-01-20
Remove eager includes from schools_controller

kevinrobinson:
These aren't needed on the precomputed path, so are probably pulling in more data than is needed.


alexsoble:
Tests pass locally and kevin says 🚢

PR #828, opened 2017-01-20
Always precompute all school views when districtwide educator is present

alexsoble:
Some of the school overview pages were not precomputed!


kevinrobinson:
🚢

PR #831, opened 2017-01-20
Fix bug with trying to save notes on older interventions

kevinrobinson:
This is happening from `NoteCard` being used in a write-only way with older interventions.  This guards for that.  Addresses https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/33/?item_page=0&#instances


kevinrobinson:
@alexsoble check it 🚢 ing

PR #834, opened 2017-01-27
Validate that student registration dates can't be in the future

alexsoble:
# Why? 

+ We had one of these come down the assembly line from X2 (probably a data entry error) and it messed with some calculations the school overview dashboard was trying to make


alexsoble:
🚢

PR #842, opened 2017-02-02
Improve rounding in school overview

### LINKED ISSUE #841: Reduce confusion around percentage rounding in school overview
# Problem

+ We always round up when calculating percentages in the school overview, so the totals often add up to more than 100% (sometimes a couple points higher than 100%)
+ Principal from West noticed this and perceives it as a miscalculation 

# Solutions 

+ Sometimes round up, sometimes round down? (dumbest/worst/easiest solution)
+ Ensure that totals always add up to 100%
+ Show no. of students instead of percentages (no rounding there)

alexsoble:
+ Implement https://github.com/studentinsights/studentinsights/issues/841#issuecomment-277070277
+ Fix #841


kevinrobinson:
:ship

PR #843, opened 2017-02-02
This refactors ProfileBarChart in prep for phaselines

kevinrobinson:
The main change is to split out `monthKeys` as a way to define the window of data that is graphed.

It also fixes some eslint errors, cleans up the Jasmine warnings, and makes Jasmine tests fail on React warnings.  And fixes a bug with removing attachments on restricted event notes that came up in the process (users can no longer remove restricted event note attachments).

No regressions:
screen shot 2017-02-02 at 4 37 48 pm

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,6 +2,11 @@
   "env": {
     "browser": true
   },
+  "globals": {
+    "_": true,
+    "$": true,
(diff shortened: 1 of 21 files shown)


alexsoble:
yay!

PR #855, opened 2017-02-12
Upgrade to Rails 5

kevinrobinson:
An attempt at upgrading to Rails 5.0.1.  This is motivated by https://twitter.com/dhh/status/808348184481124352 and https://github.com/rails/webpacker.

Tests are passing locally, and the app is running locally.

Remaining issues:
- [x] either upgrade React to 0.15.1 or fix version for now.  Jasmine failures likely related to [propTypes](https://facebook.github.io/react/warnings/dont-call-proptypes.html).
- [x] bundle exec teaspoon timeout
- [ ] FakeStudent spec
- [ ] controller deprecation warnings (from Teaspoon?)
- [x] smoke test staging deploy


alexsoble:
Damn. Awesome @kevinrobinson.

kevinrobinson:
There are some deprecation warnings when running RSpec tests, which all come from `react-rails`, since I have kept it pegged back at the version that provides React 0.14.3 to avoid upgrading that here as well.  They come from https://github.com/reactjs/react-rails/blob/v1.5.0/lib/react/rails/controller_lifecycle.rb#L8 and https://github.com/reactjs/react-rails/blob/v1.5.0/lib/react/rails/railtie.rb#L88, and we can fix those separately.

kevinrobinson:
@alexsoble FYI, one substantive change is that `lib/analysis/analyze_star_file_dates.rb` had some code that was doing some analysis (rather than defining a class).  With the Rails 5 change to remove autoloading for classes, and to eager load these instead, this led to running the analysis on app startup.  I worked around by wrapping it up in a class definition; alternately we could move these files elsewhere to move clearly indicate they're intended for manual analysis on the console.

@alexsoble @jhilde  This looks good as best I can tell.  Tests pass, everything works locally, and things look good deploying on a new Heroku env up until authentication.  Unless anyone has any concerns about merging this, I'll ship it later today.

alexsoble:
Awesome. 👍🏽👍🏽👍🏽5️⃣5️⃣5️⃣

alexsoble:
🎉  🎉  🎉

jhilde:
Let me jump in on this too!

🙌🏿🙌🏾🙌🏽🙌🏼🙌🏻

On Mon, Feb 20, 2017 at 11:54 AM Alex Soble

<
mailto:Alex Soble <notifications@github.com>
> wrote:

🎉

🎉

🎉

—

Reply to this email directly,
https://github.com/studentinsights/studentinsights/pull/855#issuecomment-281130298
, or

kevinrobinson:
@alexsoble @jhilde thanks pals!  looks good 👍

PR #860, opened 2017-02-20
Some small JS fixes for service uploads

kevinrobinson:
This fixes https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/35/?item_page=0&#instances, which can be triggered when cancelling the upload dialog and adds a spec for it.

It also passes in `currentEducator` in serializedData, which `MixPanelUtils#registerUser` expects.

Finally, this removes a bit of state for `missingFields` that needed to be managed and kept consistent and instead computes it as needed in `isMissingRequiredFields` instead.

### DIFF
--- a/app/assets/javascripts/service_uploads/service_uploads_page.js
+++ b/app/assets/javascripts/service_uploads/service_uploads_page.js
@@ -24,7 +24,6 @@
         lasidAuthorizationError: false,
 
         // Overall form validation
-        missingRequiredFields: true,
         serverSideErrors: [],
         uploadingInProgress: false,
       };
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/service_uploads/service_uploads_page.js
```diff
@@ -106,47 +105,41 @@
...
 
     onClickUploadButton: function () {
-      if (this.state.missingRequiredFields === true) return;
+      if (this.isMissingRequiredFields()) return;
 
       this.upload();
     },
 
-    verifyRequiredFields: function () {
+    isMissingRequiredFields: function () {
```
kevinrobinson: @alexsoble Switching this to a method that computes the value for "is anything missing?" means that the various other methods here don't all have to make sure to update `missingFields`.

kevinrobinson:
ff
screen shot 2017-02-20 at 11 15 26 am

alexsoble:
Thank you @kevinrobinson!

PR #861, opened 2017-02-20
Incremental adoption path for ES6, JSX, stricter linting with Webpacker

kevinrobinson:
This is path forward we could take with adoption Webpacker and incrementally migrating from the current JS setup to a build process with Webpack.  That allows ES6, JSX and potentially a good way to move incrementally to stricter eslint rules.  These things are useful to add more tooling support for developers, and in making the codebase look more like other React projects.  There's also some benefits in being able to using newer language features things like `const` and destructuring.

The plumbing works like this:
- `students_controller#show` sets `@javascript_pack_tag = :student_profile`
- the app's view layout file checks for `@javascript_pack_tag` and if there is one it renders that Webpacker pack
- the `main.js` stub for the student profile page is moved to a Webpacker pack in `app/javascript/packs/student_profile.js`
- In development, you run `bin/webpack-dev-server` as a separate process.  That builds the JS source into Webpacker packs.
- Webpacker's Rails code loads packs from the development server locally

The `student_profile` pack then uses `import` to pull in `PageContainer` module.  I updated that file to use ES6 and JSX, cleaned it up so it follows our eslint rules.  In still pulls in most of its dependencies by reaching into `window.shared` like we have been doing.  And that allows migrating components over to the Webpacker build setup incrementally.  So this pull request has no user-facing changes.

@alexsoble @hilde any thoughts on this?  This would be a fair amount of work to migrate, but is mechanical and perhaps worthwhile at least for new feature development that is JS-centric (eg., maybe analytics views?).  I'm not in any hurry to merge this, or convinced that this is the most valuable thing to spend time on, but wanted to share and see what y'all thought.

TODO:
- [ ] talk about whether this is a good idea, worthwhile, etc.
- [ ] verify that production build works correctly


alexsoble:
@kevinrobinson This is a suuuuuuper good path. I like the incremental approach.

As we saw kicking the tires today, there are a few more bugs to work out before we merge in.

My suggested path would be something like:

1. Fix bugs that crop up locally
2. Document new setup steps in README
3. Push to `somerville-teacher-tool-demo` to verify deploy
4. Merge and push to prod

kevinrobinson:
@alexsoble Yep!  All sounds great 👍

kevinrobinson:
@alexsoble Capturing what we talked about:
- This seems good, with the main benefit over `react-rails` being ES6 modules and some other smaller niceties like code splitting and live reloading.
- But, it doesn't account for compiling ES6 modules in the teaspooon testing setup.  That doesn't just work, and we'd have to look into how to get that working with nothing promising in the webpacker repo after a quick look.
- Given that, splitting the UI and Rails seems more promising, and using `create-react-app` to do that.  It'd take some work to fiddle with that and try out the setup and migration path, which is worth doing but not today.

So, closing this for now.

alexsoble:
Thanks for capturing this @kevinrobinson!

Issue #835, opened 2017-01-30
Generate fake attendance events within correct school year


jhilde:
Currently the generation of fake student data includes generation of attendance events (absences and tardies). These events are generated randomly across 5 years (from 2010) but the records are created with the current school year as the school year for ALL attendance events regardless of date.

_demo_data/fake_student.rb_

kevinrobinson:
@jhilde Good catch!  👍

Feel free to ping me or @alexsoble if you want to pair on fixing this up, or a hand reviewing anything.

Jngai:
I can dive into this next.

alexsoble:
This issue isn't relevant any more because we dropped the `school_years` and `student_school_years` tables in June: #938.

PR #862, opened 2017-02-20
Revisions of SPED report, now Student Report

jhilde:
This is the revisions of the SPED report which is now branded as the Student Report. 

Since this is my first PR, would appreciate feedback, especially on testing strategy. I'm also a verbose commenter it turns out, let me know if we'd prefer fewer comments.

Note that the report will have strange data for absences and tardies because of dummy data issues articulated in #835.

Here's a sample of the document generated: [sped_referral.pdf](https://github.com/studentinsights/studentinsights/files/788565/sped_referral.pdf)


alexsoble:
Cool, awesome @jhilde and congrats on your first PR! Will get back to you on this with a first look tomorrow and a deeper dive on Thursday! 

🎉 🎉 🎉

app/views/students/sped_referral.pdf.erb
```diff
@@ -1,61 +1,192 @@
...
-<p>
-  <h4>Current services:</h4>
-  <ul style="list-style-type:none">
+<style>
+
+* {
+  font-family: 'Open Sans', sans-serif !important;
+}
+
+.data-table  {border-collapse:collapse;border-spacing:0;border-color:#aaa;}
```
kevinrobinson: I'd vote for each rule on its own line like this:

```
.data-table  {
  border-collapse:collapse;
  border-spacing:0;
  border-color:#aaa;
}
``

if you're cool with that.  Makes it easier to edit than super wide lines, and you can see +/- rules on diffs easier.

File: app/views/students/sped_referral.pdf.erb [hunk 2]
```diff
@@ -1,61 +1,192 @@
...
+<br style="clear: left;" /><br/>
+
+<!-- Services -->
+<% right = false %>
+<div>
+<h4>Services:</h4>
+  <% if @services.any? %>
     <% @services.each do |service| %>
-      <li><%= "Since #{format_date(service.date_started)} - #{service.service_type.name}" %></li>
+      <div style="float: left; width:370px; padding:5px">
```
kevinrobinson: I was wondering if it these were `display: inline-block`, would the layout wrap on its own, without having to set and clear floats manually?  Or maybe there's something whack with the way CSS works in the PDF gem.

kevinrobinson:
@jhilde nice!  This looks great 👍 

One bit of UX feedback is that I found the two-column layout harder to parse, particularly to figure out the order of events.  But that's always easy to change around if that's confusing to actual users, so no reason to hold up shipping anything.

I left two quick comments for you, and will leave it to you and @alexsoble to merge and 🚢 this.  Excited to see how folks use this, awesome work! :)

alexsoble:
Thoughts: 

+ I think the code comments are great @jhilde! 
+ I also super appreciate the way you expanded the spec alongside expanding the feature.
+ Agreed with @kevinrobinson that we should find a solution that works with CSS only and doesn't require ruby variables like `<% right = false %>` ...
+ ... but I also don't think that should hold us back from merging and shipping this excellent work. Merging now!

kevinrobinson:
💥 awesome!

jhilde:
Beautiful!

On Wed, Feb 22, 2017 at 9:36 AM Kevin Robinson

<
mailto:Kevin Robinson <notifications@github.com>
> wrote:

💥

awesome!

—

Reply to this email directly,
https://github.com/studentinsights/studentinsights/pull/862#issuecomment-281686437
, or

PR #871, opened 2017-02-28
Tooltip roster

pauledwardmitchell:
Hey @alexsoble,

Here is what I have so far for the changes re: SpEd tooltips on the roster page.

image

image

A few things:
(1) I'd love to chat about how you all are testing react components.  I am reading the code - but could use a conversation also.
(2) I wrote the components in JSX... This is different from what the code looks like now, but I saw the PR from Kevin about moving in the direction of ES6 / JSX.  Let me know what you think!
(3) I think that I have removed all dependencies to mustache and tooltipster - but wanted to check in before I took it all out.


alexsoble:
💥 

Reviewing now!

alexsoble:
Hi @pauledwardmitchell, couple of thoughts here:

## Code

+ Excellent PR! The code is super clear, cleanly-written, and easy to review. 

+ The `react_ujs` path to making these JSX components is really nice and simple . . . @kevinrobinson, take a look at this sometime when you have a minute? Curious how this path compares/contrasts with #861.

## UI

+ Looking at this locally, one thing that pops out is how the colors match the risk level bubble colors. (Screenshot below.) 

+ Visually, this makes it seem like SPED level of need numbers are the same as or comparable to the risk level numbers, when they're really different. To me this suggests using a different color scheme (or maybe no color scheme?) for the SPED level of need numbers:

screen shot 2017-03-01 at 2 24 20 pm

+ I'm also wondering if it's accurate to say "Pluto Kenobi receives 0-2 hours of special education services per week." As I understand it, X2 contains data on student needs & demographics, but not necessarily services. The column in X2 is called  `STD_FIELDB_046`, which is of course kind of ambiguous, but it was described to me as "SPED level of need" -- which might not correspond precisely to the services a student is receiving at any particular time. I'll check in with Uri about this.

pauledwardmitchell:
RE: Risk bubble colors - Ooooo yeah good point.  I'll play with the colors for SPED bubbles a bit.

RE: SPED verbiage - Totally.  I'll wait to hear what Uri says.

Also - I forgot to mention - I used [this react method](https://facebook.github.io/react/docs/dom-elements.html#dangerouslysetinnerhtml) because the student.student_risk_level.explanation is in html... Thought I'd call it out.

alexsoble:
Oooh, thanks for calling that out @pauledwardmitchell. We can pair & look for alternatives to using `dangerouslySetInnerHTML`.

kevinrobinson:
Hi @pauledwardmitchell!  This is awesome, nice work!  👍 

Here's some general thoughts related to what you and @alexsoble were talking about for the React components, and then suggestion  question about this specific PR.

1. `react_component` looks great for "sprinkling in" bits of React components into Rails Views like this in a more Rails-centric way.  It could be a good alternative for the way we do this now (eg., [this](https://github.com/studentinsights/studentinsights/blob/c4a461e3315d9472813e829c520d56f8bd3725f4/app/assets/javascripts/roster/roster_page.js)), which is to have JS code run on particular pages and then decorate or add behavior to the HTML that the Rails views render.  https://github.com/studentinsights/studentinsights/pull/861 is similar, just a different mechanism of making that happen.
2. I think it's likely to be confusing to specify how JS code loads in two different places for the same page.  For example, this PR removes the [.risk-tooltip-circle](https://github.com/studentinsights/studentinsights/pull/871/files#diff-25b3c00b6d4d5b70ea683262c5bb9235L138) class in the Rails view, but the other JS code is still expecting it (eg., [here](https://github.com/studentinsights/studentinsights/blob/c4a461e3315d9472813e829c520d56f8bd3725f4/app/assets/javascripts/roster/roster_page.js#L1150)).  That seems tough to discover at the moment with our current setup!  I'd be in favor of moving all pages that are primarily Rails views to using the `react_component` method you're using here to add in JS behavior over the "decorate the page" since it allows developers to express everything about a page in one place (the view) and makes that more discoverable to others.  That would take separate work, so if that's an intention this seems like a great step!  But if not then modifying `roster_page` would avoid adding the complexity of another way that Ruby and JS code interact.
3. If using `react_component`, I think we should try to make the props that the Rails view passes into the React component as strict as possible (eg., only pass in the fields that the component needs), and that we describe what props the component expects using [propTypes](https://facebook.github.io/react/docs/typechecking-with-proptypes.html).

For this particular PR:

4. Could you say more about what motivates using React components here?  It looks to me like these could be expressed directly in the Rails view and there's nothing dynamic happening?  One rationale might be "let's move these bits of UI into React because we need to re-use them elsewhere" but it seems like then we'd want to isolate the styles to those components as well.  Otherwise it might be simpler to keep this as straight Rails view code.
5. Also, I see you reference removing `tooltipster` in the description, but don't see that in the diff (vs. [searching](https://github.com/studentinsights/studentinsights/search?utf8=%E2%9C%93&q=tooltipster)).  Could you clarify?

Thanks!  Let me know if this is helpful and also happy to chat on Slack or whatever else too.  We can also just merge some work here and talk about other bits separately from this PR, and I'm happy to follow whatever you and @alexsoble think is best here.

pauledwardmitchell:
Hey @kevinrobinson,

This is some fantastic feedback.  Thanks for your patience as I learn you all's intention with the UI code!

Re: 
>"`react_component` looks great for "sprinkling in" bits of React components into Rails Views"

 This is what I was thinking.  Since the view is still mostly html and interpolated ruby, I thought this `react_component` best.  Additionally (and honestly), I am relatively new to React, and need to spend more time understanding how you all have organized the JS files (like how all components in [student_profile](https://github.com/studentinsights/studentinsights/tree/c4a461e3315d9472813e829c520d56f8bd3725f4/app/assets/javascripts/student_profile) interact, for instance).  So the choice for `react_component` was one, also, of simplicity / self-preservation for my young understanding of React.

Your point of 
>it's likely to be confusing to specify how JS code loads in two different places for the same page. 

is right on.  I'd love to walk through (with you or @alexsoble) how you all have set up the organization of the JS code in the app so that I can help keep things simple.

Re: `tooltipster`

@alexsoble and I talked about removing that dependency (as well as that of mustache.js) as a way of improving the general hygiene of the app.

Thanks again for this feedback!  I'll check in with @alexsoble tomorrow and will write you again after we discuss these issues.

kevinrobinson:
@pauledwardmitchell No problem, it's awesome how you're not only shipping new features but working to make it easier for others to contribute too!  👍 

Let me know if there's anything else I can do to talk out options or help out, and otherwise I'll be excited to check out whatever you and @alexsoble decide works best here.  I'm also happy to pitch in with some migration work at some point if that's helpful too.

pauledwardmitchell:
Thanks, @kevinrobinson.  @alexsoble and I pushed back our check-in to next Friday (10 March), so I will check in after we chat!

alexsoble:
Things @pauledwardmitchell and I talked about on our call! 

## Homeroom roster page

1. Remove code that can be safely removed thanks to Paul's refactor (`jquery.tooltipster.min.js`, `mustache.min`, `handlebars.runtime` and the `roster_page.js` code that handles Risk Bubbles.
2. Re-writing homeroom view risk bubble code to not lean on React, since the logic can be easily expressed with Rails code (that way we avoid `dangerouslySetInnerHTML`)

## Student profile page (next bit of work, separate PR...)

+ Moving SPED level of need over to the "Interventions" column, since Uri explained to us that this number refers to # of hours of SPED services delivered
+ Add a tooltip with the same explanation content as the homeroom roster page, but instead of a bubble, words with a dotted underline or something else gentle & small enough to fit into the small space
+ Suggest Student Profile Page work be on a separate PR

pauledwardmitchell:
@alexsoble - Check out the homeroom roster page!  Should be good to go / feedback welcome.

I don't like the white space at the top of the sped tooltip, as you see in this pic... I can't seem to find why that is happening.

screen shot 2017-03-10 at 1 46 38 pm

alexsoble:
Looks great @pauledwardmitchell! Thanks for your flexibility on this and your patience re-writing the same feature-side code to better fit the project direction. 👍

alexsoble:
Made a small commit to get rid of the extra whitespace you mentioned @pauledwardmitchell: 26ac00a2a7a07bf7758b27b8ae5cc627494d35c2.

There might be a more elegant way to do it but just going in and deleting whitespace does work.

alexsoble:
And deleted a little more cruft that you made unnecessary by pulling out mustache/handlebars/tooltipster etc: 086aa7b3246c03afdfcc30bccb46e04dda2fbde5.

pauledwardmitchell:
Nice! Thanks, @alexsoble.

PR #873, opened 2017-03-01
Adding graphs for attendance and discipline data

jhilde:
This is phase two of the reformatted student report to add graphs for attendance (absences / tardies as a stacked column graph) and discipline incidents (column graph).

Note: This change introduces a javascript delay into the pdf generation. This allows HighCharts to fully render the graphs before the pdf is created. Initially this is set to 1000 milliseconds. If this is too short (shouldn't be) we can increase the delay to suit the needs of the server.
[sped_referral.pdf]

(https://github.com/studentinsights/studentinsights/files/811853/sped_referral.pdf)


alexsoble:
Hi @jhilde! Thanks for the PR! 

Looks like the PDF includes variations on the bar charts for Discipline Incidents and Attendance/Tardy events from the Web UI.

Could we pull in that code from `profile_bar_chart.js` and use it here, possibly with some modifications? That would keep all the JS code for this kind of chart in one place.

jhilde:
Agreed @alexsoble in concept. Unfortunately this isn't as easy as it seems since the pdf binary is outside the scope of rails. The easiest way to include javascript is through CDNs (as I did in the code). Alternatively you can link to individual files in the filesystem but it seems this doesn't play well with the asset pipeline.

I'm not sure it would make sense to try to get the profile_bar_chart.js React component to be reused since you would need to manually include the each of the javascript files. 

What we might consider doing is refactoring out some of the shared logic and using it from both profile_bar_chart.js in as a library of sorts.

Open to thoughts and opinions.

kevinrobinson:
@jhilde @alexsoble Interesting!  Systems on systems...  I poked around with this and found a few things that might be helpful, let me know what you think.

First, wicked_pdf supports a [debug param](https://github.com/mileszs/wicked_pdf#debugging) when calling render pdf that you can use to have it generate the view as HTML, before translating for debugging.  That can be useful for inspecting the HTML that the view is generating, before it gets translated to a PDF and you can't see things like what's in the actual style and script tags anymore:

screen shot 2017-03-05 at 12 46 31 pm


Second, in my tinkering it looks like you can use the `wicked_pdf_javascript_include_tag` to include JS from the asset pipeline, and that this injects the JS code inline, rather than injecting multiple script tags like it normally does in development mode:

screen shot 2017-03-05 at 12 47 14 pm

This is cool because it would let us move the JS code that's inline into a separate file, and then either refactor before shipping or ship first and refactor after.

So this looks like it could work for us, just involves changing the way the view file includes the various script tags.  And that way the HTML page gets created in much the same way normal views do, we can re-use JS in the rest of the codebase, and the wkhtmltopdf binary can read that HTML file and do its thing to convert it to a PDF.  I pushed the tinkering to https://github.com/studentinsights/studentinsights/commit/0169bb28ef9cfbc323289caff78c69cc1b6d3c7b, so check that out if it's helpful or feel free to ignore if not.

jhilde:
Very nice! I stopped at the bit about base64 encoding. Not knowing enough about the asset pipeline, decided so dismiss it. Thanks for trudging ahead.

So let's agree on the structure before I refactor?

2 options I see:

1) Use react for the PDF.

2) Extract common functionality in JavaScript included files that are used by the React code (views) and the non-react code (PDF)?

Which makes sense to the two of you?

I'll likely incorporate the changes together with phase 3 (filtering) tomorrow.

On Sun, Mar 05, 2017 at 12:58 PM Kevin Robinson

<
mailto:Kevin Robinson <notifications@github.com>
> wrote:

https://github.com/jhilde https://github.com/alexsoble
Interesting! Systems on systems... I poked around with this and found a few things that might be helpful, let me know what you think.

First, wicked_pdf supports a
https://github.com/mileszs/wicked_pdf#debugging
when calling render pdf that you can use to have it generate the view as HTML, before translating for debugging. That can be useful for inspecting the HTML that the view is generating, before it gets translated to a PDF and you can't see things like what's in the actual style and script tags anymore:

https://cloud.githubusercontent.com/assets/1056957/23589781/c41ba238-01a1-11e7-8388-d21c0e19a651.png

Second, in my tinkering it looks like you can use the

wicked_pdf_javascript_include_tag

to include JS from the asset pipeline, and that this injects the JS code inline, rather than injecting multiple script tags like it normally does in development mode:

https://cloud.githubusercontent.com/assets/1056957/23589789/d92b746e-01a1-11e7-8e81-bcde7516977f.png

This is cool because it would let us move the JS code that's inline into a separate file, and then either refactor before shipping or ship first and refactor after.

So this looks like it could work for us, just involves changing the way the view file includes the various script tags. And that way the HTML page gets created in much the same way normal views do, we can re-use JS in the rest of the codebase, and the wkhtmltopdf binary can read that HTML file and do its thing to convert it to a PDF. I pushed the tinkering to
https://github.com/studentinsights/studentinsights/commit/0169bb28ef9cfbc323289caff78c69cc1b6d3c7b
, so check that out if it's helpful or feel free to ignore if not.

—

Reply to this email directly,
https://github.com/studentinsights/studentinsights/pull/873#issuecomment-284247320
, or

kevinrobinson:
No problemo! 👍  I find the asset pipeline and connections between all the pieces quite complicated, so happy to help share some obscure experience I've gained there :)

For this pull request, I'd recommend splitting the JS code out from being inline in the view.  That seems the most important thing, since it moves JS code into the asset pipeline and enables code-reuse.  Then, you can figure out how the awesome chart code you wrote here relates to other chart code that is very similar (eg., [profile_bar_chart](https://github.com/studentinsights/studentinsights/blob/master/app/assets/javascripts/student_profile/profile_bar_chart.js)).  It seems like there's opportunity to re-use much of the existing code (maybe with some modifications as @alexsoble suggested to allow your improvements related to chart labels on the PDF).  So the way I personally would approach it would be to start from https://github.com/studentinsights/studentinsights/commit/0169bb28ef9cfbc323289caff78c69cc1b6d3c7b and then dig into the JS to see what re-use actually makes sense.  That seems like the most important bit to me.

If you think it'd be simpler to move everything to be React-rendered, that seems fine too!  The main benefit I see would be simplicity, and that a developer working on the PDF feature would only need to learn one framework, especially if the feature uses existing JS code.  The main downside I see for moving to React-rendered would be that it duplicates work a bit, as you've already shipped the initial version in Rails and clearly compartmentalized that the JS code is responsible only for the charts.

I don't see any major benefits or costs either way in React-centric versus Rails-centric for this particular feature.  From my own personal experience I tend towards doing UI things in React more than Rails but we all don't have to be the same.  The prototype version of this feature started as a Rails view initially since an awesome Rails developer wanted to help out and could move quickly in Rails.  So some of it also depends on what you personally are most comfortable with now, and where you want to invest in contributing going forward.  All are great! :)

I'm happy to talk more (or to defer to @alexsoble's original feedback), but also trust you to share your thoughts on whatever you think makes sense and then submit the PR.  I'm a big fan of shipping the feature and then refactoring right after as well :)

jhilde:
What I worry about inlining application.js is the overhead of all the javascript. The file size shoots up to 2.6 MB. It seems this is somewhat expected for rails though. So perhaps it isn't concerning?

kevinrobinson:
You mean the size of JS on the HTML page that's used to generate the PDF?
Could you say more about why you're thinking that size would be a concern?

This is my mental model of the pieces in creating a PDF: the user requests
a PDF, Rails renders a view that links to a bunch of JS or has a bunch of
JS inline, Rails then runs the wkhtmltopdf binary that loads that page in
QtWebKit.  That WebKit process on the server loads/parses/executes the JS,
and then wkhtmltopdf somehow converts that page to a PDF, and Rails sends
that PDF down to the user.

So I think the size of any inline JS could add to the JS parsing time in
WebKit on the server.  But this is quite a complex set of steps, so it's
hard for me to think about how much JS size would matter to the overall
user experience.  My assumption would be not much, but I don't really
know.  If you wanted to, you could make another top-level JS file for PDF
generation that includes all the dependencies needed just for the PDF, but
it might be good to ship it and see if there's a performance problem first
before optimizing.

And yeah, the standars Rails approach to JS is to just concat all the files
together into one big compiled file.  That'd be great to improve separately
(e.g., do code splitting by page or something like that), but probably not
necessary to ship this :)

jhilde:
Here's the update:

@kevinrobinson: I couldn't seem to get your mods with the wicked_pdf_javascript_include_tag to work with the pdf. It worked like a charm for the debug html version. Did you see it work with pdf generation?

@alexsoble: I modified my logic to bring it more inline with the code from profile_bar_chart.js to facilitate. Much of it is currently copied and pasted since I couldn't get the javascript includes working properly. It will now at least be easy to merge them into one common location.

@kevinrobinson: Can you see if the tinkering you did worked with the pdf generation, not just the debug html version?

I'm going to branch to work on the section and date filtering for the report.

Let me know what you want to do.

kevinrobinson:
@jhilde Awesome! 👍 

Sure, can you share the code that you're trying to debug?  Another thing we could try is you could see what happens with https://github.com/studentinsights/studentinsights/commit/0169bb28ef9cfbc323289caff78c69cc1b6d3c7b in your local development environment, and whether that produces a PDF with your sweet graphs.  If you can share any error messages you see or more info about what doesn't work I can take a look and see if I can help.

jhilde:
@kevinrobinson Decided to get back to basics and start with your work. Started clean but still the pdf doesn't show the graphs. Tried a few things and realized it doesn't work if gem'd libraries are included that way. If I use the wicked_pdf_javascript_include_tag for my code and use CDN's for JQuery, Highcharts (along with moment and underscore) it works.

When the pdf doesn't work, there are no errors. Kinda frustrating because it's pretty much a black box.

kevinrobinson:
@jhilde Ah yeah I must have mixed something up or missed that in https://github.com/studentinsights/studentinsights/commit/0169bb28ef9cfbc323289caff78c69cc1b6d3c7b, I see that it works with ?debug but not in the PDF.  I'll be away the rest of this week, but can take another look next week if that's helpful.  What I'd try next would be to have the JS code in `sped_referral` start writing things into the HTML page, to see what's happening when it runs in the wktopdf phase.

Also if you've got something working I'm all in favor of moving forward with something we can have school folks try out, and then tackling this after.

jhilde:
Great minds think alike. Did some basic JavaScript to see if it was workings sans other libraries.

I've been able to get it working if I include application AND separately include the new js file.

I'll play a little bit and update the pull request so we can get the features out there.

On Wed, Mar 15, 2017 at 8:10 AM Kevin Robinson

<
mailto:Kevin Robinson <notifications@github.com>
> wrote:

https://github.com/jhilde
Ah yeah I must have mixed something up or missed that in
https://github.com/studentinsights/studentinsights/commit/0169bb28ef9cfbc323289caff78c69cc1b6d3c7b
, I see that it works with ?debug but not in the PDF. I'll be away the rest of this week, but can take another look next week if that's helpful. What I'd try next would be to have the JS code in

sped_referral

start writing things into the HTML page, to see what's happening when it runs in the wktopdf phase.

Also if you've got something working I'm all in favor of moving forward with something we can have school folks try out, and then tackling this after.

—

Reply to this email directly,
https://github.com/studentinsights/studentinsights/pull/873#issuecomment-286723230
, or

jhilde:
Okay, here's the scoop. Overall lesson learned, do one thing at a time.

The wicked_pdf_javascript_include_tag('application') now works. I just had to add the new file to application.js. Forgot about that part.

As for the code hanging in test, it had nothing to do with the javascript includes. I had also switched to checking if the javascript is done (so the pdf can generate the snapshot) from a simple delay to using window.status. This works like a charm in development but chokes in test (and assuming production). Not sure what the difference is. 

Rather than spend time tracking it down, decided just to go back to a javascript delay of 1 second. Works fine. Note, I also fully disabled animations in highcharts which required the delay. So we might not even need the delay after all.

jhilde:
Added #880 to investigate further merging the graphing code.

app/assets/javascripts/service_uploads/service_uploads_page.js
```diff
@@ -188,13 +188,14 @@
...
       var headerRow = rows.shift().split(",");
 
-      if (headerRow[0] !== 'LASID') {
+      if (headerRow[0].trim() !== 'LASID') {
         this.setState({ missingLasidHeader: true });
         return;
       };
 
-      var student_lasids = rows.map(function(row) { return row.split(",")[0]; })
+      var student_lasids = rows.map(function(row) { return row.split(",")[0].trim(); })
```
kevinrobinson: @jhilde Looks like these changes got included accidentally when merging/rebasing?

app/assets/javascripts/student_profile/pdf/student_profile_pdf.js
```diff
@@ -0,0 +1,195 @@
...
+  // for the previous school year and 5 in the first part of the school year have elapsed). 
+  // We simply need to add the number of months elapsed this calendar year to that 17.
+
+  // Otherwise, we are in the first year of the current school year (i.e. 2016 in school year
+  // 2016-2017). Months 1 - 7 belong to the prior school year so we subtract 7 from the current
+  // month and add 12 for the prior school year. Combining 12 - 7 gives us 5. So we add
+  // 5 to the current month to yield the correct number of months in the current and prior
+  // school years.
+
+  function calculateMonths(nowMomentUTC) {
```
kevinrobinson: This is awesome to explain, but maybe we can figure out how to simplify a bit.  The ways I read this now, the logic is something like always start the graph from August of the current school year.   One thought is to define a method like `startOfSchoolYear(momentUTC) -> augustMomentUTC` as a way to try to express that more directly in the code.

app/assets/javascripts/student_profile/pdf/student_profile_pdf.js
```diff
@@ -0,0 +1,195 @@
...
+    return categories;
+  };
+
+  function yearAxisCaption(monthKey) {
+    return moment.utc(monthKey).format('YYYY');
+  };
+
+  function monthAxisCaption(monthKey) {
+    return moment.utc(monthKey).format('MMM');
+  };
```
kevinrobinson: This looks good, and the one more step in https://github.com/studentinsights/studentinsights/issues/880 we can remove the copy-paste duplication here next too! 👍

app/assets/javascripts/student_profile/pdf/student_profile_pdf.js
```diff
@@ -0,0 +1,195 @@
...
+
+  };
+
+  window.shared.StudentProfilePdf = {
+    load: function() {
+
+      var attendanceData = $('#serialized-data').data('attendance-data');
+
+      var allMonthKeys = monthKeys(moment.utc(), calculateMonths(moment.utc()));
+      console.log(allMonthKeys);
```
kevinrobinson: leftover console.log

app/views/students/sped_referral.pdf.erb
```diff
@@ -138,6 +138,10 @@ ul, li {
     <% end %>
   </table>
 </p>
+<p>
+  <h4>Attendance graph (current and prior school year):</h4>
+  <div id="attendance_container"></div>
```
kevinrobinson: We don't have good tooling for this now, but CSS classes are all hyphen case, so let's keep them consistent at least in the same file.

jhilde:
Because of the randomly merged bits that came in accidentally and further development, I'm going to close this and create a comprehensive pull request that references this closed PR.

PR #877, opened 2017-03-10
Make an exception for Let's Encrypt

alexsoble:
# Why?

+ We want to use a Let's Encrypt cert, but they expect us to serve it over http, so let's make an exception just for that


alexsoble:
Tests pass locally!

PR #879, opened 2017-03-10
Add codemod for converting to JSX

kevinrobinson:
Use https://github.com/cpojer/js-codemod and a tiny fork at https://github.com/kevinrobinson/react-codemod to run codemods and convert things to JSX and use const over var where possible.

@alexsoble Check out the diff, we could use this to convert things all at once.  It just needs to get smarter about renaming these files to JSX (and we could remove the "imports" for `dom` and `createEl` too.  Separately, we could do something similar for converting `window.shared` to ES6 imports as well.

More background if you're curious: https://medium.com/@cpojer/effective-javascript-codemods-5a6686bb46fb#.7dkmg6l3b

PR #881, opened 2017-03-16
Student profile sped

pauledwardmitchell:
Hey @alexsoble!

Here is the SpEd Level feature for the Student Profile page.

image

image

Happy to refactor sans JSX.

### DIFF
--- a/app/assets/javascripts/student_profile/student_profile_page.js
+++ b/app/assets/javascripts/student_profile/student_profile_page.js
@@ -367,17 +367,11 @@
     },
 
     renderSped: function(student) {
-        
-      return(
-        <div>
-          <span style={styles.spedTitle}>SpEd services</span>
-          <ul>
-            <li>
-              { this.spedLevelText(student) }
-            </li>
-          </ul>
-        </div>
-      )
+      return dom.div({},
+               dom.span({ style: styles.spedTitle }, "SpEd services"),
+                 dom.ul({},
+                   dom.li({}, this.spedLevelText(student))
+             ));
     },
 
     spedLevelText: function(student) {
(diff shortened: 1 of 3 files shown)


alexsoble:
Talking with @pauledwardmitchell now, we agreed it might make more sense to write, i.e. "6-14 hours of services" under SpEd services instead of the "Level of Need", which is a little more cryptic/opaque.

pauledwardmitchell:
@alexsoble - let me know what you think!

image

pauledwardmitchell:
Hey @kevinrobinson - 

This PR renames student_profile_page.js to student_profile_page.js.jsx ... @alexsoble and I were talking just now and I wondered if this change would interfere with codemod.  Let me know!  I'm happy to refactor sans JSX.

kevinrobinson:
@pauledwardmitchell Cool!  Thanks for the ping.  My suggestion would be that within a single file we always keep consistent about whether we use `dom.div` and `createEl` or whether we use JSX.  So my suggestion for this PR would be to either stick with the current style, or to convert the entire file to JSX, rather than using both approaches within a single file.

If you'd like to convert, you could do it all by hand, or try grabbing https://github.com/studentinsights/studentinsights/pull/879 and running that on this file.  Alternately, I'm happy to do the conversion work as part of that PR, so we can decouple that from your great work to ship new features here.

pauledwardmitchell:
This makes sense, @kevinrobinson - I will refactor without JSX to keep consistent.

pauledwardmitchell:
JSX-less! Should be good to go, @alexsoble .

kevinrobinson:
No prob @pauledwardmitchell!  👍

alexsoble:
@pauledwardmitchell 👍

PR #889, opened 2017-04-04
Add grade equiv to student charts

pauledwardmitchell:


pauledwardmitchell:
Hey @alexsoble!  

Still refactoring here to make this PR more DRY - but tagging you here in case you have any feedback!

pauledwardmitchell:
@alexsoble - Here is the failing test:

image

To get the grade level equivalent figure into the chart, I added a fifth item to the array (a "quad") that creates the x and y values that eventually populate the tooltip... this fifth item is causing the test to fail.  If you see a better way to pass in the data, I am happy to change it up.  Also happy to amend the test.

alexsoble:
@pauledwardmitchell -- Yes! If you update the test to passing we should be good to go!  👍

pauledwardmitchell:
tests passing! @alexsoble

alexsoble:
OK awesome @pauledwardmitchell! Let me import the real grade level equivalent data into Insights and then I'll merge this in.

alexsoble:
Hey @pauledwardmitchell -- one quick thing I noticed looking at the data -- we seem to be missing one digit. For example, when a student's grade level equivalent is "4.2", we're displaying it on Highcharts as "4". My intuition would be to keep the extra information after the decimal point, what do you think?

pauledwardmitchell:
Hey @alexsoble - let's totally keep the decimal.  

I'll fix it right up

alexsoble:
@pauledwardmitchell groovy!

alexsoble:
@pauledwardmitchell I'm importing the STAR grade_equivalent data into Insights now so that we can merge in your PR and push to production.

pauledwardmitchell:
👍

alexsoble:
Hey @pauledwardmitchell! I merged and pushed to master. Looks okay but there are a few things that seem weird to me... (1) The values are showing integers only, not decimals — not sure `to_f` did the trick. We should double check this in the demo data. (2) It looks like `null` values are being displayed as `0`, which isn't right — let's handle `null`. These can wait for Monday I think but we should tackle them first thing next week.

pauledwardmitchell:
@alexsoble - Hmmmmmmm - thanks for this feedback. I'll look into it presently.

PR #890, opened 2017-04-05
Fixing Student Report bugs

jhilde:
Looks like my manual jsx transformation reverted the renaming of sped_report to student_report. Quick fix.

Looking into adding an integration test for the student_report and running into some roadblocks. Will ping @alexsoble and @kevinrobinson.

I also noticed that the graphs were again only showing up in the debug version, not the pdf version. There is some strange funkiness with the wicked_pdf_javascript_include_tag. Including application.js brings in all the external stuff but doesn't seem to bring in our files. Each local js file needs to be included separately. So I added the helpers/graph_helpers which has the common graphing code.

### DIFF
--- a/app/assets/javascripts/student_profile/profile_details.jsx
+++ b/app/assets/javascripts/student_profile/profile_details.jsx
@@ -149,7 +149,7 @@
 
     onClickGenerateStudentReport: function(event) {
       const sections = $('#section:checked').map(function() {return this.value;}).get().join(',');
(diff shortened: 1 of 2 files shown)


kevinrobinson:
@jhilde 👍 on the renaming, 🚢 ing this now.  Seems like there's something deeper with the PDF problem, added https://github.com/studentinsights/studentinsights/pull/891 to see if that will help.

kevinrobinson:
Also, for the integration test you can just restart the Travis build if it seems like it timed out or something like that, that's what I did here.  That shouldn't ever happen, and if the tests are flaky more often than not we should try to fix that, but sometimes computers act computery :)

PR #891, opened 2017-04-05
Allow PDF debug param in production

kevinrobinson:
This generates an HTML file, no harm in this and it may help debug what's leading to scripts not loading in the PDF generation process.


kevinrobinson:
selfie, cc @jhilde

PR #892, opened 2017-04-05
Allow adjusting delay in PDF JS

kevinrobinson:
See if this reveals anything interesting.  Confirmed that `&debug` reliably works in production, but without it not seeing charts.


kevinrobinson:
cc @jhilde

PR #893, opened 2017-04-05
Fix typo in param for delay

kevinrobinson:
From https://github.com/studentinsights/studentinsights/pull/892


kevinrobinson:
selfie

PR #894, opened 2017-04-05
Wait for JS to finish before loading student profile PDF

kevinrobinson:
Trying this option to rule out that it's a race condition between JS rendering and PDF generation.  Fiddling with `delay_ms` in production didn't help, so this should rule that out altogether, if so then we can see what else might be going on.

### DIFF
--- a/app/assets/javascripts/student_profile/pdf/student_profile_pdf.js
+++ b/app/assets/javascripts/student_profile/pdf/student_profile_pdf.js
@@ -128,6 +128,11 @@
       if(_.flatten(disciplineMonthBuckets).length > 0) {
         generateGraph("#discipline-incident-container", "Number of Discipline Incidents", xAxisSettings, "Discipline Incidents " + GraphHelpers.dateTitle(filterToDate, monthsBack), disciplineDataSeries);
       }
+
+
+      // Set a special value that wicked_pdf can check for to know JS rendering
+      // is done.
+      window.status = 'READY';
     }
   };
 })();
--- a/app/controllers/students_controller.rb
+++ b/app/controllers/students_controller.rb
@@ -63,7 +63,7 @@ class StudentsController < ApplicationController
           pdf: 'student_report', 
           title: 'Student Report', 
           footer: { center: footer, font_name: 'Open Sans', font_size: 9}, 
-          javascript_delay: params[:delay_ms] || 1000,
+          window_status: 'READY',
           show_as_html: params.key?('debug')
         })
       end


kevinrobinson:
fyi @jhilde

kevinrobinson:
Confirmed that this reliably leads to a timeout, when adding `&debug` renders the HTML within a second or two.  So the JS just isn't running in production.  Looking at https://wkhtmltopdf.org/docs.html to see if I can learn more there.

jhilde:
Are the graphs not working in production? Should work with the patch adding the explicit javascript include.

kevinrobinson:
@jhilde Let me know what you see, I'm trying on student 777 and seeing timeouts.  My interpretation is that the Ruby code is blocking waiting for `window.status` to be set, and that this never happens since the JS code is not executing, but not sure.

Also, we're running a different version in production than I am locally, which may be another thing to check out:
```
$ wkhtmltopdf --version
wkhtmltopdf 0.12.3 (with patched qt)
```

Locally I have `wkhtmltopdf 0.9.9`.  That may be mucking with our debugging efforts too, not sure.

jhilde:
@kevinrobinson 

I'm not using window.status anymore. I don't exactly recall what the issue was. It just wasn't getting set. Would have been so much better than the delay.

You need the new wkhtmltopdf for the js delay to work as well. I think.

kevinrobinson:
@jhilde Ah yeah sorry I've been fiddling with this over the last 20 minutes and probably spamming you with mentions :)  Apologies if I've undone something here then, I thought I could help debug in production since it seemed there was some drift between dev and production.  I'll stop now, and here's my current understanding:

- Your awesome PR https://github.com/studentinsights/studentinsights/pull/890 is merged
- I added in &debug and also &delay_ms to debug (and since removed them)
- I switched to use the `window_status` option, confirmed it worked locally, and confirmed it reliably doesn't work in production
- I started looking at `wkhtmltopdf` to see if I could reproduce the problem directly.  One thing I found from [--delay-javascript](https://github.com/wkhtmltopdf/wkhtmltopdf/issues/2207) was:

```
~ $ wkhtmltopdf --debug-javascript /app/tmp/page.html -
Loading pages (1/6)
Warning: undefined:0 TypeError: 'undefined' is not a function
Warning: undefined:0 TypeError: 'undefined' is not an object
```
and [these folks](https://github.com/wkhtmltopdf/wkhtmltopdf/issues/2721#issuecomment-284621608) suggesting it's something missing in QtWebKit like Function.bind.  I patched that in and the error went away.  Piping the HTML file from `&debug` into `wkhtmltopdf` on the dyno, then piping back to open it as a PDF confirmed that the charts render with the Function.bind shim added in.  Going to try to submit that and deploy now, then have to cut out either way, we'll see if that helps. :)

jhilde:
@kevinrobinson Wow, you certainly take longer to declare defeat than I do. 🥇

kevinrobinson:
@jhilde I think https://github.com/studentinsights/studentinsights/pull/895 did it, looks good to me in production now, can you try it out for some other students and confirm whenever you get a chance?

ha!  and yeah I thought I might be able to help debug directly in production since it seemed like something weird was going on, and didn't want to have that hold up you shipping all the awesome work you did here any longer.   😄

jhilde:
I don't have access to production. 👎

kevinrobinson:
@jhilde Yep, we can change that too, although if we're doing things right hopefully we track down and fix differences like this, so that it's unusual when something works locally and then doesn't work when we deploy it.  Hopefully we've done that for the JS-in-PDF for a while :)

jhilde:
@kevinrobinson The bars for the charts aren't being drawn long enough (at least in my dev environment). I've turned off all the animation I could find with HighCharts but it must still be briefly animating the bars. Looks like the javascript is technically completing before the animation is done, causing the bars to be short.

Add attendance events so there are several (like 5) in the same month. Then take a look at the pdf vs. html report. You should see it.

kevinrobinson:
@jhilde Ah yeah good call!  So I imagine we need to either change the way `window_status` gets set so that it's after the animations finish or turn them off (either in JS or CSS, not sure how HighCharts does it).

PR #895, opened 2017-04-05
Add Function.bind polyfill to PDF rendering

kevinrobinson:
Follow on to https://github.com/studentinsights/studentinsights/pull/894#issuecomment-292015698.


kevinrobinson:
selfie

kevinrobinson:
Looks like this did it.

PR #899, opened 2017-04-07
Use Browserify Rails to allow import/export

alexsoble:
# Notes 

+ `react-rails` got us JSX but can't do `import` or `export`
+ Use browserify-rails for that (slow to compile in development)
+ Merging this in to check that it doesn't break production so that we can then go ahead and convert everything to use `import` and `export`


alexsoble:
`bundle exec teaspoon` passes locally!

PR #900, opened 2017-04-07
Fix Travis post-browserify

alexsoble:
Heya @jhilde @pauledwardmitchell! FYI for devs on the project -- we added Browserify-rails to let us do `import` and `export` on our cool ES6 JS. Things will break for you when you pull down these changes but then running `npm install` should fix everything. The first time you try to start a rails server locally it will be reaaaaally slow as Browserify builds up a cache.


jhilde:
👍

pauledwardmitchell:
Groovy. Thanks @alexsoble

PR #901, opened 2017-04-07
Add eslint, convert one spec to use import and update browserify-rails config for specs

kevinrobinson:
This adds eslint into the `package.json` so you can do:

```
$ npm run lint
...

/Users/krob/github/studentinsights/app/assets/javascripts/helpers/profile_details_style.jsx
  103:7  error  Duplicate key 'margin'        no-dupe-keys
  104:7  error  Duplicate key 'borderRadius'  no-dupe-keys

/Users/krob/github/studentinsights/app/assets/javascripts/helpers/prop_types.jsx
  4:9  error  'PropTypes' is assigned a value but never used  no-unused-vars

...

✖ 683 problems (683 errors, 0 warnings)
```

We'll chip away at fixing these (or codemod them) and then move this into the build.  We can also add docs for adding inline lint errors to Sublime too so folks can see helpful warnings like this:

screen shot 2017-04-07 at 5 42 58 pm
screen shot 2017-04-07 at 5 43 25 pm

I also ran `eslint --fix` so the noisy change here is those whitespace or semicolon fixes.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,15 +1,19 @@
 {
   "env": {
-    "browser": true
+    "browser": true,
+    "es6": true
   },
   "globals": {
     "_": true,
+    "moment": true,
     "$": true,
     "React": true
   },
-  "plugins": [
-    "react"
-  ],
+  "parserOptions": {
+    "ecmaVersion": 6,
+    "sourceType": "module"
+  },
+  "plugins": ["react"],
   "extends": [
     "eslint:recommended",
     "plugin:react/recommended"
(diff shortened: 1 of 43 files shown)


alexsoble:
Hey @kevinrobinson -- I should have left a status note about this last night -- the eslint and the style fixes are good, but this PR isn't successfully transforming the `spec/javascripts` file yet. The line we added to transform `spec/javascripts` was breaking on Travis. I commented it out last night to try to pin down the exact issue. So we have some more work to do here on browserifying specs.

kevinrobinson:
@alexsoble k, thanks!  looking at it now then

kevinrobinson:
@alexsoble This should be resolved with https://github.com/studentinsights/studentinsights/pull/903 now, but ping me if not!

PR #902, opened 2017-04-07
Handle null in star chart tooltip formatter

pauledwardmitchell:
(continuation of conversation at end of [PR 889](https://github.com/studentinsights/studentinsights/pull/889))

hey @alexsoble - I've handled null within the formatter function.  

The decimals are rendering for me locally...

screen shot 2017-04-07 at 2 12 37 pm

... is there a way we could pair on Monday to find where the decimal place is being lost on production?

### DIFF
--- a/app/assets/javascripts/student_profile/profile_chart_settings.jsx
+++ b/app/assets/javascripts/student_profile/profile_chart_settings.jsx
@@ -77,11 +77,16 @@
     },
     tooltip: {
        formatter: function () {
-        return  Highcharts.dateFormat('%A, %b %e, %Y', new Date(this.x)) +
-                '<br>Percentile Rank:<b> ' +
-                  this.y +
-                '</b><br>Grade Level Equivalent: <b>' +
-                  this.points[0].point.gradeLevelEquivalent + '</b>';
+        var date = Highcharts.dateFormat('%A, %b %e, %Y', new Date(this.x))
+        var percentileRank = this.y
+        var gradeLevelEquivalent = this.points[0].point.gradeLevelEquivalent
+
+        if ( gradeLevelEquivalent === null ) {
+          return date + '<br>Percentile Rank:<b> ' + percentileRank
+        } else {
+          return date + '<br>Percentile Rank:<b> ' + percentileRank +
+                 '</b><br>Grade Level Equivalent: <b>' + gradeLevelEquivalent
+        }
       },
       shared: true
     },


app/assets/javascripts/student_profile/profile_chart_settings.jsx
```diff
@@ -77,11 +77,16 @@
...
-        return  Highcharts.dateFormat('%A, %b %e, %Y', new Date(this.x)) +
-                '<br>Percentile Rank:<b> ' +
-                  this.y +
-                '</b><br>Grade Level Equivalent: <b>' +
-                  this.points[0].point.gradeLevelEquivalent + '</b>';
+        var date = Highcharts.dateFormat('%A, %b %e, %Y', new Date(this.x))
+        var percentileRank = this.y
+        var gradeLevelEquivalent = this.points[0].point.gradeLevelEquivalent
+
+        if (gradeLevelEquivalent === 0 || null) {
```
alexsoble: `null` is always going to evaluate as falsey, I think we want something like this instead: 

```
if (gradeLevelEquivalent === 0 || gradeLevelEquivalent === null) . . . 
```

pauledwardmitchell: Good call @alexsoble I'll fix it up presently.

alexsoble:
OK awesome! Thanks for the update @pauledwardmitchell. I might also drop the check against zero, because some of the values in production are actually zero (as opposed to null).

Here are the results of a query against the production data so you know what you're working with:

```
$ StudentAssessment.pluck(:grade_equivalent).uniq.compact.sort

=> ["0.00", "0.10", "0.20", "0.30", "0.40", "0.50", "0.60", "0.70", "0.80", "0.90", "1.00", "1.10", "1.20", "1.30", "1.40", "1.50", "1.60", "1.70", "1.80", "1.90", "10.00", "10.10", "10.30", "10.40", "10.50", "10.60", "10.70", "10.80", "10.90", "11.00", "11.10", "11.20", "11.30", "11.40", "11.50", "11.60", "11.70", "11.80", "11.90", "12.00", "12.10", "12.20", "12.30", "12.40", "12.50", "12.60", "12.70", "12.80", "12.90", "13.00", "2.00", "2.10", "2.20", "2.30", "2.40", "2.50", "2.60", "2.70", "2.80", "2.90", "3.00", "3.10", "3.20", "3.30", "3.40", "3.50", "3.60", "3.70", "3.80", "3.90", "4.00", "4.10", "4.20", "4.30", "4.40", "4.50", "4.60", "4.70", "4.80", "4.90", "5.00", "5.10", "5.20", "5.30", "5.40", "5.50", "5.60", "5.70", "5.80", "5.90", "6.00", "6.10", "6.20", "6.30", "6.40", "6.50", "6.60", "6.70", "6.80", "6.90", "7.00", "7.10", "7.20", "7.30", "7.40", "7.50", "7.60", "7.70", "7.80", "7.90", "8.00", "8.10", "8.20", "8.30", "8.40", "8.50", "8.60", "8.70", "8.80", "8.90", "9.00", "9.10", "9.20", "9.30", "9.40", "9.50", "9.60", "9.70", "9.80", "9.90"]
```

pauledwardmitchell:
Ooo that is another good observation @alexsoble - You mentioned in PR889 that:

> It looks like null values are being displayed as 0, which isn't right — let's handle null

This is why I added the check against zero... if the null values are being displayed as zero, I wondered if the fifth value in the quad _was in fact zero_ .  

But now I see that this check should be dropped.  

You said "I might also drop" - I am happy to do it!  Let me know

PR #907, opened 2017-04-21
Handle nil on server side for grade level equiv equal to nil

pauledwardmitchell:


pauledwardmitchell:
@alexsoble - It looks like the `.to_f` was changing `nil` to `0.0`.  I added logic on the server side to create the original 'quad' if the grade_level_equivalent is `nil`.

image

and handled `undefined` instead of `null` in `profile_chart_setting.jsx`.

All seems to be well.

image

PR #908, opened 2017-04-21
Handle grade level equivalent as nil on server side

pauledwardmitchell:
@alexsoble - It looks like the `.to_f` was changing `nil` to `0.0`.  I added logic on the server side to create the original 'quad' if the grade_level_equivalent is `nil`.

image

and handled `undefined` instead of `null` in `profile_chart_setting.jsx`.

All seems to be well.

image


alexsoble:
👍  👍

alexsoble:
@pauledwardmitchell Looks good in prod!

pauledwardmitchell:
huzzah! @alexsoble

PR #909, opened 2017-04-21
Delete old dom, createEl constants

alexsoble:
# What does this do?

Deletes two `window.shared ...` lines of code at the top of many of our JS files that we don't need anymore. We don't need them because we are transforming our JS to JSX through `react-rails`.

Deleted lines:

```
const dom = window.shared.ReactHelpers.dom; ...
const createEl = window.shared.ReactHelpers.createEl; ...
```


alexsoble:
`students_table.jsx` needs to be transformed to JSX, still relying on createEl.

kevinrobinson:
💥

PR #910, opened 2017-04-21
Update README -- we're using JSX now

alexsoble:
+ Remove section that talks about the old `dom.div ...` syntax.


kevinrobinson:
nice!

PR #913, opened 2017-05-01
Adding symbolic link for node's executable for browserify-rails.

jhilde:
When @alexsoble added browserify-rails in #900, it broke the docker config. Turns out the node package installs node as nodejs because of some odd name conflict. Browserify-rails seems to be doing a /usr/bin/env node so it doesn't find the executable. Fix is a simple symbolic link to create /usr/bin/node from /usr/bin/nodejs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,8 @@ VOLUME /mnt/somerville-teacher-tool
 WORKDIR /mnt/somerville-teacher-tool
 RUN bundle install
 
+RUN ln -s /usr/bin/nodejs /usr/bin/node
+
 COPY . /mnt/somerville-teacher-tool
 
 EXPOSE 3000


kevinrobinson:
@jhilde ⚡️ Thanks for fixing up! 👍 

It seems weird there'd be a `nodejs` binary, but TIL the Debian node packages only link `nodejs` https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions, while the official node package installs `node` like you'd expect.

jhilde:
I guess there's a conflict between the Node (Node.js) we know and love and the Amateur Packet Radio Node Program.

Go figure! 🤷‍♀️

kevinrobinson:
hehe naming is hard :)

PR #928, opened 2017-06-01
Adding pseudo-anonymizer class

kevinrobinson:
This is for enabling some exploratory data analysis by project collaborators without granting them full production data access.  See the comments in the code below for more caveats - this is intended only for limited use cases within other layers of security and access controls.

This isn't used in any production path, but includes a small change to factor out a method in `Service`.

### DIFF
--- a/app/models/service.rb
+++ b/app/models/service.rb
@@ -12,17 +12,18 @@ class Service < ActiveRecord::Base
                                                                   # the service isn't discontinued yet.
   end
 
-  def has_scheduled_end_date?
+  def end_date
     # Some services are scheduled to be discontinued in the future. When staff
     # enter new student services in the UI, they can't select a future end date yet.
     # But bulk-uploaded services can have a future end date. If a service has an
     # end date in the future, we don't want it to show up as "discontinued."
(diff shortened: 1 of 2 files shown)


kevinrobinson:
@alexsoble Can you take a look at this and do a sanity-check when you get a chance?  Thanks!

alexsoble:
@kevinrobinson Seems useful for folks who want to do analysis but don't need to know any info about any particular students! 

Looked good when I tried it out against demo data locally, only change is that I addd `require "csv"` to the top.

Hashing the school_id is interesting. Couldn't someone de-anonymize that fairly trivially since it's public info how many students go to each school? Though I suppose the use case here is mostly just making it so that someone doesn't "have to know" that kind of info when working with the data.

alexsoble:
🚢

kevinrobinson:
@alexsoble Yeah exactly!  It definitely doesn't fully anonymize data - we should never use this share anything publicly like folks in other contexts might do for hackathons.  It's only intended to make one layer of anonymity for people who are already working on the project so they don't need full production access.

PR #929, opened 2017-06-02
Add student foreign key to attendance and discipline events

alexsoble:
# What does this do?

+ Associate attendance and discipline events directly onto the students table so that we can eventually drop the student school years table

# Issue 

+ #611


alexsoble:
🚢 ing!

PR #932, opened 2017-06-13
Use new methods that cut out the student_school_year_model ...

alexsoble:
+ ... when computing school year absences/tardies/discipline for the school overview page and student profile page
+ Plus a bunch of Sublime text whitespace autofixes. 

# Issues

+ Implements methods from #931 
+ Works toward #611


alexsoble:
These new methods yield the exact same distribution of discipline incidents, absences, and tardies in prod as the old methods did. 👍

PR #936, opened 2017-06-16
Remove student school year association from interventions

alexsoble:
# Issue

+ Progress toward #611.


alexsoble:
There's not much feature-side code to change here because Interventions are only called on in the Student Profile page via `serialize_intervention`, which doesn't include school year or student school year in its serialization.

PR #937, opened 2017-06-16
Add database records for IEP documents (no actual file storing yet)

kevinrobinson:
This is still in progress, just pushing early steps that @alexsoble and I tested manually.

**UPDATE:**

### What's ready in this PR

+ Merge in import process that creates database rows with name, date, and student reference for each IEP (but not the actual PDF itself)
+ Display IEP document count in student profile when IEPs are present with a "coming soon" notice, like this:

screen shot 2017-07-13 at 4 38 08 pm

### What's going in the next PR

+ Actually importing IEP PDFs into the blob storage (AWS or Microsoft's version)
+ Allowing educators with access to display or download the PDFs

### DIFF
--- a/app/assets/javascripts/student_profile/page_container.jsx
+++ b/app/assets/javascripts/student_profile/page_container.jsx
@@ -46,6 +46,7 @@
         attendanceData: serializedData.attendanceData,
         access: serializedData.access,
         dibels: serializedData.dibels,
+        iepDocuments: serializedData.iepDocuments,
 
         // ui
         selectedColumnKey: queryParams.column || 'interventions',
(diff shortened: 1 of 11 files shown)


alexsoble:
# What's ready in this PR

+ Merge in import process that creates database rows with name, date, and student reference for each IEP (but not the actual PDF itself)
+ Display IEP document count in student profile when IEPs are present with a "coming soon" notice, like this:

screen shot 2017-07-13 at 4 38 08 pm

# What's gonna go in the next PR

+ Actually importing IEP PDFs into the blob storage (AWS or Microsoft's version)
+ Allowing educators with access to display or download the PDFs

PR #940, opened 2017-06-24
Double-write precomputed student hashes to migrate to shorter key

kevinrobinson:
With importing new data for the high school, this led to creating some keys for precomputed data that are longer than Postgres likes.  This led to these precompute jobs failing, and made the new high school overview page timeout on load, and therefore inaccessible to users.

Tracked this down in logentries:
```
ActiveRecord::StatementInvalid: PG::ProgramLimitExceeded: ERROR: index row size 3216 exceeds maximum 2712 for index "index_precomputed_query_docs_on_key"
168 <190>1 2017-06-24T08:02:13.064797+00:00 app scheduler.5296 - - HINT: Values larger than 1/3 of a buffer page cannot be indexed.
184 <190>1 2017-06-24T08:02:13.064800+00:00 app scheduler.5296 - - Consider a function index of an MD5 hash of the value, or use full text indexing
224 <190>1 2017-06-24T08:02:13.064801+00:00 app scheduler.5296 - - : INSERT INTO "precomputed_query_docs" ("key", "json", "created_at", "updated_at") VALUES ($1, $2, $3, $4) RETURNING "id"

...

/app/app/jobs/precompute_student_hashes_job.rb:51:in `precompute_and_write_student_hashes!' /app/app/jobs/precompute_student_hashes_job.rb:13:in `block in precompute_all!'
/app/app/jobs/precompute_student_hashes_job.rb:12:in `each' Context
/app/app/jobs/precompute_student_hashes_job.rb:12:in `each_with_index'
/app/app/jobs/precompute_student_hashes_job.rb:12:in `precompute_all!' 
/app/lib/tasks/precompute.rake:4:in `block (2 levels) in <top (required)>'
```

This pull request addresses that by using a new keying strategy that hashes the values (previously it just concatenated the student ids).  It double-writes both key strategies, and then on the read path it reads the new one first, then tries the old one.

This also adds email notifications for when errors occur writing a precomputed doc.

### DIFF
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -74,18 +74,33 @@ class SchoolsController < ApplicationController
       # so keep a buffer that makes sure the import task and precompute job
       # can finish before cutting over to the next day.
       query_time = time_now - 9.hours
-      key = precomputed_student_hashes_key(query_time, authorized_student_ids)
-      doc = PrecomputedQueryDoc.find_by_key(key)
-      return JSON.parse(doc.json).deep_symbolize_keys![:student_hashes] unless doc.nil?
+
+      # There are two keying strategies, and we're double-writing to migrate over.
+      # The older keying strategy concatenated student ids; the newer hashes them
+      # to keep the size down to avoid Postgres size limitations on indexes.
+      # Query new first, fall back to old, and if nothing then fall all the way back.
+      new_key = precomputed_student_hashes_key(query_time, authorized_student_ids, use_hashed_key: true)
+      logger.warn "load_precomputed_student_hashes querying new_key: #{new_key}..."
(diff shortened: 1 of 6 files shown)


kevinrobinson:
@alexsoble Does this seem reasonable?  I was going to ship it, verify that this doesn't break anything on the read path in production now, and then check in again tomorrow morning to make sure the new keying strategy works.  Then we can remove the old key strategy.  I also repurposed the awesome error emails you made for the import task (maybe worth making that generic afterward too).

@jhilde This was holding up the high school overview page from loading - there's a lot of data required to show that page, so the query is precomputed, but that was failing.  This PR should let us verify that the new data from the high school has been loaded successfully!

kevinrobinson:
This is good, verified in production.  We can remove the old key path tomorrow.

PR #941, opened 2017-06-24
Use generic ErrorMailer class, fixing bug in PrecomputeErrorMailer

kevinrobinson:
Follow-on from https://github.com/studentinsights/studentinsights/pull/940, from testing this manually.

There was a bug in the previous PR - the view template name didn't match the controller name.  I updated this to merge `ImportErrorMailer` and `PrecomputeErrorMailer` into a generic mailer that goes out to anyone on `ERROR_MAILER_LIST`, which is set to include both @alexsoble and me for now.


kevinrobinson:
selfie

PR #945, opened 2017-07-12
Adding courses, sections, student and educator assignments

jhilde:
**This PR will require a new sql export from X2. This isn't currently in production.**

This PR will add courses and sections to Student Insights in preparation for the high school project.

A course can have many sections. To make them unique, they need to be combined with the school. (Looks like schools are able to have duplicate course and section names).

There is a simple testing page to see the data in production. Currently this is viewable only to district-wide admins. Going to /sections will generate an index of all the courses with clickable sections underneath. Example here:

screen shot 2017-07-12 at 12 40 56 pm

Clicking on a section will give all the data about that section. Example here:

screen shot 2017-07-12 at 12 41 24 pm


kevinrobinson:
@jhilde Awesome!  👍

I apologize, I'm busy this week with teacher workshops, but can take a look at this on the weekend or next week for sure, or @alexsoble might be able to sooner.

jhilde:
Thanks @kevinrobinson. John said he would work on the queries next week. Not sure it makes sense to do it until the file is available.

jhilde:
The export files for this are now in production. Unfortunately they contain a large number of extraneous rows with school_local_id set to nil.

I've updated the code to better ignore such rows since they are invalid. I'll work with John to figure out what's going on with the queries?

@alexsoble @kevinrobinson Think we should 🚢 it as is? Really want to get those courses / sections in place. Coding for educator and student mapping to section is right behind. Thoughts?

kevinrobinson:
@jhilde My advice would be to be as strict and slow as any data quality issues as possible.  Features and user interfaces won't matter if the data isn't accurate, and as a team it's we have less ability to understand, debug and address data quality issues that software issues.

That being said, sometimes it's hard to advance data quality investigations without shipping the first version. :)  So I'd say shipping sounds good (especially so you can use the UI tools you built to help debug in addition to the awesome tests and validations you have here).

I also left some comments, mostly clarifying my own understanding of your awesome work, so if you can reply or update however you think makes sense (or doesn't) that'd be helpful for me in learning more.  It might be good to see if @alexsoble has any other suggestions on this first one too.  In general, it's up to you to decide when to :ship changes, but you'll have to own making sure the deploy works and everything looks good in production afterward.  I'm happy to do the first one with you (or I'm sure @alexsoble would be too), so you're all set to do this on your own going forward.

jhilde:
Thanks for the input @kevinrobinson. I don't see your comments?

Perhaps we could pair to do my first shipping of code?

kevinrobinson:
@jhilde Sure!  I'll email re: logistics.

jhilde:
X2 query has now been fixed to eliminate courses not tied to a school. Will work on merging when John Breslin has switched over the new queries.

jhilde:
Query changes have been implemented by John. Ready for 🚢 .

app/importers/file_importers/educator_section_assignments_importer.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def data_transformer
+    CsvTransformer.new
+  end
+
+  def filter
+    SchoolFilter.new(school_scope)
+  end
+
+  def delete_rows()
```
alexsoble: Let's ditch the parentheses if this doesn't take any arguments.

app/importers/file_importers/student_section_assignments_importer.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def data_transformer
+    CsvTransformer.new
+  end
+
+  def filter
+    SchoolFilter.new(school_scope)
+  end
+
+  def delete_rows()
```
alexsoble: Same parentheses note as above.

spec/importers/file_importers/courses_sections_importer_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+                    term_local_id:'FY', 
+                    section_schedule:'3(M-R)', 
+                    section_room_number:'232B'
+                } }
+
+        before do
+          described_class.new.import_row(row)
+        end
+
+        it 'creates a course' do
```
alexsoble: I think this should say, "does not create a course."

app/importers/file_importers/courses_sections_importer.rb
```diff
@@ -0,0 +1,29 @@
...
+    SchoolFilter.new(['HEA', 'WSNS', 'ESCS', 'BRN', 'KDY', 'AFAS', 'WHCS', 'SHS'])
+  end
+
+  def school_ids_dictionary
+    @dictionary ||= School.all.map { |school| [school.local_id, school.id] }.to_h
+  end
+
+  def import_row(row)
+    course = CourseRow.new(row, school_ids_dictionary).build
+    if course.school.present?
```
jhilde: I've been able to resolve the data issues with some fixes to queries. Would still rather not fail hard since it is simply course data. Will add a bit more logging.

app/importers/file_importers/courses_sections_importer.rb
```diff
@@ -0,0 +1,29 @@
...
+      if course.save! 
+        section = SectionRow.new(row, school_ids_dictionary, course.id).build
+        section.save
```
jhilde: Same as above.

app/importers/rows/course_row.rb
```diff
@@ -0,0 +1,29 @@
+class CourseRow < Struct.new(:row, :school_ids_dictionary)
+  # Represents a row in a CSV export from Somerville's Aspen X2 student information system.
+  # This structure represents courses ignoring the section data. Unfortunately this is run 
+  # for each section, rather than each course.
```
jhilde: @kevinrobinson and I had talked about this. The unfortunate bit is that it will re-query a course for each of its sections. No big deal though.

app/importers/file_importers/educator_section_assignments_importer.rb
jhilde: Old habits die hard!

app/importers/file_importers/student_section_assignments_importer.rb
jhilde: Old habits die hard!

spec/importers/file_importers/courses_sections_importer_spec.rb
jhilde: You are correct!

db/schema.rb
```diff
@@ -78,10 +86,24 @@
     t.datetime "updated_at"
   end
 
+<<<<<<< HEAD
```
alexsoble: Hmmm, we have merge conflict cruft here

app/importers/file_importers/courses_sections_importer.rb
```diff
@@ -0,0 +1,33 @@
...
+      else
+        log.write("Course import invalid row: #{row}")
+      end
+    else
+      log.write("Course import invalid row missing school_local_id: #{row}")
```
alexsoble: This line of code will never be reached because of the `save!` on line 22 -- it will raise an error instead of evaluating to `false`-y.

alexsoble:
We're almost there! @jhilde is working on some local Postgres configuration issues. Then he'll run `rake db:migrate:reset` to create a clean `schema.rb` with no merge conflicts in it. Then we should be ready to merge!

jhilde:
My docker has trouble creating a database and the accessing it which is what db:reset does. I'm able to drop that database, recreate, migrate, then db:schema:dump. Same effect.

alexsoble:
@jhilde great!

PR #946, opened 2017-07-17
Store IEPs securely in AWS

alexsoble:
# What's done?

+ Set up AWS S3 instance with secure permissions and private access via an IAMS user
+ Code to move IEP documents from EC2 to S3 and then to stream them to the educator via the Student Profile Overview tab

# What's TODO (on this PR)?

+ [x] Tell AWS to store IEPs encrypted at rest
+ [x] Add auth spec for `IepDocumentsController` and unit test for `IepStorer`
+ [x] Simplify IEP PDF import job so that it only loops through once (DONE: 3847db9bcbf22c5b2dfa55a3c29ff0b9c1270a20, 177f7fa42e269f013256f99f593caac63181f9c3)
+ [x] Little bit of QA on the server side encryption

### DIFF
--- a/app/assets/javascripts/student_profile/profile_details.jsx
+++ b/app/assets/javascripts/student_profile/profile_details.jsx
@@ -262,7 +262,7 @@
           </p>
           <br/>
           <p style={{fontSize: 15}}>
-            PDFs of the documents will be available in Insights soon.
+            <a href="/iep_documents/1">Download.</a>
           </p>
         </div>
       );
--- a/app/assets/javascripts/student_profile/profile_details.jsx
+++ b/app/assets/javascripts/student_profile/profile_details.jsx
@@ -249,8 +249,29 @@
       );
     },
 
+    renderIepDownloadLinks: function () {
+      const iepDocuments = this.props.iepDocuments;
+
+      return iepDocuments.map((iepDocument) => {
+        return this.renderIepDownloadLink(iepDocument);
+      }, this);
+    },
+
+    renderIepDownloadLink: function (iepDocument) {
+      const url = `/iep_documents/${iepDocument.id}`;
+
+      return (
+        <p style={{fontSize: 15}}>
+          <a href={url}>
+            Download {iepDocument.file_name}.
+          </a>
+        </p>
+      );
+    },
+
     renderIepDocuments: function () {
-      const numberOfDocuments = this.props.iepDocuments.length;
+      const iepDocuments = this.props.iepDocuments;
+      const numberOfDocuments = iepDocuments.length;
 
       if (numberOfDocuments === 0) return null;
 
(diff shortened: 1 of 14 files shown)


kevinrobinson:
@alexsoble awesome!!!

app/assets/javascripts/student_profile/profile_details.jsx
```diff
@@ -249,8 +249,29 @@
       );
     },
 
+    renderIepDownloadLinks: function () {
+      const iepDocuments = this.props.iepDocuments;
+
+      return iepDocuments.map((iepDocument) => {
```
kevinrobinson: If you like, this can be: `return iepDocuments.map(this.renderIepDownloadLink, this)`;

app/controllers/iep_documents_controller.rb
```diff
@@ -0,0 +1,37 @@
+class IepDocumentsController < ApplicationController
+
+  before_action :find_iep_document, :authorize!
+
+  def show
+    @file_name = @iep_document.file_name
+    @bucket_name = ENV['AWS_S3_IEP_BUCKET']
+
+    object = s3.get_object({
```
kevinrobinson: Does this need some kind of credentials from an ENV variable to be able to read from the bucket?

app/controllers/iep_documents_controller.rb
```diff
@@ -0,0 +1,37 @@
+class IepDocumentsController < ApplicationController
+
+  before_action :find_iep_document, :authorize!
```
kevinrobinson: This is probably quibbling, but this seems indirect versus just making a method call in `show`.

app/assets/javascripts/student_profile/profile_details.jsx
alexsoble: Ooooh yeah, I love one-liners!

app/controllers/iep_documents_controller.rb
alexsoble: I wanted to share the `@iep_document` instance variable between the `authorize!` method and the `show` method. Open to improvements!

app/controllers/iep_documents_controller.rb
alexsoble: Yes it does. The AWS SDK automatically pulls in that config from ENV: http://docs.aws.amazon.com/sdkforruby/api/#Configuration

alexsoble:
This should be in good shape to 🚢 , planning to merge in, push to prod, and run the import task tomorrow AM.

app/controllers/iep_documents_controller.rb
kevinrobinson: ah cool, thanks for learning me!

app/controllers/iep_documents_controller.rb
kevinrobinson: 👍

PR #955, opened 2017-08-04
Fix nil homerooms, they are bad

alexsoble:
# Why?

+ The way the code is currently set up generates "fake" homerooms for students who have a `nil` homeroom assignment in X2
  + This isn't really accurate, we don't want to be going around and adding data that's different from what's in X2
  + It's OK for students not to be assigned to a homeroom. They show up just fine in the Searchbar and appear on the School Overview page... their data is surfaceable

# What's the fix? 

+ This commit skips assigning a homeroom when X2 doesn't supply a homeroom name 
+ It uses .present? to catch both `homeroom name == nil` and `homeroom name == ''`, which was the error that first drew @kevinrobinson's and my attention to the issue

### DIFF
--- a/app/importers/file_importers/students_importer.rb
+++ b/app/importers/file_importers/students_importer.rb
@@ -18,17 +18,23 @@ class StudentsImporter < Struct.new :school_scope, :client, :log, :progress_bar
 
   def import_row(row)
     student = StudentRow.new(row, school_ids_dictionary).build
+    student.save!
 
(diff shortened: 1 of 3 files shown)


alexsoble:
Pushed to prod, ran migration, homeroom with `name == nil` is no more. Will talk to Uri about the other fake homerooms like "HEA HOMEROOM" etc.

Issue #965, opened 2017-08-16
Adding "NGE meeting" as SHS note option in Insights


alexsoble:
NGE == Ninth Grade Experience Team

alexsoble:
Closed via #969.

PR #969, opened 2017-08-18
Add ninth grade meeting as Event Note Type

alexsoble:
# UI

+ Made buttons bigger and more spread out to accommodate longer Event Note Type name:

screen shot 2017-08-18 at 2 23 02 pm

# Specs 

+ Substituted in `match_array` expectation so that updating array-like expectations is simpler and demands even less detail attention (doesn't care about sort order, which in this case isn't important)

# Issue 

+ #965


alexsoble:
Thinking out loud a little bit ... right now adding a new Event Note Type is a bit of a pain because it involves updating multiple fixtures in both ruby and JS specs. Maybe that's OK because making sure you have the right constants is important? But I feel like having one server side and one client side spec for this should be enough.

Also, more importantly, some of these Event Note Types are Somerville-specific. So one of the changes we'll need to make to accommodate the second district is make Event Note type configurable. One route would be have all district's Types hanging out in the database together and then configure which district gets sent which Event Note Type option at the Rails level. Another way would be to do `SchoolDistrict has_many EventNoteTypes` and store the relationship in the database.

PR #974, opened 2017-08-21
WIP – Start calling MixpanelUtils with import/export instead of window

alexsoble:
# Notes 

+ Nothing special about MixpanelUtils in particular, just a very straightforward chunk of code -- a good place to start moving slowly towards using `import` instead of `window.shared` and begin turning that into an easily repeatable task


alexsoble:
Pushed to prod, Mixpanel calls are working! Mixpanel live view updated correctly when I poked around the site in prod.

PR #975, opened 2017-08-22
Improve specs fake data generator specs, move methods around for readability

Jngai:
I was working on facestudent.rb and then noticed that the spec isn't well tested. I added a couple of more tests to the spec. I moved some methods around to be defined earlier before its actually called. I noticed that `add_services` and `add_events` in fakestudent.rb actually isn't saving for each student but only some of them. There is also a duplicate line at line 230.

### DIFF
--- a/spec/demo_data/fake_student_spec.rb
+++ b/spec/demo_data/fake_student_spec.rb
@@ -23,7 +23,7 @@ RSpec.describe FakeStudent do
   end
 
   it "sets student race attributes" do
-    expect(student.enrollment_status).not_to be_nil
+    expect(student.race).not_to be_nil
   end
 
   it 'adds student assessments' do
--- a/spec/demo_data/fake_student_spec.rb
+++ b/spec/demo_data/fake_student_spec.rb
@@ -18,6 +18,14 @@ RSpec.describe FakeStudent do
     expect(student.home_language).not_to be_nil
   end
 
+  it "sets student enrollment status to be either 'Active' or 'Transferred'" do
+    expect(student.enrollment_status).to eq('Active').or(eq('Transferred'))
+  end
+
+  it "sets student race attributes" do
+    expect(student.enrollment_status).not_to be_nil
+  end
+
   it 'adds student assessments' do
     expect(student.student_assessments).not_to be_empty
   end
--- a/spec/demo_data/fake_student_spec.rb
+++ b/spec/demo_data/fake_student_spec.rb
@@ -30,6 +30,11 @@ RSpec.describe FakeStudent do
     expect(student.student_assessments).not_to be_empty
   end
 
+  it 'adds student attendance events' do
+    expect(student.absences).not_to be_empty
+    expect(student.tardies).not_to be_empty
+  end
+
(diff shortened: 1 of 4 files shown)


Jngai:
I also moved the `next` method to the end since it gets passed around a lot thus it is the continuation of the next step.

spec/demo_data/fake_student_spec.rb
```diff
@@ -18,10 +18,23 @@
     expect(student.home_language).not_to be_nil
   end
 
+  it "sets student enrollment status to be either 'Active' or 'Transferred'" do
+    expect(student.enrollment_status).to eq('Active').or(eq('Transferred'))
+  end
+
+  it "sets student race attributes" do
+    expect(student.enrollment_status).not_to be_nil
```
alexsoble: The comment mentions race attributes, but the spec checks enrollment status -- looks like a simple copy-paste thing, easy fix.

alexsoble:
Hey there @Jngai, thanks for the PR! I left one comment about what looks like a simple copy-paste thing, otherwise this looks fine! Adding more specs is good. If you're looking for areas that are in particular need of refactor/code quality improvement, check out the [issues](https://github.com/studentinsights/studentinsights/issues) or the [CodeClimate CI](https://codeclimate.com/github/studentinsights/studentinsights). But yeah improving the fake data is a wonderful way to dive in and get started improving the codebase! Thanks!

alexsoble:
Cool, thanks for fixing that typo! 

I'm not sure what's going on with that schema check-in diff, since the schema on your branch seems identical to the schema in `student-insights/master`. 

Also I spotted an issue with one of the new specs, the one that caused the CI to fail on c9793e3679974cddfe0e8273db0c17c82db21f1d. Will add a comment.

Jngai:
@alexsoble I just fix the typo. After I fix the typo, the test failed and that was because I forgot to check in my schema. The schema has a slightly different format but I did generate it from the migrations files. I had to check in the schema for other projects in the past for the tests to run smoothly. I notice that there are lots of specs for areas that are important. However I just have a habit of improving test coverage. Its good for another look. Thanks for your review.

spec/demo_data/fake_student_spec.rb
```diff
@@ -18,10 +18,23 @@
...
+  it "sets student race attributes" do
+    expect(student.race).not_to be_nil
+  end
+
   it 'adds student assessments' do
     expect(student.student_assessments).not_to be_empty
   end
 
+  it 'adds student attendance events' do
+    expect(student.absences).not_to be_empty
```
alexsoble: This expectation isn't accurate. Some students in the fake data have no absences or tardies, just like some students in the real production data. Check out the distribution function in the method that generates fake attendance events.

Jngai:
I think we commented on the same time and you posted earlier.

alexsoble:
@Jngai OK. Check out my note about the attendance event spec you added. Thanks!

Jngai:
@alexsoble I remove the test because the `add_attendance_events` doesn't necessarily generate both attendance and tardies for a student.

PR #983, opened 2017-08-23
WIP – Switch to database-level dependent: destroy cascades for three event columns

### LINKED ISSUE #246: Switch to database-level dependent: destroy cascades
From looking at https://robots.thoughtbot.com/referential-integrity-with-foreign-keys, it seems we can also move `dependent: destroy` expressions here too.  I don't think we're super rigorous about that at the moment, but would be a good improvement after we land https://github.com/studentinsights/studentinsights/pull/241.

alexsoble:
# Issue

+ Fix #246

# Notes 

+ This is a super destructive migration. In order to add the right database-level cascades, we have to remove the foreign key associations. That means we need to delete all the records in question and then re-run the import process. 

+ The right way to do this would be to run the migration at night or when we're confident folks aren't looking at the dashboard, and be ready to do a database backup in case it doesn't go well.

### DIFF
--- a/db/migrate/20170823184540_use_database_level_destroy_cascades_for_events.rb
+++ b/db/migrate/20170823184540_use_database_level_destroy_cascades_for_events.rb
@@ -1,0 +1,27 @@
+class UseDatabaseLevelDestroyCascadesForEvents < ActiveRecord::Migration[5.1]
+  def change
+    district_name = ENV['DISTRICT_NAME']
+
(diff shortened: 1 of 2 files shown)


alexsoble:
Want to test this on some demo data before trying it on prod!

alexsoble:
Talking this through with @kevinrobinson ... this is super destructive and we don't have a high-priority reason to do this, so let's 🚫  the PR.

Issue #993, opened 2017-08-29
Fix student names endpoint timeout


alexsoble:
Getting this error from Logentries: 

```
at=error code=H12 desc="Request timeout" method=GET path="/students/names" host=somerville.studentinsights.org 
```

alexsoble:
Strategy . . . gonna cache JSON that powers the searchbar in the database instead of checking authorization for every single student against the current educator.

alexsoble:
Still throwing lots of errors, let's see if this helps:

```
heroku run:detached rake chores:update_searchbar_data_for_educators_who_log_in
```

jhilde:
@alexsoble Wanted to checkin to see how this is coming. Looks like you were moving the updating of the students for the search bar out of the importer and into a separate job. 

Looking into the issue #1023, it appears to me that it is this code that's taking 3+ hours to execute, causing the timeout. 

Unfortunately, this means that high school teachers who had students transfer into our out of sections don't see the correct roster for their section in Student Insights. 

Looks from your code that the call to Educator.save_student_searchbar_json for each educator should be eliminated? And the rake chores:update_searchbar_data_for_educators_who_log_in should be added to the scheduled nightly jobs?

alexsoble:
Hey @jhilde! Hmmmm, what data source are you looking at that shows it takes 3+ hours to execute?
 I'm hoping this will give us a little more data: https://github.com/studentinsights/studentinsights/pull/1050

But overall, yes you're 100% right . . . now that we have a task that's precomputed for most of the educators, we don't need to run `save_student_searchbar_json` for every educator on every import. Good catch, thanks for keeping an eye out here!

jhilde:
Hey @alexsoble, I just looked through logentries to see when each importer began and subtracted. Nothing fancy. I have lots of ideas on doing some more logging on imports but I'll save that for another day soon.

IMHO, it looks like we're trying to reuse the logic in educator.is_authorized_for_student by calling it once per student per educator. I think the reuse of the logic is noble and definitely makes sure the permissions are in sync. For example when I added section permissions, I didn't need to know anything about the json for search bar. But for performance, I might argue that the logic should be duplicated to reduce the overhead of LOTS of redundant queries per educator.

jhilde:
[...]

PR #1001, opened 2017-09-01
Cache student searchbar data to database

alexsoble:
# Issue 

+ #993

### DIFF
--- a/app/controllers/admin/educators_controller.rb
+++ b/app/controllers/admin/educators_controller.rb
@@ -13,6 +13,11 @@ module Admin
       super
     end
 
+    def update
+      # TODO (ARS): re-compute student searchbar JSON once permissions are updated
+      super
+    end
+
     def resource_params
       params["educator"]["grade_level_access"] = params["educator"]["grade_level_access"].try(:keys) || []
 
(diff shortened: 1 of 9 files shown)


alexsoble:
Note on `administrate`'s `update` method, which we'd like to extend so that it precomputes educator searchbar JSON after update (`administrate/app/controllers/administrate/application_controller.rb`):

```
    def update
      if requested_resource.update(resource_params)
        redirect_to(
          [namespace, requested_resource],
          notice: translate_with_resource("update.success"),
        )
      else
        render :edit, locals: {
          page: Administrate::Page::Form.new(dashboard, requested_resource),
        }
      end
    end
```

alexsoble:
# Notes 

+ Did manual QA with educator permissions dashboard in local environment; cached searchbar data updated correctly after educator permissions were changed
+ This PR includes unit tests for saving student searchbar JSON correctly and controller tests for making sure it's parsed well at the controller level 
+ I'm curious about adding Rails new system tests to exercise this kind of functionality from end-to-end for JS-less pages like the admin dashboard, but at this point it looks like `rspec-rails` hasn't shipped a version or documentation that wraps Rails system tests yet

Issue #695, opened 2016-09-20
Ability to select an end date for services


alexsoble:
Classroom teacher mentioned that this would be useful when back-filling services data.

awalker4:
I noticed that when you discontinue a service it shows duration down to the minute. Does that mean that we could satisfy the back-fill use case with a duration field? If we just add an end date we wouldn't get any hour/minute data.

alexsoble:
Closed by @Jngai in #1105!

PR #1004, opened 2017-09-07
Add ability to select end date for services

Jngai:
serviceendfeature

This is good for back filing previous services that ended. Basically I added an extra datepicker with a  dateended state. I notice that the prop data heads to the api which eventually does a post to the controller. I did the check there and create a discontinued service for that service if it has already ended.


Jngai:
Issue #695. Future end date probably isn't expiring and I already tried setting datepicker 2 to empty/null. @alexsoble Please review and let me know your input.

alexsoble:
Hey there @Jngai, thanks for this PR! I'm currently on the road so I may not be able to give this a thorough review until this weekend or next week — apologies for the delay. In the meantime it looks like rebasing onto master will be helpful. Thanks again, you are really powering through issues here Judy!

Jngai:
Hey @alexsoble No worries about the review time, have fun with your trip and just review it when you are free. Hopefully I get the chance to clear up all of the issues currently outstanding. Have a good weekend :)

alexsoble:
Hi there @Jngai! Thanks again for this excellent PR, your spirit and vision around clearing up all the issues is wonderful!

The code side of this PR looks good. On the UI side, this is a quite challenging issue to tackle because the issue is just one sentence and has no written spec or design -- that makes it a tough one but also very useful because your PR is nudging me to take a much closer look at this feature. 

I watched the GIF you provided and poked around the app. I think we need to break this feature into three main cases and three smaller cases:

## 1. Start in past, end in past
+ This is backfilling a service. 
+ The UI here looks great in your PR: 

start-past-end-past

## 2. Start in past, end in future
+ This is probably going to be the most common use case.
+ Here’s what it looks like right now: 

start-past-end-future

+ There are a couple of issues here:
  + First of all, it says “10 days”, but it should be more explicit that there will be a future end date. Maybe alongside the “Started” we need an “Scheduled to end September 30, 2017” I think that’s less cryptic for the user/reader.
  + There is a Discontinued button, but no visual for the user showing that there is already a future Discontinue planned . . . in this case I think it would be clearer for the user for the button to show "Discontinue Early," especially if they see a scheduled end date.

## 3. Start in future, end in future
+ There's seems to be a real bug here, which I verified in production: when a user creates a Service with a future start date, it says "Started Today" even if it's meant to start in the future. We should fix this:

start-future-end-future

+ Thank you for bug catching in addition to feature building!

In addition to those three main cases, there are a few "edge cases" concerning features that were started/ended the same day the user records them: 

1. Start in past, end today
2. Start today, end today
3. Start today, end in future

We don't necessarily need to write code to cover these cases right away but I do think it would be nice for the Service UI to say "Started today" instead of "Started [AT DATE]" at some point.

What do you think @Jngai? Does breaking it into these three main cases and three smaller cases help clarify?

alexsoble:
Quick notes from conversation with @Jngai: let's skip the edge cases at the bottom ("Start in past, end today", "Start today, end today", "Start today, end in future") and just focus on the main three use cases above.

Jngai:
see #1097

PR #1005, opened 2017-09-08
Section View (high school)

jhilde:
This PR will release the section view for educators. 

**What does it look like?**
screen shot 2017-09-07 at 8 16 38 pm

**Who will see this page?**
This will be the default homepage for any educator assigned to a high school (SHS). District Admins still have access to the kludgy /sections/ index page. Others who have access to sections (see permissions for who can see what), could manually access it by entering the url. They would need to know the Student Insights id of one of their sections (not likely).

In the future, a few directions:
1. We'll likely add section information to the student profile so educators could click on that section information to see the section view (if they had permission to access the section).
2. We should figure out a way for elementary specialists (e.g. art, music, etc.) to access the section view.
3. We may decide to collapse the concept of homeroom into the section view.

**Permissions**
The primary use case is a high school educator who will have permission to see any section they've been assigned as primary staff in Aspen X2. [As mentioned above, this will work for educators assigned to an elementary school as well but will need to manually go to the url and know an internal section id.]

School-wide admins will be able to see all sections in a particular school. (This currently would result in a very long select box)

**Flexible Roster**
In an attempt to make our code DRYer, this PR includes the FlexibleRoster React Component. It is largely based on the roster used in the school view with some styling from the homeroom view. Configuration inspiration from [arqex/react-json-table](https://github.com/arqex/react-json-table). My hope is this component can be used by our pages that require a roster view.

The FlexibleRoster requires three props: rows (the data in an array), columns (the configuration), and initialSortIndex.

Each item in the columns (configuration) array can have:
**label** (required): string that will be used as the column header.
**group**: string that will be used as a super (group) column header. Repeated groups of the same name will be combined in a shared super (group) column header.
**key** (required): string that will be the key into the data elements in the rows.
**cell**: callback function to modify the content of a cell. (The section view uses this to add a link and click handler for the student name.) If cell is omitted, the raw value for the key is used.
**sortFunc**: callback function to compare values to sort them. (The section view uses this to sort numbers or custom enums like fluency). If sortFunc is omitted, a basic string sort is used.


app/assets/javascripts/components/flexible_roster.jsx
```diff
@@ -0,0 +1,158 @@
+import SortHelpers from '../helpers/sort_helpers.jsx';
```
jhilde: This is the FlexibleRoster component. Reuses a bunch of code from students_table.jsx

app/controllers/application_controller.rb
```diff
@@ -21,6 +21,8 @@ def homepage_path_for_role(educator)
       educators_districtwide_url
     elsif educator.schoolwide_access? || educator.has_access_to_grade_levels?
       school_url(educator.school)
+    elsif educator.school.school_type == 'HS'
+      default_section_path(educator)
```
jhilde: If an educator is tied to a high school, they will be defaulted to the section view. We may want to consider defaulting to section view for elementary school educators who have no homerooms (this would allow art, music, etc teachers to see their students).

app/models/educator.rb
```diff
@@ -67,6 +70,7 @@ def is_authorized_for_student(student)
     return true if self.schoolwide_access? || self.admin? # Schoolwide admin
     return true if self.has_access_to_grade_levels? && student.grade.in?(self.grade_level_access) # Grade level access
     return true if student.in?(self.students) # Homeroom level access
+    return true if student.in?(self.section_students) # Section level access
```
jhilde: Allows educators to see student profiles for students in their assigned sections.

app/models/educator.rb
```diff
@@ -76,6 +80,16 @@ def is_authorized_for_school(currentSchool)
...
+  def is_authorized_for_section(section)
+    return true if self.districtwide_access?
+
+    return false if self.school.present? && self.school != section.course.school
+
+    return true if self.schoolwide_access? || self.admin?
+    return true if section.in?(self.sections)
+    false
+  end
+
```
jhilde: Districtwide can see all sections
Schoolwide can see all sections in their school
Otherwise, educators can see sections they are assigned to.

We may want to revisit the logic that limits educators to only seeing students in the educator's assigned school unless they are district-wide. I know there are elementary specialists (library and music) who teach at multiple elementary schools.

app/models/student_risk_level.rb
```diff
@@ -55,7 +55,10 @@ def update_risk_level!
   def calculate_level
     # As defined by Somerville Public Schools
 
-    if mcas_or_star_at_level(3) || limited_english_proficiency_risk_level == 3
+    if student.school.school_type == "HS"
+      #High School risk levels have not been defined
+      level = nil
+    elsif mcas_or_star_at_level(3) || limited_english_proficiency_risk_level == 3
```
jhilde: This effectively removes risk levels from high school students. We haven't determined the logic but it will be different than at the elementary school level.

spec/javascripts/helpers/sort_helpers_spec.jsx
```diff
@@ -0,0 +1,39 @@
...
+        if(sortData[i]['testElement'] != targetData[i]['testElement']) {
+          match = false;
+        }
+      }
+
+      expect(match).toBe(true);
+      
+    });
+  });
+});
```
jhilde: I was having a bone-headed problem with the enum sorting so I decided to add a spec. Wasn't the code, was my calling it but decided to leave the spec in to exercise that code.

alexsoble:
Hi @jhilde, thanks for the pull request! I will aim to review this weekend or very early next week.

alexsoble:
Hey @jhilde! I had a little more time than I thought, so I went ahead and reviewed the PR! I especially like the gradual progression from flat HTML views to React-powered views for sections. I ran the new seed data, logged in, and did some manual clicking around as well.

I'm going to leave a few notes on areas that could use another pass. Once those are addressed I think this should be good to merge.

app/assets/javascripts/components/flexible_roster.jsx
```diff
@@ -0,0 +1,158 @@
...
+  },
+
+  renderSuperHeaders() {
+    var currentCount = 0;
+    const columns = this.props.columns;
+
+
+    var superHeaders = [];
+
+    for (var i=0; i<columns.length; i++) {
```
alexsoble: This `for` loop is hard to read, as is a some of the code within it:

```
if(i+1 == columns.length || columns[i].group != columns[i+1].group) 
```

I would suggest you refactor using lodash, a helper library for these sorts of things that we use throughout the JS to make complex functions simpler and easy to read: https://lodash.com/docs/4.17.4

spec/controllers/sections_controller_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+        end
+      end
+    end
+    
+    context 'districtwide educator logged in' do
+      let(:dw_educator) { FactoryGirl.create(:educator, districtwide_access: true, school: school) }
+  
+      before { sign_in(dw_educator) }
+
+      context 'when requesting a section inside their school' do
```
alexsoble: District-wide educators don't need to be assigned to a school; see `Educator#has_school_unless_districtwide` under `educator.rb`. Can you adjust these context blocks to account for that?

app/assets/javascripts/components/flexible_roster.jsx
jhilde: @alexsoble lodash is cool, not sure how it would help in this instance though since I'm trying to do something pretty specific. Essentially, I'm collapsing the structure down to the number of sequentially repeated elements. 

I thought about adding more structure / hierarchy to the configuration (columns) but it doesn't seem worth it.

See if this code is more readable:

for (let i=0; i<columns.length; i++) {
      // group for the current column
      let itemGroup = columns[i].group;

      // group for the next column for comparison
      // set to null if this is the last column
      let nextItemGroup = columns.length > i+1 ? columns[i+1].group : null;

      // count of items with the same group
      // increment in the beginning since colSpan starts at 1
      currentCount++;
      
      // if the current item doesn't equal the next
      // push the super header with a length of currentCount
      // and reset currentCount for a new column group
      if(itemGroup != nextItemGroup) {
        superHeaders.push({label: itemGroup, span: currentCount});
        currentCount = 0;
      }
    }

alexsoble: @jhilde Yes, that's helpful! Having a `nextItemGroup` variable, plus a simpler inequality comparison at the end makes it simpler to read. 

With lodash I was thinking that `_.forEach` instead of `for (let i=0; i<columns.length; i++)` might make the code read even more simply, but I don't think it's essential or a blocker. Sorry for not being more specific there!

jhilde: No worries! I think that since I'm traversing the array a bit more than just as an iterable, an explicit for loop might be best.

spec/controllers/sections_controller_spec.rb
jhilde: @alexsoble Changed it a bit to remove the school for district wide admins. Point is regardless of whether they have a school assigned, they always see all sections districtwide.

jhilde:
Just want to check in to see if @alexsoble has more comments or @kevinrobinson has anything? Would love to ship this soon.

alexsoble:
@jhilde Looks good to me! Thanks for adding code comments to the `for` loop. Let's :ship:!

alexsoble:
@jhilde Pushed to prod, looks great! Looking at https://somerville.studentinsights.org/sections/ when logged in as an admin educator.

Issue #1011, opened 2017-09-11
Numeric Sorts broken for IE11


jhilde:
The sorting helper code shared by all of our rosters (school, homeroom, and section) uses Number.isInteger as it sorts numeric fields (attendance count, test score, etc). 

[Number.isInteger isn't supported in IE11](https://docs.microsoft.com/en-us/scripting/javascript/reference/number-isinteger-function-number-javascript)

Not sure what our preference is on polyfill vs. change the code.

Any thoughts @alexsoble or @kevinrobinson ?

kevinrobinson:
@jhilde Nice report!  That probably means that each of these Rollbar messages means a user is clicking on those headers and it's not sorting.

If this isn't supported IE, then https://lodash.com/docs/4.17.4#isInteger should do the trick.

jhilde:
When in doubt, lodash! 😆

jhilde:
Fixed with #1013

PR #1013, opened 2017-09-12
Moving isInteger from Number.isInteger to lodash's isInteger for brow…

jhilde:
Moving isInteger from Number.isInteger to lodash's isInteger for browser compatability.


jhilde:
Issue #1011

alexsoble:
Excellent and good catch!

PR #1020, opened 2017-09-12
Import Student Section Grades

jhilde:
This will add another importer (StudentSectionGradesImporter) to import the current average for students in sections. Unfortunately, this is largely redundant to StudentSectionAssignmentsImporter. This is a proprietary file that can't easily be merged into the other csv files.

**What's interesting?**
-- This proprietary file doesn't include headers, so I've updated the CSVTransformer to allow you to set the headers manually. Default is true which means there's a header row in the CSV, use it.

-- Turns out rails doesn't like when you don't have a primary key on join tables. Can't figure out a way to update the record without a primary key. So this adds both grade and a primary key (id) to the student_section_assignments table. 

-- Since the StudentSectionAssignmentImporter deletes the data and imports fresh, I didn't want to have the sequence for the ids get huge over time. So the StudentSectionAssignmentImporter now resets the sequence after it deletes the data.

**When to deploy?**
The file currently exists on the ftp site BUT the grade itself is broken (i.e. all grades are blank). So deploying won't hurt anything, but likely good to hold off until the vendor fixes the proprietary feed. (John Breslin is working with them on it).


alexsoble:
Hey there @jhilde! I see your rubocop comments!

Quick tip: running this command will autofix almost every issue that rubocop raises:

```
rubocop -a
```

Thanks for bearing with this new linter. Now that we have a growing team I figure it makes sense to have more common & enforced code style standards across the Ruby side as well as the JS side so that the code base stays unified.

jhilde:
I've created hashes for students (local_id => id) and sections (sections.section_number|sections.term_local_id|schools.local_id|courses.course_number => section.id).

These are both populated by a call in the initializer of the importer and passed around. 

Students is easy and leverages rails activerecord pluck which skips the overhead of creating the models.

Sections is more difficult because there are 4 fields across 3 related table that defines a unique key (as described above). While I could have leveraged activerecord's pluck to get each of the fields, I would then have needed to concatenate them in rails. So I opted to execute a sql statement that includes the concat in the select. 

@alexsoble @kevinrobinson I'd value your feedback on the caching in general and specifically on executing sql vs. hacking activerecord a bit.

jhilde:
The file is finally populating from the vendor, turns out the grade field is a float and a string separated by a space (e.g. "85.0 B"). So I've separated them out into two fields.

jhilde:
Hey @alexsoble and @kevinrobinson would love a quick sanity check so I can push this for data validation?

app/importers/data_transformers/csv_transformer.rb
```diff
@@ -4,8 +4,12 @@ class CsvTransformer
 
   attr_accessor :pre_cleanup_csv_size
 
+  def initialize(headers=true)
```
kevinrobinson: For optional arguments, accepting an `options = {}` hash helps make a bit more explicit what's happening.  The call site then will read like `CsvTransformer.new(headers: whatever)` instead of the more cryptic `CsvTransformer.new(whatever)`.  And this particularly helps keep it clear what's happening when you add or remove optional arguments over time.

app/importers/file_importers/student_section_assignments_importer.rb
```diff
@@ -14,6 +13,7 @@ def filter
 
   def delete_rows
     StudentSectionAssignment.delete_all
+    ActiveRecord::Base.connection.reset_pk_sequence!('student_section_assignments')
```
kevinrobinson: Can you help me understand this change?  It seems like this would change the core feature of primary keys, where they always refer to the same entity over time.  And so I'm wondering what makes that worth doing here.

I saw your note about not wanting ids to get huge, but doing some napkin math:
- 10k records imported each day
- with 365 days each year
- that's 3.7M records every year
- for 10 years we get 37M records

If I follow then we'd still have two orders of magnitude left until we hit the 4B ceiling in 32-bit integer space.

app/importers/file_importers/student_section_grades_importer.rb
```diff
@@ -0,0 +1,31 @@
+class StudentSectionGradesImporter < Struct.new :school_scope, :client, :log, :progress_bar
+  def initialize(*)
+    super
+    self.student_lasid_map = Student.all.pluck(:local_id,:id).to_h
+    self.section_number_map = ActiveRecord::Base.connection.execute("select CONCAT(sections.section_number, '|', sections.term_local_id, '|', schools.local_id, '|', courses.course_number), sections.id  from sections inner join courses on courses.id = sections.course_id inner join schools on schools.id = courses.school_id").values.to_h
```
kevinrobinson: Can you say more about writing SQL here rather than Ruby?  It seems like adding `.includes(course: school)` and then mapping over them to concat like you mentioned would be just fine.  All things being equal, I think it's good to default to the idioms of the language, framework and the codebase, but of course we should also do what makes sense if there are good reasons to go our own way!

app/importers/file_importers/student_section_grades_importer.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
```
kevinrobinson: To clarify my understanding, this is because Aspen doesn't have a primary key for sections.

Given that's the case, is the tuple of `(section_number, course_number, term_local_id)` not unique?  if we also need the school `local_id` to disambiguate, then should that be updated in the unique constraints on the `Section` model?

app/importers/file_importers/student_section_grades_importer.rb
```diff
@@ -0,0 +1,31 @@
...
+  def data_transformer
+    CsvTransformer.new(%w[section_number student_local_id school_local_id course_number term_local_id grade])
+  end
+
+  def filter
+    SchoolFilter.new(school_scope)
+  end
+
+  def import_row(row)
+    student_section_assignment = StudentSectionGradeRow.new(row, student_lasid_map, section_number_map).build
```
kevinrobinson: I think it'd help clarify to move the lookups into these maps to this method, and not pass them to the row class.  That leaves the mapping as this importer's job, and also moves the code making the unique student/section key that's in the row class back here into the importer where it's all next to each other together.

kevinrobinson:
Hi @jhilde!  I was online and working a bit today so thought I could help.  After looking at that I realize that might not be particularly helpful since you and @alexsoble have more context here and were working on this earlier.  Sorry about that, please let me know if this is more distracting than helpful.

This looks great!  I've left some specific feedback on this PR as comments inline as you asked, and here are some thoughts for thinking about after this particular PR.

If I'm following, shipping this feature involved some back-and-forth between you, John and folks working at Aspen.  That's awesome, much respect for working across all those layers to get the data flowing that we need!  One thing to keep in mind is that this may be challenging to adapt for other districts; if we need to get humans to write custom code on their end in order to get this data, the cost of maintaining this in the product goes up quite a bit.  That's not to say we shouldn't do this, just a thing to consider.

One thing I pick up on here is you're being thoughtful about finding ways to optimize the run-time of this code.  This is an offline job though, and so the runtime isn't super important but the way that the data gets cut over is.  The approach here introduces risk to the end-user experience of data consistency - all the sections are deleted, and then they're imported, and users actively using the site at the time will experience what appear to be bugs around data disappearing and appearing.  We can sort of get away with this because of usage patterns and when we run the job, but this isn't ideal.  Are there ways we could focus our optimization efforts towards improving the user experience there?  No need to hold this pull request up, but managing tht seems like the major design decision for this part of the import (which I know was in a previous PR and not directly related to this change).

jhilde:
Thanks for the deep dive @kevinrobinson!

In terms of the grades...it's the one piece of data that seems to provide the most value for the high school educators. I believe there are ways to calculate the current grade based on data available in Aspen's database. It's just that the vendor already had implemented this for Somerville. The feed is custom but the Aspen functionality is standard. We should give some thought to other areas where there is customized data. For instance, from the column names, it seems like several of the student attributes are customizations of Aspen and may vary from district to district.

In terms of caching ids for the imports...initially, I was just considering caching the ids to speed things up, if necessary. As it stands though, the import takes > 4 hours so several of the imports aren't getting called. (I've pinged @alexsoble because I think most of that time is caused by caching the json for the name searchbar). Just wanted to do my part and not increase the import duration too much. ;-)

And you're correct in terms of the user experience during the import of student and educator assignments to sections, and adding grades to the student section assignment. The challenging thing is to accomodate deletes like when a student drops a section. Deleting the existing rows and doing a fresh import seems like the most straight-forward way to accomplish this within the current import paradigm. Thinking out loud, we could do this using student_section_assignments as an example:

--process the import file as usual
--for each record saved (new or updated), add the id to an array valid_ids
--StudentSectionAssignment.where.not(id: valid_ids).delete_all

Sounds like a better idea. Thanks for pushing this. I'll do that work under a separate PR since, as you mentioned, it isn't introduced in this PR.

app/importers/file_importers/student_section_assignments_importer.rb
jhilde: You're right, I'm too sensitive. The number of imported records for the assignments is currently 30,000. But yes, even with a 3 fold increase the id space would still accomodate this for a long time.

This also becomes moot when I change to the isolated deleting scheme I put in the comments anyways.

app/importers/file_importers/student_section_grades_importer.rb
jhilde: Yeah, it will be fine to map over them to concat.

app/importers/file_importers/student_section_grades_importer.rb
jhilde: Correct but not sure about adding it to the unique constraints of the Section model? The importer enforces the uniqueness. Is it possible to include a field from a parent table in the uniqueness constraint of a model?

jhilde:
I considered just removing the caching code but doing an extra 22,000 database hits every day for little reason still doesn't sit well with me.

alexsoble:
Hey there @jhilde! I've noticed that the import task is taking a long time to run as well! You mentioned that you think most of that time is caused by caching JSON for the name searchbar. I think that's quite possible, but I think it would be good for us to have data on how long each section of the import process is taking. Is there any tooling or logs you're looking at here? I was going to build some reporting around how long things are taking so we can start to debug but I want to make sure I'm not missing anything that already exists!

app/importers/file_importers/student_section_grades_importer.rb
kevinrobinson: Ah my bad, the Section uniqueness constraints work because on the `Section` model the Insights `course_id` contains both the (course_number, school local_id) bits.

As for uniqueness constraints across tables, you can definitely do this in Rails/ActiveRecord if not at the database level, which I think would be good to capture.  But here `course_id` does that so it's already covered.

kevinrobinson:
@jhilde Thanks for sharing what you're thinking, all sounds great and awesome to see all the work you've put in to getting this shipped!

re: the import job taking too long, here's how I'd triage.  The import task is currently broken (looks like it has been for over a week, I missed that earlier).  So that's user-impacting and means we have data quality issues right now, depending on how far the job gets when it gets killed.  Users may be entering things in Aspen and other users may not be seeing them in Student Insights.  So I'd vote to treat that like a production incident and drop everything to swarm on that first, and shipping https://github.com/studentinsights/studentinsights/pull/1051 seems like a good step!  I'll add some other quick thoughts on https://github.com/studentinsights/studentinsights/issues/1023.

Yeah, I think either 1) "update the existing data in-place" or 2) "fresh batch import each night and then cut over to the new data all at once" seem like two general approaches.  You idea sounds like it moves towards #1 which seems cool, and main complexity will be around missing records and sounds like a good way to tackle that!

jhilde:
From the logs it looks like @alexsoble's work to get the import to complete worked this morning.

I'm going to push this patch and rerun the importers to make sure it works before the Monday scheduled process. 

To be clear, this PR only gets the data into the database. Users will not be able to see grades yet. I'd like to make sure the data is accurate, etc before exposing it.

PR #1025, opened 2017-09-13
Move to eslint Airbnb JS Standards, take two

### LINKED ISSUE #1022: Use Airbnb eslint config?
The Airbnb JavaScript/React Style Guide is somewhat of a community standard: https://github.com/airbnb/javascript

It's also much stricter than our current `.eslintrc` and does a lot more autofixing around formatting and spacing, so it results in a cleaner and more consistent front-end codebase. I think this would be a good thing as our team expands.

Running it against our current codebase yields more than a thousand errors, even after autofixing. So I'd suggest a strategy of fixing a few easy ones up front, modifying the config to that we get to green, and gradually move closer to complying with the Style Guide over time.

# npm package
https://www.npmjs.com/package/eslint-config-airbnb

# branch where I started messing around with this
https://github.com/studentinsights/studentinsights/tree/eslint-airbnb

alexsoble:
# Notes 

+ Install `eslint-airbnb-config`
+ Check in large amounts of autofixes
+ Disable large amount of eslint/airbnb rules so we can get to green
+ Once this merges in, start peeling off the disables one by one to put our codebase onto the standard
+ Close #1022 
+ QA checks out alright locally


alexsoble:
Looks like precompile is breaking so we can't push to prod . . . got to revert and make sure precompile works before we merge!

PR #1028, opened 2017-09-14
Make Uglifier behave nicely when compressing ES6 JS

alexsoble:
Reverts studentinsights/studentinsights#1027.

Before I merge, need to fix this issue blocking deploy to Heroku:

```
remote:        rake aborted!
remote:        ExecJS::RuntimeError: SyntaxError: Unexpected token: punc ())
remote:        JS_Parse_Error.get ((execjs):3538:621)
remote:        (execjs):4060:47
remote:        (execjs):1:102
remote:        Object.<anonymous> ((execjs):1:120)
remote:        Module._compile (module.js:570:32)
remote:        Object.Module._extensions..js (module.js:579:10)
remote:        Module.load (module.js:487:32)
remote:        tryModuleLoad (module.js:446:12)
remote:        Function.Module._load (module.js:438:3)
remote:        Module.runMain (module.js:604:10)
remote:        /tmp/build_78e3eab438d06055df53be90ba0b637a/vendor/bundle/ruby/2.3.0/gems/execjs-2.7.0/lib/execjs/external_runtime.rb:39:in `exec'
remote:        /tmp/build_78e3eab438d06055df53be90ba0b637a/vendor/bundle/ruby/2.3.0/gems/execjs-2.7.0/lib/execjs/external_runtime.rb:21:in `eval'
remote:        /tmp/build_78e3eab438d06055df53be90ba0b637a/vendor/bundle/ruby/2.3.0/gems/execjs-2.7.0/lib/execjs/external_runtime.rb:46:in `call'
```

PR #1031, opened 2017-09-14
Remove arrow function from .js files

kevinrobinson:
@jhilde I'll leave it for you to decide what you think is best, and would vote for reverting immediately, but this is the other thing I found poking around.

Errors in IE:
screen shot 2017-09-14 at 7 08 25 pm
screen shot 2017-09-14 at 7 08 33 pm

And grepping finds three arrow functions in the compiled output, so this just manually removes them.  Fixing the build process is a better long-term idea, just trying to help with getting the site back up ASAP.

@jhilde I'm stopping and won't take any action here, but let me know if I can help with anythign else!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ install:
  - bundle _1.10.6_ install --retry=3
 
 script:
-  - npm run lint-cli
+ # - npm run lint-cli #turning off linting for the moment to correct the production issue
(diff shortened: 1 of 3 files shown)


kevinrobinson:
@jhilde @alexsoble clarifying: this was just to share what I found debugging, not intended to be merged.  I think we should revert https://github.com/studentinsights/studentinsights/pull/1028 to bring the site up, then investigate how to fix the issue in 1028 while still compiling down to ES5.

PR #1036, opened 2017-09-15
Isolate Rollbar loading from the rest of the application code

kevinrobinson:
So that if there are fatal things like parse errors, monitoring still works.  This has three changes aimed at improving JS error reporting.

First, it inlines the Rollbar monitoring code into the application template, so that it runs and loads independently of what our application code does.  It's not compiled in the build process and the intention is that it will almost never change.  In order to respect environmental config (eg don't report local development errors), it reads those values from application code on-demand at error time.  If it can't read the environment config, it defaults to reporting.  https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/47/ Shows an example of this that was generated by reproducing the error with IE parsing Harmony code on the demo site.

Second, it adds an environment variable for `SHOULD_REPORT_ERRORS` and I set that to be true for the demo site as well.  Previously it had just checked `SHOULD_REPORT_ANALYTICS` so this splits that out with the idea that error reporting is always on, and deploying this would add monitoring to the demo site.

Third, since the demo site runs in "production" mode, this also reports the `deploymentKey` from the environment config so that we can distinguish between errors coming from demo or production site (you can also tell by the URL).  The intent here is similar to the deploy script improvements, to treat the demo site as a first class deployment.


kevinrobinson:
need to test this locally with https://github.com/studentinsights/studentinsights/pull/1028 and see if it would catch it (working around the `Env` check).

kevinrobinson:
@alexsoble @jhilde Does this seem reasonable?

alexsoble:
@kevinrobinson Yeah seems really smart!

kevinrobinson:
Confirmed logging errors works like normal: https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/48/

PR #1037, opened 2017-09-15
Add note to docs on using IE11 to test

kevinrobinson:
With links to free VMs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ Our presentation at [Code for Boston demo night](docs/readme_images/Student%20In
   - [3. Start Rails](#3-start-rails)
   - [4. Run the tests](#4-run-the-tests)
   - [5. Write code!](#5-write-code)
+  - [6. Use the product locally!](#6-use-product)
 - [Browser/OS Targeting](#browseros-targeting)
 - [Deployment](#deployment)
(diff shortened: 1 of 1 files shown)


kevinrobinson:
@alexsoble @jhilde What do you pals think?

alexsoble:
@kevinrobinson I think these are great resources! To be honest, I tried to use IE (via Browserstack) to test and develop earlier in the project, but I fell off the horse eventually because I use Chrome for everything else. I'm trying to think of ways to make it more difficult for us devs to develop on Chrome or at least give us big flashing warnings about it. Like maybe a big red banner on the top when you're logged in as the admin user reminding you that "this site is meant to use in IE"? This PR should definitely ship, just thinking about how to make it stick!

kevinrobinson:
Cool, yeah I used BrowserStack a bit before too but also pretty inconsistently.  I think the VM workflow can help, since then it's just running right on your computer and basically no different than Chrome.  The developer tools are 85% as good in IE11 now too.

Warnings sound good too, https://www.npmjs.com/package/detect-browser could be a warning as well, that runs only in development or for a particular user.

alexsoble:
👍

PR #1040, opened 2017-09-16
Update HTML rendering to force closing HTML tags

kevinrobinson:
Addressing console warnings on Windows 7 IE11:
screen shot 2017-09-16 at 12 01 01 pm

This also removes some templates that were redundant with `shared/serialized_data` and makes a small changes to remove rendering a "back" link in the restricted notes page (there's a back button in the browser, so unless there's a particular reason to keep it seems redundant).

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -22,4 +22,10 @@ module ApplicationHelper
     @devise_mapping ||= Devise.mappings[:user]
   end
 
+  # IE11 reports HTML1500 warnings on the console if tags are not explicitly
+  # closed (like happens if you used `tag`).  Here we're rendering tags with
+  # attributes and no content to be able to parse the JSON in JS.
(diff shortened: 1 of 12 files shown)


app/helpers/application_helper.rb
```diff
@@ -22,4 +22,10 @@ def devise_mapping
     @devise_mapping ||= Devise.mappings[:user]
   end
 
+  # IE11 reports HTML1500 warnings on the console if tags are not explicitly
+  # closed (like happens if you used `tag`).  Here we're rendering tags with
+  # attributes and no content to be able to parse the JSON in JS.
+  def json_div(options = nil)
```
kevinrobinson: I think we should bias towards shipping, but @alexsoble if you happen to know a better way to do this in Rails that's awesome too.  You have much knowledge of the Rails-y ways :)

alexsoble:
👍

Issue #976, opened 2017-08-22
Service Gets Recorded Even if No Teacher is Selected


Jngai:
Under services, you don't have to specify the teacher for the form to post. I am wondering if this is a bug or was this intended by design?

At the moment, the form only checks if date and service type id exist before allowing access to record service.

![hello 1](https://user-images.githubusercontent.com/1547934/29580468-1cb20072-8744-11e7-8ec6-3399854da2cc.gif)

kevinrobinson:
@Jngai Thanks!

I think the current behavior was intentional, but seems like it'd be a good idea to make it stricter.  Checking the production database, there are some records that do in fact have no educator entered, but it's only ~1% of all services recorded.

The changes in https://github.com/studentinsights/studentinsights/pull/1041/files#diff-6e6d6d92b78948abf256f6dfaaa28bb4R95 change this to be stricter (but still allow empty string in the database for the older records), and adds a test for this as well.

PR #1041, opened 2017-09-16
Fix bug with date state when recording services

kevinrobinson:
Addressing https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/41/ and usability issues with the date picker.  If users tried to edit the text manually (ie, not using the datepicker with the mouse), on changes it would change the date to unexpected values (as the component parsed the text into a moment and then rendered it back out to text).

This updates the `RecordService` component to track the raw text as state, and then parses that as needed into a moment or particular format.  If it fails to parse into a valid date (just the format, not looking at when the date is in time), then the UI shows a message and the user is prevented from saving.  This also adds some tests and refactors some test helpers.

Screenshot showing input allowing arbitrary text, but warning on invalid date and disabling save:
screen shot 2017-09-16 at 2 57 00 pm

### DIFF
--- a/app/assets/javascripts/student_profile/record_service.jsx
+++ b/app/assets/javascripts/student_profile/record_service.jsx
@@ -26,6 +26,10 @@
       padding: 5,
       width: '50%'
     },
+    invalidDate: {
+      color: 'red',
+      padding: 5
+    },
     serviceButton: {
       background: '#eee', // override CSS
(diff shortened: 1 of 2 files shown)


kevinrobinson:
This also resolves https://github.com/studentinsights/studentinsights/issues/976.

PR #1042, opened 2017-09-16
Relax propType check for RiskBubble risk level

kevinrobinson:
Warning in development mode:
screen shot 2017-09-16 at 3 20 46 pm

`null` is a valid value according to the component code and what's in the database, so making this prop required doesn't make sense.  What we really want is "you must pass the prop as a key, even if the value is null or undefined," but that's not what `.isRequired` means.


kevinrobinson:
selfie

PR #1045, opened 2017-09-16
Testing mention bot #4

kevinrobinson:
testing...

PR #1046, opened 2017-09-16
Making .mention-bot valid JSON

kevinrobinson:
so that it will work :)

see https://github.com/studentinsights/studentinsights/pull/1045#issuecomment-329992968


kevinrobinson:
selfie

PR #1048, opened 2017-09-16
Add note about studentinsights-bot in CONTRIBUTING

kevinrobinson:
So that there's a note about how this happens.


kevinrobinson:
@alexsoble @jhilde I added y'all to the Heroku app running this.  It's set to ping on changes to `production.rb` and the `Procfile` and can be configured here: https://github.com/studentinsights/studentinsights/blob/master/.mention-bot#L8

Does this seem useful or silly?

alexsoble:
Seems useful -- I'd suggest we add `.eslintrc` as well since changing that is what brought down the site in this PR: https://github.com/studentinsights/studentinsights/pull/1028/files.

I'll open a PR to do that.

alexsoble:
Let's 🚢 !

PR #1055, opened 2017-09-17
Add ImportRecord when emailing about failed import job

kevinrobinson:
So we can see it right in the email, and also use `<pre>` to format stack traces.


alexsoble:
I love this change. I'm going to remove the `timing_log` from the mailer because it's sending too much sensitive information over email. :/

PR #1059, opened 2017-09-18
Only parse/save recent attendance events

alexsoble:
# Notes

+ We're spending a lot of time and energy in the attendance import doing lookups on attendance events that happened far in the past, are already in the database, and haven't changed
+ Skip attendance rows that are more than four days in the past
+ Technically we could set it to 1 or 2 days since the import task runs nightly, but I'm thinking that if the import task fails one or two nights we don't want to risk skipping any event rows

### DIFF
--- a/app/importers/file_importers/attendance_importer.rb
+++ b/app/importers/file_importers/attendance_importer.rb
@@ -13,9 +13,7 @@ class AttendanceImporter < Struct.new :school_scope, :client, :log, :progress_ba
   end
 
   def import_row(row)
-    occurred_at = DateTime.parse(row[:event_date])
-
-    return if Time.current - 90.days > occurred_at
+    return if Time.current - 90.days > row[:event_date]
 
     AttendanceRow.build(row).save!
   end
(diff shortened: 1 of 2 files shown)


jhilde:
Makes sense. The other strategy I used for Grades is to cache lookups with a single DB call in the initializer. For grades, it caches student lasid -> students.id and a uniqueness hash -> sections.id. 

But yeah, that's 400,000+ rows in the attendance file. Ugh. 

My performance (and data quality) mind wants to investigate using triggers or snapshotting to get deltas from Aspen. Would greatly reduce import time and would allow more frequent imports.

alexsoble:
@jhilde Cool! Could you say more about the triggers or snapshotting strategies and how you would implement them?

jhilde:
@alexsoble Triggers would likely be out of the question since, rightfully so, John wouldn't want to modify the Aspen Database. Essentially, you create a trigger on create, update, delete that calls a stored proc which logs the delta, likely in a separate table. You can then query that table to get deltas.

There's a data_audit table in Aspen, I'm wondering it this already exists. (Much longer term though).

Not so sure about snapshotting, back in the day I think Oracle had a mechanism to keep two databases in sync. Not sure about MySQL.

alexsoble:
@jhilde Cool, thanks!

alexsoble:
Note from Kevin: consider changing the time limit to something more like 90 days, since we're not sure about the human processes around data entry -- people could be entering attendance issues late, we would have no idea.

alexsoble:
Did some local QA on this, it imports the right things and nothing breaks.

PR #1060, opened 2017-09-18
Sorting section student names by last name

jhilde:
Based on feedback from the high school teachers, they really wanted to see the student names alphabetized by last name, not first name. Makes sense.


jhilde:
Tested with IE11...and 🚢 ing.

PR #1072, opened 2017-09-20
Turn on grades in section view for all educators

jhilde:
This is a continuation of #1070 which added the grade column for section view. Initially this was only turned on for educators with districtwide_access set to true for data validation. With validation complete, I'm now turning the column on for all educators (with permissions to the section of course).


jhilde:
Checked on IE11 ✔️

Issue #1074, opened 2017-09-21
Remove redundant StudentRiskLevel records and add constraint/validation


kevinrobinson:
`StudentRiskLevel` is the largest table in the database, and there are ~100x times as many records there as there are students.

Here's what I see with a sample of 100 student records:
```
# get a sample
ss = Student.all.sample(50)

# how many student risk levels are there for each student in the sample?
ss.map {|s| StudentRiskLevel.where(student_id: s.id).size }.sort
=> [1, 1, 2, 2, 2, 3, 3, 3, 11, 13, 14, 15, 16, 17, 17, 17, 27, 28, 30, 38, 118, 122, 124, 131, 133, 137, 138, 138, 139, 140, 140, 140, 141, 142, 143, 144, 145, 146, 155, 157, 165, 168, 170, 172, 173, 181, 182, 184, 188, 194]
```

So this means:
- there is an active bug in the import process, or there was in the past and we didn't clean up the extra records
- there's no constraints or validations ensuring that each student only has one risk level

kevinrobinson:
It seems like the behavior of `#create_student_risk_level!` is a bit complicated, and not what I expected.  Rails injects this method when the association is created.  Calling it when there already is a risk model raises, but first it deletes the existing `StudentRiskLevel`:

```
irb(main):001:0> s = Student.first.reload
=> #<Student id: 1, grade: "3", ... risk_level: 3, gender: "M">
irb(main):002:0> s.student_risk_level
=> #<StudentRiskLevel id: 91, student_id: 1, level: 3, created_at: "2017-09-24 16:52:25", updated_at: "2017-09-24 16:52:25", mcas_math_risk_level: 3, star_math_risk_level: 1, mcas_ela_risk_level: 0, star_reading_risk_level: 1, limited_english_proficiency_risk_level: nil>
irb(main):003:0> s.create_student_risk_level!
   (0.4ms)  BEGIN
  StudentRiskLevel Exists (0.6ms)  SELECT  1 AS one FROM "student_risk_levels" WHERE "student_risk_levels"."student_id" = $1 LIMIT $2  [["student_id", 1], ["LIMIT", 1]]
   (0.3ms)  ROLLBACK
   (0.2ms)  BEGIN
  SQL (0.5ms)  DELETE FROM "student_risk_levels" WHERE "student_risk_levels"."id" = $1  [["id", 91]]
   (2.0ms)  COMMIT
[...]

PR #987, opened 2017-08-25
Skip students with future registration dates

alexsoble:
# Issue

+ #986

PR #1075, opened 2017-09-21
Fix bug with student importer creating more risk level records

kevinrobinson:
Addressing part of https://github.com/studentinsights/studentinsights/issues/1074.

I think there are two bugs here.  One is that the importer calls `create_risk_level!` for every row that's processed during import, even for `Student` records that already exist and are just being updated (or haven't changed at all).  The other is that the check for registration dates in the future only applies to the `student.save!` call, and so other actions like assigning a homeroom or creating a risk level would still run.  I don't think this is what we want based on my read of https://github.com/studentinsights/studentinsights/pull/987.

### DIFF
--- a/app/importers/file_importers/students_importer.rb
+++ b/app/importers/file_importers/students_importer.rb
@@ -18,14 +18,14 @@ class StudentsImporter < Struct.new :school_scope, :client, :log, :progress_bar
 
   def import_row(row)
     student = StudentRow.new(row, school_ids_dictionary).build
+    return if student.registration_date_in_future
 
-    student.save! unless student.registration_date_in_future
+    student.save!
+    student.create_student_risk_level!
 
     if row[:homeroom].present?
       assign_student_to_homeroom(student, row[:homeroom])
     end
-
-    student.create_student_risk_level!
   end
 
   def assign_student_to_homeroom(student, homeroom_name)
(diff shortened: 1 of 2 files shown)


alexsoble:
Did some rails consoling and confirmed that there are way more `StudentRiskLevel` objects in the database than we want/need:

```
irb(main):012:0> Student.count
=> 5845
irb(main):013:0> StudentRiskLevel.count
=> 570800
```

We only need one per student since we're not doing anything with historical student risk levels. Great catch @kevinrobinson.

alexsoble:
I might be missing something, but I think this PR so only fixing one of the bugs: the one where database-affecting code still runs even when students are registered in the future. 

To test the second bug, we'd need to run `import` multiple times in the spec to see if multiple student risk level records get created; this spec only runs `import` once.

I think we should merge in this code and then tackle the second bug (with creating multiple student risk level records) in a separate PR, happy to handle that one.

kevinrobinson:
@alexsoble thanks for checking this closely! 👍 

In https://github.com/studentinsights/studentinsights/pull/1075/files#diff-762dc5b20df365311d4f9b4535ea4956R60, I'm trying to create a record that is about to be imported, and then check that only three new records are created on import.  So I think that'll test what we want.

But running `import` twice is a simpler better idea!  I'll update that next week and deploy, or if you want to push this along sooner feel free to do that too.

kevinrobinson:
@alexsoble I added the "run import twice" test case https://github.com/studentinsights/studentinsights/pull/1075/files#diff-762dc5b20df365311d4f9b4535ea4956R73, thanks!

PR #1076, opened 2017-09-21
Remove risk level chart computation from homeroom

kevinrobinson:
This came up from the same line of inquiry in https://github.com/studentinsights/studentinsights/issues/1074.  Looking at index usage by table with `heroku pg:index-usage`, it looks like `student_risk_levels` is only using an index 68% of times.  I grepped around and found this usage in the homerooms page.

This isn't used anywhere in the product so this removes it.  The bug in https://github.com/studentinsights/studentinsights/issues/1074 means this was also doing a table scan on ~500k records.

### DIFF
--- a/app/controllers/homerooms_controller.rb
+++ b/app/controllers/homerooms_controller.rb
@@ -10,10 +10,6 @@ class HomeroomsController < ApplicationController
 
     @rows = eager_students().map {|student| fat_student_hash(student) }
 
-    # Risk level chart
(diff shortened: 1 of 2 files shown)


alexsoble:
Right, this is the color chart that used to be at the top of the Homeroom Roster page. Really good catch Kevin.

alexsoble:
Let's 🚢 .

kevinrobinson:
👍 thanks @alexsoble!

PR #1084, opened 2017-09-22
Use 3.10.1 lodash functions consistently

alexsoble:
# New thing

Fix lodash method inconsistency:

We're using `_.pluck`, which is exists in 3.10.1 but not 4.17.4.

And we're using `_.isInteger`, which exists in 4.17.4 but not 3.10.1.

And use yarn to manage lodash instead of a static copy in the `vendor/assets` folder. 

# Old thing 

~~Hey @jhilde and @kevinrobinson!~~

~~Do either of you get this error from Browserify when you run `rails server` locally?~~

~~I've been getting it and it's blocking me from developing on my machine:~~

```
Error: Cannot find module 'lodash/isInteger' from '/Users/alexsoble/somerville-teacher-tool/app/assets/javascripts/helpers'
```

~~Maybe it's overkill but I decided to just pull lodash into the yarn pipeline, add to package.json, and use `import` everywhere. If you were having the same error as me, does this seem like a good fix? If not, does this look like a good change anyway?~~


kevinrobinson:
@alexsoble I can't reproduce on a clean build.  If you `rm -rf node_modules` and reinstall, do you still see it?

alexsoble:
@kevinrobinson Yep, that fixed the error! 

Was curious about why the Travis build failed for this and looking into it showed me that we're using lodash inconsistently. 

We're using `_.pluck`, which is exists in 3.10.1 but not 4.17.4.

And we're using `_.isInteger`, which exists in 4.17.4 but not 3.10.1.

`_.isInteger` is only used to check for non-numeric values in numeric table sorts, which is a rare edge case and isn't actively throwing errors in the app. So it's not a high-priority fix. 

Not urgent at all but I'll aim to land this PR to clean up the inconsistency and guard against those non-numberic values that might creep into number columns and break the sort.

alexsoble:
Features to double-check:

+ `graph_helpers.js`
+ Student Report PDF features

alexsoble:
Oy upgrading to 4.17.4 is a nightmare, let's say at 3.10.1 and use `isNumber` instead of `isInteger`.

kevinrobinson:
@alexsoble Yeah if I remember right there are actually two separate copies right now because of the way we're building both through the asset pipeline and through browserify.  The best long-term solution is to keep chipping away at migrating things away from the asset pipeline (or do a sprint to finish that), and that should get simpler.  lodash does have breaking changes but most I think are mechanical (eg `_.pluck(things, 'field')` is now just supported by `_.map(things, 'field')`.

alexsoble:
Agreed on the mechanical changes! 

Tested this branch on an IE VM, looks like everything works good with only one copy of lodash (managed via yarn) including graphs and PDF reports.

alexsoble:
Looks good in prod.

Issue #1071, opened 2017-09-20
Fix data integrity job


kevinrobinson:
It has been failing integrity checks, and we haven't been notified :(

From last night:
```
20 Sep 2017 03:23:36.518128 <190>1 2017-09-20T07:23:35.148025+00:00 app scheduler.2954 - - Validating Assessments...
20 Sep 2017 03:23:36.518126 <190>1 2017-09-20T07:23:35.202258+00:00 app scheduler.2954 - - Validating Educators...
20 Sep 2017 03:23:41.442125 <190>1 2017-09-20T07:23:41.142957+00:00 app scheduler.2954 - - Validating Students...
20 Sep 2017 03:23:42.028116 <190>1 2017-09-20T07:23:41.744696+00:00 app scheduler.2954 - - rake aborted!
20 Sep 2017 03:23:42.028178 <190>1 2017-09-20T07:23:41.744710+00:00 app scheduler.2954 - - ActiveRecord::RecordInvalid: Validation failed: Grade must be a valid grade
20 Sep 2017 03:23:42.028220 <190>1 2017-09-20T07:23:41.745001+00:00 app scheduler.2954 - - /app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.3/lib/active_record/validations.rb:78:in `raise_validation_error'
20 Sep 2017 03:23:42.028203 <190>1 2017-09-20T07:23:41.745003+00:00 app scheduler.2954 - - /app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.3/lib/active_record/validations.rb:50:in `save!'
20 Sep 2017 03:23:42.028215 <190>1 2017-09-20T07:23:41.745003+00:00 app scheduler.2954 - - /app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.3/lib/active_record/attribute_methods/dirty.rb:43:in `save!'
20 Sep 2017 03:23:42.028214 <190>1 2017-09-20T07:23:41.745004+00:00 app scheduler.2954 - - /app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.3/lib/active_record/transactions.rb:313:in `block in save!'
20 Sep 2017 03:23:42.028242 <190>1 2017-09-20T07:23:41.745005+00:00 app scheduler.2954 - - /app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.3/lib/active_record/transactions.rb:384:in `block in with_transaction_returning_status'
[...]

PR #1089, opened 2017-09-24
Add validation and tests on StudentRiskLevel, migrate to destroy extra records

kevinrobinson:
Trying to close out https://github.com/studentinsights/studentinsights/issues/1071

The primary changes are to:
- add a validation on `StudentRiskLevel` to enforce that it's unique on `student_id`
- add a migration to delete stray `StudentRiskLevel` records

Incidentally, this updates `Student#update_risk_level` is two ways.  One is to make the `self.save` call raise on an error so that validation errors bubble up, and to replace `create_student_risk_level!` with more explicit calls since that method had unexpected behavior.

### DIFF
--- a/app/importers/file_importers/students_importer.rb
+++ b/app/importers/file_importers/students_importer.rb
@@ -21,7 +21,7 @@ class StudentsImporter < Struct.new :school_scope, :client, :log, :progress_bar
     return if student.registration_date_in_future
 
     student.save!
-    student.create_student_risk_level!
+    student.update_risk_level!
 
     if row[:homeroom].present?
       assign_student_to_homeroom(student, row[:homeroom])
--- a/app/models/student.rb
+++ b/app/models/student.rb
@@ -192,25 +192,27 @@ class Student < ActiveRecord::Base
 
   ## RISK LEVELS ##
 
-  def self.update_risk_levels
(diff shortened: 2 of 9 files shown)


kevinrobinson:
@alexsoble I think you're right about your comment in https://github.com/studentinsights/studentinsights/pull/1075#issuecomment-331568152 too and I missed that twice, even when you pointed it out.  I got a bit mixed up looking at this, particularly with https://github.com/studentinsights/studentinsights/issues/1074#issuecomment-331726052, do you might checking this out and seeing if it makes sense?

alexsoble:
Will dig into this and review later today @kevinrobinson!

kevinrobinson:
@alexsoble sure, that sounds good to me.  Updated it.

kevinrobinson:
oops, I messed up and didn't push the commit moving this to a class.  There's no harm as a migration so I think fine to leave it as is now.

I ran this in production and verified that running `student.update_risk_levels!` works as expected, and that the migration did was it was trying to do.

alexsoble:
@kevinrobinson 

```
irb(main):002:0> StudentRiskLevel.count
=> 5867
irb(main):003:0> Student.count
=> 5867
```

Woohoo!

PR #1090, opened 2017-09-24
Fix bug in StudentServiceImporter

kevinrobinson:
This was introduced in https://github.com/studentinsights/studentinsights/pull/1066/files and came up in manually testing that those changes worked as expected.

@alexsoble it looks to me like this job doesn't run and it's intended to be run as a one-off.  The folder doesn't exist on the box either.  I figure those are both as expected, but FYI in case not :)

### DIFF
--- a/app/importers/student_services/student_service_importer.rb
+++ b/app/importers/student_services/student_service_importer.rb
@@ -29,7 +29,7 @@ class StudentServiceImporter
   end
 
   def files_in_remote_server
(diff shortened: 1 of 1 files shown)


alexsoble:
Hey @kevinrobinson! 

This is indeed a leftover type class. I used it to trigger imports of Services CSV files from the command line before I built the Student Services Upload UI so that Uri could upload services himself through the web app.

PR #1093, opened 2017-09-25
Increase delayed_job max run time again

alexsoble:
+ Since it's timing out at 28800 seconds/8 hours


alexsoble:
Lol at @studentinsights-bot, thank you! :D

PR #1097, opened 2017-09-25
WIP - Ability to select an end date for services

Jngai:
Answering all the original scenarios originally mentioned here 

https://github.com/studentinsights/studentinsights/pull/1004

### DIFF
--- a/app/assets/javascripts/student_profile/api.jsx
+++ b/app/assets/javascripts/student_profile/api.jsx
@@ -48,7 +48,7 @@
         service: {
           service_type_id: serviceParams.serviceTypeId,
           date_started: serviceParams.dateStartedText,
-          date_ended: serviceParams.dateEndedText,
+          estimated_end_date: serviceParams.estimatedEndDateText,
           provided_by_educator_name: serviceParams.providedByEducatorName,
           student_id: studentId
         }
--- a/app/assets/javascripts/student_profile/api.jsx
+++ b/app/assets/javascripts/student_profile/api.jsx
@@ -48,6 +48,7 @@
         service: {
           service_type_id: serviceParams.serviceTypeId,
           date_started: serviceParams.dateStartedText,
+          date_ended: serviceParams.dateEndedText,
           provided_by_educator_name: serviceParams.providedByEducatorName,
           student_id: studentId
         }
--- a/app/assets/javascripts/student_profile/record_service.jsx
+++ b/app/assets/javascripts/student_profile/record_service.jsx
@@ -68,7 +68,8 @@
       return {
         serviceTypeId: null,
         providedByEducatorName: "",
-        dateStartedText: nowMoment.format('MM/DD/YYYY')
+        dateStartedText: nowMoment.format('MM/DD/YYYY'),
+        dateEndedText: nowMoment.format('MM/DD/YYYY')
       };
     },
 
@@ -77,8 +78,17 @@
       return moment.utc(dateStartedText, 'MM/DD/YYYY', true); // strict parsing
     },
 
+    dateEndedMoment: function() {
+      const {dateEndedText} = this.state;
+      return moment.utc(dateEndedText, 'MM/DD/YYYY', true); // strict parsing
+    },
+
     onDateTextChanged: function(dateStartedText) {
       this.setState({dateStartedText});
+    },
+
+    onEndDateTextChanged: function(dateEndedText) {
+      this.setState({dateEndedText});
     },
 
     onProvidedByEducatorTyping: function(event) {
@@ -101,12 +111,14 @@
       const {serviceTypeId, providedByEducatorName} = this.state;
       const {currentEducator} = this.props;
       const reformattedDateText = this.dateStartedMoment().format('YYYY-MM-DD');
+      const reformattedEndDateText = this.dateEndedMoment().format('YYYY-MM-DD');
 
       // Get the value of the autocomplete input
       this.props.onSave({
         serviceTypeId,
         providedByEducatorName,
         dateStartedText: reformattedDateText,
+        dateEndedText: reformattedEndDateText,
         recordedByEducatorId: currentEducator.id
       });
     },
@@ -175,6 +187,7 @@
 
     renderWhoAndWhen: function() {
       const isValidDate = this.dateStartedMoment().isValid();
+      const isValidEndDate = this.dateEndedMoment().isValid();
 
       return (
         <div>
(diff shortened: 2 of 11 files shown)


Jngai:
Hey @alexsoble I am almost done with this. I am getting a lint error above. Please see my logs and my terminal below. Have you seen this error?

```
23 error Exit status 1
24 error Failed at the studentinsights@0.1.0 lint-cli script 'eslint --ext jsx --ext js -c .eslintrc app/assets/javascripts spec/javascripts'.
24 error Make sure you have the latest version of node.js and npm installed.
24 error If you do, this is most likely a problem with the studentinsights package,
24 error not with npm itself.

Judys-MacBook-Pro:_logs judyngai$ vim 2017-09-25T20_35_24_547Z-debug.log
Judys-MacBook-Pro:_logs judyngai$ node --version
v7.10.1
Judys-MacBook-Pro:_logs judyngai$ npm --version
4.2.0
Judys-MacBook-Pro:_logs judyngai$ node.js --version
-bash: node.js: command not found
Judys-MacBook-Pro:_logs judyngai$ node --version
v7.10.1
Judys-MacBook-Pro:_logs judyngai$ brew info node
node: stable 8.5.0 (bottled), HEAD
Platform built on V8 to build network applications
https://nodejs.org/

```

alexsoble:
Hey @Jngai! 
 
The key lines are directly above the snippet that you posted: 

```
/home/travis/build/studentinsights/studentinsights/app/assets/javascripts/student_profile/services_list.jsx
  180:7  error  Unexpected console statement  no-console
  181:7  error  Unexpected console statement  no-console
  182:7  error  Unexpected console statement  no-console
```

This is the linting tool reminding you to delete `console.log` statements that shouldn't be in the production build.

Jngai:
Great! it passed thanks @alexsoble

alexsoble:
@Jngai Great! Reviewing now. If you could add a little more context to the description for this PR up above (like brief description of the changes, link to the issue fixed, and screenshot of UI changes) — that would be fantastic. Also let's be sure to QA any UI changes in Internet Explorer.

alexsoble:
Hey @Jngai! My highest-level question I have for you about this PR is on the addition of a `date_ended` field to Service. 

In my mind, adding a `date_ended` field to Service competes with the existing DiscontinuedService model. If I want to know the truth about when a service ended, do I look at the `date_ended` or at the service's DiscontinuedServices? Normally these should be the same, but no codebase is perfect and over time differences will creep in — and if those differences creep in, which value should the developer trust? This is the "single source of truth" idea.

Overall my impression is that using a DiscontinuedService model instead of a simple `date_ended` field adds more complexity than we need. It does let us capture which educator is responsible for ending the service, but we could also capture that by adding an `ended_by` educator field on the Educator model. The `DiscontinuedService` model also lets educators discontinue services multiple times, which I think is more complexity than we need.

So overall, for this PR I'd like us to pick one strategy and stick with it all the way — either add a `date_ended` and `ended_by` migration to Service and make DiscontinuedServices obsolete, or refrain from adding `date_ended` and leaving DiscontinuedService as the sole source of truth about when services end. 

What do you think @Jngai? Going with the second option will make it way simpler for us to land this PR and we could then simplify later. The first option will require running a fairly high-stakes migration to remove the DiscontinuedServices model, so my instinct is that we should hold on that for now.

Jngai:
Hey @alexsoble When I have more final UI changes I will put it in the description above. In the mean time here is my progress.

I close the original pr because I noticed there were lots of changes to the original files that I was working on since I worked on that old pr. I had to rewrite it but I carried the ideas over. Right now I am trying to answer all the scenarios as listed in our last conversation which I linked above. 

The changes in the code is mostly in record services and is mostly about building a new box, capturing the value of the box, passing it through to the api, getting it white listed and stored correctly. The natural flow of it already exist and I reuse similar architecture that is already in place.

**1. Start in the Past, End in the Past** 
pastpast

This is relatively the same as my last pr. Here during the post to the controller, it does a check to see if date_ended has already ended and if its yes then we create a discontinued service object passing in the service id, current educator id, and discontinued at etc. 

**2. Start in Past, End in Future** 
startinpastendinfuture

-  You mentioned we should add a "Scheduled to end [datetime]". I added a new field `date_ended` to services for this particular purpose and reused your "renderDateStarted" function and created my own "renderDateEnded" function to show the date ended for this particular UI. In the "When did/will they end?" box I added the `date_ended` field  as an extra params to be posted to the api and stored through the controller. 

- For the second comment about the button text "Discontinue Early". For the past tense, there won't be a button display. So if its started in the past then end in future or whether its start in future end in the past then its still "discontinue early" for both scenarios. I just changed the text. 

**3. Start in Future, End in Future** 

startinfutureendinfuture

I am having a little bug going on with the `start today/end today` section. I will work on this later but in time I commented it out and made the datetime explicit. 

I think we should continue using discontinued service model. For the 3rd scenario it really can't be discontinued because in discontinue model there is `must_be_discontinued_after_service_start_date` validator. Tests start breaking when I comment this out. 

From what I remember Discontinue Service explicitly creates a new row object that stores a service that _has already ended_. hence we can't really use this to schedule something to end. But for the second scenario the discontinue button function is intact and it will still discontinue. I see the `date_ended` as another information gathering. I do think discontinue service adds complexity. I think this is beyond this pr at the moment. I will work more on it today.

Jngai:
@alexsoble Also what led me to add date_ended to the table is because I was passing it through the original input box and it was a matter of storing it or not.

Jngai:
Also @alexsoble I am having a small issue with this commit
https://github.com/studentinsights/studentinsights/pull/1097/commits/04160a820019051eae5186509a5173e4262e8a0e

I just couldn't get the operators (< and >) right.

Jngai:
I will explore some more if I can stop relying on `date_ended`.

Jngai:
Hey @alexsoble I decided to change `date_ended` to `estimated_end_date` instead. I think from an user point of view, the future end date selection that they pick may not be exactly the date that actually ends so calling it an `end_date` is not completely correct. Its more of an estimated date when service ends. The `DiscontinuedService` model can then handles the absolute `discontinued_at` scenario. Also the option to display `service.discontinued_services.discontinued_at ` for scheduled at end in the UI isn't there without already creating a Discontinuedservice object and making the discontinue button obsolete since a Discontinuedservice object already exist. This way DiscontinuedService is still the sole source of truth.

alexsoble:
@Jngai — OK, that makes a lot of sense and really does help clarify! 

I did some more manual exploring and I came across one edge case that causes a bug: when the user tries to delete a service with a future start date: 

impossible-to-discontinue

I checked in the demo app in production and verified that this bug exists in production right now and was not introduced by your PR at all. So maybe improving the UI for this experience is the next chunk of work to go into your queue since now you're very familiar with this part of the codebase!

alexsoble:
Overall this PR looks great, thanks @Jngai, merging now!

Jngai:
@alexsoble I am glad that is good for a merge. For that edge case, its because of this method `must_be_discontinued_after_service_start_date` as a validator in Discontinuedservice model. I think its written because you don't need to discontinue something that haven't started. I can definitely look into the discontinued service side and put it in my queue.

Issue #1082, opened 2017-09-22
School Administrator Dashboard (built from Power BI)


snoopyuri:
We mocked up a dashboard based on feedback from principals and based on PowerBI mock-ups that Uri and Jill did with Microsoft fellow.  Eric will be working on the initial stage of this by starting with attendance graphs.  PPT and screenshots attached.

# Example graph: Attendance
![screen shot 2017-09-22 at 9 44 29 am](https://user-images.githubusercontent.com/3209501/30749961-b4c6edde-9f7a-11e7-83cc-86d7de25f881.png)

alexsoble:
👍 👍 

Uri's second issue on GitHub!

kevinrobinson:
we made some of these!

PR #1098, opened 2017-09-25
WIP - School Absence Dashboard #1082

edavidsonsawyer:
This is the first of several school level dashboards designed to give principals and assistant principals an overview of trends in attendance, discipline, and academic performance. 

This PR implements an absence dashboard which shows administrators the average daily attendance both for the school and per homeroom as well as providing a list of absences per student. By default it will show this data from the beginning of August through the current date. However, a user can change the time covered by the charts like so: 

admindash


kevinrobinson:
@edavidsonsawyer I submitted https://github.com/edavidsonsawyer/studentinsights/pull/1, which is a pull request into your branch here that you can merge into your work here to resolve the merge conflicts introduced in https://github.com/studentinsights/studentinsights/pull/1182.

kevinrobinson:
@edavidsonsawyer I pushed another commit that you might want to check out as well, see the comment https://github.com/edavidsonsawyer/studentinsights/pull/1#issuecomment-339867048 (not sure if you get notifications on those since they're in another repo).

edavidsonsawyer:
@kevinrobinson I should have commented here, but I [responded to your comment.](https://github.com/edavidsonsawyer/studentinsights/pull/1#issuecomment-340000311) Basically that all looks good, except I'd need to modify the highcharts wrapper before I could use it in the dashboard.

app/assets/javascripts/school_administrator_dashboard/dashboard_bar_chart.jsx
```diff
@@ -0,0 +1,126 @@
+//import _ from 'lodash';
```
alexsoble: @edavidsonsawyer Let's remove this line if we're not using lodash.

app/assets/javascripts/school_administrator_dashboard/dashboard_bar_chart.jsx
```diff
@@ -0,0 +1,126 @@
...
+    tooltipTemplateString: React.PropTypes.string.isRequired, // Underscore template string that displays each line of a tooltip.
+    titleText: React.PropTypes.string.isRequired,
+    nowMomentUTC: React.PropTypes.instanceOf(moment),
+    monthKeyFn: React.PropTypes.func,
+    phaselines: React.PropTypes.array
+  },
+
+  getDefaultProps: function(){
+    return {
+      tooltipTemplateString: "<span><%= moment.utc(e.occurred_at).format('MMMM Do, YYYY')%></span>",
```
alexsoble: Two things surprised/confused me about `tooltipTemplateString`: first, it looks like an `erb` template in a `jsx` file. Second, it's referencing a variable `e` which I don't see anywhere else in this function. @edavidsonsawyer Can you help me understand what's going on with these two things? I think this could use some code commenting to make this easier to understand for future reviewers as well.

app/assets/javascripts/school_administrator_dashboard/dashboard_bar_chart.jsx
```diff
@@ -0,0 +1,126 @@
...
+      nowMomentUTC: moment.utc(),
+      monthKeyFn: defaultMonthKeyFn
+    };
+  },
+
+  render: function() {
+    const {GraphHelpers} = window.shared;
+    const {HighchartsWrapper} = window.shared;
+    const monthKeys = GraphHelpers.monthKeys(this.props.nowMomentUTC, this.props.monthsBack);
+    //const monthBuckets = GraphHelpers.eventsToMonthBuckets(monthKeys, this.props.events);
```
alexsoble: Remove commented-out line.

app/assets/javascripts/school_administrator_dashboard/dashboard_bar_chart.jsx
```diff
@@ -0,0 +1,126 @@
...
+            },
+            {
+              offset: 50,
+              linkedTo: 0,
+              categories: yearCategories,
+              tickPositions: Object.keys(yearCategories).map(Number),
+              tickmarkPlacement: "on"
+            }
+          ]}
+          title={{text: ''}}
```
alexsoble: I'm wondering, any reason this chart doesn't need a title? Will users of Insights automatically understand what it's about and what it represents? I don't know the answer to this myself, this line just struck me as an interesting override.

app/assets/javascripts/school_administrator_dashboard/dashboard_bar_chart.jsx
```diff
@@ -0,0 +1,126 @@
...
+              data: this.props.seriesData
+            }
+          ]} />
+      </div>
+    );
+  },
+
+  renderHeader: function() {
+    const nYearsBack = Math.ceil(this.props.monthsBack / 12);
+    const title = this.props.titleText + ', last ' + nYearsBack + ' years';
```
alexsoble: This looks like a good use case for ES2015 string concatenation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

app/assets/javascripts/school_administrator_dashboard/date_slider.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import Slider from 'rc-slider';
```
alexsoble: Woohoo! You got this working @edavidsonsawyer!

app/assets/javascripts/school_administrator_dashboard/page_container.jsx
```diff
@@ -0,0 +1,122 @@
...
+import SchoolAbsenceDashboard from './school_absence_dashboard.jsx';
+
+export default React.createClass({
+  displayName: 'DashboardPageContainer',
+
+  propTypes: {
+    attendanceData: React.PropTypes.object.isRequired  //should rename more generically
+  },
+
+  //necessary?
```
alexsoble: Let's implement this or remove this.

app/assets/javascripts/school_administrator_dashboard/page_container.jsx
```diff
@@ -0,0 +1,122 @@
+import React from 'react';
+import _ from 'lodash';
+import SchoolAbsenceDashboard from './school_absence_dashboard.jsx';
+
+export default React.createClass({
+  displayName: 'DashboardPageContainer',
+
+  propTypes: {
+    attendanceData: React.PropTypes.object.isRequired  //should rename more generically
```
alexsoble: @edavidsonsawyer I see a TODO here, is this something you want to do before or after merging?

app/assets/javascripts/school_administrator_dashboard/page_container.jsx
```diff
@@ -0,0 +1,122 @@
...
+  //returns map of homerooms to absence events
+  absencesByHomeroom: function() {
+    //Group data by homeroom
+    const homeRoomGroups = _.groupBy(this.props.attendanceData.absences, 'homeroom');
+    let homeroomAbsences = {};
+    //Simplify homeroom data into map of homerooms to total homeroom absences along with homeroom size
+    Object.keys(homeRoomGroups).forEach( (key) => {
+      const size = homeRoomGroups[key].length;
+      let key2 = key;
+      if (key == "null") key2 = "No Homeroom";
```
alexsoble: @edavidsonsawyer Is this the same as `(key == null) ? "No Homeroom" : key`?

app/assets/javascripts/school_administrator_dashboard/page_container.jsx
```diff
@@ -0,0 +1,122 @@
...
+  averageDailyAttendance: function(dailyAbsences, size) {
+    let averageDailyAttendance = {};
+    //mapping each homeroom to an array of day buckets containing all absences for each day w/in the homeroom
+    Object.keys(dailyAbsences).forEach((day) => {
+      averageDailyAttendance[day] = (size - dailyAbsences[day].length)/size*100;
+    });
+    return averageDailyAttendance;
+  },
+
+  mergeDateLists: function(list1, list2) {
```
alexsoble: Could we use lodash `_.merge` instead of writing our own merge? @edavidsonsawyer

app/assets/javascripts/school_administrator_dashboard/school_absence_dashboard.jsx
```diff
@@ -0,0 +1,162 @@
...
+  studentListContainer: {
+    padding: '30px',
+    float: 'left',
+    display: 'flex',
+    flexDirection: 'row',
+    alignItems: 'stretch',
+    marginLeft: '100px',
+    marginRight: 'auto'
+  },
+  dateRangeContainer: {
```
alexsoble: @edavidsonsawyer Empty object?

app/assets/javascripts/school_administrator_dashboard/students_table.jsx
```diff
@@ -0,0 +1,50 @@
...
+  },
+
+  render: function() {
+    return(
+      <div className= 'StudentsList'>
+        <table className='students-list' style={{width: '100%'}}>
+          <thead>
+            <tr>
+              <td>
+                {"Last Name"}
```
alexsoble: @edavidsonsawyer Small thing here but no need to use the curly brackets if you're just returning a string.

app/controllers/school_administrator_dashboard_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+      }
+    )
+  end
+
+  def students_for_dashboard(school)
+    school.students.active
+  end
+
+  def authorize_for_school
+    unless current_educator.is_authorized_for_school(@school)
```
alexsoble: @edavidsonsawyer Like I mentioned in our email thread, this is going to block educators who are districtwide from viewing the graphs.

config/routes.rb
```diff
@@ -47,6 +47,7 @@
   resources :sections, only: [:index, :show]
   resources :import_records, only: [:index]
   resources :iep_documents, only: [:show]
+  resources :school_administrator_dashboard, only: [:show]
```
alexsoble: A Rails-y pattern here would be to have `administrator_dashboard` as a `member` route on the school resource.

alexsoble:
Hey there @edavidsonsawyer! I left comments above on areas of the codebase where I saw loose ends, comments from you that we should address before merging, and opportunities to use good Rails and JS conventions. 

I also looked through the UI of the feature you built. Overall it looks great and I can see the usefulness of this immediately! And it's very cool that you're building this based off of PowerBI charts that @snoopyuri created. As I explored the feature I noticed two things in the UI that made me pause:

## Bar chart trouble

screen shot 2017-10-30 at 3 04 55 pm

First, when I hovered over the bars in the chart, I saw some Highcharts defaults that didn't seem to make sense in the context. It wasn't clear what "Series 1" meant, and there were no units or truncation on 97.2692... Both of these seem like important things for us to clarify before we merge this work in and show it to educators. 

## Date range trouble

screen shot 2017-10-30 at 3 04 50 pm

I also paused at the Date Range, because I didn't know what the start and end dates of the range were. This meant I wasn't sure what the defaults were and what the outer limits were.

I only discovered the labels when I hovered:

screen shot 2017-10-30 at 3 10 53 pm

We should always show users the default values they're working with, the user shouldn't need to take any actions to discover those. 

## Overall

On a "Must Have," "Very Good To Have," "Nice To Have", and "Not Important" scale: I would rate the labels and units on the bar charts as a "Must Have" for this PR and the labels on the date range as  in "Very Good To Have" but not super far from "Must Have," since we should always show users the default values they're working with.

What do you think @edavidsonsawyer / @snoopyuri? How would you rate these two issues on the scale? @edavidsonsawyer, are there other issues you've been wondering about as maybe being somewhere on this scale? 

@edavidsonsawyer, I'd also love to learn from your experience with solving two things. First, we talked on one of our team meetings about the challenge of making these computations without knowing explicitly how many school days in a month there are. I'd love to hear about how you solved that or worked around that. You also had some issues installing a yarn dependency, I'd love to know what the problem was and how you ultimately solved it!

edavidsonsawyer:
@alexsoble This is great, thank you. Looks like this PR is still a little rougher than you'd want for getting it into production, which is totally fine. I'm only looking to do that so I can see how it handles realistic data, especially from a performance perspective. I agree completely on the UI. However, before putting it in front of user's, I'd say:

**Must have**: Correct user authentication. Labels on chart, tooltip values, date range. Sorting for the student absence chart. Agreed upon standard for handling active school days.

**Nice to Have**: Manual input box for date range. Better UI for student absence chart. 

Regarding your two questions, I still don't have a great solution for active school days. Locally I have something where I'm extrapolating from overall school attendance data, but I'm not sure how close that will be. One of the big reasons I'd like to work with the production dataset is to test my assumptions here. 

For the yarn dependancies, I just bit the bullet and installed react and react-dom with yarn and imported them into the relevant files. I'm not totally satisfied with this solution since it introduces the possibility of different components using different versions of those frameworks.

alexsoble:
Hey there @edavidsonsawyer! 

I'm wondering about the slice of semi-anonymized data I sent you. Questions: 

+ Were you able to load the semi-anonymized in to your local environment and test how the feature looked with real data? 

+ Was it unclear how to load it into your local environment, since I don't think we have an import-from-arbitrary-local-CSV code path checked into our codebase? 

+ If you were able to load it in — Would it be better if I sent you a bigger slice of data? Or did you have enough to work with?

edavidsonsawyer:
Hi @alexsoble !

The anonymized data worked just fine, and it's the basis for the code I mentioned earlier. In short, from your data slice it looks reasonable to assume that there are no schoolwide perfect attendance days, and we can treat the dates on which an absence occurred as the dates school was in session. I'm going to clean up a couple other pieces from your notes and include that in the updated PR. 

If you're getting at whether I need this code in production to move forward, the answer to that is no. I'd like to see how it plays with the real data set, but I'm not blocked immediately. It should only be a little more time before the issues we discussed above are all addressed so it's fine to wait until that point. I'll ping you when it's ready for a second look. Thanks again!

app/assets/javascripts/school_administrator_dashboard/dashboard_bar_chart.jsx
edavidsonsawyer: This was left over from the hatchet job I did on the profile bar chart from the student profile page. I've revised it since it's unnecessary for this chart.

app/assets/javascripts/school_administrator_dashboard/page_container.jsx
edavidsonsawyer: I don't think so, b/c I have to use the original key value to access the object later on.

app/assets/javascripts/school_administrator_dashboard/page_container.jsx
edavidsonsawyer: For some reason that didn't work the way I expected it to. I'll look at it again and see about simplifying this.

app/controllers/school_administrator_dashboard_controller.rb
edavidsonsawyer: I'm confused about this. Are districtwide educators not authorized for their schools? My understanding was the existing permissions are too broad and we want to restrict to principals and assistant principals.

alexsoble: You're totally correct @edavidsonsawyer and I was wrong about this — `is_authorized_for_school` checks for `districtwide_access` first and returns `true` for districtwide educators, who aren't assigned to school. I double-checked to see who wrote that method and turns out I did lol.

alexsoble: In terms of permissions being too broad, correct. @snoopyuri mentioned restricting to principals and assistant principals. While Insights doesn't know exactly who's a principal/AP based on the data we get from X2, checking for `current_educator.admin == true` should be an excellent proxy. I sent around a list of educators where `admin == true` called "insights_admins_for_review.txt" so @snoopyuri can review.

alexsoble:
@edavidsonsawyer Sounds good! These dashboards are looking great and I want us to be able to merge them in as soon as possible. 

I also want the path-to-merge to be super clear and transparent to you so that you're not left wondering, "what do I have to do to get this merged and deployed?" 

Here's my sense of what's a "Must Have" and what's "Nice to Have", let me know if you think this list makes sense @edavidsonsawyer:

## "Must Have"

+ [ ] Update auth code to check for `current_educator.admin == true`, assuming this permissions solution works well for @snoopyuri 
+ [ ] Add some rspec Rails-side specs to check that the authorization code is working as expected

## "Nice to Have" (but NOT blockers to merging)

+ [ ] Add some description and screenshots/GIFs to the "description" section of this PR so that folks know what this dashboard is about and what it looks like — show off your great work here! 🌈 
+ [ ] Use the Rails-y routes pattern I mentioned in my code review comment

## Really not blockers at all

+ [ ] Any of the remaining comments I left about JS-side improvements like using existing `_.merge`, seeing if we can use more concise syntax, etc. — these are maybe improvement opportunities but they shouldn't stand in the way of shipping this feature

edavidsonsawyer:
Sounds good! I'll get some rspec specs going and make the description more useful. Let me know if @snoopyuri is happy with the auth solution and I'll implement that. Otherwise, the rest of these should be resolved with the last few commits.

snoopyuri:
That permission works for me.  Getting very excited that this is coming to fruition!

Thanks,

Uri

edavidsonsawyer:
Alright, @alexsoble, I think this should be ready for another look. I'm working on a couple things to spruce up the UI, but nothing that should hold up merging.

alexsoble:
@edavidsonsawyer Merging and shipping now! I enjoyed the "aspire to be more than wild beasts" commit comment! :D

PR #1102, opened 2017-09-26
WIP – Nightly IEP import job

alexsoble:


alexsoble:
Good learning from conversation with John Breslin — I had the data model wrong. There are *not* many IEP documents per student, just one active IEP document per student. So when we get new IEPs for a student we don't need to create a new unique record for it if an IEP already exists, just override and use the new document (and delete the old one).

alexsoble:
Travis is refusing to start the build for some reason, but the specs pass locally and local QA looks good. Going to 🔀  merge, 🚢  ship, and trigger a build in prod.

screen shot 2017-10-03 at 11 55 34 am

Issue #1078, opened 2017-09-21
Add student schedule and grades to student profile (high school only)


jhilde:
This is where Uri and I came to so we can add student schedule and current grades to student profile. This is the next biggie for high school educators.

All changes will only affect students in schools with school_type of HS. This is currently just Somerville High School until we add Next Wave or other districts. **_Students profiles in for students in all other schools will remain the same._**

**Collapse Reading and Math to one panel**
The new panel will have two graphs: MCAS Scores with 2 plotlines (ELA & Math), MCAS SGP (ELA & Math) _[Rationale is to keep the scale the same]._

There will be text indicating most recent STAR scores _[Rationale is that STAR tests are only used in the elementary schools so we don't need trends in the summary, just the most recent (like 8th grade scores).]_

When a user clicks on the Reading / Math panel, the details from the panel will be combined.

**Create a schedule panel**
The summary panel will include all courses in the current semester (so current term + full year courses). Essentially <section_number (first n characters)> <grade_numeric>. 

If the educator has permission to view a section, that section will be a link to section view. Otherwise, they will just see text. [To be explicit, educators who have permission to view a student's profile will be able to see the student's grade in all current year sections, regardless of whether or not they have permission to view the sections.]

Clicking on the schedule panel will change the details pane to a new schedule pane. The pane will include all courses for the current school year. 

So: <section_number> <grade_numeric> <educators> <schedule> (maybe also room)

**Outstanding Questions**
- Still need to understand how the grades file is generated. (It's a "proprietary feed" from the vendor written in Java. My imagination is that it is pulling assignment grades and grading rubrics from Aspen then calculating a current average. It's unclear to me what happens when grades (done on a quarter system) are posted. From Aspen's interface, it looks like the average is eliminated and a posted grade is created. 

**Things to note**

- To accomplish the summary only showing current semester, we'll need to know the semester start dates. Unfortunately it doesn't seem this information exists in Aspen. So initially, the dates for this year will be hard coded into the database. In the future, we made add an interface to change them in the admin dashboard. 

- Some of this will be Somerville specific. Terms in Aspen are configurable, some districts have trimesters, quarters, some use different identifiers for different terms.

jhilde:
A bit of original history in #1038.

kevinrobinson:
@jhilde Sweet!  It seems like the big thing here is the product and UI changes.

[...]

PR #1107, opened 2017-09-28
Add section data to Student Profile for high school students

jhilde:
This PR adds sections to the overview area of the Student Profile implementing #1078.

**Who does it affect?** All high school students. Any student in a school type other than "HS" remains as is.

**What does it look like?**
capture

**What do I need to know?**
If an educator has permission to view a student's profile (via districtwide, schoolwide, homeroom, grade level, section, etc), they can see the student's current sections with current average grade. 

If an educator has access to a section in which the student is currently enrolled, they will see the section number as a link (goes to section view). Otherwise, the section number will be plain text.

### DIFF
--- a/app/assets/javascripts/helpers/profile_details_style.jsx
+++ b/app/assets/javascripts/helpers/profile_details_style.jsx
@@ -19,6 +19,14 @@
       marginBottom: 10,
       width: '100%',
     },
+    roundedBox: {
+      border: '1px solid #ccc',
+      padding: 15,
+      marginTop: 10,
+      marginBottom: 10,
+      width: '100%',
+      borderRadius: '5px 5px 5px 5px',
+    },
     header: {
       display: 'flex',
       flexFlow: 'row',
@@ -59,6 +67,11 @@
       marginBottom: 20
     },
     fullCaseHistoryTitle: {
+      color: 'black',
+      display: 'inline-block',
+      flex: 'auto',
+    },
+    sectionsRosterTitle: {
       color: 'black',
       display: 'inline-block',
       flex: 'auto',
--- a/app/assets/javascripts/helpers/profile_details_style.jsx
+++ b/app/assets/javascripts/helpers/profile_details_style.jsx
@@ -25,7 +25,7 @@
       marginTop: 10,
       marginBottom: 10,
       width: '100%',
-      borderRadius: '5px 5px 5px 5px',
+      borderRadius: 5,
     },
     header: {
       display: 'flex',
--- a/app/assets/javascripts/student_profile/profile_details.jsx
+++ b/app/assets/javascripts/student_profile/profile_details.jsx
@@ -1,4 +1,5 @@
 import FlexibleRoster from '../components/flexible_roster.jsx';
+import SortHelpers from '../helpers/sort_helpers.jsx';
 import _ from 'lodash';
 
 (function() {
@@ -22,6 +23,7 @@ import _ from 'lodash';
       chartData: React.PropTypes.object,
       iepDocuments: React.PropTypes.array,
       sections: React.PropTypes.array,
+      currentEducatorAllowedSections: React.PropTypes.array,
       attendanceData: React.PropTypes.object,
       serviceTypesIndex: React.PropTypes.object,
       currentEducator: React.PropTypes.object,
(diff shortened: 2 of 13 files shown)


jhilde:
@alexsoble, @kevinrobinson let me know if you want to take a look. Otherwise, I'll deploy. Should be solid. Would be great to get into production by the end of the week.

Tested on IE11 ✅

app/assets/javascripts/helpers/profile_details_style.jsx
```diff
@@ -19,6 +19,14 @@
       marginBottom: 10,
       width: '100%',
     },
+    roundedBox: {
+      border: '1px solid #ccc',
+      padding: 15,
+      marginTop: 10,
+      marginBottom: 10,
+      width: '100%',
+      borderRadius: '5px 5px 5px 5px',
```
kevinrobinson: You can just do `borderRadius: 5` here if these are all the same.

app/assets/javascripts/student_profile/profile_details.jsx
```diff
@@ -164,6 +170,38 @@ import _ from 'lodash';
       return _.sortBy(events, 'date').reverse();
     },
 
+    styleEducators: function(section, column) {
+      const educatorNames = section.educators.map(educator => {
```
kevinrobinson: You can do `_.map(section.educators, 'full_name')` if you like.

app/assets/javascripts/student_profile/profile_details.jsx
```diff
@@ -197,6 +238,39 @@ import _ from 'lodash';
...
+    renderSectionDetails: function() {
+      const sections = this.props.sections;
+      
+      // If there are no sections, don't generate the section
+      if (!sections || sections.length == 0) return null;
+
+      // If this student is not a high school student, don't generate the section
+      if (this.props.student.school_type != 'HS') return null;
+      
+      const columns = [
```
kevinrobinson: This and the related methods in this component seem like they might split out nicely to a `SectionsTable` component or something like that, to keep it isolated and keep this component from getting huger :)

app/controllers/students_controller.rb
```diff
@@ -37,6 +37,9 @@ def show
       educators_index: Educator.to_index,
       access: student.latest_access_results,
       iep_documents: student.iep_documents,
+      sections: student.sections.select('sections.*, student_section_assignments.grade_numeric')
```
kevinrobinson: This is a lot of complexity inline with serialization code, factor out into a method?

FWIW `educators_index` is sent to the browser so you can serialize just ids and join client side if that were simpler, no need to change here just FYI.  And TIL that `#select` doesn't just work as "filter a list" but also in the SQL sense to pluck fields. :)

app/models/educator.rb
```diff
@@ -151,6 +151,10 @@ def allowed_sections
     end
   end
 
+  def allowed_sections_index
+    allowed_sections.map { |section| section.id }
```
kevinrobinson: Not sure the value of adding to the `Educator` model here?  You could inline `current_educator_allowed_sections: `educator.allowed_sections.map(&:id)` in the controller or a helper there.

spec/javascripts/student_profile/student_profile_page_spec.jsx
```diff
@@ -95,8 +106,27 @@ describe('StudentProfilePage integration test', function() {
...
+      it('renders student with 1 section', function() {
+        const el = this.testEl;
+        helpers.renderStudentProfilePage(el, '10', [], 0, 1, 'HS');
+        expect(el).toContainText('1 section');
+      });
 
+      it('renders student with 5 sections', function() {
+        const el = this.testEl;
+        helpers.renderStudentProfilePage(el, '10', [], 0, 5, 'HS');
+        expect(el).toContainText('5 sections');
```
kevinrobinson: ❤️ these simple tests!

jhilde:
Thanks for the great code suggestions @kevinrobinson!

jhilde:
Looking good :ship ing.

jhilde:
Looks good in production on IE11.

Logged one bug #1108 to fix grade sorting. Not worth rolling back, no errors thrown, just weird sorting by grade.

Issue #1108, opened 2017-09-28
Fix Grade Sorting in section and student section rosters


jhilde:
Looks like grade sorting doesn't like sparsely populated grades (e.g. a student doesn't have grades in several sections). This is likely caused by the comparison with nulls in the sort. 

Currently doesn't throw any errors, just sorts weird and non-deterministically if there are several nulls. Don't think an immediate patch or rollback is required.

alexsoble:
@jhilde We've had issues in this area before, maybe it's time for a `sort_helper_spec.js`!

jhilde:
@alexsoble 😃 I added one when I created the FlexibleRoster. Unfortunately I didn't have "" as any numbers in that test. Fixed that up.

Looks like it was the switch from _.isInteger to _.isNumber had a bit of a side effect. isInteger(NaN) is false while isNumber(NaN) is true. 

I think there may also have been a logic issue so I made it a little simpler as well.

jhilde:
Fixed with #1110, will be deployed with some other little fixes.

jhilde:
In prod, looks good.

PR #1110, opened 2017-09-28
Fix numeric sorting

jhilde:
PR for #1108 

It looks like a combination between some faulty logic and a change from isInteger to isNumber so we could use lodash 3.10.1 caused numbers not to be properly sorted.

Modified the test to include nulls so we could exercise this case in the tests.

Plan on batching this with other little PRs for deploy to production.

### DIFF
--- a/app/assets/javascripts/helpers/sort_helpers.jsx
+++ b/app/assets/javascripts/helpers/sort_helpers.jsx
@@ -22,15 +22,14 @@ export default {
   },
 
   sortByNumber: function (a, b, sortBy) {
-    const numA = parseInt(a[sortBy]);
-    const numB = parseInt(b[sortBy]);
+    // Parsing all numbers by floats to retain the decimal in sorting
+    const numA = parseFloat(a[sortBy]);
+    const numB = parseFloat(b[sortBy]);
 
-    if (!_.isNumber(numA) && !_.isNumber(numB)) return 0;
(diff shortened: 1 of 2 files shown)


alexsoble:
Oh no! @jhilde Did I introduce a bug by moving from _.isInteger to _.isNumber? Would love to dig in and understand better.

jhilde:
@alexsoble I think there was already a logic problem. Yeah, isInteger and isNumber are very similar except for their treatment of NaN which is what is returned when you try to convert a string to an int with parseInt. 

Was good to look at it though because it wasn't going to sort grades properly anyways. It converted to integers rather than floats.

alexsoble:
Oy! Thanks for spotting and fixing @jhilde!

alexsoble:
Yeah, just tried it out in my browser to confirm:

```
=> _.isNumber(NaN)
true
=>_.isInteger(NaN)
false
```

Thanks for teaching me this @jhilde!

jhilde:
@alexsoble There are many layers to lodash! 😜  Really, all we needed was isNaN because the numbers are either numbers or NaN, nothing else. Fortunately isNaN has been around since the beginning of lodash.

alexsoble:
In the beginning . . .

jhilde:
Checked in IE11, looking good! ✅

PR #1111, opened 2017-09-28
comment out rake tasks

Jngai:
@alexsoble lets comment out those rake tasks so those won't run every time somebody deploys.


alexsoble:
Hey @Jngai! My preference would be simply to remove them since they were one-time-only tasks.

Jngai:
@alexsoble yup removed.

PR #1114, opened 2017-09-28
Implement new IEP data model: A student has one active IEP document, not many 

alexsoble:
# Notes 

+ Updated my understanding of the IEP data model after a conversation with John Breslin: https://github.com/studentinsights/studentinsights/pull/1102#issuecomment-332923369


alexsoble:
Checked in the production database; there are some students with more than one IEP so will need to write a migration to destroy all except the most recent.

alexsoble:
Tested the migration in production without actually destroying any records (commented out the line that actually destroyed database records); looks good.

Issue #1099, opened 2017-09-26
Remove Low Income Status from Teacher View


snoopyuri:
In speaking the school lawyer, he stated there was not enough of a compelling interest for teachers to have this information, like they would need around contact information, home language, or disability.  He recommended that we remove this column from the Class Roster and from the Student Profile.  The information should still remain in the School Overview filter page for people with school-wide access.

jhilde:
Hey @snoopyuri, I've removed free & reduced lunch info from the following places:

- student profile
- student report (pdf)
- homeroom
- section

I'm wondering if it might make sense to leave it in student profile but only for people with districtwide or schoolwide access?

Let me know you thoughts.

snoopyuri:
That's great.  Much appreciated.  I actually don't think it's needed in the student profile anymore.  I think it's information that's more needed by principals and district admin in the aggregate view rather than individual.

jhilde:
Excellent, will push to prod soon.

jhilde:
In prod, looks good.

PR #1115, opened 2017-09-29
Remove free / reduced lunch (income status) from all views except school

jhilde:
This PR implements #1099.

Removed column from:
Homeroom view
Section View

Removed demographic attribute from:
Student Profile
Student Report (pdf)


jhilde:
Checked in IE11, looking good! ✅

PR #1118, opened 2017-09-29
Bump up delayed job timing even further

alexsoble:
# Notes 

+ Caching student searchbar data to educators is making the job take more than 12 hours
+ Raise the limit now so the job succeeds, figure out how to slim down that job later

### DIFF
--- a/config/initializers/delayed_job_config.rb
+++ b/config/initializers/delayed_job_config.rb
@@ -1,1 +1,1 @@
-Delayed::Worker.max_run_time = 12.hours
+Delayed::Worker.max_run_time = 16.hours


alexsoble:
Lolol @studentinsights-bot . . . how about @kevinrobinson?

kevinrobinson:
@alexsoble seems fine for now, but I'm surprised that it could take 12 hours to compute this ever?  In looking at the code briefly, and running a few commands on the console, it seems like it takes >30 seconds to run `SearchbarHelper.names_for(educator)`, which would come out to about 6 hours total for all educators if that were average.  That's basically just bouncing every one off of `Educator.is_authorized_for_student`, but perhaps that's more work now with the rules around sections?  cc @jhilde

Issue #1080, opened 2017-09-22
SImplify authorization definitions?


kevinrobinson:
This is a thought from https://github.com/studentinsights/studentinsights/issues/1078.

re: authorization... It might be helpful to clarify some of the authorization in code (trying to remove some of the nuance there like "educators who have permission to view a student's profile will be able to see the student's grade in all current year sections, regardless of whether or not they have permission to view the sections".  It looks like this was modeled after the homerooms code, and I think with increasing complexity in these rules maybe a good opportunity to simplify.

One simplification might be: authorization is done on a `(educator, student)` basis, full stop ([here](https://github.com/studentinsights/studentinsights/blob/master/app/models/educator.rb#L63)).  And other things like "can they see a homeroom?" or "can they see a section?" are not authorization or security features, but convenience.  In other words, what if https://github.com/studentinsights/studentinsights/blob/8e7da6300f40109c1e953606308422af85ec29d1/app/controllers/sections_controller.rb#L27 showed anyone who asked information overall about the session (including the total number of students), but then for each student in that session, asked if the educator was authorized and only showed those students' data.

This same approach would apply to Homerooms as well, [here](https://github.com/studentinsights/studentinsights/blob/master/app/controllers/homerooms_controller.rb#L58).  Basically whenever we're looking up student data in controller code, we run it through the authorization check based on the current educator.  We could do a sneaky Rails thing to make that happen automagically, but I'd be in favor of doing it explicitly.

Concretely, anyone could look at any section/homeroom, but they'd only see students they're authorized to see.  Asking for "their sections" or "their homerooms" would be a convenience question and not an authorization and data security question.  The main benefit is a simpler model for authorization, where we have only one method that controls access to student data instead of three that are slightly duplicative and likely to drift out of sync.  The cost might be having to index or precompute some of that data if it's two slow to compute the access for each student at read time, but that's just a guess.

jhilde:
I kinda see where you're going but I'm not sure about everyone being able to see all homerooms and sections with only students they have permission to see. When you view a section or homeroom, you're likely not thinking about a particular student...you're thinking about a grouping of students. Showing some but not all section or homeroom students seems incomplete.

Also, the majority of users gain permission not through a relationship with students but a relationship to a section or homeroom. 

[...]

PR #1127, opened 2017-10-02
Improve performance of student searchbar JSON computing

### LINKED ISSUE #1120: Improve performance of Educator.save_student_searchbar_json
It looks like the component of the nightly imports taking the longest time is likely caching the json of students available in the student_searchbar.

Likely strategy to decrease the time a bit without taking the authorization logic out of Educator#is_authorized_for_student will be to limit the initial query of students to the educator's school (unless the educator is a districtwide_admin). 

See #1118.

alexsoble:
# What does this do?

+ When precomputing searcbar JSON, if educator is school-based, only query students at their school. (Very good catch by @jhilde.)
+ When precomputing searcbar JSON, if educator is districtwide, use an instance variable cached to the job so that we don't precompute JSON for all students more than once per job.
+ Pull all this logic into a `PrecomputeSearchbarJson` service object. 

# Issue 

+ Fix #1120.

### DIFF
--- a/app/helpers/searchbar_helper.rb
+++ b/app/helpers/searchbar_helper.rb
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 module SearchbarHelper
   def self.names_for(educator)
-    authorized_students = Student.with_school.select do |student|
+    authorized_students = educator.school.students.select do |student|
       educator.is_authorized_for_student(student)
     end
+
     authorized_students.map do |student|
       {
         label: "#{student.first_name} #{student.last_name} - #{student.school.local_id} - #{student.grade}",
--- a/app/helpers/searchbar_helper.rb
+++ b/app/helpers/searchbar_helper.rb
@@ -11,4 +11,13 @@ module SearchbarHelper
       }
     end
   end
+
+  def self.names_for_all_students
+    Student.with_school.map do |student|
+      {
+        label: "#{student.first_name} #{student.last_name} - #{student.school.local_id} - #{student.grade}",
+        id: student.id
+      }
+    end
+  end
 end
--- a/app/importers/file_importers/educators_importer.rb
+++ b/app/importers/file_importers/educators_importer.rb
@@ -18,10 +18,11 @@ class EducatorsImporter < Struct.new :school_scope, :client, :log, :progress_bar
 
   def import_row(row)
     educator = EducatorRow.new(row, school_ids_dictionary).build
+    searchbar = PrecomputeSearchbarJson.new
 
     if educator.present?
       educator.save!
-      PrecomputeSearchbarJson.for(educator)
+      searchbar.for(educator)
 
       homeroom = Homeroom.find_by_name(row[:homeroom]) if row[:homeroom]
       homeroom.update(educator: educator) if homeroom.present?
--- a/app/importers/file_importers/educators_importer.rb
+++ b/app/importers/file_importers/educators_importer.rb
@@ -21,7 +21,7 @@ class EducatorsImporter < Struct.new :school_scope, :client, :log, :progress_bar
 
     if educator.present?
       educator.save!
-      educator.save_student_searchbar_json
+      PrecomputeSearchbarJson.for(educator)
 
       homeroom = Homeroom.find_by_name(row[:homeroom]) if row[:homeroom]
       homeroom.update(educator: educator) if homeroom.present?
--- a/app/models/precompute_searchbar_json.rb
+++ b/app/models/precompute_searchbar_json.rb
@@ -10,7 +10,7 @@ class PrecomputeSearchbarJson
 
   def self.for(educator)
     if educator.districtwide_access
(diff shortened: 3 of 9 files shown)


app/helpers/searchbar_helper.rb
```diff
@@ -1,14 +1,24 @@
 # frozen_string_literal: true
 module SearchbarHelper
   def self.names_for(educator)
-    authorized_students = Student.with_school.select do |student|
+    authorized_students = educator.school.students.select do |student|
```
alexsoble: This is the key line that implements the school-restricting solution @jhilde outlined.

app/importers/file_importers/educators_importer.rb
```diff
@@ -18,10 +18,11 @@ def school_ids_dictionary
 
   def import_row(row)
     educator = EducatorRow.new(row, school_ids_dictionary).build
+    searchbar = PrecomputeSearchbarJson.new
 
     if educator.present?
       educator.save!
-      educator.save_student_searchbar_json
+      searchbar.for(educator)
```
jhilde: This actually needs to be a separate after_imports item. Calculating which students are accessible for an educator takes into account sections. Those are imported after the educators have been imported. 

Seems like the proper place for this would be import.thor after all the imports have completed.

alexsoble:
Ah, great point @jhilde! So basically the code as-is is always going to be 1 extra day behind because it will be using slightly stale authorization data.

In that case, the current code in production is also 1 day stale.

Since this PR isn't a regression on that front, and will decrease the time by a bunch (we'll see exactly how much), how do you feel about merging this, deploying, and then working on the after_imports improvement you describe @jhilde?

jhilde:
Sounds like a plan!

On Mon, Oct 02, 2017 at 10:04 PM Alex Soble

<
mailto:Alex Soble <notifications@github.com>
> wrote:

Ah, great point
https://github.com/jhilde
! So basically the code as-is is always going to be 1 extra day behind because it will be using slightly stale authorization data.

In that case, the current code in production is also 1 day stale.

Since this PR isn't a regression on that front, and will decrease the time by a bunch (we'll see exactly how much), how do you feel about merging this, deploying, and then working on the after_imports improvement you describe
https://github.com/jhilde
?

—

Reply to this email directly,
https://github.com/studentinsights/studentinsights/pull/1127#issuecomment-333717901
, or

app/importers/file_importers/educators_importer.rb
kevinrobinson: Agree!  The searchbar isn't about importing, it's about precomputing an index for a feature.

app/models/precompute_searchbar_json.rb
```diff
@@ -0,0 +1,33 @@
...
+  def for_all_educators
+    Educator.find_each { |educator| self.for(educator) }
+  end
+
+  def for_educators_who_log_in
+    educators_who_log_in.find_each { |educator| self.for(educator) }
+  end
+
+  def for(educator)
+    if educator.districtwide_access
```
kevinrobinson: @alexsoble @jhilde I'm concerned this approach in general, and about moving authorization code into `PrecomputeSearchbarJson` in particular.

If the purpose of this class is to help precompute some data that's slow, then it shouldn't be reaching into educator models to read their roles and determine who's authorized to see students.  If the problem is that it's expensive to compute authorization, can we look at speeding that up?  I submitted https://github.com/studentinsights/studentinsights/pull/1129 as a step towards that.

spec/models/precompute_searchbar_json_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    context 'educator has permissions for no students' do
+      let(:educator) { FactoryGirl.create(:educator) }
+
+      it 'saves the correct JSON' do
+        PrecomputeSearchbarJson.new.for(educator)
+        expect(educator.student_searchbar_json).to eq("[]")
+      end
+    end
+
+    context 'multiple districtwide_access educators' do
```
kevinrobinson: This is a smell too - we're testing authorization code within the `PrecomputeSearchbarJson` class.

alexsoble:
Good feedback @jhilde and @kevinrobinson! I'll close this PR since it looks like we want to go with a different strategy. 

@kevinrobinson, can you help me understand your ideas about authorization? The reason the precomputing code I wrote asked about educator authorization is so that it could surface only students the educator is authorized to see in the searchbar. 

Are you saying we don't want the searchbar code to be aware of authorization at all? So that (for example) any educator would get the entire list of students in the district when they type in the searchbar, but get a "You are not authorized" warning from the server side if they try to visit the profile of a student they're not authorized for?

kevinrobinson:
@alexsoble oops maybe I never responded to this, but If I remember right, we talked about your comment in https://github.com/studentinsights/studentinsights/pull/1127#issuecomment-333886223 at some point offline, but if not we could talk more over here! https://github.com/studentinsights/studentinsights/issues/1080#issuecomment-334588617

PR #1128, opened 2017-10-02
Modify the way we handle stale assignments for students and educators

jhilde:
Background:
The files we import to create the assignments of students and educators to sections includes the full import every time rather than providing us with deltas. That makes it difficult to determine which assignment have been deleted in Aspen. (Think student drops a course). 

Previous Strategy:
Initially I simply deleted all rows in the student_section_assignments and educator_section_assignments tables then imported all the fresh data. While that works, it is very sensitive to importer failure (as was seen in production today because of a timeout) and educators won't see data while each importer is running.

New Strategy:
As each of the two assignment importer imports rows, the id is added to an array. This becomes the list of fresh data. After all the importing is done for the importer, delete_rows is called. This deletes all rows from the table whose ids are not in the array (i.e. the stale rows).

### DIFF
--- a/app/importers/file_importers/educator_section_assignments_importer.rb
+++ b/app/importers/file_importers/educator_section_assignments_importer.rb
@@ -1,4 +1,8 @@
 class EducatorSectionAssignmentsImporter < Struct.new :school_scope, :client, :log, :progress_bar
+  def initialize(*)
+    super
+    @imported_assignments = []
+  end
 
   def remote_file_name
     'educator_section_assignment_export.txt'
@@ -13,7 +17,8 @@ class EducatorSectionAssignmentsImporter < Struct.new :school_scope, :client, :l
   end
 
   def delete_rows
-    EducatorSectionAssignment.delete_all
+    #Delete all stale rows no longer included in the import
+    EducatorSectionAssignment.where.not(id: @imported_assignments).delete_all
   end
 
   def import_row(row)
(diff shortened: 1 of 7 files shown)


jhilde:
@alexsoble I'm going to deploy once Travis stops choking on the teaspoon checks. Let me know if you have comments.

alexsoble:
Hey @jhilde! This looks like a good temporary fix. One longer term question I have: why not update data every night instead of deleting/recreating data? Is there a reason we can't do this for assignments?

jhilde:
@alexsoble That's actually what it's doing now. The delete is to delete only the stale data. 

If you run the importer twice with the same input file, it should not be deleting anything. It also shouldn't be updating anything either since grades are a separate importer. 

If, however, a student drops a section of a course from their schedule, there is no indication of that action in the export except for the row to be missing. So once it imports all rows in the relevant file, it deletes any assignment not in the db but not in the file.

That seems to be the strategy with the best user experience that fits into our current importer paradigm. 

Perhaps I could better state the new strategy as:
As each of the two assignment importer imports rows, the id **(of the existing row or the newly created row)** is added to an array. This becomes the list of fresh data. After all the importing is done for the importer, delete_rows is called. This deletes all rows from the table whose ids are not in the array (i.e. the stale rows).

Does that make sense?

alexsoble:
Oooh thanks for the explanation @jhilde! I need to spend a little more time
playing with the code before it sinks in fully but this explanation is
super helpful & puts me on the right track!

Issue #1088, opened 2017-09-22
Push EventNoteTypes into the database


alexsoble:
https://github.com/studentinsights/studentinsights/pull/1085#issuecomment-331514545

alexsoble:
This duplicates #970 but doesn't really make sense, closing it in favor of the better version.

PR #1130, opened 2017-10-03
WIP - beginning work to expand props to catch bugs

Jngai:


Jngai:
@alexsoble I read the comments in the other issue and decided to go with Kevin's alternative way. I just started building the props out. Will you glance at it to make sure I am heading to the right direction? The data shape is pretty complicated from `studentscontroller show`.

alexsoble:
Hey @Jngai! Why don't we schedule a check in meeting for this Thursday or Friday? During last Friday's meeting I think we surfaced a couple of very concrete ways we can keep improving the services feature. These prop types look useful and way stricter than what we had before, but I think we should keep going with shipping features around Services before we invest your time + energy in refactoring things like prop types.

Jngai:
Hey @alexsoble I would love to continue to build out the service feature. Lets keep this pr open for now. I will email you privately for a time on Thursday.

alexsoble:
👍🏽👍🏽

alexsoble:
Talking this through with @Jngai. Because we're going to pull EventNoteTypes into the database once we add the next districts to the app, these hard-coded IDs are going to change district-to-district (#1088), so let's avoid baking them directly into the front end.

Issue #1135, opened 2017-10-05
Handle discontinuing services that will start in the future 


alexsoble:
# GIF of issue 

![impossible-to-discontinue](https://user-images.githubusercontent.com/3209501/31232535-d0b262c8-a9af-11e7-95a7-9fa8cc3d3a5f.gif)

kevinrobinson:
@alexsoble The original intent was to capture the historical log of who has entered what and what changes have been made.  That allows looking back to see who has edited what (as in auditing) and also to look back historically at what services students had in past years.  So it wouldn't just be storing the current state in the database, it'd be storing it such that we capture the complete history.  This might be different than other data we store in the database, since Student Insights fully owns this.

My overall mental model is: somewhere we want a log of all the things that have happened.  This could be done directly in the data model as it was initially (eg., when users create a service we make a record for that, and then when they discontinue it we make a DiscontinuedService for that).  The downside is at read time you have to merge those things together to understand the current state.  Or the data model could model the current state, and we write the set of changes somewhere else where we could recreate what happened.

It sounds like the change to add "end date" complicates this, and that you all found a bunch of edge cases.  @alexsoble @Jngai you all should figure out whatever you think makes sense, and feel free to ignore what the past intents were too as you're figure out how to improve this!  If changing it to just capture the current state is simpler, and persist a log of changes to another table keeps the history, that seems great too!

Jngai:
Check out #1138

Jngai:
@alexsoble @kevinrobinson I think the edge cases are very useful. I do find myself coding to adjust for each edge case. I think the bigger picture is coming soon since I have to investigate #1112 . Overall I think what we added help answer multiple ways that a user would go about using the service feature. @alexsoble I just notice that the input box says **"when did they start"** so a teacher seeing that is less likely to enter a future date however it still could be possible.

alexsoble:
Hey @kevinrobinson! I like the "somewhere we want a log of all the things that have happened" framing, that seems correct for sure.

Here's my mental model of all the things that can happen to a service:

+ **First**, educator creates the service *or* Uri bulk uploads the service.
  + `recorded_by_educator_id` captures who created the service
  + `date_started` specifies when the service started (or was meant to start)
  + `estimated_end_date` specifies when they think it will end
+ **Later**, another educator or the same educator ends the service
  + Right now we have DiscontinuedService object, but that same data could be flattened down into just two fields: a `discontinued_by_educator_id` and a `date_discontinued`

[...]

PR #1138, opened 2017-10-05
Disallow educators from entering Services with future start dates

Jngai:
Please refer to #1135


Jngai:
servicepastpastdelete

I am using a linux command to generate the gif, please disregard the speed.

app/controllers/services_controller.rb
```diff
@@ -10,18 +10,22 @@ def authorize!
...
-      service_id: service_id,
-      recorded_by_educator_id: current_educator.id,
-      discontinued_at: Time.now
-    })
-
-    serializer = ServiceSerializer.new(Service.find(service_id))
-
-    if discontinued_service.save
-      render json: serializer.serialize_service
+    if Service.find(service_id).date_started > Time.now
```
Jngai: Basically I am destroying the service completely if its a future scenario. This is where the discontinued button eventually points to.

app/controllers/students_controller.rb
```diff
@@ -97,7 +97,7 @@ def service
       :estimated_end_date
     ])
 
-    estimated_end_date = params.require(:service).permit(*[:estimated_end_date])
+    estimated_end_date = clean_params["estimated_end_date"]
```
Jngai: I noticed that I am getting some `unpermitted params` error in my server log. I simplify things a bit here.

app/assets/javascripts/student_profile/services_list.jsx
```diff
@@ -85,6 +85,7 @@ import _ from 'lodash';
 
       this.props.onClickDiscontinueService(serviceId);
       this.setState(this.getInitialState());
+      window.location.reload();
```
Jngai: So when I completely delete a service record after clicking on discontinue, the button would be stuck at updating. I notice that I have to refresh the page for that deleted item to disappear. I think the gif makes it feel like its not incredibly recognizable that the page actually refreshed which is good since you don't notice it.

Jngai:
Restricting datepicker start to be future dates. 

screen shot 2017-10-05 at 4 00 13 pm

Jngai:
@alexsoble would you say this is a go then? I moved on to another issue. Let me know if I need to revisit this one.

alexsoble:
Hey @Jngai, did local QA and looked through the code, this looks good to me! I think this was definitely the quickest and simple way to eliminate the bug we saw around the future start date edge case and I'm glad we went this direction!

I'm curious to know if we'll hear anything back from teachers about it. This would be a good topic to watch the next time we sit in on an SST meeting!

@Jngai Why don't you remove "WIP" from the PR name, since this is no longer a work in progress?

Jngai:
Yup @alexsoble I removed it.

app/assets/javascripts/student_profile/record_service.jsx
```diff
@@ -211,7 +211,8 @@
             datepickerOptions={{
               showOn: 'both',
               dateFormat: 'mm/dd/yy',
-              minDate: undefined
+              minDate: undefined,
+              maxDate: new Date
```
Jngai: btw @alexsoble this just set the maximum date, its straight from https://api.jqueryui.com/datepicker/#option-maxDate.

Jngai:
@alexsoble I notice that there are a lot of changes in the master branch every time I update myself. Will you merge this if its good because the longer this sits here the more likely it would collide with the weekly changes in master. 

The test is green for this on CI and not locally. I think I just have errors in my local environment.

alexsoble:
@Jngai Yup!

alexsoble:
@Jngai Another way to pull in changes from the master branch without creating a new commit is to run `git rebase master`. If there's no conflicts between your branch and master, the rebase will update your branch automatically. Let me know if you want to chat or pair on git rebase this week!

Jngai:
@alexsoble I will start using `git rebase master`. Thanks for the tip. I know how to do it. I was using another way of setting the repo as `upstream` and then periodically fetch and merge `upstream/master`. This works too and I don't think I fetch and merge incorrectly. More like I just need to start paying attention to other people's work getting merged in and their changes.

alexsoble:
@Jngai The change you shipped looks good in production!

Jngai:
@alexsoble great I am glad that works out.

PR #1141, opened 2017-10-05
Show big red warning on Demo Site so folks don't think it's the real site

### LINKED ISSUE #1133: Disabling Old Insights Link But Still Google-Able
Tracked down the source of a few users who haven't been able to log in in the last couple of weeks.  Instead of going to Favorites or using our link, they Google "Student Insights" and come up with the old site that looks just the current site:

https://somerville-teacher-tool-demo.herokuapp.com/

Can we disable this?

alexsoble:
# Issue 

+ Close #1133.

# Screenshot (with warning)

screen shot 2017-10-05 at 2 04 01 pm

# Screenshot (without warning)

screen shot 2017-10-05 at 2 05 08 pm

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -11,10 +11,12 @@
   </head>
   <body class="<%= controller_name %> <%= action_name %> <%= if educator_signed_in? then "educator-signed-in" end %>">
     <div class="nav">
-      <div class="big_warning" style="padding: 20px; background-color: red; color: white; text-align: center; margin-bottom: 10px; font-size: 30px;">
-        <div>This is the demo site! It's filled with fake data.</div>
-        <div>The real site is <a style="font-size: 30px;">somerville.studentinsights.org</a>.</div>
-      </div>
+      <% if ENV['SHOW_DEMO_SITE_WARNING'] === true %>
+        <div class="big_warning">
+          <div>This is the demo site! It's filled with fake data.</div>
+          <div>The real site is <a>somerville.studentinsights.org</a>.</div>
+        </div>
+      <% end %>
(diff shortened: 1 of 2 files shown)


app/views/layouts/application.html.erb
```diff
@@ -11,6 +11,12 @@
   </head>
   <body class="<%= controller_name %> <%= action_name %> <%= if educator_signed_in? then "educator-signed-in" end %>">
     <div class="nav">
+      <% if ENV['DEMO_SITE_WARNING'] === 'SHOW' %>
```
kevinrobinson: Alternately I think there's a `DEPLOYMENT_KEY` value or something like that already set that's used to indicate that it's the demo site.

app/views/layouts/application.html.erb
```diff
@@ -11,6 +11,12 @@
...
+        <div class="big_warning">
+          <div>This is the demo site! It's filled with fake data.</div>
+          <div>The real site is <a>somerville.studentinsights.org</a>.</div>
```
kevinrobinson: Does this need an href?

alexsoble: Ya! Good catch. :D

app/views/layouts/application.html.erb
alexsoble: Ooooooh! I thought we had something like this but I couldn't remember what it was called! Thank you @kevinrobinson!

alexsoble:
OK @snoopyuri @kevinrobinson the big red warning is now live at https://somerville-teacher-tool-demo.herokuapp.com/.

alexsoble:
Big and red and hard to miss.

PR #1143, opened 2017-10-06
Make sure one importer blowing up doesn't stop the whole train

### LINKED ISSUE #1068: Flatten importer classes so that one error doesn't stop the whole train

alexsoble:
+ Fix #1068


alexsoble:
screen shot 2017-10-06 at 2 52 32 pm

Issue #1146, opened 2017-10-06
Remove FileImport wrapper class


alexsoble:
Important insight from @kevinrobinson: the `FileImport` wrapper class obscures the control flow of importing data and makes it hard to tell what's happening, when. Also uses `delegate` which is a code smell.

alexsoble:
Heyyyyyyyy closed via #1147.

PR #1147, opened 2017-10-06
Remove FileImport wrapper class

alexsoble:
# Notes 

+ Taking a crack at #1146 to improve the legibility and error-resiliency of the data import process by removing the `FileImport` class.  `FileImport` "wraps" a bunch of other import classes and obscures the flow of control those classes use to import data.


app/importers/file_importers/downloader.rb
```diff
@@ -0,0 +1,32 @@
+class Downloader
```
kevinrobinson: Is this a `CSVDownloader`?  It looks like it expects there be a `transformer` and that has a method `pre_cleanup_csv_size` etc.  Adding a comment about what the class is for could help.

app/importers/file_importers/downloader.rb
```diff
@@ -0,0 +1,32 @@
...
+    data = @transformer.transform(file)
+
+    CleanupReport.new(
+      @log, @remote_file_name, @transformer.pre_cleanup_csv_size, data.size
+    ).print
+
+    return data
+  end
+
+  def download_file
```
kevinrobinson: Is this intended to be public or private?

kevinrobinson:
@alexsoble This looks good to me!

app/importers/file_importers/downloader.rb
alexsoble: @kevinrobinson Good points, will implement.

app/importers/file_importers/downloader.rb
alexsoble: @kevinrobinson — Private, I'm thinking `get_data` should be the only public method. Will add that.

Issue #1134, opened 2017-10-05
Only services with an estimated_end_date should say "Discontinue Early"


alexsoble:
If there's no `estimated_end_date`, the button should just say "Discontinue."

For example, this service from the demo site has no estimated end date, but it still says "Discontinued Early", which could be confusing: 

![screen shot 2017-10-05 at 9 25 44 am](https://user-images.githubusercontent.com/3209501/31232472-9333e08e-a9af-11e7-9ac9-fa2b40a60347.png)

alexsoble:
This is a "backfilling" feature, because all future services added through the Student Profile UI will have an `estimated_end_date`.

alexsoble:
Note to self: do a quick database check for how many Services are in our database without any discontinue records / estimated end date.

alexsoble:
Fixed by @Jngai!

PR #1148, opened 2017-10-09
Discontinue early

Jngai:
Text change for issue #1134. 

screen shot 2017-10-09 at 10 19 25 am


app/assets/javascripts/student_profile/services_list.jsx
```diff
@@ -252,9 +252,20 @@ import _ from 'lodash';
       const isHovering = (this.state.hoveringDiscontinueServiceId === service.id);
       const isPending = (this.props.discontinueServiceRequests[service.id] === 'pending');
 
-      const buttonText = (isPending)
-        ? 'Updating...'
-        : (isConfirming) ? 'Confirm' : 'Discontinue Early';
+      var Text = 'Confirm';
+
+      if (isPending) {
```
Jngai: Button text was already a nested ternary operator. I think writing out the statements is more readable than an even more nested ternary operator. Also `const buttontext` is read only so I added a `text` variable to be modified then assigned to `const buttontext`.

Jngai:
@alexsoble this is good too. Basically I expanded out `buttontext` and added a new condition.

Jngai:
@alexsoble I want to double check this. Will reopen soon.

alexsoble:
@Jngai OK!

PR #1151, opened 2017-10-10
Remove obscure pass-through classes: SomervilleStarImporters, SomervilleX2Importers

alexsoble:
# Notes 

+ These classes are mostly just passing config through an extra layer, better to move that up to the main `import.thor` which is really responsible for config
+ The fewer the layers of classes, the easier to analyze and debug the import process


lib/tasks/import.thor
```diff
@@ -68,8 +90,18 @@ class Import
         @record ||= ImportRecord.create(time_started: DateTime.current)
       end
 
-      def importers(sources = options["source"])
-        sources.map { |s| SOURCE_IMPORTERS.fetch(s, nil) }.compact.uniq
+      def file_import_class_to_client(import_class)
+        return SftpClient.for_x2 if import_class.in?(X2_IMPORTERS)
```
alexsoble: Is there any meaningful difference between creating one X2 client and passing it in to each of the X2 importers, as opposed to creating a new X2 client for each importer?

lib/tasks/import.thor
```diff
@@ -97,9 +129,14 @@ class Import
     def connect_transform_import
       timing_log = []
 
-      file_importers = importers.flat_map { |i| i.new(options).file_importers }
+      file_import_classes.each do |file_import_class|
+        file_importer = file_import_class.new(
```
alexsoble: This initialization is ugly and messy, as a first step to cleaning it up I'd like to add Ruby required keyword arguments.

lib/tasks/import.thor
```diff
@@ -36,12 +62,8 @@ class Import
       desc: "Fill up an empty database"
     class_option :source,
       type: :array,
-      default: SOURCE_IMPORTERS.keys,
-      desc: "Import data from the specified source: #{SOURCE_IMPORTERS.keys}"
-    class_option :x2_file_importers,
-      type: :array,
-      default: SomervilleX2Importers.file_importer_names,
-      desc: "Import data from the specified files: #{SomervilleX2Importers.file_importer_names}"
+      default: FILE_IMPORTER_OPTIONS.keys,
```
alexsoble: TODO: Add a code comment about what this default is doing.

alexsoble:
Did local QA, looks good!

PR #1152, opened 2017-10-11
expand button text to include discontinue for services with no estimated end date 

Jngai:
Please see #1134

### DIFF
--- a/app/assets/javascripts/student_profile/services_list.jsx
+++ b/app/assets/javascripts/student_profile/services_list.jsx
@@ -257,7 +257,7 @@ import _ from 'lodash';
         text = 'Updating...';
       } else if (isConfirming) {
         text = 'Confirm';
-      } else if (service.estimated_end_date == null) {
+      } else if (service.estimated_end_date === null) {
         text = 'Discontinue';
       } else {
         text = 'Discontinue Early';
--- a/app/assets/javascripts/student_profile/services_list.jsx
+++ b/app/assets/javascripts/student_profile/services_list.jsx
@@ -251,19 +251,19 @@ import _ from 'lodash';
       const isConfirming = (this.state.discontinuingServiceId === service.id);
       const isHovering = (this.state.hoveringDiscontinueServiceId === service.id);
       const isPending = (this.props.discontinueServiceRequests[service.id] === 'pending');
-      var Text = 'Confirm';
+      let text = 'Confirm';
 
       if (isPending) {
-        Text = 'Updating...';
+        text = 'Updating...';
       } else if (isConfirming) {
-        Text = 'Confirm';
+        text = 'Confirm';
       } else if (service.estimated_end_date == null) {
-        Text = 'Discontinue';
+        text = 'Discontinue';
       } else {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/student_profile/services_list.jsx
```diff
@@ -251,11 +251,19 @@ import _ from 'lodash';
...
+
+      if (isPending) {
+        Text = 'Updating...';
+      } else if (isConfirming) {
+        Text = 'Confirm';
+      } else if (service.estimated_end_date == null) {
+        Text = 'Discontinue';
+      } else {
+        Text = 'Discontinue Early';
+      }
```
Jngai: @alexsoble This is good for a review. Basically I rewrote the ternary operator into if else statements to include the extra option for discontinue. The original ternary operator was already nested. It would harm readability if I make it even more nested. This way makes it more readable.

Jngai:
screen shot 2017-10-11 at 10 13 06 am

app/assets/javascripts/student_profile/services_list.jsx
alexsoble: @Jngai OK cool, let me check it out!

alexsoble:
@Jngai Functionality-wise this looks great! JS-style wise there are just a few small changes I'd recommend, I'll make some inline comments. Not huge things but just little practices we do to keep the code style consistent.

app/assets/javascripts/student_profile/services_list.jsx
```diff
@@ -251,11 +251,19 @@ import _ from 'lodash';
       const isConfirming = (this.state.discontinuingServiceId === service.id);
       const isHovering = (this.state.hoveringDiscontinueServiceId === service.id);
       const isPending = (this.props.discontinueServiceRequests[service.id] === 'pending');
+      var Text = 'Confirm';
```
alexsoble: Since we're using ES6, let's prefer `const` or `let` over `var` because `const` or `let` will say more about our intention for the variable.

alexsoble: Let's also use camelcase names for variables and have them start with a lowercase letter, like `text` instead of `Text`.

Jngai:
@alexsoble I pushed a commit to change the js style. I am using `let` and not `const` for `text` because `const` is read-only.

app/assets/javascripts/student_profile/services_list.jsx
```diff
@@ -251,11 +251,19 @@ import _ from 'lodash';
       const isConfirming = (this.state.discontinuingServiceId === service.id);
       const isHovering = (this.state.hoveringDiscontinueServiceId === service.id);
       const isPending = (this.props.discontinueServiceRequests[service.id] === 'pending');
+      let text = 'Confirm';
+
+      if (isPending) {
+        text = 'Updating...';
+      } else if (isConfirming) {
+        text = 'Confirm';
+      } else if (service.estimated_end_date == null) {
```
alexsoble: Let's use `===` here because it's behavior is more consistent than `==`.

PR #1153, opened 2017-10-11
Implement no-var and eqeqeq JavaScript rules

alexsoble:
# Notes 

+ Let's prefer const or let now that we're working in an ES6 codebase

### DIFF
--- a/app/assets/javascripts/components/flexible_roster.jsx
+++ b/app/assets/javascripts/components/flexible_roster.jsx
@@ -25,7 +25,7 @@ export default React.createClass({
 
     return sortedRows;
   },
-  
+
   sortedRows () {
     const rows = this.props.rows;
     const columns = this.props.columns;
@@ -38,7 +38,7 @@ export default React.createClass({
     else {
       return rows.sort((a, b) => SortHelpers.sortByString(a, b, key));
     }
-    
+
   },
 
   headerClassName (sortByIndex) {
@@ -59,7 +59,7 @@ export default React.createClass({
       this.setState({ sortByIndex: sortByIndex});
     }
   },
-  
+
   render () {
     return (
       <div className='FlexibleRoster'>
@@ -90,11 +90,11 @@ export default React.createClass({
       // count of items with the same group
       // increment in the beginning since colSpan starts at 1
       currentCount++;
-      
+
       // if the current item doesn't equal the next
       // push the super header with a length of currentCount
       // and reset currentCount for a new column group
-      if(itemGroup != nextItemGroup) {
+      if(itemGroup !== nextItemGroup) {
         superHeaders.push({label: itemGroup, span: currentCount});
         currentCount = 0;
       }
@@ -104,7 +104,7 @@ export default React.createClass({
       <tr className='column-groups'>
         {superHeaders.map((superHeader, index) => {
           return (
-            <th key={index} className={superHeader.label == null ? '' : 'column-group'} colSpan={superHeader.span}>
+            <th key={index} className={superHeader.label === null ? '' : 'column-group'} colSpan={superHeader.span}>
               {superHeader.label}
             </th>
           );
@@ -131,7 +131,7 @@ export default React.createClass({
   renderBodyValue(item, column) {
     if ('cell' in column) {
       return column.cell(item,column);
-    } 
+    }
     else {
       return item[column.key];
     }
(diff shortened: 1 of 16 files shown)


alexsoble:
Any feedback on this one @kevinrobinson / @jhilde? Adding to push more of the reviewing-JS work over to the CI, since "eqeqeq" and "no-var" issues came up on a PR earlier this week. I like the idea of enforcing strict equals so that we can avoid situations like:

```
1 == '1'
true
```

app/assets/javascripts/components/flexible_roster.jsx
```diff
@@ -104,7 +104,7 @@ export default React.createClass({
       <tr className='column-groups'>
         {superHeaders.map((superHeader, index) => {
           return (
-            <th key={index} className={superHeader.label == null ? '' : 'column-group'} colSpan={superHeader.span}>
+            <th key={index} className={superHeader.label === null ? '' : 'column-group'} colSpan={superHeader.span}>
```
kevinrobinson: We might want `undefined` to have the same behavior here (as it would with double equals)?  Could flip this to be "if not a string" if that's what we really mean.

app/assets/javascripts/components/star_charts_page.jsx
```diff
@@ -117,13 +117,13 @@ $(function() {
       return _.flatten(students.map(function(student) {
         const studentFields = _.omit(student, 'star_results');
         return student.star_results.filter(function(result) {
-          return result.percentile_rank != null;
+          return result.percentile_rank !== null;
```
kevinrobinson: Might want the same for undefined here?

app/assets/javascripts/components/star_charts_page.jsx
```diff
@@ -117,13 +117,13 @@ $(function() {
...
         }).map(function(result) { return merge(result, { student: studentFields }); });
       }));
     },
 
     resultsDelta: function(student) {
-      const results = student.star_results.filter(function(result) { return result.percentile_rank != null; });
+      const results = student.star_results.filter(function(result) { return result.percentile_rank !== null; });
```
kevinrobinson: undefined?  I'll stop commenting on those, but I suspect this might be common.

app/assets/javascripts/helpers/graph_helpers.js
```diff
@@ -10,10 +10,10 @@ import _ from 'lodash';
   window.shared.GraphHelpers = {
     // Returns a list of monthKeys that are within the time window for this chart.
     monthKeys: function(nowMomentUTC, monthsBack) {
-      var lastMonthMomentUTC = nowMomentUTC.clone().date(1);
+      let lastMonthMomentUTC = nowMomentUTC.clone().date(1);
       return _.range(monthsBack, -1, -1).map(function(monthsBack) {
-        var monthMomentUTC = lastMonthMomentUTC.clone().subtract(monthsBack, 'months');
-        var monthKey = monthMomentUTC.format('YYYYMMDD');
+        let monthMomentUTC = lastMonthMomentUTC.clone().subtract(monthsBack, 'months');
+        let monthKey = monthMomentUTC.format('YYYYMMDD');
```
kevinrobinson: Should these all be `const`?

app/assets/javascripts/helpers/graph_helpers.js
```diff
@@ -28,7 +28,7 @@ import _ from 'lodash';
     // Given a list of monthKeys, map over that to return a list of all events that fall within
     // that month.
     eventsToMonthBuckets: function(monthKeys, events){
-      var eventsByMonth = _.groupBy(events, this.defaultMonthKey);
+      let eventsByMonth = _.groupBy(events, this.defaultMonthKey);
```
kevinrobinson: Should this be a `const`?

kevinrobinson:
@alexsoble Yeah, ++ for adding more linting, either all at once in a big bang using airbnb or whatever else or incrementally.

These seem like the ==/=== changes might not do exactly what we want re: undefined (which would match the == null before, but won't match the === null in changes here).  Might be worth checking those, or changing them to be more strict?

re: let/const, that seems awesome but many of those `let`s seem like they should be `const`s?

alexsoble:
@kevinrobinson Hey thanks for the review, what I'm thinking here is that these two rules should be broken to separate PRs.

PR #1156, opened 2017-10-13
Remove educator searchbar precomputing from nightly task

alexsoble:
# Notes 

+ @jhilde had the right idea about this from the beginning; this should run as its own nightly task
+ The educator searchbar precomputing sometimes takes more than 20 hours to run for reasons we don't fully understand yet
+ That throws an import error email which is high-urgency; put precomputing this JSON to optimize the searchbar endpoint is not high-urgency because it doesn't block educator login. So this should get its own, less urgent error report.

### DIFF
--- a/lib/tasks/import.thor
+++ b/lib/tasks/import.thor
@@ -121,7 +121,6 @@ class Import
 
     def run_update_tasks
       begin
-        Educator.save_student_searchbar_json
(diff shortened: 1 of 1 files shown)


alexsoble:
Just set this up as a scheduled task on Heroku; merging in now.

Issue #1086, opened 2017-09-22
Filter by House and Counselor in Overview (SHS)


snoopyuri:
SHS is broken down into 4 houses (Broadway, Elm, Highland, etc.) - Housemasters will need to be able to filter out the kids in their overview tab and then be able to further filter by absences, discipline, or last SST.  Same for the counselors in each house.  Could possibly fit under tardies.

jhilde:
Have we decided how to approach this, since house is definitely Somerville specific and counselor is drawn from a custom field in Aspen. Thoughts @alexsoble @kevinrobinson @snoopyuri

jhilde:
If we decide yes, I was trying to centralize the student data import add-ons a bit in #1012.

kevinrobinson:
@jhilde Am I following, is the idea here is that there are two new user personas: house masters, and counselors, and we'd be re-purposing the school overview page for them but with different sets of students?

It sounds like both bits of data are stored in custom ways in Aspen, so one option could be to make a separate import that is totally isolated, put all that data in its own table/tables (with keys into the educators or students tables as needed), and then make a UI that reads from those?  That seems like it captures that this is a very specific import, and isolated what relies on it.

Another question might be, how much do we want to support housemasters and counselors as first-class user personas with features built specifically for them?  If we had ideas about why this would be meaningful for students, or how it would support the workflows, meetings or decisions they're regularly making... that might help clarify too.

snoopyuri:
I've got a little more perspective after sitting in the SST meetings at the high school this week.

Each house has its own SST meeting.  Their ideal workflow for the meeting is:

1) Pull up the overview page
2) Click on the filter for their House
3) Sort by attendance - discuss and dig into those kids with high absences/tardies
4) Sort by discipline incidents - do the same

Counselors go through the same process in their work and meetings.

So my gut is that I don't think we need to go as far as crafting the UI for them.  We can do it within the existing filtering system in the overview.  

As far as the custom field part and expansion to other schools, I think that every high school will have a counselor field.  Every high school might not have the "housemaster role" (wish we could get rid of that terminology), but they usually have a system for organizing students into smaller units (whether it's a house system, by grade, by program).

jhilde:
[...]

PR #1158, opened 2017-10-13
Adding House & Counselor to School Overview for high schools

jhilde:
**What does it do?**

1. Imports house and counselor from students_export.txt (generated by Aspen)

- House is one of (Beacon, Elm, Highland, Broadway) for most, if not all, SHS students. It is an empty string for all other schools.

- Counselor is <LASTNAME>, <FIRST_INITIAL> X-<EXTENSION> for SHS students. It is nil for all other schools. 

- Counselor is transformed to split on the first "," (if the feel isn't nil). So just leaving LASTNAME.

2. Adds House and Counselor as filters and table columns on School Overview. This is only added for schools that are high schools. (I contemplated hard-coding it for SHS but it made me too sad 😿.) 

Screen Shot (SHS):
screen shot 2017-10-13 at 1 32 43 pm

Screen Shot (other school):
screen shot 2017-10-13 at 1 34 40 pm

3. Adds House and Counselor as fields on the csv export of School Overview. This is, again, only added for schools that are high schools.

**Things to remember**
As mentioned this is very **Somerville-specific**. When we move to other schools a more configurable strategy for student sub-groupings (like house) will need to be determined.


jhilde:
Accomplishes short-term strategy #1086.

alexsoble:
Hey @jhilde, looks like a great start! Once idea in terms of generalizing: what do you think about, instead of passing down `school` to the React components and making decisions based on school type === HS, passing down some config like "show counsellors" or "show Houses"? That would make the code a bit more flexible. Other districts might want to include counselor data but not Houses, or they might have counselor data for elementary schools. That way the School Overview JS components can be agnostic about schools and just take students + config.

jhilde:
@alexsoble I was actually kind of surprised that the school overview didn't know what school it was giving an overview of. 

Where would you envision show counselors or show houses to be set? It seems like the controller shouldn't be responsible for knowing how to render the UI. 

Eventually we could make the StudentOverview component make the decision and pass filter and view information down to the StudentTable. In a similar way that I did with the Generic Roster component for sections. Not sure that make sense to do as a flag long term.

alexsoble:
@jhilde Yeah, I'd envision the controller passing in the configuration, that's a Rails-y type of MVC pattern where the model stores data, the controller decides which data should flow to the view, and the view renders stuff. 

If you want to save that config type stuff for a later day and just go ahead and ship let's do that for sure! Don't want to block you on shipping at all. I noticed one small Ruby thing but that should take you like 0.5 seconds to fix.

app/reports/students_spreadsheet.rb
```diff
@@ -23,14 +23,26 @@ def csv_string(students)
...
-  def flat_row_hash(student, all_service_types)
-    student_fields = student.as_json.except(*[
-      'created_at',
-      'updated_at',
-      'student_address',
-      'hispanic_latino',
-      'race'
-    ])
+  def flat_row_hash(student, all_service_types, school)
+    if school.school_type === "HS" then
```
alexsoble: This is a tricky one -- since this is Ruby code the triple equals isn't the same as the JS triple equals, use double equals to mean equality in Ruby. https://stackoverflow.com/questions/3422223/vs-in-ruby

jhilde: Thanks! I can never remember which, when, and why.

jhilde:
@alexsoble Yeah, I'm just going to go with this for now. But interesting conversation. 

I can imagine the controller pushing down the view semantics to react. So for instance, each school district is permitted to customize the columns of the roster. That configuration would be stored in the database and the controller would be responsible for pushing the proper json configuration to the React component. 

It definitely muddies things when the V is split between Rails and React.

alexsoble:
@jhilde OK sounds good! Is this PR in a good state to 🚢  or do you want to add/change anything else?

jhilde:
@alexsoble It is, feel free to 🚢 .

app/reports/students_spreadsheet.rb
alexsoble: As a rule of thumb I always default to `===` in JavaScript and `==` in Ruby.

alexsoble:
👍

alexsoble:
@jhilde Woohoo! Deploy/migration went smooth. Right now there's no house/counselor data since the import process hasn't run yet so it looks like this: 

screen shot 2017-10-16 at 2 33 41 pm

But by tomorrow we should be looking at legit production House/Counselor data! Nice work on this and I like the bias toward shipping. :D

PR #1160, opened 2017-10-13
Implement no-var .eslintrc rule

alexsoble:
# Note

+ Pulling this chunk of work out of #1153.


alexsoble:
Clicked around in Chrome to double check the charts and graphs and Student Report, need to do the same in IE.

alexsoble:
IE looks alright too although I wasn't able to open the Student Report PDF in the Windows VM.

alexsoble:
Double checked prod, looks good.

PR #1162, opened 2017-10-16
Make order of importers configurable

### LINKED ISSUE #1144: Make order of importers configurable
# Notes 

+ Some importers govern things like educator access to the tool, so they should probably run first — builds off of #1143 to make sure critical data is always imported

alexsoble:
# Issue

+ Fix #1144.

### DIFF
--- a/lib/tasks/import.thor
+++ b/lib/tasks/import.thor
@@ -117,7 +117,9 @@ class Import
                                 .flatten
                                 .compact
                                 .uniq
+      end
 
+      def sorted_file_import_classes(import_classes = file_import_classes)
         import_classes.sort_by do |import_class|
           [PRIORITY[import_class], import_class.to_s]
         end
@@ -152,7 +154,7 @@ class Import
 
       timing_log = []
 
-      file_import_classes.each do |file_import_class|
+      sorted_file_import_classes.each do |file_import_class|
         file_importer = file_import_class.new(
           school,
           file_import_class_to_client(file_import_class),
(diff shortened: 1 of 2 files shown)


alexsoble:
TODO: Handle case where we add an importer but don't add it to the `ORDER` list. Shouldn't blow up but should probably go last.

lib/tasks/import.thor
```diff
@@ -52,6 +52,20 @@ class Import
...
+      EducatorSectionAssignmentsImporter => 2,
+      StudentsImporter => 3,
+      StudentSectionAssignmentsImporter => 4,
+      BehaviorImporter => 5,
+      AttendanceImporter => 5,
+      StudentSectionGradesImporter => 5,
+      X2AssessmentImporter => 6,
+      StarMathImporter::RecentImporter => 6,
+      StarReadingImporter::RecentImporter => 6,
+    }
```
kevinrobinson: Why not just a list?

jhilde: Why not a json configuration that could be transitioned to be per school district?

alexsoble: @kevinrobinson — I want to express the idea that some importers are app-critical and have dependencies — like educators which go before educator sections which go before students ... — while other importers are the same in terms of priority, like X2 assessments and STAR assessments. A list doesn't express that. 

@jhilde — My vote would be let's see what Fall River's data looks like before making this configurable. :D

kevinrobinson: @alexsoble Sounds great!  Using `priority` instead of `order` might capture that.

alexsoble:
@kevinrobinson 👍🏽👍🏽

alexsoble:
Adding some sorting-specific specs, those are 💚 . Also exercised the import process locally, it ran good and in the correct order. 🎉

PR #1164, opened 2017-10-16
Add "avoid ===" rubocop rule

alexsoble:
# Notes 

+ Based on chat with @jhilde in #1158
+ Ruby style guide notes: https://github.com/bbatsov/ruby-style-guide#no-case-equality


jhilde:
Thanks @alexsoble!

Issue #1121, opened 2017-09-29
Beef up logging from imports


jhilde:
Plan to:

-- Use ImportRecord as a record of the overall import run, as it is currently.

-- Add ImportRecordDetail as a record of each importer (or job) that the import runs. This will have number of rows processed, rejected, and deleted. For most importers only the rows processed will be populated. For the new sections related importers, they will have the number of rejected rows. Eventually, they will also have the number of deleted rows once I change the way stale data is deleted from student_section_assignment and educator_section_assignment.

-- Add low level performance data for generating the searchbar json since we suspect that is causing the timing issue. (Turned on with an environment variable, so we can profile in production).

alexsoble:
Hey, this sounds great @jhilde. Really thorough and well-thought-out. 

I had a branch in my local codebase that tackles this issue in a quick-and-dirty way: https://github.com/studentinsights/studentinsights/pull/1123

It stores some JSON in the database about import timing and prints it in the ErrorReport email.

If you don't mind, we could merge this first for some quick-and-dirty easier-to-access info about timing and then you can implement your more comprehensive solution, which I think will serve us really well especially as we add more districts. Let me know what you think!

jhilde:
Sounds great!

On Sun, Oct 01, 2017 at 4:40 PM Alex Soble

<
mailto:Alex Soble <notifications@github.com>
> wrote:

a, pre, code, a:link, body { word-wrap: break-word !important; }

Hey, this sounds great
https://github.com/jhilde
. Really thorough and well-thought-out.

I had a branch in my local codebase that tackles this issue in a quick-and-dirty way:
https://github.com/studentinsights/studentinsights/pull/1123

It stores some JSON in the database about import timing and prints it in the ErrorReport email.

If you don't mind, we could merge this first for some quick-and-dirty easier-to-access info about timing and then you can implement your more comprehensive solution, which I think will serve us really well especially as we add more districts. Let me know what you think!

—

You are receiving this because you were mentioned.

Reply to this email directly,
https://github.com/studentinsights/studentinsights/issues/1121#issuecomment-333405117
, or

alexsoble:
OK, merging the ugly thing now! @jhilde

kevinrobinson:
we did this!

PR #1166, opened 2017-10-18
WIP - Improve Importer Logging

jhilde:
The basis for this is to beef up importer logging #1121. This is a work in progress and I would value your feedback, especially @alexsoble since you've been doing some work in the importers.

**Current State**
There is an ImportRecord created for each run of the importers. This record contains overall timing information along with some specific timing information @alexsoble added to give immediate insight into our importer performance issues.

There is a basic page off of the admin homepage that reports when each run was started and whether it completed.

**This PR (so far)**
In this first addition to the PR, I've added ImportRecordDetail which represents a run of an individual importer. The import_record_details has the following columns:

- _**:import_record_id**_ (id of the related import_record)
- _**:importer**_ (class name of the file_importer for this record)
- _**:time_started**_ (time the import started)
- _**:time_ended**_ (time the importer ended, should be set with most failures as well)
- _**:status**_ (current status of the importer: pending, started, failed, completed. this is updated int the database when the status changes)
- _**:error_message**_ (if a failure was detected, this should contain the error message)
- _**:rows_processed**_ (#of rows that import_row was called on.)
- _**:rows_excluded**_ (#of rows excluded by the school fiter primarily) _(the sum of rows_processed and rows_excluded should equal the number of lines in the CSV file)_
- _**:rows_created**_ (#of processed rows resulting in a new record in the database)
- _**:rows_updated**_ (#of processed rows resulting in an updated record in the database)
- _**:rows_deleted**_ (#of rows deleted from the database. Currently used by StudentSectionAssignments and EducatorSectionAssignments for stale assignments)
- _**:rows_rejected**_ (#of processed rows that are rejected. For example, an EducatorSectionAssignment where the Educator or Section doesn't exist in the database)

I've also pushed the refactoring of the importers in the direction of a base class to reduce the amount of boilerplate code in each importer.

Specifically I've created the BaseCsvImporter from which all current importers inherited. This base class provides the following:

- An implementation of #import that essentially: logs the start, calls a before_import hook in the derived class (if it exists), reads the CSV file, and for each row...filters school appropriately and calls import_row on the derived class, calls an after_import hook in the derived class (if it exists), and finally logs the importer completion for ImportRecordDetail.

- It proxies the logging functions from ImportRecordDetail through method_missing.

**Work to do**
- Adding specs to thouroughly test ImportRecordDetail and BaseCsvImporter in isolation.
- Updating the import record front end to add ImportRecordDetail
- Test end to end with import.thor.
- Think through adding some sort of ImportEngineDetail for the import post processors like creating the json for the student searchbar.


app/importers/file_importers/base_csv_importer.rb
```diff
@@ -0,0 +1,37 @@
...
+    @data.each.each_with_index do |row, index|
+      if filter.include?(row)
+        import_record_detail.log_processed
+        import_row(row)
+      else
+        import_record_detail.log_excluded
+      end
+      ProgressBar.new(log, remote_file_name, @data.size, index + 1).print if progress_bar
+    end
+    after_import if self.respond_to?(:after_import)
```
alexsoble: Hey @jhilde! I definitely see where you're coming from with the `BaseCsvImporter` class. It feels tidy and clean to wrap up shared logic into one place! Check out #1147 though, we used to have all this logic in one place called `FileImport`, but then we decided to duplicate it across all the importers. 

Why? We wanted to avoid splitting up flow-of-control code across multiple classes. Checks like `after_import if self.respond_to?(:after_import)` make it hard to read through the flow-of-control for each importer class, because you have to skip around in the code to discover what the import process looks like. 

Especially when we have patterns like "import rows, then delete some rows", we want it to be really discoverable and clear at the top of each class how the flow of control pattern works.

app/importers/file_importers/base_csv_importer.rb
```diff
@@ -0,0 +1,37 @@
...
+  def data_transformer
+    CsvTransformer.new
+  end
+
+  def filter
+    SchoolFilter.new(school_scope)
+  end
+
+  # Mirror import_record_detail functionality to simplify importers
+  def method_missing(name, *args, &block)
```
alexsoble: Let's avoid using `method_missing`, it's not easy to read and makes the code harder to debug. Check out the ruby style guide comments: https://github.com/bbatsov/ruby-style-guide#user-content-no-method-missing

db/migrate/20171017011510_create_import_record_details.rb
```diff
@@ -0,0 +1,20 @@
+class CreateImportRecordDetails < ActiveRecord::Migration[5.1]
+  def change
+    create_table :import_record_details do |t|
+      t.integer :import_record_id
+      t.string :importer
+      t.datetime :time_started
+      t.datetime :time_ended
+      t.string :status
+      t.string :error_message
+      t.integer :rows_processed
```
alexsoble: I like this data structure a lot! @jhilde can you say more about the difference between rows_rejected and rows_excluded? Let's document that as a code comment as well.

spec/importers/file_importers/courses_sections_importer_spec.rb
```diff
@@ -31,8 +33,63 @@
...
+            expect(Section.count).to eq(1)
+          end
+
+          it 'updates the section' do
+            section.reload
+            expect(section.room_number).to eq("NEW ROOM")
+          end
+
+          it 'logs one row processed, one row updated' do
+            expected_statuses = [
```
alexsoble: Interesting data structure, kind of an array/hash combination? Can you say a little more about this?

When I run it in the console I get:

```
2.3.0 :001 >           expected_statuses = [
2.3.0 :002 >                 "processed": 1,
2.3.0 :003 >                 "excluded": 0,
2.3.0 :004 >                 "created": 0,
2.3.0 :005 >                 "updated": 0,
2.3.0 :006 >                 "deleted": 0,
2.3.0 :007 >                 "rejected": 1
2.3.0 :008?>             ]
 => [{:processed=>1, :excluded=>0, :created=>0, :updated=>0, :deleted=>0, :rejected=>1}] 
2.3.0 :009 > 
```

So it looks like it becomes an array with a hash inside. Would it be simpler to just use a hash? If not, should we explicitly show the hash within the array to make it clear what the data structure is?

alexsoble:
Cool, thanks @jhilde! Just reviewed.

db/migrate/20171017011510_create_import_record_details.rb
alexsoble: Ooooh now I see you left an explanation about this in the PR summary -- let's add it as a comment to the codebase too. Probably belongs in `app/models/import_record_detail.rb`.

app/importers/file_importers/base_csv_importer.rb
jhilde: Yes I completely appreciated the removal of `FileImport'! For me it was mostly because it used a bit of convoluted composition rather than inheritance. 

My theory for using inheritance here is:
1) Lots of duplicate code. 
2) A desire to push toward increasing configuration over code (a common theme with me) as we anticipate expanding our base of districts.
3) Creating a base class that is thoroughly tested rather than expect each importer to repeat the same test with different fields.

May not be the time which is fine...just got really tired of copying and pasting. ;-)

app/importers/file_importers/base_csv_importer.rb
jhilde: Yeah, so many strange features of Ruby. I'll switch use explicit delegation.

db/migrate/20171017011510_create_import_record_details.rb
jhilde: Yup. Sounds like a plan. 

Another part of my importer vision is to make them deal with data inconsistency in a way that doesn't simply abort the import in an indeterminate state.

So yes, to be concrete:
`rows_excluded` right now is the count of rows not in the school filter.
`rows_rejected` is incremented for a row that fails parsing, referential integrity, something like that.

spec/importers/file_importers/courses_sections_importer_spec.rb
jhilde: Yeah, sorry got my Ruby datastructures a bit mixed up. ;-)

jhilde:
Many thanks @alexsoble. So overall, would you be more comfortable with a return to importers that inherit from `Struct`? No worries, I'm fine either way.

alexsoble:
@jhilde I think even with the inheritance, we have a couple signals in the code that @kevinrobinson taught me to watch out for as a "code smells":

+ Use of `delegate`: means that when you call a method it's not obvious which class to look in for the method definition
+ Use of `if self.respond_to?`: means that you have to check in two class files to read the flow of control

If we can use inheritance in a way that avoids `delegate` and `if self.respond_to?` and keeps the flow of control in one file, I'm all for it! Might be a little less D.R.Y. than using `delegate` and `if self.respond_to?` but worth it for what we gain in readability of flow-of-control.

jhilde:
@alexsoble Great! 

To get rid of the delegate: I'll either just have importers call `import_record_detail.log_*` directly. 

To get rid of the `self.respond_to?`, I'll create no-ops for before_import and after_import. If they're overridden in the individual importers, that code will be called. Otherwise it will just be a quick no-op.

Thanks for helping me think through this!

alexsoble:
@jhilde Cool, curious to see what that will look like. If it keeps all the flow of control code clear to read in one place it should be 👍 .

jhilde:
@alexsoble I made those changes. Will be iterating on the ui to show the `import_record_details` and specs that exercise `BaseCsvImporter`.

alexsoble:
Hey @jhilde! I'm going to suggest @kevinrobinson review this when he gets a chance to. 

The `before_import`, `after_import` strategy is an interesting approach. It definitely makes what's happening more explicit, it still splits up the import process across two files though. 

I'll defer to @kevinrobinson. Thanks @jhilde!

jhilde:
@alexsoble Sounds good.

My chief reason for the inheritance is to DRY out the import method. Just seems really problematic to have copy and pasted code in each of the importers.

When I added the deletes for the Assignment importers, I was really frustrated that I had to include a list of importers that import.thor needs to call delete on. In my OO view of the world, import.thor shouldn't have any knowledge of importer internals. It should simply know that it can call import on importers and they work.

That's just my opinion though, happy to flatten them out and put all the code back to the individual importers.

kevinrobinson:
@alexsoble @jhilde Sure, I'm happy to help.  From skimming quickly, it sounds like some great programming discussion :)  I'll take a closer look later tonight, and then let's get this better logging shipped!

app/importers/file_importers/base_csv_importer.rb
```diff
@@ -0,0 +1,36 @@
+class BaseCsvImporter < Struct.new :school_scope, :client, :log, :progress_bar, :import_record_detail
```
kevinrobinson: Question 2: This looks like you've awesomely factored some bits around processing CSV files into what's essentially a concrete class that does a clear, well-defined thing.  But here it's implementing as an abstract base class that's intended to be subclassed, and that adds some more scope creep.

In particular, I'm noticing the `filter` and `before_import` and `after_import` hooks.  Adding hooks like that are really good if you have to give up control of what's happening to another class.  But for most code, unless you really want to describe things declaratively and commit to those abstractions, adding hooks like this creates  barriers to change.  Here we're taking a step towards essentially creating a small DSL for programming primitives (filter, before, after), which is usually a smell that we don't have the abstraction we need - if we did we'd just write these things in plain Ruby :)

So I'm wondering about what it would look like if we made this a concrete class.  `CsvProcessor` might take a bunch of parameters `(school_scope, client, log, progress_bar, import_record_detail, remote_file_name, data_transformer)`.  It has some defaults, and callers can pass parameters instead of subclassing and overriding methods.  It's single method, `#process` could take a block, which it then yields to for each transformed row.  The caller can do whatever it wants with it with the row, filtering, etc.  With a concrete class like that, individual importer classes can call it in their `#import` methods.  That removes the indirection with the hooks.  Sketching this out, it might look something like this:

```
class SectionsImporter
  def import
    csv_processor = CsvProcessor.new({ remote_filename: 'sections.csv', ... })
    # do something before like initialize a set of records we're tracking
    csv_processor.process {|row| process_row(row) }
    # do something after, like delete stale rows
  end

  private
  # This receives the transformed row, and then we do whatever you want with it, skip it, update, import
  def process_row(row, import_record_detail)
    # should it skip it?
    # import it and log
    nil
  end
end
```

The benefit being that we've removed the concepts around filter, before and after (since we can say them in plain Ruby now), and the control flow is more explicit within each importer.

app/models/import_record_detail.rb
```diff
@@ -0,0 +1,124 @@
+class ImportRecordDetail < ActiveRecord::Base
```
kevinrobinson: Question 1: This adds a great new feature for understanding what happens during individual import tasks, especially when something goes wrong.  An assumption I have is that the most immediately useful thing for improving logging would be to just see the data for the importer jobs in one place in the admin page.  The other would be to see the output from [print_initial_report](https://github.com/studentinsights/studentinsights/blob/master/lib/tasks/import.thor#L141) and `print_final_report`, which isn't included here.

I'm wondering about the benefits that adding this structure provides over writing to the `log` that's already passed and then persisting that (or some similarly minimal API).  The first benefit to this structure that I see is that we can show explicit columns in the admin panel, and we can sort them.  That doesn't seem like it adds much, are there other benefits I'm not thinking of?  It does feel like this is capturing some generally true things about records-based imports (records get processed, skipped, created, updated, deleted) that are unlikely to change.

A downside of adding more structure like this is that it's harder for me to see how we could add more logging for other things.  For example, we might want to track things like how many students are null when importing from STAR? or how many courses were nil for section records?  Here I think we'd have to decide about adding a column, but maybe there are other ways to enable this kind of logging.  A benefit for capturing logs, or just asking each importer to log directly or to return a hash of data is that import tasks can track whatever they want (or they can all track the same things too!).

app/importers/file_importers/base_csv_importer.rb
jhilde: @kevinrobinson You're correct, I was creating what is essentially an abstract class (needs to be subclassed, would throw an error in compiled languages if you try to instantiate it). Typically abstract classes do provide some functionality, otherwise you would use an interface. 

Of course, none of this is very rubyish. I'm likely trying to solve a problem that doesn't yet exist, namely importing a wider variety of files for different districts.

So I think for now, it's best to leave things as they are and add some good logging without changing the object structure of the importers.

kevinrobinson: @jhilde Yeah I think I'd have the same suggestion in Java too - using collaborator objects is another way to share functionality too, and usually has benefits over abstract base classes.  By providing an abstract base class as an API, you're really committing to an abstraction and callers are forced to give up control to the base class, who "owns" the flow of what is happening.  The subclass can only answer specific questions that the abstract base class anticipated in advance, it can't change how it uses the base class' functionality, or compose it in new ways.  And changing the base class becomes increasingly hard, since you can't breaking it apart and recompose it as easily as you can when composing via collaborators and normal method calls.

If you're aiming for reducing duplication, you can often do that more directly by factoring the duplication out into a concrete class that you then use in a few different places with plain method calls.  That makes the control flow more explicit - callers are always in control except for when they call methods on collaborators.  This is most important when you want to preserve flexibility and optionality to compose those pieces in different ways in the future - if you have a few small pieces with clear lines between them, it's easier to build new things from it versus a fatter base class that does all the things.

Anyway, I think it's awesome you're looking at this and at ways to improve factoring!  👍 But agreed that it drifts a bit from improving logging.

Issue #1171, opened 2017-10-19
Revised 2017 MCAS Scaled Scores & Level Names


snoopyuri:
The State just released the MCAS results for 2017 in grades 3-8 and they look very different.  They got into Insights this morning with a few kinks:

+ SGP: Good news is that this is the same and is showing up in student graphs.  Yes (Thought I'd start with good stuff first :)

+ Proficiency Levels in grades 3-8:   Advanced (Old) = Exceeding Expectations (New): Proficient (Old) = Meeting Expectations (New); Needs Improvement (Old) = Partially Meeting (New); Warning (Old) = Not Meeting Expectations (New).    It seems that this would affect the y axis of the graphs in the student profile.  Hopefully, this can just be dealt with by renaming them.

+ Scaled Score - "The Biggie": Old MCAS is scaled 200 to 270.  New MCAS is in the 400 to 600 range.  I'm working on getting some conversion tables, which I'll attach here when I get them.  So when you look at a graph in the student profile, you'll see a student who has been hovering in the 200 range for years and then shooting up to the top of the axis.    Not sure how to tackle.  Do we make it a whole new graph?  Do we convert the new score to its old score equivalent so it fits on the graph?  Also is complicated by the fact that high school is staying the same, so those graphs don't change.

alexsoble:
Thanks for investigating the impact of this change and writing up a detailed summary @snoopyuri! Hmmmm yeah that's some "off the charts" growth!
![screen shot 2017-10-19 at 1 16 18 pm](https://user-images.githubusercontent.com/3209501/31786987-ea4ee938-b4cf-11e7-9b03-c0699d8d7762.png)

My preferred approach would be to make a whole new graph for 2017-and-beyond-MCAS, and a whole new database model. Converting obscures the original data point which puts us into murky territory. I can start tackling this.

alexsoble:
@snoopyuri How would you describe this new MCAS score? "New MCAS"? "2017 MCAS"? Is there an official name for this change?

snoopyuri:
Here they’re calling it MCAS 2.0 or Next Generation MCAS.

alexsoble:
@snoopyuri Thanks Uri! Helpful. Naming things is weirdly one of the hardest things in programming...

alexsoble:
@kevinrobinson @jhilde Let me know if you think I'm traveling down a solid path with #1173 and if you have any suggestions or comments! Thanks!

alexsoble:
[...]

PR #1172, opened 2017-10-19
WIP – Remove old assessment-related tables

alexsoble:
# Notes 

+ Checked to see if the Assessments table was normalized by assessment family and subject or not to start work on #1171; found that Assessments are not normalized, but those old tables are still hanging around the schema/database
+ This PR removes the cruft tables to clean things up for #1171


alexsoble:
Ran the migration against the demo app and clicked around to triple-sanity-check this, because the PR looks scary even though it's not really since these tables are unused. 😨 👍

PR #1173, opened 2017-10-19
WIP – Handle new flavor of MCAS data

alexsoble:
# Issue 

+ #1171

# TODO 

+ [x] Add new model for "Next Generation" MCAS
+ [x] Add fake data generator with new "Next Generation" scaled scores
+ [x] Add new chart to Student Profile if relevant data is present 
+ [ ] Think about what impact this should have on the School Overview page 
+ [ ] Think about what impact this should have on the Homeroom page 
+ [ ] Student Report
+ [ ] Update X2AssessmentImporter code to correctly classify incoming MCAS data from X2
+ [ ] Write migration to re-assign "Next Generation" student assessments to the correct parent `Assessment`

### DIFF
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -4,6 +4,7 @@ class Assessment < ActiveRecord::Base
   validate :has_valid_subject
 
   VALID_MCAS_SUBJECTS = [ 'ELA', 'Mathematics' ].freeze
+  VALID_MCAS_FLAVORS = [ 'Next Generation MCAS', 'Regular' ].freeze
   VALID_STAR_SUBJECTS = [ 'Mathematics', 'Reading' ].freeze
   VALID_ACCESS_SUBJECTS = [
     "Composite", "Comprehension", "Literacy", "Oral",
@@ -28,6 +29,8 @@ class Assessment < ActiveRecord::Base
     Assessment.create!([
       { family: "MCAS", subject: "Mathematics" },
       { family: "MCAS", subject: "ELA" },
+      { family: "MCAS", subject: "Mathematics", flavor: "Next Generation MCAS" },
+      { family: "MCAS", subject: "ELA", flavor: "Next Generation MCAS" },
       { family: "STAR", subject: "Mathematics" },
       { family: "STAR", subject: "Reading" },
       { family: "ACCESS", subject: "Composite" },
--- a/app/assets/javascripts/student_profile/ela_details.jsx
+++ b/app/assets/javascripts/student_profile/ela_details.jsx
@@ -80,9 +80,40 @@
         <div className="ELADetails">
           {this.renderNavBar()}
           {this.renderStarReading()}
+          {this.renderMCASELANextGenScores()}
           {this.renderMCASELAScores()}
           {this.renderMCASELAGrowth()}
         </div>
+      );
+    },
+
+    renderMCASNextGenLink() {
+      const data = this.props.chartData.next_gen_mcas_ela_scaled;
+
+      if (!data || data.length === 0) return null;
+      return (
+        <span>
+          <a style={styles.navBar} href="#NextGenScores">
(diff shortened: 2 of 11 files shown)


app/models/assessment.rb
```diff
@@ -28,6 +29,8 @@ def self.seed_somerville_assessments
     Assessment.create!([
       { family: "MCAS", subject: "Mathematics" },
       { family: "MCAS", subject: "ELA" },
+      { family: "MCAS", subject: "Mathematics", flavor: "Next Generation MCAS" },
+      { family: "MCAS", subject: "ELA", flavor: "Next Generation MCAS" },
```
kevinrobinson: Adding `flavor` seems like it adds complexity to the whole data model for this particular assessment.  I'm wondering how it compares to using the `family` field and calling these "Next Generation MCAS" - are there benefits to adding another field for this particular assessment?

alexsoble: @kevinrobinson Yeah, the benefit I have in mind is keeping the code path simple for all of these attributes on the student model:

```
+  most_recent_mcas_math_growth
+  most_recent_mcas_ela_growth
+  most_recent_mcas_math_performance
+  most_recent_mcas_ela_performance
+  most_recent_mcas_math_scaled
+  most_recent_mcas_ela_scaled
```

For all these attributes, we don't care if the latest MCAS result is the old kind or the new kind — we're just surfacing a number or a string, and we don't necessarily need to add a lot of context around it (for example, in the homeroom view).

If we make a new family called "Next Generation MCAS" we'll have to update the code paths for all of these attributes and merge together Next Generation MCAS results with old-style MCAS results.

kevinrobinson: Yeah, that makes sense!  I was wondering if we wanted to render the two kinds of assessments the same way if they were on different scales and had different performance levels.  In a table like on the homeroom view, would we want those two kinds of scores to look the same and be in the same column?

My assumption is that it might be important to add that context, so that when you ask how a student did on MCAS, you probably need to know whether it was MCAS or MCAS 2.0 and how it was scaled in order to understand what that number means.

alexsoble: Let's ask @snoopyuri what his assumption is? Since the two ranges don't overlap at all (200 to 270 vs 400 to 600) my assumption is that educators won't need an explicit label to know if they're looking at a 1.0 or a 2.0 score.

alexsoble:
# How will the School Overview need to change? 

To the end user, the School Overview looks like it's just filtering by performance category, a string value (fake distributions below): 

screen shot 2017-10-19 at 6 02 36 pm

But behind the scenes in `SlicePanels.jsx`, it's reading an integer value and converting it to performance category:

screen shot 2017-10-19 at 6 08 21 pm

This data is coming from precomputed `Student#most_recent_mcas_math_scaled` and `Student#
most_recent_mcas_ela_scaled`.

One approach would be to test the integer score and surface Next Gen MCAS categories if we detect scores above 400.

Another would be to ditch testing integers altogether in this slice panel and just use the strings that surface from X2 to let users filter. 

One advantage to the strings-only approach is that the categories could naturally bubble up into the School Overview page from X2 as students go through the system. For example, Uri mentioned that the high school is going to use the old system. Students who go to HS next year might get different MCAS performance categories than they're getting this year.

alexsoble:
# How will the Homeroom Roster page need to change? 

screen shot 2017-10-19 at 6 04 37 pm

If we go with the "flavors" data structure, I don't think it will need to change at all. As students take the Next Generation MCAS tests, their most_recent_mcas_math_scaled / most_recent_mcas_ela_scaled scores will update from the old performance categories and 200-280 scale to the new categories and the 400-600 scale. And when students graduate to HS (assuming the HS stays with the old MCAS for at least 1 more year), their scores will naturally revert to the older scales.

alexsoble:
Closing this in favor of #1178.

Issue #1180, opened 2017-10-20
Use ES6 modules and import/export everywhere to create single entrypoint to JS


alexsoble:
# Notes

+ Currently our JS uses a mix of `import`/`export` and stuffing JS definitions into `window.shared`.
+ To move to a more modern JS tooling like webpacker or create-react-app, we'll need to create a single point of entry to our JS using modern ES6 `import`/`export`.
+ This is a dependency to unblock #1161.

# What does this look like?

+ See #974 as an example of how to move a JS file away from `window.shared` and into an ES6 module. 

# Files to convert to ES6 modules with import/export
**top level:**
+ [x] datepicker_config.js (https://github.com/studentinsights/studentinsights/pull/1182)
+ [x] session_timeout_warning.js (https://github.com/studentinsights/studentinsights/pull/1182)
+ [x] student_searchbar.js (https://github.com/studentinsights/studentinsights/pull/1182)

**components:**
+ [x] collapsable_table.jsx (https://github.com/studentinsights/studentinsights/pull/1224)
+ [x] fixed_table.jsx (https://github.com/studentinsights/studentinsights/pull/1224)
+ [x] flexible_roster.jsx (https://github.com/studentinsights/studentinsights/pull/1224)
+ [x] slice_buttons.jsx (https://github.com/studentinsights/studentinsights/pull/1224)
+ [x] slice_panels.jsx (https://github.com/studentinsights/studentinsights/pull/1224)
+ [x] star_charts_page.jsx ( (https://github.com/studentinsights/studentinsights/pull/1224 and https://github.com/studentinsights/studentinsights/pull/1223)
+ [x] students_table.jsx (https://github.com/studentinsights/studentinsights/pull/1224)

**helpers:**
+ [x] feed_helpers.jsx (https://github.com/studentinsights/studentinsights/pull/1483)
+ [x] filters.jsx (https://github.com/studentinsights/studentinsights/pull/1224)
+ [x] from_pair.jsx (https://github.com/studentinsights/studentinsights/pull/1483)
+ [x] graph_helpers.js (https://github.com/studentinsights/studentinsights/pull/1483)
+ [x] mixpanel_utils.jsx
+ [x] profile_details_style.jsx (https://github.com/studentinsights/studentinsights/pull/1564)
+ [x] prop_types.jsx (https://github.com/studentinsights/studentinsights/pull/1608)
+ [x] react_helpers.jsx (https://github.com/studentinsights/studentinsights/pull/1608)
[...]

Issue #1161, opened 2017-10-16
Replace Browserify?


jhilde:
In our team meeting on 10/13, a couple of us were complaining about the very long build times browserify takes as we're developing javascript. So we decided to look a bit at webpack.

Turns out Rails is moving toward a more modern understanding of the place of javascript on the front end. [Evidenced by a javascript folder now at the app level, not relegated to being in the app/assets]. Rails now provides the [webpacker gem](https://github.com/rails/webpacker) to facilitate webpack in rails apps. It appears that as of Rails 5.1, this is the recommended method to building and packaging javascript assets. 

The dev experience includes a wrapper around webpack-dev-server that allows hot recompiling of javascript. It also includes all the goodness of webpack in terms of code splitting and dependency mapping. (I haven't tried it with the complete app but the speed looks promising. I often get caught in Browserify hell. Webpack at least shows you progress unlike browserify).

From a first look, my short term recommendation is that we create one large entry point (bundle) for all the application code. This will be the easiest migration path because of the way our javascript is currently setup.

As we migrate to React 16 and hopefully move to using import rather than window.shared, it may make sense to create an entry point for the vendor javascript that is common to most of our pages and perhaps an entry point for each page. 

@alexsoble and @kevinrobinson, have you used webpacker in other projects? Any thoughts on stragegy?

alexsoble:
Yes I use webpacker in a Rails 5.1 project! 👍

alexsoble:
@jhilde In terms of prioritizing code-quality features, would you rank this one as more important or #1121?

kevinrobinson:
Sounds good!  Here's what we learned trying to do this last time: https://github.com/studentinsights/studentinsights/pull/861#issuecomment-284084767, which might be outdated, or we might find other ways of tackling it.

It'd be awesome if this is a drop-in replacement, but my assumption is that it won't be and that iterating on the production config will require testing in the demo or staging site.  If not, then building towards JS-centric tooling that's separate from Rails seems like a better long-term path (eg `create-react-app`).

jhilde:
Note to self, search the issues. ;-)  I recalled there being some conversation but wasn't up to speed on webpack, etc. 

I'm a big fan of create-react-app backed with a restish API.

[...]

PR #1178, opened 2017-10-20
Handle Next Gen MCAS as a separate assessment

alexsoble:
# What does this do? 

+ Handles new "Next Gen" MCAS data. This new format has a new range for scaled scores (400 to 600) and new performance levels.
+ This PR includes a migration to detect and update "Next Gen" MCAS records by checking which range their scaled scores fall into.
+ It also updates the `McasRow` importer class to detect future "Next Gen" MCAS records that come through from X2.

# How does the UI change?

## School Overview 

+ If the majority of students in the filters have most recent "Next Gen" MCAS scores, the School Overview shows the "Next Gen" performance levels as filters. For example, 9th graders will have "Next Gen" scores as their most recent since the last time they took MCAS was at the end of 8th grade, and "Next Gen" was administered in grades 3-8.

screen shot 2017-10-24 at 12 57 56 pm

+ If the majority of students in the filters have most recent old MCAS scores, the School Overview shows the old performance levels as filters.

screen shot 2017-10-24 at 12 58 07 pm

## Student Profile 

+ New charts to show "Next Gen" scores, scales, and categories when a student has "Next Gen" results. (Since there's only one Next Gen score per student in the database these graphs show points right now instead of lines.)

screen shot 2017-10-24 at 1 03 33 pm

## Homeroom

+ The homeroom view doesn't change. 

# Issue 

+ Fix  #1171.

# Change list 

+ [x] Add new model for "Next Generation" MCAS
+ [x] Add fake data generator with new "Next Generation" scaled scores
+ [x] Add new chart to Student Profile if relevant data is present 
+ [x] Update School Overview page 
+ [x] Update Homeroom Roster page 
+ [x] Update Student Report
+ [x] Update X2AssessmentImporter code to correctly classify incoming MCAS data from X2
+ [x] Write migration to re-assign "Next Generation" student assessments 
+ [x] Fix spec
+ [x] Add more spec coverage for areas that need it
+ [x] QA against real data locally
+ [x] Add screenshots of UI and explanations about changes to PR
+ [ ] Deploy, run migrations, and triple-check in IE

PR #1165, opened 2017-10-17
Flatten out DiscontinuedService

Jngai:
#1112 

# Steps toward merge (reviewer: @alexsoble)

+ [x] Code review
+ [x] Test data migration against demo data locally
+ [x] Test data migration against demo data on Heroku demo site
+ [x] Demo site/Internet Explorer: Test record/discontinue UI on Internet Explorer, make sure `discontinued_at` and `discontinued_by_educator_id` are set correctly 
+ [x] Test service upload UI (demo site/Chrome — filesystem not working in IE VM)
+ [x] Test Student Report PDF rendering  (demo site/Chrome — filesystem not working in IE VM)
+ [ ] Capture a backup of database before deploying
+ [ ] Deploy and run `heroku run rake services:move_discontinued_at_to_services`

PR #1182, opened 2017-10-21
Remove browserify-rails and Teaspoon, use Webpack and Jest outside of Rails

kevinrobinson:
This has two major changes, aimed at improving the developer experience for front end work (from discussion in https://github.com/studentinsights/studentinsights/issues/1161).  First, it removes browserify-rails and builds most JS instead with Webpack outside of Rails altogether.  Second, it removes Teaspoon and adds Jest as the test runner.

For developers, this means running `yarn start` instead of `bundle exec rails s` and then just working like normal.  And running `yarn test` to run tests instead of using Teaspoon.  The compilation time in development is ~500ms (on save), a test run takes ~10 seconds, and a production build takes ~15 seconds.

In this pull request, almost all the complexity is in changes to the tests, and so the core change for development and production mode is best seen in the [first commit](https://github.com/studentinsights/studentinsights/commit/8fddf48b25e61fc14961fcb935bc59dbafd1ae7b).  The third commit cleans out the Travis build a bit too, since we don't need phantom binaries (or particular old versions of bundler).

### What would it take to deploy this?
- [x] get feedback on this approach
- [x] verify this works for other folks (including PDF rendering)
- [x] fix problem with Highcharts
- [x] fix PDF rendering
- [x] check production build bundle sizes
- [x] check react-select and styles
- [x] manually test locally in IE11
- [x] test deploys
- [x] verify gzipping works in production
- [x] verify product works in production in IE
- [x] verify PDF rendering in production
- [x] update README docs

If this seems good, we could ship this next week.

### How would development work?
Running `yarn start` spins up both Rails and Webpack.  Rails runs as normal, but there's a new script tag in `application.hml.erb`, that loads `public/dev/bundle.js`.  Webpack runs in watch mode, building all the JS and putting it there for Rails to read.  There's no live or hot reloading, but JS is recompiled on save and ready on the next page load.

The entrypoint is a file that replicates the ordering that was in `application.js`, and imports all those files as side-effects, where they rebuild the same `window.shared` object.  This has the same effect as just concatenating all the files like Sprockets does (and @alexsoble I should have thought of this earlier!).  We can still keep incrementally migrating to ES6 modules in like we could with browserify-rails for https://github.com/studentinsights/studentinsights/issues/1180.

### How would production deploys work?
There's a separate Webpack config for a production build, and there's a Ruby hook to run that after `rake assets:precompile` and puts it in `public/build`.  The script tag in `application.html.erb` loads that instead of the development bundle in production.  I didn't test it in staging yet, but this should require no changes to deployment.

#### Teaspoon
I got a bit stuck here.  I started trying to make Teaspoon work with browserify-rails to start, and can't remember why that didn't work.  Removing it, the tests didn't work anymore since they use import/export.  They were getting compiled to ES5 somehow, but there was no module loader so `require` calls were failing.  This didn't make much sense to me, so I switched gears.

#### Jest
Adding Jest required two main changes related to the test runner.  First, we use the `toContainText` matcher a bit, and that's a `jquery-jasmine` matcher.  The way to add matchers is Jest seems clunky (you can't do it globally, and have to add them inside a `beforeEach` block).  We could do that but here I dropped the matcher and switched these to `expect($(el).text()).toContain('foo')`.  Second, we use instance variables in a few places, most importantly in the `SpecSugar.withTestEl` method.  `this` isn't passed around in Jest, so I changed this to pass a `container` object that the helper code injects the test element into later.  Others I just changed to lexically scoped variables.

After that, there were a bunch of changes needed to the tests, mostly around dependencies that were still in application.js and so not present in the test environment.  There are things like jQuery plugins (autocomplete, datepicker, Highcharts), Sanitize, Cookie, moment and d3.

One thing that's lost in Jest is being able to use Web Inspector to debug tests, but Jest has [support for node-inspector](http://facebook.github.io/jest/docs/en/troubleshooting.html) if we want to add it.

### Other notes
There's a separate `student_report_pdf.js` bundle for `wkhtmltopdf` too.

#### Could we just make browserify-rails faster?
I looked at browserify-rails and added some logging, but hard a hard time understanding where the slowdown was coming from.  `browserify-rails` is a Sprockets postprocessor, and there were ~2 second delays happening in some other code in between the calls into the Sprockets postprocessor ([logs](https://gist.github.com/kevinrobinson/5dd90c4fde0dd61c969b7a1255b3c4d3)).  Commenting out the body of the class cut the time down to zero, so I don't really understand what was slowing it down.  I read a bit of Sprockets to see if I could understand, but then stopped.

#### Why not use Webpacker?
This would be great!  But it didn't just work last time we tried.  Separately, there's benefits from removing the Ruby/Rails middleman in terms of being able to debug and configure, so I'm probably a bit biased towards that too.  The build steps are just plain JS.

####  What about create-react-app?
There's a bunch of other work to make a single-page app outside of Rails (eg, authentication, routing, styles).  I wanted to avoid that here, while taking a step in that general direction.


jhilde:
Whoa @kevinrobinson, the elves were at work last night!

I had been trying a similar switch the rails application.js into an entry point with lots of random imports strategy with webpacker. Was getting some traction but certainly not complete.

Is the structure in jest similar enough to jasmine that fixing some more global dependencies will make all of our tests work? 

Will dig in more at some point. I need to find some of those elves you have though!

kevinrobinson:
@jhilde Yeah, tests are all green.  jest is mostly a drop-in replacement, there were two changes related to the test runner (see above).  Most of the work was around moving the dependencies that are still managed in Rails to work in Jest (which doesn't know about them).  That's just an intermediate step to not have to empty out application.js all in one go.

I also noticed this morning that `jasmine.createSpy` still worked, seems that's a niceness for migration ([code](https://github.com/facebook/jest/blob/6f7ece619b2dc722d27038c6077864109fd920f7/flow-typed/npm/jest_v20.x.x.js#L512)) but I converted those fully now too.

alexsoble:
Ayeeeeee @kevinrobinson this is great! Congrats + thanks for putting your brainpower into figuring out how to make this migration happen. And in one PR too!

At first I was confused about how we could `import` files that don't `export` anything, like in `legacy.js`, then I did a little StackOverflow digging and found this: 

https://stackoverflow.com/questions/41179828/es6-import-equivalent-of-require-without-exports/41179861#41179861 

... which helped me understand how this works. This strategy of importing all the legacy files as side-effects is awesome and lets us bypass soooo much mechanical work! :D

I was able to verify `yarn start` locally, and could click around and use the web app with the new build process! 🎉 🎉 

One piece that didn't work for me yet was the PDF rendering. Clicking "generate student report" just gave me an infinite spinny circle in my browser tab. You mentioned that there's a separate student_report_pdf.js bundle for wkhtmltopdf too... does that need any explicit loading or starting on the developer's end?

I also tried deploying this branch to the `somerville-teacher-tool-demo` Heroku app, the push failed and here's why:

```
remote:        $ NODE_ENV=production webpack --config ui/config/webpack.prod.js
remote:        sh: 1: webpack: not found
remote:        error Command failed with exit code 127.
remote:        info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
remote:  !
remote:  !     Precompiling assets failed.
remote:  !
remote:  !     Push rejected, failed to compile Ruby app.
remote: 
remote:  !     Push failed
remote: Verifying deploy...
remote: 
remote: !	Push rejected to somerville-teacher-tool-demo.
remote: 
```

Looks like we need to tell Heroku to download webpack before it does anything else to build the rest of the JS? I can dig into this some more tonight or Monday.

Overall, looks like we're within striking distance! @kevinrobinson I like your checklist and I agree that we should be able to ship this next week!

kevinrobinson:
@alexsoble Ah thanks!

For the deploy, `webpack` is specified in package.json as a devDependency.  So a first thing to check is whether Heroku runs `yarn install` on a Rails deploy (I think it will).  If not we can run it ourselves [here](https://github.com/studentinsights/studentinsights/pull/1182/files#diff-52cfe83d7f9f1c952a7995dfb0f99341R5).  One wrinkle is that in a production node project you typically wouldn't install `devDependencies`, so we may need to add a flag that tells yarn to install those as well.

For the PDF, yep I think this just isn't working yet and is the last substantive bit.  I thought the other day I checked back on master and had the same problem too, but confirmed now it's this branch.  I don't understand what's up yet, or how to debug.  When we ask `wicked_pdf` to render the PDF, we pass an option telling it to wait for a [magic window.status value](https://github.com/studentinsights/studentinsights/blob/master/app/controllers/students_controller.rb#L71).  That's how it knows to wait until the render is done, so I thought that might be the problem.  But I looked at this just now by adding `&debug` at the end and I can see it's set just fine.  But without `&debug` the `wkhtmltopdf` process hangs and doesn't even ever timeout, I have to kill the process.

kevinrobinson:
Added https://github.com/studentinsights/studentinsights/pull/1182/commits/dfd6db7d87b09bc4dc7dbea231bb28f5ee44d553, which removes all of application.js except for jquery, jquery_ujs (not sure if we use that), and jquery-ui.

kevinrobinson:
Okay, https://github.com/studentinsights/studentinsights/pull/1182/commits/3779ff221c440b55d8370fe44c6f5a0a93d45a14 addresses the PDF issue.  It was from two things: the charset wasn't set and so UTF characters like π would throw, and I removed the `Function.bind` polyfill since I assumed any shims would be handled by the `react-app` Babel preset.  This doesn't go that far back though, so I added back `es5-shim` and imported that only in the PDF file.

FWIW, debugging this worked best by grabbing the `whhtmltopdf` command on the Rails output, running that directly on the command line, and then manually updating the HTML file with new JS while iterating.  There's a `javascript-delay` option which delays before exiting, and a `debug-javascript` option that lets your `console.log` output show up on the command line when running `wkhtmltodpf` directly.

kevinrobinson:
### Development
Aside from compilation times, this also speeds up the local development experience quite a bit because of reducing the number of files the browser has to fetch (it only makes 6 requests in parallel and so has to block even requesting them all).  Check out the bottom line here: the number of files, total MB and time.  This is a hard reload of a page that's already been compiled and nothing has changed.

Before:
screen shot 2017-10-24 at 7 50 22 am

After:
screen shot 2017-10-24 at 7 49 41 am

### Production
This also improves the production build's bundle size.  Before, it was:
- application.js 2.7MB (909kb gzipped)

After it is:
- bundle.js 974kb (310kb gzipped)
- application.js 795kb (206k gzipped)

There's plenty of ways to optimize that more (right now we include all of jQuery UI in application.js when we only need the datepicker and autocomplete).  But I think this is an improvement in both dev and production and good to ship in this respect.

alexsoble:
👏 👏 👏  Fantastic improvements right off the bat @kevinrobinson!

I started working on deploying to Heroku (demo site) on this branch: https://github.com/studentinsights/studentinsights/tree/deploy-webpack-jest-to-heroku

If it makes sense to you, once I get the branch to deploy to the demo site I can open a PR from my branch into this one. Or happy to go with a different deploy strategy / merge strategy if you prefer!

alexsoble:
Deployed to https://somerville-teacher-tool-demo.herokuapp.com/! Deploy is live, cool to see `<script src="/build/bundle.c8c8326f17126602d8b8.js"></script>` at the bottom of the page. 

And maybe it's just placebo but the app definitely feels ... faster.

Opened #1187 to merge the deployment branch into this PR.

kevinrobinson:
@alexsoble Awesome, thanks!  Do you think this is good enough to check off "verify deploys" in the issue's checkbox then?

kevinrobinson:
Also your demo warning banner is amazing :)

alexsoble:
@kevinrobinson Thank you! Tried to keep it nice and subtle!

alexsoble:
@kevinrobinson I'll leave it up to you to decide if we want to check off "verify deploys". Clearly we can get it deployed by moving dependencies out of devDependencies but I'm happy to work towards a more elegant solution so that we fix it in a 👌 way.

kevinrobinson:
Added `Object.assign` polyfill with https://github.com/studentinsights/studentinsights/pull/1182/commits/b7b716b9de2ceed9fde16d1bb1f81cebd305a1ee to address bug affecting IE11.  This is tricksy since the call to Object.assign isn't in our code but is produced in the Babel transpilation process, and so our eslint rule from https://github.com/studentinsights/studentinsights/pull/1002 can't detect it.  This also removes the need for the rule in https://github.com/studentinsights/studentinsights/pull/1002 altogether.

kevinrobinson:
Manual testing in IE surfaced a bunch of small UI issues (https://github.com/studentinsights/studentinsights/issues/1189 and https://github.com/studentinsights/studentinsights/issues/1190 https://github.com/studentinsights/studentinsights/issues/1191), but these are unrelated and not going to address them here.

It also revealed that the STAR charts have been broken for a week and a half, but that's unrelated and opened https://github.com/studentinsights/studentinsights/pull/1158 for that, not going to address it here either.

kevinrobinson:
Verified demo deploys work with the change in https://github.com/studentinsights/studentinsights/pull/1187#issuecomment-339048677, added in https://github.com/studentinsights/studentinsights/pull/1182/commits/bc87b46620d174fb9a47ffb5d0307204ed182334.

kevinrobinson:
Okay, verified that PDFs are good in the production demo site.  There are some minor style differences, but not important enough to look further right now.  Note there's one extra label and the legend shape is different.  Maybe I grabbed the wrong minor version (we're two major versions behind).

### Before
screen shot 2017-10-25 at 9 01 11 am

### After
screen shot 2017-10-25 at 9 01 06 am

kevinrobinson:
@alexsoble @jhilde This should be ready to go now, just needs documentation updates.  The gist is `yarn start` in one terminal to start Rails and Webpack and `yarn test` in another to run the tests.  And `rm -rf node_modules && yarn install` to setup the first time (or on dependency changes).

Can you both check this out at some point (the first commit https://github.com/studentinsights/studentinsights/commit/8fddf48b25e61fc14961fcb935bc59dbafd1ae7b is the important one) and share feedback or give a thumbs up when you feel good about this?

After that I can ping everyone with a heads up about this, and can help resolve merge conflicts on any open PRs (but I don't think we will have many).

lib/tasks/assets.rake
```diff
@@ -2,5 +2,7 @@
 # production Webpack build is run after `assets:precompile`.
 # Monkey patching, from https://github.com/rails/webpacker/commit/d2a6537e6ab0896a9bf14b6eddc5c0961db2c802
 Rake::Task['assets:precompile'].enhance do
-  exec 'yarn run build'
+  # Run yarn install first to force devDependencies to be installed, 
+  # so that we can run the build process
+  exec 'YARN_PRODUCTION=false yarn install --frozen-lockfile && yarn run build'
```
alexsoble: 👍

alexsoble:
I feel good about this. Pulled the latest changes, ran `yarn start` again, and explored the build. I also wrapped my mind around the legacy.js `import` strategy. 👍 

Open PRs that may have merge conflicts—especially in the JS specs, since many of the-features side `.jsx` files are amazingly untouched by this PR—are #1178, #1165, and #1154.

kevinrobinson:
Docs are updated in https://github.com/studentinsights/studentinsights/pull/1182/commits/e533d7494c258cc554ce663504c80d503b8b5bbf.  I made progress on Docker in https://github.com/studentinsights/studentinsights/pull/1182/commits/d972ca47087a7ea597e2ad08c21bedca2340fcc9, but this branch can't yet build the container, `yarn install` fails saying "Yarn requires Node.js 4.0 or higher to be installed."

kevinrobinson:
The Docker setup is working in https://github.com/studentinsights/studentinsights/pull/1182/commits/60ddf27367bb9bff98efe82980067f8729d7f6f7, but performance is not good on my computer.  I'm not sure if this is a regression or just related to the large number of files.  For me, `docker-compose build` took 10m.  The initial Webpack build running `yarn start` inside the container took 71246ms.

I added a `:cached` hint which helped a little, bringing a page load from ~12 seconds to ~9 seconds (with no changes or anything, that's just time to load the page).

Going to push this up, consider this done, and ask about removing Docker support altogether in https://github.com/studentinsights/studentinsights/issues/1209.

kevinrobinson:
🚢ing cc @alexsoble

PR #1184, opened 2017-10-21
Remove old Travis env variables

kevinrobinson:
The env variables are from the old AWS tasks in here: https://github.com/studentinsights/studentinsights/pull/1183, and I don't think we have any Webhooks going on anymore.  @alexsoble maybe you remember more about that?


alexsoble:
@kevinrobinson Yep, this is good to remove. The webhook was a CFA thing that showed the CI status of all the CFA projects on a screen in the main office -- but now when you go to http://project-monitor.codeforamerica.org/ it throws a Heroku error page saying "There's nothing here, yet."

kevinrobinson:
@alexsoble thanks!

PR #1187, opened 2017-10-24
Unblock webpack Heroku deploys

alexsoble:
# Notes 

+ Unblock webpack Heroku deploys by moving webpack-required dependencies out of  `devDependencies`
+ From the Heroku/nodejs docs: "...since you usually don’t want all development dependencies in your production builds, it’s preferable to move only the dependencies you actually need for production builds (bower, grunt, gulp, etc) into dependencies." (https://devcenter.heroku.com/articles/nodejs-support)

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,24 +28,24 @@
     "react": "15.4.2",
     "react-addons-test-utils": "15.4.2",
     "react-dom": "15.4.2",
-    "react-modal": "^3.0.4"
+    "react-modal": "^3.0.4",
+    "webpack": "^3.8.1",
+    "webpack-manifest-plugin": "^1.3.2",
+    "webpack-merge": "^4.1.0",
+    "clean-webpack-plugin": "^0.1.17",
+    "compression-webpack-plugin": "^1.0.1"
   },
   "devDependencies": {
     "babel-jest": "^21.2.0",
     "babel-loader": "^7.1.2",
     "babel-preset-react": "^6.24.1",
     "babel-preset-react-app": "^3.0.3",
-    "clean-webpack-plugin": "^0.1.17",
-    "compression-webpack-plugin": "^1.0.1",
     "concurrently": "^3.5.0",
     "eslint": "^3.3.1",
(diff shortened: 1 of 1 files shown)


kevinrobinson:
@alexsoble Awesome! 👍 

Would you be open to trying one other thing first, which is to tell yarn "install both sets of dependencies?"  That keeps the `package.json` accurate about which things are for development and which things are for the actual app.  I think adding `YARN_PRODUCTION=false yarn install --frozen-lockfile` before `yarn run build` in the `assets.rake` hook might do the trick.

If that doesn't work, it'd be good to also see if moving these to `devDependencies` doesn't increase the bundle size.  I don't think it will, and even if it does I think we should still ship to unblock it and tackle later, but running a building locally could quickly confirm/disconfirm.

kevinrobinson:
@alexsoble Yeah I think in general we want NODE_ENV=production, and if we want to change the npm/yarn config to behave in a non-production way in production, we should do that in a limited specific way (eg., on the command line rather than as an environment variable in general).

The Heroku docs also apply to when you don't directly run `yarn install`, but here we do since we're a Ruby dyno and we're manually running `yarn install` in `assets.rake`.  Normally folks can't just add this in a more limited way.

alexsoble:
@kevinrobinson OK that makes sense! I tried force-pushing the tip of the `experiment/webpack-jest` branch to the demo site, expecting the deploy to fail since webpack etc. are in `devDependencies`, but the deploy succeeded! My guess it that something must have been cached from my earlier deploys. 😅

kevinrobinson:
Ah that is scary if deploys are not isolated and repeatable!

I can poke around at this more later this week too then, but will leave it
for last. Would rather ship to clear the large PR and revisit this after.

kevinrobinson:
This is working now in https://github.com/studentinsights/studentinsights/pull/1182#issuecomment-339215449

kevinrobinson:
Thanks @alexsoble!

PR #1195, opened 2017-10-25
Remove student_searchbar_json from Educator#as_json

kevinrobinson:
In looking at the production bundle sizes for https://github.com/studentinsights/studentinsights/pull/1182, I noticed that there was a ~600kb difference between the HTML files in production and the demo site.  It looks like this is from serializing `current_educator`, which includes the `student_searchbar_json` field introduced in https://github.com/studentinsights/studentinsights/pull/1001.  This is inadvertently adding 200-400kb of data to most HTML pages (student profile, school overview, homeroom).

This overrides `as_json` for now, since this will reduce load times immediately, and there's no case I can think of where we'd want this to be serialized with an educator.

Alternately, we could move this to a different table, since it's really an index or derived view of something we compute (who the educator is authorized to see), rather than a source of truth about the educator.  That would let us remove the `as_json` override.

### DIFF
--- a/app/models/educator.rb
+++ b/app/models/educator.rb
@@ -26,6 +26,14 @@ class Educator < ActiveRecord::Base
 
   VALID_GRADES = [ 'PK', 'KF', '1', '2', '3', '4', '5', '6', '7', '8' ].freeze
 
+  # override
+  # The `student_searchbar_json` field can be really heavy (~500kb), and
+  # there's no circumstances where we want to include it when serializing
+  # an educator model.  So override `as_json to omit it by default.
+  def as_json(options = {})
+    super(options.merge({ except: [:student_searchbar_json] }))
+  end
+
   def has_school_unless_districtwide
     if school.blank?
       errors.add(:school_id, "must be assigned a school") unless districtwide_access?
--- a/spec/models/educator_spec.rb
+++ b/spec/models/educator_spec.rb
@@ -2,6 +2,12 @@ require 'rails_helper'
 
 RSpec.describe Educator do
 
+  describe '#as_json' do
+    it 'does not include student_searchbar_json' do
+      educator = FactoryGirl.build(:educator)
+      expect(educator.as_json.has_key?('student_searchbar_json')).to be false
+    end
+  end
   describe '#has_school_unless_districtwide' do
     context 'no school assigned' do
       context 'has districtwide_access' do


kevinrobinson:
A school (which does have a lot of other data already for reals):
screen shot 2017-10-25 at 1 13 34 am

A student in Pre-K, who probably doesn't have much other real data:
screen shot 2017-10-25 at 1 14 03 am

Also noticed that these responses aren't gzipped, so opened https://github.com/studentinsights/studentinsights/issues/1196

alexsoble:
Site note: I looked for where we're serializing `current_educator` in the school overview. The code looks like this: 

```
    @serialized_data = {
      students: merged_student_hashes,
      school: @school,
      current_educator: current_educator,
      constant_indexes: constant_indexes
    }
```

That gets rendered in `serialized_data.html.erb`: 

```
<%= json_div(id: "serialized-data", data: @serialized_data) %>
```

`json_div` creates a `content` tag... which then encodes to JSON under the hood:

> "Data attribute values are encoded to JSON, with the exception of strings, symbols and BigDecimals. This may come in handy when using jQuery's HTML5-aware .data() from 1.4.3."
> https://apidock.com/rails/ActionView/Helpers/TagHelper/content_tag

So we never explicitly call `as_json` on `current_educator` anywhere. 

Not sure if this means we should do anything differently — let's ship this! — but wanted to document my own process trying to trace how `as_json` would have an impact here since we're not calling `as_json` directly in our code. 

I guess if Rails ever changes how `content_tag` encodes JSON under the hood this change would revert? But that's hypothetical, not a blocker for this PR.

kevinrobinson:
@alexsoble Yeah `as_json` is the method Rails calls to serialize things, although I think there are newer things like `Serializers` now that you can use to cut that out of the model altogether.  But `as_json` is still what Rails controller or view code falls back to.

The way we write data into the page is not great, since the escaping means each of those quotes in JSON take up more four bytes as `&quot;` or whatever.  We should definitely improve that (we looked at this a while back when we moved this to a JSON endpoint, and that's probably the next point at which we could pull this, since we need a different authorization method for those requests).  But either way it'll still be `#as_json` getting called if we do something like `render json: { educator: current_educator }` in the controller.

alexsoble:
@kevinrobinson Makes sense. I'm good to 🚢  if you are! :D

kevinrobinson:
🚢 👍

alexsoble:
📐📏📐📏📐
Deploying and measuring...
📐📏📐📏📐

alexsoble:
@kevinrobinson 

Before (prod, Healey School overview):
screen shot 2017-10-25 at 10 45 05 am

After (prod, Healey School overview):
screen shot 2017-10-25 at 10 51 25 am

Looks like a 25% reduction in the size of the HTML file! 👍 📉 🎉  

Amazing work @kevinrobinson, thank you for spotting this! 👍

PR #1199, opened 2017-10-25
Add script for rebuilding database on demo site

kevinrobinson:
I almost dropped the real database twice doing this (thanks to Rails and Heroku for the guardrails), so this adds a script with no args to pass so you can't pass the wrong app, etc.


alexsoble:
👍 👍

PR #1200, opened 2017-10-25
WIP - Finish up dropping discontinue services after #1165 merged and after migrating data succeeded.

Jngai:
This is to be merged right after #1165 . #1165 needs the model and schema to run rake task so we remove the model and drop the table after #1165. I branch this off from #1165 branch so the only commit to be merged is actually 86840ec since all the rest of the commits will be in master after the prev one get merge in.


alexsoble:
Hey @Jngai! Hmmm, this PR has a title which signal that merging and deploying it would be very destructive. Which is true! 

Why don't we add a "WIP" in front to indicate that it's still in progress and not finalized yet? And maybe something like "...after migrating data" to the end of the PR title so that folks know when and why we'd be running such a destructive migration?

Jngai:
@alexsoble all set.

Jngai:
@alexsoble please see #1206.

Issue #1194, opened 2017-10-25
Columns aren't sorting on admin permissions page


kevinrobinson:
Headers look like they should sort, but don't work.

<img width="1216" alt="screen shot 2017-10-25 at 12 10 16 am" src="https://user-images.githubusercontent.com/1056957/31980240-1b1f9dda-b919-11e7-8283-a1ebf1b99291.png">

The URL updates with `sort_order` but `direction` doesn't.  Updating it manually in the URL doesn't work either.

alexsoble:
Ah, looks like what happened is that I overwrote the `index` action of the administrate controller, then administrate's API introduced a breaking change that I missed.

I see a solution with lots of upvotes here: https://github.com/thoughtbot/administrate/issues/442 so I'll try introducing that.

alexsoble:
Looks like there's a second problem here which is that administrate doesn't support sorting by `belongs_to` relationships out-of-the-box.

alexsoble:
Sorting is now fixed for every column except the Homeroom column, so I'm going to close this issue and open a more specific one about Homeroom.

PR #1202, opened 2017-10-25
Fix most sorting on educator permissions dashboard

alexsoble:
# Notes 

+ Unbreak sorting for string/numeric/boolean columns by removing outdated override
+ Re-set default sort order using better code path
+ Write custom code to make sorting by school work

# GIF (demo data) 

sort-sort

# Issue 

+ #1194

# What's left? 

+ Sorting is still broken on the Homeroom column. That's because the `administrate` project doesn't include an out-of-the-box way to enable sorting for a `HasOne` relationship. I wasn't able to figure out how to solve this in a custom way during the time I had to work on this today. I'll take another look tomorrow. If I don't fix it we could ship this as-is since it unbreaks most columns and tackle Homeroom sorting as a separate piece of work. Any suggestions or ideas for Homeroom are welcome!

### DIFF
--- a/app/fields/school_name_field.rb
+++ b/app/fields/school_name_field.rb
@@ -2,6 +2,10 @@ require "administrate/field/base"
 
 class SchoolNameField < Administrate::Field::Base
   def to_s
-    data.try(:name) || data.try(:local_id)
+    return 'None' unless data
+
+    school = School.find(data)
+
+    school.try(:name) || school.try(:local_id)
   end
 end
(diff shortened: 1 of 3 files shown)


app/fields/school_name_field.rb
```diff
@@ -2,6 +2,10 @@
 
 class SchoolNameField < Administrate::Field::Base
   def to_s
-    data.try(:name) || data.try(:local_id)
+    return 'None' unless data
```
kevinrobinson: :)  Calling `object.to_s` and having it make database queries is pretty surprising, but this is limited to this one admin sorting thing, and if this kind of thing isn't supported in administrate, shipping a workaround and moving on seems great. 👍

alexsoble: Tried a less surprising way of doing this but it didn't fix sorting. I'm going to go with the "ship ugly/weird but working" approach for now to fix as much as possible as quickly as possible.

kevinrobinson: Yeah sounds great! 👍

Issue #1208, opened 2017-10-26
Make house display for only SHS, not any high school.


jhilde:
Since we're adding Full Circle (the alternative high school in Somerville), we now have two high school that each structure themselves in unique ways.

Somerville High School uses house to sub-group students, Full Circle does not.
Both schools (and likely all high schools) have the concept of counselor. Currently, elementary and middle schools in Somerville do not have this concept.

Logic will be:
Show House: local_id = 'SHS'
Show Counselor: school_type = 'HS'

jhilde:
This is currently implemented in the `patch/full-circle-no-houses` branch. The starting point is @kevinrobinson's awesome work in #1182. For some reason, the current browserify setup isn't so happy with the export. 

Will wait till #1182 is in master then create pull.

jhilde:
🚢 with #1212

Issue #1201, opened 2017-10-25
Activate Capuano and Full Circle Schools


snoopyuri:
I think we are now ready to bring Capuano (K/PreK) and Full Circle (alternative school) on board.  

The benefit would be that if Capuano SST started using it, we would have details about them when they went to Grade 1 in the 7 schools the following year.  With Full Circle on board, they could see all the notes on kids before they got referred there.

What kind of lift is that?

alexsoble:
@snoopyuri Fairly straightforward! Before we do that I'd like to run a quick scan of the Capuano/Full Circle associated educators and send them to you to make sure the default-assigned permissions look right. We'll also need to see if the data import process runs smoothly for those schools or if they are different in some important way that might trigger a data import error. But to change the actual import process is just one line of code. 😄

alexsoble:
## progress 

+ [ ] check default permissions 
+ [ ] check data import
+ [ ] turn on in prod

jhilde:
Taking a look at how Full Circle will work since it is a high school but operates differently from Somerville High School.

Looks like courses, sections, educator / student section assignments are coming through in the feeds. (I'll send @snoopyuri an email with educators missing lasids).

Educators will be directed to the section page (like SHS educators are). They will see the grade column on both section roster and student profile but it will be blank. (I don't think Full Circle uses the same assessment paradigm). 

The one issue that I'll fix today is that house should not be shown, since they are not organized in that way. I will leave the counselor capability since each student is assigned a counselor.

snoopyuri:
That makes sense….they don’t really use the gradebook, but they do have kids who move back and forth between the high school depending on how they’re doing so that information would work.

PR #1212, opened 2017-10-27
Enhance customization of house and counselor for SHS and Full Circle.

jhilde:
This PR implements the change for #1208 making #1201 better for the end user.

When I added the house and counselor fields for Somerville High School, we were a bit wary of adding too many district-specific and school-specific attributes. That still stands, but for the moment we're being very cautious about it.

I added house and counselor conditionally for all schools that are high schools. At the time, the only one was SHS. As we add Full Circle (Somerville's alternative high school), we need to better support the fields it needs. Namely, Full Circle does not have houses but does have counselors.

Rather than replicating the logic in each place where that data is exposed, I decided to create a simple helper in `app/assets/javascripts/helpers/customization_helper.js` There's currently one function `shouldDisplay` that takes an `element` and a `school`.  

It's pretty conservative in that it returns false for elements it doesn't know about, for a null school, or for missing attributes on school.

This is a very small step toward encapsulating this customization logic. As we discover more about future districts, we will handle this differently. But for now this is a first step.


app/assets/javascripts/helpers/customization_helpers.js
```diff
@@ -0,0 +1,17 @@
+function shouldDisplay(element, schoolContext) {
```
alexsoble: Oooooh I like this approach! If I'm reading it right this is similar to the the "centralize decision-making methods into one place" pattern that @kevinrobinson has talked about for authorization.

app/assets/javascripts/helpers/customization_helpers.js
```diff
@@ -0,0 +1,17 @@
...
+      return "SHS" === schoolContext.local_id;
+    }
+    return false;
+  case 'counselor': 
+    if (schoolContext && schoolContext.school_type) {
+      return "HS" === schoolContext.school_type;
+    }
+    return false;
+  }
+  return false;
```
alexsoble: What is this second layer of `return false` doing? Is it kind of like a default?

jhilde: Yup, forgot about that default thing. :-)

app/assets/javascripts/helpers/customization_helpers.js
jhilde: Yeah, depending on customization needs, we likely need a more robust solution that lives in a data store, not code. But for the moment this gets us a step in a good direction.

kevinrobinson: 👍 👞 👠 <--- those are steps :)

app/assets/javascripts/helpers/customization_helpers.js
```diff
@@ -0,0 +1,18 @@
+function shouldDisplay(element, schoolContext) {
+  switch(element) {
+  case 'house':
+    if (schoolContext && schoolContext.local_id) {
+      return "SHS" === schoolContext.local_id;
+    }
+    return false;
+  case 'counselor': 
+    if (schoolContext && schoolContext.school_type) {
```
kevinrobinson: Is `schoolContext` the same as `school`?

alexsoble:
Hey there @jhilde! #1204 is ready to merge/ship. On our call last week it would be best to merge this PR in first. Looks like @kevinrobinson has a comment open. Let's prioritize merging this in over refactoring work so that we can unblock turning on Capuano and Full Circle.

jhilde:
Changed the name in response to @kevinrobinson's comment. Made sense in my mind at the time, not sure why. Keep it simple!

Will deploy @alexsoble once build passes.

jhilde:
🚢 
✔️ ✔️  on ie11

alexsoble:
👍 👍  Thanks @jhilde!

Issue #1192, opened 2017-10-25
Fix or remove STAR charts


kevinrobinson:
The changes to the overview page adding counselor and HS house filters in https://github.com/studentinsights/studentinsights/pull/1158 introduced changes that broke the STAR chart pages (specifically, requiring the school to be passed and expecting properties on it https://github.com/studentinsights/studentinsights/commit/bfbf9983d86db8af8b13edc67a76299e16cb61fc#diff-66cf517f6e4bb7cc45a680baba80ee1dR276).  (cc @jhilde, it's a good idea to grep and find all the call sites for a component when adding new required props)

Monitoring shows that no one tried to use these pages, and we caught this before it affected anyone.   Let's either fix this and add a smoke test for these charts, which is straightforward and would have caught the bug in https://github.com/studentinsights/studentinsights/pull/1158.  Or, since we haven't really been supporting these charts as first-class features, we could just remove them altogether if they're weight we have to carry that no educators benefit from.  The code is always in source control if we want to add them back.

@alexsoble @snoopyuri Any thoughts on this?

snoopyuri:
Since there isn’t any usage, maybe we should pull out remove them for now and bring them back on another iteration (maybe with the dashboard where we’ll be training administrators to use this deeper level of data analysis)

jhilde:
Ughhh...sorry about that. You'd think there would be an easy way to validate all component usage against the prop-types in eslint but I don't think there is.

Grepping it is.

kevinrobinson:
@jhilde hehe sounds like you want a type system :)  Yeah no worries, it's a totally normal mistake and it's my fault there's not even any smoke tests to catch this.  Grepping won't be so bad here since the codebase isn't that big.  I don't know of linter rules to catch this (the linter would have to have global knowlege of all modules since you can't tell the required props without looking in the file) but one-line smoke tests catch changes like this this easily.  We could do a pass to add those in for every component now that tests are faster, which I'd be up for helping with.  Alternately, if you were excited about static typing you could experiment with https://flow.org/  In my limited experience, it's been more trouble then it's worth (maybe that's changed as it's matured), but its "weak mode" adds some level of type checking without any changes from developers.

For this issue, sounds like the consensus here is to remove, so let's do that then!

jhilde:
@kevinrobinson this will be corrected with the #1212. I intentionally made the school is null simply return false (i.e. don't display this element).

[...]

PR #1216, opened 2017-10-27
Experimenting with Jest's snapshot feature

jhilde:
The bug I introduced described in #1192 has been annoying me. Essentially, I added a mandatory prop to a component. I didn't realize it was being used by another page, so I didn't change it. Tests were clear because that page isn't well tested. 

As Kevin pointed out, perhaps Flow or Typescript would alleviate my craving of some sort of type safety that's compiler enforced. _We won't be using Flow or Typescript anytime soon._

So I started to think about how we can ensure that each of the React components are tested. 

I'm thinking using Jest snapshots for very basic testing of React components could go a long way. Essentially, Jest takes a snapshot the first time the tests are run. Those snapshots are committed alongside jest test js.

So I wanted to test a couple of components. Two with existing tests `SectionPage` and `FlexibleRoster` and one without `SectionHeader`.

Would love your thoughts @kevinrobinson and @alexsoble. Note I had to downgrade react-test-renderer to the version of react we're using. Doesn't look like that was used anywhere else.

### DIFF
--- a/spec/javascripts/section/section_page_spec.jsx
+++ b/spec/javascripts/section/section_page_spec.jsx
@@ -1,4 +1,5 @@
 import SpecSugar from '../support/spec_sugar.jsx';
+import renderer from 'react-test-renderer';
 
 describe('SectionPage', function() {
   const ReactDOM = window.ReactDOM;
@@ -79,5 +80,29 @@ describe('SectionPage', function() {
       const firstDataRows = dataElements.eq(0).find('td');
       expect(firstDataRows[0].innerHTML).toEqual('<a href="/students/2">Duck, Donald</a>');
     });
+  });
+
+  it('renders correctly', () => {
+    const props = {
+      students: [
+        { id: 1, first_name: 'Minnie', last_name: 'Mouse', program_assigned: 'Regular Ed', disability:'', plan_504: '', limited_english_proficiency: 'FLEP', home_language: 'Spanish', free_reduced_lunch: 'Free Lunch', most_recent_school_year_absences_count: 3, most_recent_school_year_tardies_count: 5, most_recent_school_year_discipline_incidents_count: 0},
+        { id: 2, first_name: 'Donald', last_name: 'Duck', program_assigned: 'Special Ed', disability:'Motor', plan_504: '504 Plan', limited_english_proficiency: 'ELL', home_language: 'English', free_reduced_lunch: 'Reduced Lunch', most_recent_school_year_absences_count: 1, most_recent_school_year_tardies_count: 0, most_recent_school_year_discipline_incidents_count: 0},
+      ],
+      educators: [
+        { }
+      ],
+      sections: [
+        { id: 1, section_number: 'Art-1'},
+        { id: 2, section_number: 'Art-2'},
+        { id: 3, section_number: 'Art-3'}
+      ],
+      section: { id: 1, section_number: 'Art-1', course_number: 'Art', course_description: 'Awesome Art Class', term_local_id: '9', schedule: '3(M-R)', room_number: '304' },
+      currentEducator: { districtwide_access: false }
+    };
+    
+    const tree = renderer.create(
+      <SectionPage {...props} />
+    ).toJSON();
+    expect(tree).toMatchSnapshot();
   });
 });
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "eslint-plugin-react": "^6.1.2",
     "jest": "^21.2.1",
     "react-addons-test-utils": "15.4.2",
-    "react-test-renderer": "^16.0.0",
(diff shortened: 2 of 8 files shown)


kevinrobinson:
@jhilde This seems great to explore!  👍 

I'm not sure how this improves the case you ran into though, as compared to just adding smoke tests like [this](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/src/App.test.js#L5).  Missing props is more about just having a test at all rather than depth of coverage, so adding one-line tests seem the best immediate bang for the buck.

One other downside for this might be fatigue around snapshots changing on every change, or that we're not actually good at catching bugs by reviewing snapshot files.  That being said, I'd vote to just try it out and see how we all like it.

spec/javascripts/section/section_page_spec.jsx
```diff
@@ -80,4 +81,28 @@ describe('SectionPage', function() {
       expect(firstDataRows[0].innerHTML).toEqual('<a href="/students/2">Duck, Donald</a>');
     });
   });
+
+  it('renders correctly', () => {
+    const props = {
```
kevinrobinson: One pattern I like in test files is:

```
function testProps(props) {
  return {
    ...props,
     foo: 'default'
  };
}
...
const props = testProps({ bar: 'cfdsya' });
````

Comment (kevinrobinson) [reply spec/javascripts/section/section_page_spec.jsx]: (as a way to modify default fixture values so you don't have to duplicate them). 
 There's also a `fixtures` folder or file for common ones (like a default `currentEducator`).

**Comment by kevinrobinson** [comment]
Also, to clarify https://github.com/studentinsights/studentinsights/issues/1192#issuecomment-339755769 I think it'd be totally reasonable to experiment with weak mode Flow and see how we all find it (similar to what you're proposing here).

**Comment by jhilde** [comment]
@kevinrobinson Agree that a simple test can do the job. The challenge is more in creating props (providing fixtures for common things will work well). 

In looking at our tests, the majority are asking, does this look right. Some count the number of columns, some search for text, etc. It seems like automated snapshots would actually do most of that testing, with less code, and more completely. Agreed, that we update a core React Component, several snapshots could change. But with auto recreation of snapshots, we can then look at the diffs to make sure the output changed in the expected way.

Mostly it was just a, I have a few minutes, what should I play with. And also excited we're on more modern tooling now. ;-)

**Comment by kevinrobinson** [comment]
👍 For adding smoke tests, here are some ways fixtures are used in other places in the project: [local to the test](https://github.com/studentinsights/studentinsights/blob/adedc3bc701fdf926d45c3962e5ea44505463da5/spec/javascripts/student_profile/services_list_spec.jsx#L23), [shared across a product feature](https://github.com/studentinsights/studentinsights/blob/master/spec/javascripts/student_profile/fixtures.jsx) and [shared across the app](https://github.com/studentinsights/studentinsights/tree/master/spec/javascripts/fixtures).  If you need to make new ones that are complex, grabbing data from the console in development mode is a good way to re-use the Ruby demo data generators.

For snapshots, let's just try it!  It'll definitely be more complete in some sense, but it relies on us humans to look at those snapshots and diffs to know if the feature is working correctly. :)  In this PR here, does the roster choose the initial sort column correctly?  For me, it's harder to see that in the snapshot than in:

```
expect($(el).find('th:first').hasClass('sort-down')).toEqual(true);
```

and not much more pain to write.  For other components for things like visualizations, where TDD makes no sense and you want to verify you don't regress visual layout or styling, snapshots seem like the perfect test API.  I'm optimistic that snapshots would be helpful enough to be worth trying out!

But if we want to improve catching errors for missing props, we should just do the plain boring thing and add some smoke tests to components that don't have them :)

kevinrobinson:
Moving over from email.  I'd vote to experiment and try using snapshots like you're doing here, I think I've said that in all of my comments :)

> It seems like automated snapshots would actually do most of that testing, with less code, and more completely

I agree with this!  Certainly if we added snapshots for all components, those tests would be quite complete and fail anytime the product changes.  I'm not sure that this helps our users, or even us. :)

The cost that snapshots introduces is higher false-positive test failures, since snapshot tests break on any product change.  And that means test fatigue and friction for every developer on the team for every product improvement they make.  In a weird way, that approach would encourage us to make refactoring improvements that were invisible to users, but discourage us from making product improvements.  I've certainly worked on systems where tests functioned this way.  If you have other perspectives you can share that'd be helpful for me in understanding!

To re-iterate, I'd vote to experiment and try using snapshots like you're doing here, and to experiment with them on new things we build.  This seems great!  And I'd be excited about pull requests that added smoke tests to all components proactively.  But I'd vote against spending time to add snapshot tests to every component pro-actively.

jhilde:
I'm going to close this without merging...and perhaps do a bit more with it. This was originally meant as an experiment.

Issue #1196, opened 2017-10-25
gzip HTML responses


kevinrobinson:
We aren't doing this now for the HTML views that Rails renders (or for JSON like the names endpoint).  We should so that we send less bytes and pages load faster (assuming the CPU impact is low).  Looks like this might be as simple as dropping in a gem: https://github.com/romanbsd/heroku-deflater

kevinrobinson:
This is an example of what Web Inspector's performance audit tool suggests might be the gains from doing this - an 83% reduction in bytes:

<img width="696" alt="screen shot 2017-10-25 at 1 22 50 am" src="https://user-images.githubusercontent.com/1056957/31981732-18410acc-b923-11e7-8878-592771629a8a.png">

kevinrobinson:
Adding `config.middleware.use Rack::Deflater` works on JSON responses, but doesn't seem to reduce the bytes transferred for HTML responses (in fact increases it by a tiny amount).  The `Rack::Deflater` middleware does work on JSON responses, and the names endpoint shrinks to ~67% of its size.

To sanity check that gzip would reduce the file size for this kind of data, I ran gzip on the command line for that HTML response and it compresses it to ~10% of the size, and with a production overview HTML file it cuts it to 9% (!) of the size.  Rails has `ActiveSupport::Gzip`, which also gets similarly great compression when run on that plain HTML file in a Rails console.  

So it seems like there must be something particular about the middleware and HTML responses.  I added logs into Rack::Deflater and confirmed that the gzip path [here](https://github.com/rack/rack/blob/master/lib/rack/deflater.rb#L64) executes, and can verify the response headers indicate gzipping on the HTML response (even though the size isn't any smaller).  
  I can also see that the middleware ultimately calls Zlib::GzipWriter, the same class that ActiveSupport::Gzip calls.  The arguments are slightly different, but changing `Rack::Deflater` to match the `ActiveSupport::Gzip` default arguments still doesn't reduce the number of bytes.

So I'm stuck here, feels like I'm missing something but can't see what.  Other options altogether could be: reworking the deployment setup to put nginx in front of Rails (against the grain for Heroku deploys, risky) or splitting out the data from HTML responses and requesting them as JSON endpoints (gets compression at the cost of a second request, probably a good idea longer-term anyway).

kevinrobinson:
[...]

PR #1221, opened 2017-10-29
Enable gzip for JSON, move school overview to JSON endpoint

kevinrobinson:
This enables a Rails middleware to gzip responses for JSON endpoints and update the school overview page to request JSON.  See https://github.com/studentinsights/studentinsights/issues/1196 for more background.  This PR aims to work around not being able to get compression in HTML responses, and also to avoid making large changes to drive routing from JS.

edit: I updated this to leave the current #show endpoint as is, and to ship another #overview endpoint so we can more safely compare directly in production.

Here are some numbers.  Since this is dependent on environment because of the actual data payload sizes and interaction with other loading behaviors like fetching scripts.  It also may be different across browsers, and this testing is only with Chrome.  We won't be able to exactly measure the actual impact on load time until it's deployed.  So here's some data from development and the demo site, and then estimates about impact in production.

## names endpoint
In production, for developer user (which is an outlier in terms of the size of this response), there's no compression (although this is cached in the browser):
screen shot 2017-10-29 at 9 07 30 am

In the demo site, for the demo user, compression cuts the size to 67% of what it was.
screen shot 2017-10-29 at 9 07 50 am

If we get a similar compression rate in production, that'd reduce the size to 190kb.

## School overview
The main tradeoff here is by moving the JSON out of the HTML response, we get better compression and reduce the network transfer size, but we add another network request to get the data.

### development
Here's local development, the HTML is 61kb and takes ~1000ms to finish loading everything.
screen shot 2017-10-29 at 9 00 51 am

With this PR, it becomes:
screen shot 2017-10-29 at 9 20 19 am

And so we're sending 48% of the bytes we used to send.

### demo
On the demo site, this looks like:
screen shot 2017-10-29 at 11 30 48 am

And the page takes roughly ~1100ms to "finish" the requests with a warm cache, just based on reloading the page a few times.

After on the demo site:
screen shot 2017-10-29 at 11 24 25 am

So that reduces the overall bytes transferred to 25% of what it was previously, with the tradeoff that it's now split across two serial requests.  Time to "finish" loading is ~900ms, similarly roughly estimated.

### production
In production, this is the current size of the HTML page:
screen shot 2017-10-29 at 9 17 49 am

If we get a similar compression rate, this would reduce the total bytes transferred from 1.1MB to ~275kb.  With the larger data payloads, the compression should be more meaningful (for this page in particular) than having to make the second request.  That may not be true for other pages.

### DIFF
--- a/app/assets/javascripts/school_overview/main.jsx
+++ b/app/assets/javascripts/school_overview/main.jsx
@@ -4,14 +4,17 @@ export default function renderSchoolOverviewMain(el) {
   const SchoolOverviewPage = window.shared.SchoolOverviewPage;
   const Filters = window.shared.Filters;
 
-  const serializedData = $('#serialized-data').data();
-  MixpanelUtils.registerUser(serializedData.currentEducator);
-  MixpanelUtils.track('PAGE_VISIT', { page_key: 'SCHOOL_OVERVIEW_DASHBOARD' });
-
-  window.ReactDOM.render(<SchoolOverviewPage
-    allStudents={serializedData.students}
-    school={serializedData.school}
-    serviceTypesIndex={serializedData.constantIndexes.service_types_index}
-    eventNoteTypesIndex={serializedData.constantIndexes.event_note_types_index}
-    initialFilters={Filters.parseFiltersHash(window.location.hash)} />, el);
+  const {schoolSlug} = $('#serialized-data').data();
+  fetch(`/schools/${schoolSlug}/overview.json`, { credentials: 'include' })
+    .then(r => r.json())
+    .then(serializedData => {
+      MixpanelUtils.registerUser(serializedData.currentEducator);
+      MixpanelUtils.track('PAGE_VISIT', { page_key: 'SCHOOL_OVERVIEW_DASHBOARD' });
+      window.ReactDOM.render(<SchoolOverviewPage
+        allStudents={serializedData.students}
+        school={serializedData.school}
+        serviceTypesIndex={serializedData.constant_indexes.service_types_index}
+        eventNoteTypesIndex={serializedData.constant_indexes.event_note_types_index}
+        initialFilters={Filters.parseFiltersHash(window.location.hash)} />, el);
+    });
 }
--- a/app/assets/javascripts/school_overview/main.jsx
+++ b/app/assets/javascripts/school_overview/main.jsx
@@ -1,20 +1,34 @@
 import MixpanelUtils from '../helpers/mixpanel_utils.jsx';
 
-export default function renderSchoolOverviewMain(el) {
+// Load data from inline on the page or with another request
+export default function renderSchoolOverviewMain(el, options = {}) {
+  if (options.json) {
+    const {schoolSlug} = $('#serialized-data').data();
+    fetch(`/schools/${schoolSlug}/overview_json`, { credentials: 'include' })
+      .then(r => r.json())
+      .then(json => render(el, json));
+  } else {
+    const serializedData = $('#serialized-data').data();
+    const {students, school} = serializedData; // undo outer camelcase
+    render(el, {
+      students,
+      school,
+      current_educator: serializedData.currentEducator,
+      constant_indexes: serializedData.constantIndexes
+    });
+  }
+}
+
+function render(el, json) {
   const SchoolOverviewPage = window.shared.SchoolOverviewPage;
   const Filters = window.shared.Filters;
 
-  const {schoolSlug} = $('#serialized-data').data();
-  fetch(`/schools/${schoolSlug}/overview.json`, { credentials: 'include' })
-    .then(r => r.json())
-    .then(serializedData => {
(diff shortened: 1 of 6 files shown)


kevinrobinson:
If other folks have free cycles and are able to help with this (or tackle https://github.com/studentinsights/studentinsights/issues/1196 more directly), that'd be great!  I probably can't look more in depth in the next week or so, and this is my attempt at the simple short improvement but it still needs verification that it improves the user experience and load times in production.

If not I can pick it up again and finish it off down the line.

kevinrobinson:
I pushed some updated commits so that we can ship this without changing the user experience.  This PR would turn on the gzip middleware, and add the new endpoint separately (where no users would use it).  And it would filter the compression so it only applies to `application/json` mime types.  This will avoid doubly-compressing assets which are already compressed on disk, and avoid running the deflater middleware for other responses (see https://github.com/studentinsights/studentinsights/issues/1196#issuecomment-340269968).  I verified this works as expected locally, and also confirmed that the deflater middleware runs but also doesn't actually reduce the byte size for CSV and PDF responses (I don't know why).

So as is, we could deploy this pull request and expect to see an overall reduction in latency in the names endpoint, where we're trading more CPU usage to gzip for reduced response size in bytes.  That's a good test case to let us see what this looks like.

At the same time, this deploys a new `/schools/:id/overview` endpoint which will let us compare the time it takes to paint the overview data between that and the existing `/schools/:id/show` endpoint.  I'm optimistic that this will meaningfully reduce that time.

kevinrobinson:
@alexsoble mention bot already tagged you, but can you review when you get a chance?  This is similar to the splitting we did before, but here it's driven by wanting to take advantage of gzipping the responses.  (Even better would be gzipping the HTML responses but I can't figure that out.)

alexsoble:
Looking at this now @kevinrobinson! 👍 😎

alexsoble:
Hey @kevinrobinson! Overall, fantastic PR — I had no idea we could optimize by compressing like this.

I really like the "shipping dark" pattern with the `overview` endpoint here. Curious to see how that changes the user's experience in production — I think I remember that the last time we experimented with using JSON endpoints we saw a flash of white screen before the page loaded, but maybe the compression will make that less noticeable. 

On the CPU usage tradeoff, do you have an expect on what the consequences of that will mean? For example, do you think we'll see more memory errors than in the past? Will it just take the server more time to make the JSON (but then the client side will take way less time because the JSON is compressed)? Or is it hard to tell at this point?

I can 🚢  this this afternoon and verify the deploy if you think this is in good shape to merge!

alexsoble:
Did some manual QA, I did notice a brief visual "jump" on the (currently non-end-user-discoverable) `overview` endpoint before the page loaded the JSON. 

I don't think it's a big regression from the user experience in development, curious to see what it looks like in prod with real data. Might benefit from a spinning loader placeholder in the future and then some kind of empty-state skeleton UI for while it waits for the JSON to load.

kevinrobinson:
@alexsoble re: CPU tradeoff, it takes some CPU work to gzip the response, but I assume it'll be negligible as compared to the time it takes to ship those extra bytes over the network.  I also assume the way `Rack::Deflater` does this might not be super optimized (normally nginx or Apache would do this kind of compression, but Heroku's standard Rails building doesn't use a web server and that'd be a bigger change).

re: user experience, yeah the frame loads first and then the data comes in.  We could make a loading spinner, or change that in other ways, but I think for now let's just see if this makes the time-to-seeing-data faster and go from there?  If it's 50% faster then a loading indicator might not be that bad, but if it's only 5% faster then it might actually feel slower with the extra loading step.

kevinrobinson:
And if you have time to 🚢 and investigate that's awesome, go for it! 🌮 👍

alexsoble:
@kevinrobinson I prioritized reviewing Gabi's, Judy's, and Eric's feature PRs first and have to run to a doctor's appointment, so I didn't get to 🚢  and investigate this today, sorry.

alexsoble:
@kevinrobinson Looking this over I realized it adds a "profile" resource to the students routing but never defines profile in `students_controller.rb` or does anything with that route. Added a commit that removes the route before I merge.

kevinrobinson:
@alexsoble oops, probably leftover from an earlier iteration.  thanks!

alexsoble:
@kevinrobinson No problem! Also I created merge conflicts by merging in the STAR-removal PR first . . . my bad! I can rebase merge and deploy if you're ok with that?

alexsoble:
@kevinrobinson I ran the rebase, didn't want to rewrite your commits on your branch so I opened up a new rebased PR #1228 and QA'ed it on the live demo site.

kevinrobinson:
@alexsoble Got it I'll close this then?  Please re-open if I misunderstood!

FWIW you can also open a PR into a branch, which the merges changes (like resolving conflicts from other things that have since been merged into master) into an existing PR before it's merged to master.

PR #1224, opened 2017-10-30
Refactor components folder to ES6 imports, add shims for some helpers

kevinrobinson:
Progress towards https://github.com/studentinsights/studentinsights/issues/1180, also converting each React component to use class syntax and adding in smoke tests and class-level comments for all components touched.  This includes a few helper files as well, with shims to put those on `window.shared` so legacy components can still read from there until they are converted.  This is an almost entirely mechanical change, even though git isn't matching the diffs.  The main substantive changes is adding constructors and binding callback methods (a change for React class syntax, not ES6 modules).

This also renames app and test files to look like:
screen shot 2017-10-30 at 9 09 49 am
screen shot 2017-10-30 at 9 09 45 am

The intention here is to move towards "we write React the same as as create-react-app" over time, and doing a component-by-component migration is maybe a good time to migrate those conventions.


ui/legacyHelpers.js
```diff
@@ -0,0 +1,21 @@
...
+import '../app/assets/javascripts/helpers/sort_helpers.jsx';
+import * as Routes from '../app/assets/javascripts/helpers/Routes';
+import {colors, styles} from '../app/assets/javascripts/helpers/Theme';
+
+// Export these so legacy components can use this, and
+// new components can pull in components
+window.shared || (window.shared = {});
+window.shared.Routes = Routes;
+window.shared.colors = colors;
+window.shared.styles =styles;
```
kevinrobinson: This adds the shim for helpers that are now ES6 modules, but where legacy components still expect to find them on `window.shared`.  This has to be in a separate file from `legacy.js`, since import statements are all hoisted and so these assignments run last regardless of where they are in the file.

ui/legacy.js
```diff
@@ -1,34 +1,10 @@
...
-
-// shared react compoents:
-import '../app/assets/javascripts/components/fixed_table.jsx';
-import '../app/assets/javascripts/components/collapsable_table.jsx';
-// (unordered)
-import '../app/assets/javascripts/components/flexible_roster.jsx';
-import '../app/assets/javascripts/components/slice_buttons.jsx';
-import '../app/assets/javascripts/components/slice_panels.jsx';
-import '../app/assets/javascripts/components/star_charts_page.jsx';
-import '../app/assets/javascripts/components/students_table.jsx';
```
kevinrobinson: This is the main progress - these components are all removed from here since they're now imported.

alexsoble:
Hey @kevinrobinson, very cool! I noticed your last few commits are about a git case problem, is this something that will require some shared git configuration across devs? Or just figuring out some weirdness with how git handles file case?

kevinrobinson:
@alexsoble This should be fixed up now.  The first test failure was from git not recognizing a change in filename case, the second was me assuming it was that but it was a case typo in the test.  Neither of those failed for me locally, so must be some difference between how Jest/Babel/something handles on Mac vs. Linux in Travis.

kevinrobinson:
hmm travis got stuck, restarted manually.  Looks like they have been having issues all day: https://twitter.com/traviscistatus/status/925106931970203648

kevinrobinson:
@alexsoble I rebased this, want to take a look and see if this seems good to merge?  cc @Jngai

alexsoble:
@kevinrobinson I pulled down the branch and QA'ed against an Internet Explorer in a VM, looked good! Some of the mechanical changes (removing trailing commas after functions, fat arrow syntax) look similar to that time I tried to move us over to `eslint/airbnb` but clearly you're not touching `production.rb` here so this is way safer.

alexsoble:
@kevinrobinson So yeah I can 🚢!

alexsoble:
@kevinrobinson Prod looks good!

kevinrobinson:
@alexsoble awesome, thanks for the help! 👍

Issue #1207, opened 2017-10-26
Risk Levels Need Re-Calculating with New MCAS


snoopyuri:
Just got a heads up from a teacher that the Risk Levels might be off because of the new MCAS levels, since we're not using Proficient, Needs Improvement, and Warning anymore.  And I think that is the same for the tool-tips that Paul Mitchell created that still refer to those categories.

jhilde:
Started to look at this a bit @snoopyuri. Doesn't look like the risk levels were changed to the new levels. Happy to take a look unless @alexsoble wants to continue since you've been doing the Next Gen MCAS work.

alexsoble:
Go for it @jhilde! 👍

alexsoble:
@jhilde is going to dive into the current state of risk levels for students with "Next Gen" scores + think about ways we can start to extract some of this as patterns!

jhilde:
🚢 with #1225

PR #1225, opened 2017-10-31
Add Next Gen MCAS to risk level calculation

jhilde:
Addresses #1207. 

With the switchover to Next Gen MCAS scores, @alexsoble added a new assessment type. At that time, the risk calculation wasn't addressed. So a student's Next Gen MCAS scores weren't taken into account for the risk calculation.

I've made it so the risk calculation looks for the most recent Next Gen MCAS score. If the student has no Next Gen MCAS score for that topic, the calculation looks for the most recent MCAS (legacy) score. If the student has neither, a `MissingStudentAssessment` is returned. My assumption is that a student wouldn't have a legacy MCAS test taken after the Next Gen MCAS. (Is this valid @snoopyuri? Otherwise, I could compare the dates). I also assumed that the risk level for Next Gen MCAS would be similar to the legacy MCAS. (Specifically NME=>3, PE=>2, ME=>1, EE=>0). Any thoughts @snoopyuri ?

I've updated the reasons for the risk level to use both legacy and next gen MCAS performance levels. This minimized that additional work of tracking whether the MCAS score was legacy or next gen.

I've also flattened the logic a bit to remove the `McasRiskLevel` and `StarRiskLevel` and centralized it in the `Assessment` class. Mostly because I found it really confusing. 

At some point, we'll need to figure out what assessments we'll support and how we calculate risk from them in generic or district-specific ways.


snoopyuri:
Just read through it.  Actually, makes a lot of sense and will make it easier to customize in the future for us and other districts.

Meeting tomorrow with the high school so should get a better sense of what their risk level calculation will look like.

jhilde:
@alexsoble or @kevinrobinson Want to sanity check me on this?

app/risk_levels/mcas_risk_level.rb
```diff
@@ -1,18 +0,0 @@
-class McasRiskLevel < Struct.new :student_assessment
```
alexsoble: I'm learning that `delegate` is a bad smell so I think it's great that you're getting rid of these mini-classes.

app/models/student_risk_level.rb
```diff
@@ -128,16 +138,16 @@ def risk_factor_names_to_explanations
         3 => " is limited English proficient"
       },
       mcas_math_risk_level: {
-        3 => "'s MCAS Math performance level is Warning",
-        2 => "'s MCAS Math performance level is Needs Improvement",
-        1 => "'s MCAS Math performance level is Proficient",
-        0 => "'s MCAS Math performance level is Advanced"
+        3 => "'s MCAS Math performance level is Warning / Not Meeting Expectations",
```
alexsoble: This confused me when I first read it. It sounded like there was a risk level named "Warning / Not Meeting Expectations." But really we mean the student's performance level was either "Warning" or "Not Meeting Expectations." Using the word "or" instead of a slash would be clearer here.

app/models/student_risk_level.rb
```diff
@@ -6,6 +6,11 @@ class StudentRiskLevel < ActiveRecord::Base
 
   # Use most recent assessments to calculate risk
   def mcas_math
+    # First look for the newer MCAS assessment
+    next_gen = student.latest_result_by_family_and_subject("Next Gen MCAS", "Mathematics")
+    return next_gen if next_gen
+
+    # If there are none, look for the original MCAS
```
alexsoble: Another way to do this which might be cleaner: 

```
# student.rb

151: latest_result_by_family_and_subject(["Next Gen MCAS", "MCAS"], "Mathematics")
```

app/risk_levels/mcas_risk_level.rb
jhilde: Yeah, not a big fan of the Struct.new shortcut. Figured it was just the Rails way. 😄

app/models/student_risk_level.rb
jhilde: Ah yes, forgot that it wouldn't care if you passed a single value or an array. Good call.

app/models/student_risk_level.rb
jhilde: Agreed

spec/models/student_risk_level_spec.rb
```diff
@@ -139,7 +139,24 @@
...
+    context 'has Next Gen MCAS and MCAS, not STAR' do
+      context 'has Next Gen MCAS math and MCAS math' do
+        context 'has a W value for MCAS math and an EE for Next Gen MCAS Math' do
+          let(:student) { FactoryGirl.create(:student) }
+          let(:next_gen_mcas_math_ee) {FactoryGirl.create(:next_gen_mcas_math_exceeds_expectations_assessment, student: student)}
+          let(:mcas_math_w) {FactoryGirl.create(:mcas_math_warning_assessment, student: student)}
+          let!(:student_risk_level) { StudentRiskLevel.create!(student: student) }
+          it 'has a correct explanation' do
+            expect(student_risk_level.explanation).to eq({
+              intro: "This student is at Risk N/A because:",
```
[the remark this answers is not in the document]
jhilde: When I switch to using the `latest_result_by_family_and_subject(["Next Gen MCAS", "MCAS"], "Mathematics")` that will be the case. I had assumed that Next Gen would always be newer than legacy but your suggestion below will remove that assumption. 

I'll add dates to this test case to make it more deterministic.

jhilde:
Thanks @alexsoble. I'll likely deploy this and the importer delete fix on Thursday when I'll have more time to support...just in case.

jhilde:
And you would be correct @alexsoble. 😧 Now I remember thinking, huh, why does 0 give N/A. Skipped over it and convinced myself it was right. Fortunately, code was working correctly, just bad test case. 

Passes because I left out the `!` on the lets for `next_gen_mcas_math_ee` and `mcas_math_w`. 

Fixed the test in the branch. Should I merge into master? Deploy to prod (though this is just a test case).

Issue #1205, opened 2017-10-26
Fix 10/26 monitoring fail


alexsoble:
This morning educators were blocked from accessing Insights. Somerville's physical LDAP server was down because of a power outage. Educators were not able to log in for around 3 hours because the offline LDAP server could not authenticate them.

One aspect of this issue is that the dev team didn't get clear or urgent enough monitoring messages about this failure. I got alert emails from Logentries — did you @jhilde @kevinrobinson? — but I get frequent emails from Logentries and many of them are about less urgent issues, so I didn't realize these were critical. 

In particular it would be good to have alerts distinguish absolutely critical failures (educator cannot login) from bad but less critical failures (a slow JSON endpoint times out).

This issues is open ended right now and could go in a couple of directions. I'd like to surface a couple of possibilities for how to address this: 

## 1. Improve Logentries settings

This is something @kevinrobinson and I have looked at before. Logentries has an error tagging system, and we dug into it before in order to turn off alerts for some of the Memory Errors that made Heroku unhappy but weren't blocking user experience and weren't immediately addressable: 

![screen shot 2017-10-26 at 11 29 58 am](https://user-images.githubusercontent.com/3209501/32065016-29770214-ba41-11e7-8a2d-e226b6735c29.png)

A couple of limitations to this strategy. "Request timeout" can mean very different things if it's a login request versus a JSON endpoint request — that can be the difference between "educator can't login" and "educator's experience on a specific page is disrupted." So that's a very blunt category. 

Another issue is that Logentries doesn't have some of the bug-fix-collaboration features that Rollbar does, like including all collaborators on the same email and allowing collaborators to mark bugs as fixed.

## 2. Move away from Logentries and toward Rollbar for backend error monitoring

Currently we use Rollbar for client-side error monitoring and reporting and Logentries for server-side monitoring and reporting. Rollbar has some good bug-fix-collaboration features that I mentioned above, like including all devs on error reporting emails instead of messaging devs separately. 

Here's the Rollbar ruby gem: https://github.com/rollbar/rollbar-gem. It was last updated 2 months ago. It's CI build on the README is listed as "cancelled", which is unfortunate. 

## 3. Use a different backend error monitoring service

If we're wary of leaning all the way into Rollbar for backend error monitoring and folks have had great experiences with other error monitoring systems, we could go with a totally different option. I've had good experiences using Airbrake for backend (and front-end) error monitoring. 

##  4. Write more custom error monitoring emails (and/or texts)

[...]

PR #1226, opened 2017-11-01
Use Rollbar for server-side error tracking

alexsoble:
# Notes 

Talked online with @kevinrobinson and @jhilde, consensus is that setting up Rollbar for server-side error tracking is a good first step. This PR adds the code to set that up; I added the required ROLLBAR_ACCESS_TOKEN config to our production Heroku app already. 

Let's see how this does in the wild for a couple of weeks or months. My prediction is that it will be a significant improvement over Logentries. The next step in my mind would be some kind of SMS alerts to the team for critical errors like LDAP login errors.
 
# Issue 

+ #1205

### DIFF
--- a/config/initializers/rollbar.rb
+++ b/config/initializers/rollbar.rb
@@ -1,0 +1,57 @@
+Rollbar.configure do |config|
+  # Without configuration, Rollbar is enabled in all environments.
+  # To disable in specific environments, set config.enabled=false.
+
+  config.access_token = ENV['ROLLBAR_ACCESS_TOKEN']
+
+  # Here we'll disable in 'test':
+  if Rails.env.test?
+    config.enabled = false
+  end
+
+  # By default, Rollbar will try to call the `current_user` controller method
+  # to fetch the logged-in user object, and then call that object's `id`,
+  # `username`, and `email` methods to fetch those properties. To customize:
+  # config.person_method = "my_current_user"
+  # config.person_id_method = "my_id"
+  # config.person_username_method = "my_username"
+  # config.person_email_method = "my_email"
+
+  # If you want to attach custom data to all exception and message reports,
+  # provide a lambda like the following. It should return a hash.
+  # config.custom_data_method = lambda { {:some_key => "some_value" } }
+
+  # Add exception class names to the exception_level_filters hash to
+  # change the level that exception is reported at. Note that if an exception
+  # has already been reported and logged the level will need to be changed
+  # via the rollbar interface.
+  # Valid levels: 'critical', 'error', 'warning', 'info', 'debug', 'ignore'
+  # 'ignore' will cause the exception to not be reported at all.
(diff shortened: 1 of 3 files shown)


alexsoble:
Want to review @jhilde / @kevinrobinson or should this be a selfie?

config/initializers/rollbar.rb
```diff
@@ -0,0 +1,57 @@
...
+  # config.use_sidekiq
+  # You can supply custom Sidekiq options:
+  # config.use_sidekiq 'queue' => 'default'
+
+  # If you run your staging application instance in production environment then
+  # you'll want to override the environment reported by `Rails.env` with an
+  # environment variable like this: `ROLLBAR_ENV=staging`. This is a recommended
+  # setup for Heroku. See:
+  # https://devcenter.heroku.com/articles/deploying-to-a-custom-rails-environment
+  config.environment = ENV['ROLLBAR_ENV'] || Rails.env
```
kevinrobinson: Is it worth adding the DEPLOYMENT_KEY or whatever else we do for UI analytics?

config/initializers/rollbar.rb
```diff
@@ -0,0 +1,57 @@
+Rollbar.configure do |config|
+  # Without configuration, Rollbar is enabled in all environments.
+  # To disable in specific environments, set config.enabled=false.
+
+  config.access_token = ENV['ROLLBAR_ACCESS_TOKEN']
+
+  # Here we'll disable in 'test':
+  if Rails.env.test?
+    config.enabled = false
+  end
```
kevinrobinson: Should this be disabled in development mode too?

config/initializers/rollbar.rb
```diff
@@ -0,0 +1,57 @@
+Rollbar.configure do |config|
+  # Without configuration, Rollbar is enabled in all environments.
+  # To disable in specific environments, set config.enabled=false.
+
+  config.access_token = ENV['ROLLBAR_ACCESS_TOKEN']
```
kevinrobinson: Will this come out in a different bucket of errors in the Rollbar UI, or same as JS?  I suspect it'll be obvious which one it's coming from just never tried it before! :)

kevinrobinson:
@alexsoble Yeah I'd say 🚢 and just verify after or make some intentional errors and see how it works (notification levels, etc. too).

config/initializers/rollbar.rb
alexsoble: Yes great catch!

config/initializers/rollbar.rb
alexsoble: Yeah we'll see!

config/initializers/rollbar.rb
alexsoble: Added using their `custom_data_method` thing.

Issue #999, opened 2017-09-01
Look at ways to automate scanning JS codebase for Internet Explorer issues


alexsoble:
# Notes 

+ Internet Explorer is our main targeted browser so this is important

alexsoble:
Here's a thing: https://github.com/amilajack/eslint-plugin-compat

alexsoble:
Gave `eslint-plugin-compat` a whirl, unfortunately it wasn't able to detect the `Object.assign` issue that caused a bug in #995.

alexsoble:
Next step: rig up Teaspoon to BrowserStack Automate: https://github.com/jejacks0n/teaspoon/wiki/Using-BrowserStack-WebDriver

alexsoble:
Another good resource: https://docs.travis-ci.com/user/browserstack/

alexsoble:
Also: https://github.com/browserstack/capybara-browserstack

kevinrobinson:
https://github.com/babel/babel-preset-env looks like it might help.

FWIW I looked at babel-polyfill, which would add a layer of safety for some of these problems with "I expected the runtime to support this and tested in Chrome and it worked there."  I read through the issues in create-react-app, where they don't support that (concerns about file size).  For us in https://github.com/studentinsights/studentinsights/pull/1182 it brings the bundle size from 310kb to 348kb, so a bit more than a 10% increase.  That seems like a lot for a hypothetical layer of safety, and when production monitoring will just catch this immediately anyway.

kevinrobinson:
We could also add it and define it as a separate Webpack bundle that's basically cached forever.

PR #1228, opened 2017-11-02
[rebased] Enable gzip for JSON, move school overview to JSON endpoint

alexsoble:
# Notes 

This PR takes the fantastic optimization work from @kevinrobinson's #1221 (cuts `names` endpoint JSON to 67% of what it was using demo data; adds new compressed-JSON-powered `overview` endpoint for School Overview) and rebases it on top of master to remove conflicts with #1223.

I deployed this rebased PR to the demo site to do some manual QA and it looked good.


alexsoble:
Oooooh thanks for the reminder @studentinsights-bot; I need to double check on Internet Explorer since this touches `production.rb`.

alexsoble:
Tested this on an IE VM. The `overview` pages wouldn't render in IE and threw and error, `fetch is not defined`.

kevinrobinson:
@alexsoble thanks and good catch!

I think it'd make sense to add the Promise and fetch polyfills to match: https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#supported-language-features-and-polyfills (the overall intent being "on this project you write React code like you do in `create-react-app`).

We could also look at catching this in a more automated way, or of just polyfilling all these issues away as in https://github.com/studentinsights/studentinsights/issues/999#issuecomment-339202761, but it has a size cost.

kevinrobinson:
Pushed polyfills for this and tested they work in IE11 locally.  This will increase the bundle size - I think it's negligible but didn't check.

More importantly, it's not clear whether gzip encoding works as expected - it should be the IE devtools don't give enough information to confirm:
screen shot 2017-11-02 at 10 37 07 pm
screen shot 2017-11-02 at 10 36 57 pm

We can investigate more, but since this is a dark endpoint seems worth just shipping and investigating whether we even want this at all by roughly measuring time-to-first-data in IE in production first.

alexsoble:
Yup let's 🚢! Looks like a couple more conflicts have crept in, I'll go ahead and fix those up.

alexsoble:
Rebasing again! :D

alexsoble:
Arrrrgh. @kevinrobinson. I totally messed up the rebase process and overrwrote your excellent polyfill commits. Afraid I've done more harm then help on this PR. When you have the chance, could you `git push feature/gzip-rebased -f` to restore your useful polyfilling commits? Thanks and sorry!

alexsoble:
I fixed my own problem here and re-added the polyfills. Force-pushed to demo and verified that both the vanilla school overview page and the newer JSON-powered page work in Internet Explorer 11 ✅ .

alexsoble:
This is live in production. Scanned the app in Internet Explorer. The old-school school overview loads fine. The `overview` JSON-powered page loads significantly slower.

kevinrobinson:
@alexsoble Thanks!  Yeah, it'll be helpful to talk about what is slower.  This is a sample from what I see in Chrome just refreshing a bunch of times with the cache disabled.  There's a bunch of variability but this is a good example:

**/schools/:id**
screen shot 2017-11-09 at 4 04 07 pm

**schools/:id/overview**
screen shot 2017-11-09 at 4 04 17 pm

So forget that we have two make requests in serial - here the response time for the 1.2MB HTML response is 2.83s while the response time for the 91k JSON response is 3.44s.  There's so much variability and noise in response times it's hard to tell which of those two would have a better distribution of response times.  I'm surprised it didn't just make a big different, but I guess that's why we measure things :)

We could try simulating more responses and see if the JSON endpoint has better response distributions.  Or we could look at what's making the JSON endpoint take so long (as a guess it just might take time to do the gzipping).  But without a clear win I'd say let's just revert this.  @alexsoble seem good?

Issue #1229, opened 2017-11-02
Sections missing for most high school students


jhilde:
Educator texted @snoopyuri today indicating high school students were missing sections. 

Upon investigation, I realized that when the `StudentSectionAssignmentImporter` and `EducatorSectionAssignmentImporter` are run with a school scope, all assignments for other schools are deleted as being stale.

@alexsoble ran the imports to bring in the data for Capuano which inadvertently caused the assignment data for all other schools to be deleted.

As a stopgap, @alexsoble ran the full import to restore the assignment data. Thank you!

alexsoble:
This is healed so I'm going to close the issue.

jhilde:
🚢 the permanent change with #1230.

PR #1230, opened 2017-11-02
Fix delete on Assignment Importers to properly handle school scope.

jhilde:
In response to #1229. Assignment importers were deleting all assignments not imported in the current run as stale. This should only be done for the schools being imported.

For instance, if the school scope doesn't include SHS, the `educator_section_assignments` and `student_section_assignments` for SHS shouldn't be deleted.

### DIFF
--- a/app/importers/file_importers/educator_section_assignments_importer.rb
+++ b/app/importers/file_importers/educator_section_assignments_importer.rb
@@ -32,7 +32,10 @@ class EducatorSectionAssignmentsImporter < Struct.new :school_scope, :client, :l
 
   def delete_rows
     #Delete all stale rows no longer included in the import
-    EducatorSectionAssignment.where.not(id: @imported_assignments).delete_all
+    #For the schools imported during this run of the importer
+    EducatorSectionAssignment.joins(:section => {:course => :school})
+                             .where.not(id: @imported_assignments)
+                             .where(:schools => {:local_id => school_scope}).delete_all
   end
 
   def import_row(row)
--- a/spec/importers/file_importers/educator_section_assignments_importer.rb
+++ b/spec/importers/file_importers/educator_section_assignments_importer.rb
@@ -133,5 +133,21 @@ RSpec.describe EducatorSectionAssignmentsImporter do
           expect(EducatorSectionAssignment.count).to eq(1)
         end
       end
+
+      context 'delete only stale assignments from schools being imported' do
+        before do
+          FactoryGirl.create_list(:educator_section_assignment,20)
+          FactoryGirl.create(:educator_section_assignment, educator_id: educator.id, section_id: section.id)
+
+          esa_importer = described_class.new
+          esa_importer.school_scope = ['SHS']
+          esa_importer.import_row(row)
+          esa_importer.delete_rows
+        end
+
+        it 'deletes all student section assignments for this school except the recently imported one' do
+          expect(EducatorSectionAssignment.count).to eq(21)
+        end
+      end
     end
   end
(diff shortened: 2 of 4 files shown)


spec/importers/file_importers/educator_section_assignments_importer.rb
```diff
@@ -133,5 +133,21 @@
...
+          FactoryGirl.create(:educator_section_assignment, educator_id: educator.id, section_id: section.id)
+
+          esa_importer = described_class.new
+          esa_importer.school_scope = ['SHS']
+          esa_importer.import_row(row)
+          esa_importer.delete_rows
+        end
+
+        it 'deletes all student section assignments for this school except the recently imported one' do
+          expect(EducatorSectionAssignment.count).to eq(21)
```
kevinrobinson: I'm confused - this looks to me like it creates 20 records, then a 21st, and then it imports a row and deletes in the test.  I can't tell what to expect to happen - is the idea that the record on line 140 is stale, or can you help me understand?

app/importers/file_importers/educator_section_assignments_importer.rb
```diff
@@ -32,7 +32,10 @@ def filter
 
   def delete_rows
     #Delete all stale rows no longer included in the import
-    EducatorSectionAssignment.where.not(id: @imported_assignments).delete_all
+    #For the schools imported during this run of the importer
+    EducatorSectionAssignment.joins(:section => {:course => :school})
+                             .where.not(id: @imported_assignments)
+                             .where(:schools => {:local_id => school_scope}).delete_all
```
kevinrobinson: If I read right, this is expecting caps for `local_id` (versus lowercase for the slug).  Any downside to upcasing here to catch mistakes like `heroku run:detached thor import:start --school shs` etc.?

kevinrobinson:
nice fix! 👍

app/importers/file_importers/educator_section_assignments_importer.rb
jhilde: If we did upcase, I think it would make sense in the import.thor since a lowercase school would affect all the other importers as well.

kevinrobinson: 👍 yeah good idea, seems like a more logical spot to put it as well.  The name is `school_scope` in importers so it's ambiguous whether that's supposed to be a slug or local_id and you have to trace the execution to tell (no need to change that now).

spec/importers/file_importers/educator_section_assignments_importer.rb
jhilde: Yeah, it's a little confusing the way I added it in for school specific. The original test, doesn't test school scope at all, so it defaults to all schools. Create 20 random assignments, create 1 specific one. The one specific one is imported. All the others are stale since they weren't imported.

The new test, strikingly similar but with a school scope added won't touch the 20 random assignments because they are for a random school and the importer's `school_scope` is set to SHS.

kevinrobinson: Are the 20 assignments literally random with what school they refer to?  That seems like it would lead to non-deterministic test runs if one happened to be in SHS.

To clarify, if it's simple the generator could make 20 assignments with the school to be FC so it's clear in the test that those records shouldn't be deleted, or the expectation could check that "there's 20 non-SHS records" and "there's only 1 SHS records since it was updated."

jhilde: Agreed, will make more explicit. Actually, the school created by the factory would just have null for the local_id. I'll make it explicitly create them in another school.

alexsoble:
Yeah, good use of `school_scope` @jhilde!

alexsoble:
@jhilde 👍 👍

PR #1231, opened 2017-11-03
Revert "WIP - School Absence Dashboard #1082"

alexsoble:
Reverts studentinsights/studentinsights#1098.

I wasn't able to deploy this to the demo environment, got the following error:

```
remote: -----> Node.js app detected
remote: 
remote: -----> Build failed
remote:  !     Two different lockfiles found: package-lock.json and yarn.lock
remote: 
remote:        Both npm and yarn have created lockfiles for this application,
remote:        but only one can be used to install dependencies. Installing
remote:        dependencies using the wrong package manager can result in missing
remote:        packages or subtle bugs in production.
remote: 
remote:        - To use npm to install your application's dependencies please delete
remote:          the yarn.lock file.
remote: 
remote:          $ git rm yarn.lock
remote: 
remote:        - To use yarn to install your application's dependences please delete
remote:          the package-lock.json file.
remote: 
remote:          $ git rm package-lock.json
remote:     
remote:        https://kb.heroku.com/why-is-my-node-js-build-failing-because-of-conflicting-lock-files
remote: 
remote:  !     Push rejected, failed to compile Node.js app.
remote: 
remote:  !     Push failed
remote: Verifying deploy...
remote: 
remote: !	Push rejected to somerville-teacher-tool-demo.
remote: 
```

Reverting so that we can keep the app deployable.


alexsoble:
Hey @edavidsonsawyer — see the error message above. Heroku wouldn't accept my deploy of the PR because it had both a package-lock.json and yarn.lock. One of these comes from yarn and the other comes from NPM, and Heroku needs to use either one or the other to know what JS dependencies to use. We're using yarn in the project so we should get rid of package-lock.json so that we can deploy the your work. 

Could you open up a new PR without package-lock.json? Could you also make sure all the project dependencies work correctly when installed and managed by yarn only? 

We're super close!

edavidsonsawyer:
My bad, @alexsoble. Fix is as easy as #1234

PR #1232, opened 2017-11-03
Add CI script to detect package-lock.json

alexsoble:
# Notes 

+ If `package-lock.json` is present it blocks deploys to Heroku. Heroku won't accept a project with both a `package-lock.json` and a `yarn.lock` file present.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,3 +25,4 @@ script:
   - rubocop
   - bundle exec rake db:create db:migrate DATABASE_URL=postgres://localhost/student_insights_test
   - bundle exec rspec spec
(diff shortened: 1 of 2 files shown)


alexsoble:
@edavidsonsawyer @kevinrobinson @jhilde Any thoughts on this? This is basically some CI that enforces yarn versus npm for the JS dependency lock because if both are present Heroku won't accept deploys. Basically CI to make sure reviewers don't forget to check this, especially when new contributors join the project and start writing JS.

edavidsonsawyer:
@alexsoble checklists save time! I like it.

PR #1234, opened 2017-11-03
remove npm dependencies

edavidsonsawyer:
This is the administrator absence dashboard without the conflicting lock files. Should make this addition to #1082 deployable.


alexsoble:
Hey @edavidsonsawyer! I think the best way to get your changes merged in would be to revert my revert and then add your new commits on top of it: 

https://github.com/studentinsights/studentinsights/pull/1231

Otherwise you're not including all the great dashboard code in your PR, just the npm related changes.

You might not have the necessary permissions to revert my revert, so why don't I start things off by doing that and you can add your commits from there?

alexsoble:
@edavidsonsawyer We want to work off of this PR or a PR that looks like it: https://github.com/studentinsights/studentinsights/pull/1235

PR #1237, opened 2017-11-04
(For discussion) - Move authorization code to Authorizer, add authorized_only API

kevinrobinson:
This is for discussion with https://github.com/studentinsights/studentinsights/issues/1080.  This has two commits.

The first commit adds an Authorizer class, with a minimal implementation of what was called the "blocks-based API" in https://github.com/studentinsights/studentinsights/issues/1080.  The comments and tests there are consistent, and sketch the idea and usage.

The second commit is aimed at moving authorization code into the Authorizer class.  It doesn't update the call sites, and so Educator methods become shims into the Authorizer method calls for now.  This reveals in the Authorizer methods how there's different implementations of the authorization rules (even just from gathering together the methods in Educator, without looking at controller code yet).

The intention here is to give folks something concrete to look at and give feedback on in regards to https://github.com/studentinsights/studentinsights/issues/1080.  If this seems like a good direction, the next step might be "How would this work for the students_controller?" or "How would this affect performance?"  Those might be two additional commits, and then a natural point to decide about merging a chunk of work, and listing out what else would be needed to finish this off altogether.

### DIFF
--- a/app/lib/authorized_only.rb
+++ b/app/lib/authorized_only.rb
@@ -1,0 +1,58 @@
+# Dispatches different questions about authorization
+# to specific methods for specific models.
+class AuthorizedOnly
+  def initialize(authorizer)
+    @authorizer = authorizer
+  end
+
+  # Branches by type (Relation, Array, Model)
+  # and by the Model class.
+  def dispatch(&block)
+    return_value = block.call
+
+    # Can't duck type here, since other things like hashes
+    # are iterable or mappable in ways we don't want to
+    # support.
+    if return_value.class < ActiveRecord::Relation # subclass?
+      filter_array(return_value.to_a)
+    elsif return_value.class == Array # subclass?
+      filter_relation(return_value)
+    else
+      filter_model(return_value)
+    end
+  end
+
+  private
+  # This is a seam to optimize for authorization filtering
+  # that stays lazy when possible (only possible when
+  # authorization can be applied with simple `where`
+  # clauses and not by filtering based on a Ruby method.
+  def filter_relation(relation)
+    filter_array(relation)
+  end
+
+  def filter_array(array)
+    array.map {|model| filter_model(model) }.compact
+  end
+
+  def filter_model(model)
+    if model.class == Student
+      check_value(model, @authorizer.is_authorized_for_student?(model))
+    else
+      unchecked_value(model)
+    end
+  end
+
+  # Just a hook for logging, noop
+  def check_value(model, should_allow)
+    if should_allow then model else nil end
+  end
+
+  # This can log, raise or otherwise notifier developers.
+  # If it does return, it should return the "null" value for something
+  # the educator doesn't have access to.
+  def unchecked_value(model)
+    message = "AuthorizedOnly#unchecked_value for class: #{model.class.to_s}"
+    raise Exceptions::EducatorNotAuthorized.new(message)
+  end
+end
--- a/app/lib/authorized_only.rb
+++ b/app/lib/authorized_only.rb
@@ -13,10 +13,10 @@ class AuthorizedOnly
     # Can't duck type here, since other things like hashes
     # are iterable or mappable in ways we don't want to
     # support.
-    if return_value.class < ActiveRecord::Relation # subclass?
-      filter_array(return_value.to_a)
-    elsif return_value.class == Array # subclass?
+    if return_value.class < ActiveRecord::Relation # if subclass
       filter_relation(return_value)
+    elsif return_value.class == Array
+      filter_array(return_value)
     else
       filter_model(return_value)
     end
@@ -28,7 +28,7 @@ class AuthorizedOnly
   # authorization can be applied with simple `where`
   # clauses and not by filtering based on a Ruby method.
   def filter_relation(relation)
-    filter_array(relation)
+    filter_array(relation.to_a)
   end
 
   def filter_array(array)
--- a/app/lib/authorizer.rb
+++ b/app/lib/authorizer.rb
@@ -4,23 +4,23 @@
 # Drop down to calling individual methods if you need to.
 #
 # This class should be fully tested, and all changes should be reviewed.
-# 
+#
 # Controller usage:
 #   student = authorized_only { Student.find(params[:id]) }
 #   homeroom = authorized_only { Homeroom.(params[:id]) }
-#   students = authorized_only { homeroom.students } 
+#   students = authorized_only { homeroom.students }
 #
 # Standalone usage:
 #   authorizer = Authorizer.new(educator)
 #   student = authorizer.authorized_only { Student.find(params[:id]) }
 #   homeroom = authorizer.authorized_only { Homeroom.(params[:id]) }
-#   students = authorizer.authorized_only { homeroom.students } 
+#   students = authorizer.authorized_only { homeroom.students }
 #
 # Instead of:
 #   Homeroom.find(id).students # no authorization check
 #   Section.find(id).students # no authorization check
-#   current_educator.homeroom.students # doesn't consider other roles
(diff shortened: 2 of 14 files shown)


app/models/educator.rb
```diff
@@ -128,39 +70,17 @@ def default_section
...
-      Section.all
-    elsif schoolwide_access?
-      Section.joins(:course).where('courses.school_id = ?', school.id)
-    else
-      sections
-    end
-  end
-
-  def allowed_homerooms_by_name
-    allowed_homerooms.order(:name)
```
kevinrobinson: Removing this.

app/models/educator.rb
```diff
@@ -17,13 +17,14 @@ class Educator < ActiveRecord::Base
 
   validates :email, presence: true, uniqueness: true
 
-  validate :has_school_unless_districtwide,
-           :admin_gets_access_to_all_students,
-           :grade_level_access_is_array_of_strings,
-           :grade_level_strings_are_valid,
-           :grade_level_strings_are_uniq,
-           :grade_level_access_is_not_nil
```
kevinrobinson: These look like they're all for validation (some read like authorization), so I moved them to private and prefixed them all with `validate`.

app/lib/authorizer.rb
```diff
@@ -0,0 +1,113 @@
...
+    return true if @educator.districtwide_access?
+
+    return false if @educator.school.present? && @educator.school != section.course.school
+
+    return true if @educator.schoolwide_access? || @educator.admin?
+    return true if section.in?(@educator.sections)
+    false
+  end
+
+  # TODO(kr) remove implementation
```
kevinrobinson: These intention for these methods is that we'd be able to remove them, and defined them in terms of the other methods above.  That for this one in particular we'd say:

```
students = authorized_only { School.find(params[:id]).students }
```

in the controller.

spec/lib/authorizer_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  it 'test context is setup correctly' do
+    expect(School.all.size).to eq 12
+    expect(Homeroom.all.size).to eq 2
+    expect(Student.unscoped.all.size).to eq 2
+    expect(Educator.all.size).to eq 3
+  end
+
+  describe 'Student' do
```
kevinrobinson: This shows intended usage of `authorized_only`.

spec/lib/authorizer_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    end
+
+    it 'limits access for arrays' do
+      expect(authorized_only(healey_teacher) { [shs_student, healey_kindergarten_student] }).to eq [healey_kindergarten_student]
+    end
+  end
+
+  describe 'other models' do
+    it 'raises on calls with unchecked models' do
+      expect { authorized_only(districtwide_admin) { Educator.all } }.to raise_error(Exceptions::EducatorNotAuthorized)
```
kevinrobinson: A catch for if someone passes a model or query that isn't supported by the authorization layer.

alexsoble:
Hey @kevinrobinson, this is awesome. I love the way the `Authorizer` class pulls together all the authorization-logic related methods like `is_authorized_for_student?`, `is_authorized_for_school?`, `is_authorized_for_section?`. 

It feels useful, centralized, and very object-oriented. One specific use case: we should add `authorizer.rb` to the notification bot, or whatever tool we may use to look for PRs that touch auth and need careful review. And now new contributors should never have to write new auth code from scratch! We'll want to make "how to use the authorization API" an important part of onboarding. 

I'll add a few inline code comments where I have questions. Overall I am super supportive of and excited about this pull request! I think this is one of our highest priorities right now.

I was thinking about how folks could best support this PR, maybe one way is by helping test/QA the performance changes this may introduce so we can make sure it doesn't introduce any large regressions.

app/models/educator.rb
alexsoble: Yes, these are all for validation. Question here: does prefixing with `validation` automatically run these as validations? Is this some ancient Rails magic you're invoking here? 🔮

app/lib/authorized_only.rb
```diff
@@ -0,0 +1,58 @@
...
+  # and by the Model class.
+  def dispatch(&block)
+    return_value = block.call
+
+    # Can't duck type here, since other things like hashes
+    # are iterable or mappable in ways we don't want to
+    # support.
+    if return_value.class < ActiveRecord::Relation # subclass?
+      filter_array(return_value.to_a)
+    elsif return_value.class == Array # subclass?
```
alexsoble: These look swapped to me. It calls `filter_relation` for array and `filter_array` for relation.

app/lib/authorizer.rb
```diff
@@ -0,0 +1,113 @@
...
+# Standalone usage:
+#   authorizer = Authorizer.new(educator)
+#   student = authorizer.authorized_only { Student.find(params[:id]) }
+#   homeroom = authorizer.authorized_only { Homeroom.(params[:id]) }
+#   students = authorizer.authorized_only { homeroom.students } 
+#
+# Instead of:
+#   Homeroom.find(id).students # no authorization check
+#   Section.find(id).students # no authorization check
+#   current_educator.homeroom.students # doesn't consider other roles
```
alexsoble: @kevinrobinson What do you mean by "doesn't consider other roles"?

app/lib/authorized_only.rb
kevinrobinson: thanks!

app/lib/authorizer.rb
kevinrobinson: In other words, for a given homeroom page, this doesn't do what you'd want for all user types (eg, district admin).

app/models/educator.rb
kevinrobinson: No I just renamed them :)

kevinrobinson:
@alexsoble I'm traveling this week so won't be able to do much more here until next week at the earliest.  In terms of a todolist I think it's:
[ ] Does this seem like a good path?
[ ] How would this work for the students_controller?
[ ] How would this affect performance?
[ ] Pause, reflect, merge a chunk of work, make the next list of todos

app/lib/authorizer.rb
alexsoble: Ooooooh right, that makes sense. For example, if `current_educator` is a district admin s/he won't be assigned a `homeroom`.

alexsoble:
@kevinrobinson Thoughts a couple of your TODO items: 

**Does this seem like a good path?** 

Yes!

**How would this work for the students_controller?**

My first reaction is that we would call something like `student = authorizer.authorized_only { Student.find(params[:id]) }`, but I think you mean something deeper than "mechanically, what code will we write to implement this in the students controller?" Could you say a little more about what you mean?

kevinrobinson:
Pulled off part of this work into: https://github.com/studentinsights/studentinsights/pull/1274.

@alexsoble re: controller code, I started doing this on the `students_controller` as a spike, you can check it out [here](https://github.com/studentinsights/studentinsights/pull/1237/files#diff-11751d644b56a1cce6d8dfc73e331b91L1).  There's some messiness that came up (event_note authorization is nested and works differently, partial models don't work with the authorization method), so I cut all that out for now in favor of making progress on https://github.com/studentinsights/studentinsights/pull/1274, but if you have any feedback let me know!

kevinrobinson:
There is additional spike code on this branch https://github.com/studentinsights/studentinsights/compare/refactor/authorization, using the work in https://github.com/studentinsights/studentinsights/pull/1274 to rework the `students_controller`.  That ended up being too large of a change so I split out https://github.com/studentinsights/studentinsights/pull/1274, but that's there for looking at for future work in this direction.

PR #1243, opened 2017-11-06
Resolves #1240

### LINKED ISSUE #1240: Fix IE bug in School Administrator dashboard
Throws error in IE 11: 

```
Object doesn't support property or method "from"
```

edavidsonsawyer:
Issue #1240 was evidently caused [babel transforming the spread operator](https://github.com/babel/babel/issues/4922). It's unnecessary in this case so I rewrote the methods to sidestep this issue.

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/page_container.jsx
+++ b/app/assets/javascripts/school_administrator_dashboard/page_container.jsx
@@ -64,13 +64,15 @@ export default React.createClass({
     const schoolAttendance = this.attendanceBySchool();
     let monthlyAttendance = {};
     Object.keys(schoolAttendance).forEach((day) => {
+      //'date' here is the first day of the month in which 'day' occurs
       let date = moment(day).date(1).format("YYYY-MM-DD");
-      if (monthlyAttendance[date] === undefined) {
+      if (monthlyAttendance[date] === undefined) { //if there's no data for this month yet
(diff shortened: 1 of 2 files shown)


alexsoble:
Hey @edavidsonsawyer, are you using a virtual machine to test in IE and debug?

Check out https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/, we have some documentation on this in the README too. Happy to jump on a call to work on this together as well.

edavidsonsawyer:
I am now, @alexsoble . I just forgot that was the standard use case an neglected it in my first pass. Won't happen again.

alexsoble:
No problem! Clearly I forgot too!

kevinrobinson:
@edavidsonsawyer This seemed surprising to me at first, since we have a bunch of instances of the spread operator in the codebase already.  It seems like this issue is with the array spread operator in particular though, where it compiles down to use `Array.from`.

In this particular code, it seems like `x.concat(y)` is better than `x.push(...y)` anyway 👍   But also seems like an easy way for us to thing this compiles but then realize that it doesn't.  If we think this is a common enough mistake we could polyfill as well (https://github.com/studentinsights/studentinsights/issues/999#issuecomment-339202761).

edavidsonsawyer:
@kevinrobinson I sure wasn't expecting it. A polyfill was another option, but like you noticed .concat has better performance and is safer anyway so this was the way to go here.

Issue #1244, opened 2017-11-07
Fix slowness on absence dashboard


alexsoble:
@edavidsonsawyer, wanted to leave some ideas and notes here before our call this afternoon! 

The first step should be to gather baseline data on how long this page takes to load using demo data or semi-anonymized data. 

Check out the `log_timing` method in `application_controller.rb`, useful tool for gathering some baseline data on how long the queries run before we optimize them. Without baseline data there's no way to know if the optimizing worked! You can change the demo scripts to make more absences if you feel that would get you slightly more dense data to work with.

Looks like we have N+1 issues to optimize. These lines in particular look an N+1 problem to me:

```
dashboard_students = active_students.map { |student| individual_student_dashboard_data(student) }

def individual_student_dashboard_data(student)
  HashWithIndifferentAccess.new({
    first_name: student.first_name,
    last_name: student.last_name,
    homeroom: student.try(:homeroom).try(:name),
    absences: student.absences.order(occurred_at: :desc)
    }
  )
end

```

Check out the Rails Guides to solving N+1 problems by using eager loading: http://guides.rubyonrails.org/active_record_querying.html#eager-loading-associations

alexsoble:
Suggestion: maybe add 450 demo students to the fake school as you record baseline speeds to simulate the ~450 Healey students in this view.

edavidsonsawyer:
@alexsoble #1246 should help. Locally that has the query 300% faster using 500 students. Which makes sense because it's using 1/3 the queries.

edavidsonsawyer:
#1254 is the biggest gain that can easily be made on the client side, since it was just an oversight. Locally I have a bunch of changes that slightly improve performance, but they require some extensive rework of the dashboard component. Let's see how this change plays first.

edavidsonsawyer:
@alexsoble, can you give this a look when you get the chance?

alexsoble:
Heyyyy we fixed the slowness and now the page is loading pretty normally, so time to close this issue and have a mini online party @edavidsonsawyer! 🎉 🎉 🎉 🎉

Issue #1245, opened 2017-11-07
Remove yarn-error.log from source control


alexsoble:
@edavidsonsawyer, looks like your `yarn-error.log` got checked into source control. That's a local debugging file so we should keep it out of git. I added it to .gitignore already, could you remove the copy that's checked in?

alexsoble:
Fixed by @edavidsonsawyer!

PR #1246, opened 2017-11-08
Improving efficiency of db queries for students

edavidsonsawyer:
To help address #1244, this took 1/3 the time to execute the student queries with 500 fake students.


edavidsonsawyer:
Also removing that errant yarn-error log for #1245

alexsoble:
👍  let's see what kind of improvement this creates in prod!

alexsoble:
Hey @edavidsonsawyer! I pushed this to the live tool and tested it a few times with an in-browser stopwatch. There's definitely an improvement, and it does seem to be in the neighborhood of 1/3 the original time. That puts us from 60 second load times to 20 second load times, which is better but still not in the neighborhood we want to be in.

My suggestion for a next step: @kevinrobinson has been doing some excellent work compressing JSON in #1228. This strategy involves moving from an HTML-rendering path for putting data on the page to a JSON-oriented path where an HTML page loads, the JS runs, and it requests data via a JSON endpoint that can be compressed and help optimize. 

I'm going to work on merging in @kevinrobinson's #1228 (looks like it has a couple of merge conflicts) so that we can take advantage of the compression/optimization he's unblocked for us. 

If you have other suggestions or strategies we should try @edavidsonsawyer please post them! Let's keep working on this until the load time gets pushed down into the neighborhood we want!

edavidsonsawyer:
Thanks, @alexsoble! The stuff @kevinrobinson is doing looks promising. I'll add further suggestions to #1244. In the meantime, would you be able to check on the prod site and see the effect of changing the dates via the slider? I suspect the slowness isn't coming from the render but I want to make sure that's the case.

edavidsonsawyer:
Hey @alexsoble and @kevinrobinson! I'm noticing the school view (which I'd imagine has some of the same problems the dashboard does) takes advantage of a precompute job. Is there any objection to my adding some dashboard specific tasks there?

alexsoble:
@edavidsonsawyer When I move the sliders it takes about five seconds (counting out loud) for the data to update.

edavidsonsawyer:
@alexsoble thanks. There's probably some optimizations I can do in there.

alexsoble:
@edavidsonsawyer Ya, my feeling is let's make sure we've tackled the productive optimizations available to us in the existing code before adding a precompute jobs. Precompute jobs are awesome but sometimes they fail; they need to be debugged when they fail; there needs to be a fallback strategy for when they fail . . . fixing the N+1 query cut the load time down to a third, if we can find a couple of more low-hanging optimizations like that we'll be in a business. Also I'd say the slowness in the slider response indicates we could cut several seconds out in the JS alone.

edavidsonsawyer:
@alexsoble makes sense!

PR #1247, opened 2017-11-08
Change strategy for rebuilding demo site

alexsoble:
# Notes 

+ I'm having trouble getting the demo Heroku site into a clean state so that I can manually QA #1165 against it
+ This PR changes the strategy we use to rebuild the demo site from `db:schema:load` to `rake db:migrate`; based on the normal flow I use for doing this I think this will be be a more reliable path
+ Also clean up a few small changes in the rebuild script and one of the migrations


alexsoble:
Force-pushed this to the demo app and verified that it fixed the database state errors I was getting.

alexsoble:
Merging this in now as a selfie because I have to QA against demo again.

PR #1255, opened 2017-11-11
Limiting dashboard data to school year

edavidsonsawyer:
# Who is this PR for?
School Administrators

Slow loading times on the administrator dashboard, cf #1244  

# What does this PR do?
Limits the student data to what occurred within the school year.

# Checklist

+ [X] Tested latest version in an Internet Explorer virtual machine? *(If the PR adds Javascript.)*

### DIFF
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -38,7 +38,10 @@ class SchoolsController < ApplicationController
       redirect_to not_authorized_path and return #TodDo: determine whether there's a more appropriate action here
     end
 
-    dashboard_students = students_for_dashboard(@school).includes(:absences, :homeroom).map { |student| individual_student_dashboard_data(student) }
+    dashboard_students = students_for_dashboard(@school).includes(:homeroom, :absences)
+                                                        .where('absences.occurred_at > ?', '2016-08-01')
+                                                        .references(:absences)
+                                                        .map { |student| individual_student_dashboard_data(student) }
 
     @serialized_data = {
       absences: dashboard_students.to_json
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -39,7 +39,7 @@ class SchoolsController < ApplicationController
     end
 
     dashboard_students = students_for_dashboard(@school).includes(:homeroom, :absences)
-                                                        .where('absences.occurred_at >= ?', school_year_start)
+                                                        .where('absences.occurred_at >= ?', 1.year.ago)
                                                         .references(:absences)
                                                         .map { |student| individual_student_dashboard_data(student) }
 
@@ -130,11 +130,6 @@ class SchoolsController < ApplicationController
 
   def students_for_dashboard(school)
     school.students.active
-  end
-
-  def school_year_start
-    current_year = Date.today.year
-    Date.today.to_s > current_year.to_s + "-08-01" ? current_year.to_s + "-08-01" : (current_year -1).to_s + "-08-01"
   end
 
 end
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -39,7 +39,7 @@ class SchoolsController < ApplicationController
     end
 
     dashboard_students = students_for_dashboard(@school).includes(:homeroom, :absences)
-                                                        .where('absences.occurred_at >= ?', '2017-08-30')
+                                                        .where('absences.occurred_at >= ?', school_year_start)
                                                         .references(:absences)
                                                         .map { |student| individual_student_dashboard_data(student) }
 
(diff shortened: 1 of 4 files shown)


edavidsonsawyer:
@alexsoble, I'd be obliged if you'd merge this. Should help a lot.

alexsoble:
Hey @edavidsonsawyer, can I ask one question before we merge this in? 

I noticed this has a hardcoded date in it:

```
where('absences.occurred_at >= ?', '2017-08-30')
```

This seems like the kind of change that's really easy to forget to update when the next school year comes around. How do you feel about replacing the hard-coding with dynamically generated dates so that this dashboard will always pull data for the current school year?

edavidsonsawyer:
@alexsoble yeah! When we talked about this with the group it sounded like hard coded school year start dates were the way it was done since there doesn't seem to be anything we pull that gives us an official start date. Two thoughts:

1. I updated the PR so that it sets the year start date to august 1st of whatever year the school should be on. That date is totally arbitrary. It shouldn't be too late, but could potentially be too early if someone were looking at summer school data (or didn't want to see summer school data). Maybe mid August is better. 

2. If we did need the school year start set somewhere (say, by an admin), then it should probably be in a central location that other pieces of the app access when they need to know where to set the start date. That might be more complicated than necessary.

alexsoble:
Oooh, @edavidsonsawyer, looks like there might have been some miscommunication about what we meant by "hard coded."

The way we handle school years in certain parts of the app is "hard coded" in the sense that we make "hard coded" assumptions about when the start and end dates of each year (approximately) occur. 

But there are no dates in the app "hard coded" to always show a particular year like 2016, 2017, 2018, etc. Sorry for any confusion there.

To get a sense of how this is currently done in the app, check out `student_school_year_storter.rb` and in particular its `find_school_year_name` private method. I did some searching and it looks like this sorter class is only called once, by the Student Report PDF. So this is a fairly limited API and I'd wager it's more likely to leave the codebase with time than to expand. 

From what I've seen in this project, trying to bucket events into school years on the server side is not a good practice. It can be a time suck; it can slow down performance. It's usually much simpler and better to ignore the concept of the school year altogether and show data-over-time. Features like summer breaks between school years will emerge naturally from the data.

When we need to add a concept of school year, I think it's better done on the client side. If the server side just sends over data with timestamps, the client side can be flexible and slice up the data in different ways — by school year, by year, by month, etc. Done that way, changes the time axis only require changing code in 1 place (the client side) instead of 2 places (client side and server side). 

Let me know if this makes sense and if you have questions about this that you want to talk about!

app/controllers/schools_controller.rb
```diff
@@ -129,4 +132,9 @@ def students_for_dashboard(school)
     school.students.active
   end
 
+  def school_year_start
+    current_year = Date.today.year
+    Date.today.to_s > current_year.to_s + "-08-01" ? current_year.to_s + "-08-01" : (current_year -1).to_s + "-08-01"
```
alexsoble: Looks like we're comparing strings to strings here. This makes me pause because comparing numbers (or date times) as strings can work unpredictably: 

```
2.3.0 :001 > '11111' > '11'
 => true 
2.3.0 :002 > '11111' > '22'
 => false 
```

If we work with dates here, I think we should compare them as dates. We could also just remove the concept of student school year from the server side altogether and just send the last 1 year or 2 years of data to the client. Then the client can slice things up into school years if it wants to.

edavidsonsawyer:
@alexsoble gotcha. I'm used to relying on the server for administratively important dates, but it's easy enough to do it the other way around. In that case, though, it's probably best to just show the past year's data. I've updated the PR accordingly.

@snoopyuri, if you think this chart gets confusing unless the absence data is restricted to the current school year, let me know and I can change it to start the display in late August. 

Meanwhile, I've also cleaned up a number of spots here where the dashboard was using placeholder dates and centralized some of the logic to help a little with the load times. Now everything just comes from the server side events, as it ought to.

snoopyuri:
I think the default can be the current school year and then when I train principals I can show them how to then use the slider to look at trends and anticipate future problems.

edavidsonsawyer:
Alright then! @alexsoble, if this PR looks good to you I think we should merge it as is and I'll address the appropriate initial date range on the client side in a separate issue. I don't want to confuse the performance enhancements, which are more critical, with fiddling with the range.

app/controllers/schools_controller.rb
edavidsonsawyer: It's moot now that the school year logic is moved out of here, but I wanted to point out that while I do agree it's better generally to compare dates as dates, as long as they're formatted in descending order (eg. YYYY-MM-dd), it should be alright. In this case it seemed faster to construct the strings. But you're right, it's safer to just compare dates.

alexsoble:
👍

alexsoble:
@edavidsonsawyer Huge performance improvement here. Time to perceived load around 4-5 seconds.

snoopyuri:
YES!!!!!  You think I can show it to a few principals now?

edavidsonsawyer:
@snoopyuri I think that's a good idea! This is an excellent time to incorporate any feedback you or they might have

PR #1265, opened 2017-11-17
Change discontinued services migration to fail noisily

alexsoble:
# Notes 

+ Good catch by @kevinrobinson -- if any of the services in this migration fail to save, they'll fail silently and we'd never know. 
+ Switching to use the bang version of save so that the migration will let us know if any of the services fail to update. 
+ After this merges I'll re-run the migration.
+ cc @Jngai, good learning here, wish I had caught this myself!


alexsoble:
not waiting for ci since this is a one character change in an untested rake task 😆

alexsoble:
Re-running `heroku run rake services:move_discontinued_at_to_services` now...

alexsoble:
@Jngai @kevinrobinson The new, noisier migration ran without blowing up! 👍 😄 👍 😄

PR #1256, opened 2017-11-11
Enforce foreign key contraints

gabivoicu:
# Who is this PR for?

Add foreign key constraints for all models

# What problem does this PR fix?
Issues around data integrity with DiscontinuedServices being connected to deleted ServiceUploads

# What does this PR do?
Add the [immigrant](https://github.com/jenseng/immigrant) gem and add the migration it creates that will add foreign keys

# Screenshot (if adding a client-side feature)

# Checklist

+ [x] Tested latest version in an Internet Explorer virtual machine? *(If the PR adds Javascript.)*

# Questions/Concerns

+ [x] it appears I've committed a different version of bundle, should I reverse that?
+ [x] Immigrant recommends adding a rake task to our CI to ensure we continue adding keys (`rake immigrant:check_keys`). Is that something we want to do?
+ [x] [1275](https://github.com/studentinsights/studentinsights/pull/1275) needs to be merged before we can safely merge this one
- I've downloaded VirtualBox and the two Windows images and if I am correct, it is currently installing on my machine. If however it is not, can someone help me kick off the installation process or point me to some good installation instructions? Google is not being helpful.

PR #1267, opened 2017-11-17
Drop discontinued services table

alexsoble:
# What does this PR do?

+ Now that we've moved data on when services end over to the `services` table, we can remove the old table that used to store this data (`discontinued_services`).
+ Finishes up #1165.
+ Unblocks #1256.


alexsoble:
Hey, specs pass even when the table is removed! That's good. Making a little list of things to do before deploying: 

+ [ ] Remove discontinued service model 
+ [ ] Remove association on service model
+ [ ] QA locally
+ [ ] Capture a database backup

PR #1272, opened 2017-11-21
Allow user to see students in homeroom when selecting on chart

edavidsonsawyer:
# Who is this PR for?
Administrators with bonus refactoring for code legibility and performance

Users will want to see the students in a homeroom when clicking on that homeroom's attendance column

# What does this PR do?
Updates the student list to show the students in the selected homeroom

Coming soon

# Checklist

+ [X] Tested latest version in an Internet Explorer virtual machine? *(If the PR adds Javascript.)*


alexsoble:
@edavidsonsawyer Very cool, diving in now!

app/assets/javascripts/school_administrator_dashboard/main.jsx
```diff
@@ -1,9 +1,9 @@
-import PageContainer from  './page_container.jsx';
+import SchoolwideAttendance from  './schoolwide_attendance.jsx';
```
alexsoble: @edavidsonsawyer Love this naming change! Much more precise. This is great, and as they say, naming things is one of the hardest problems in programming! 🌈

app/assets/javascripts/school_administrator_dashboard/dashboard_bar_chart.jsx
```diff
@@ -47,18 +50,35 @@ export default React.createClass({
     };
   },
 
+  shouldComponentUpdate: function(nextProps) {
```
alexsoble: @edavidsonsawyer Can you say a little more about what this is doing? How is this different than the normal `shouldComponentUpdate` behavior? I tried removing it in my local machine and the dashboard chart seemed to work the same as before, unless I'm missing a detail that's changed.

app/models/student.rb
```diff
@@ -14,7 +14,11 @@ class Student < ActiveRecord::Base
   has_many :event_notes, dependent: :destroy
   has_many :services, dependent: :destroy
   has_many :tardies, dependent: :destroy
+  has_many :dashboard_tardies, -> { where('occurred_at >= ?', 3.months.ago) },
```
alexsoble: @edavidsonsawyer I'm curious about the different scopes on tardies versus absences here. Also wondering if this is something the student model needs to be concerned about, if I want to change the schoolwide absences dashboard I think I'd be surprised if I needed to change `student.rb`.

alexsoble:
@edavidsonsawyer Functionality looks great! Added a few comments and questions, give 'em a look and let me know what you think!

app/assets/javascripts/school_administrator_dashboard/dashboard_bar_chart.jsx
edavidsonsawyer: Right, I meant to add a comment here. This is to prevent all the charts from redrawing when a user clicks on the column to get the list of students by homeroom. The list is the only thing that needs to change so it shouldn't trigger a render of the charts.

app/models/student.rb
edavidsonsawyer: This deserves some discussion, yeah. The problem is, if we send all the attendance records to the dashboard, filtering that to a sensible range becomes slow, especially since the dashboard isn't restricted to active students. I had thought I could handle that in the school controller by restricting the included records, but it turns out that approach didn't work. Unless there's something I'm missing, you can't only bring in a subset of records with includes. 

I'm open to alternatives! This was the only way I come up with to get the range of dates the dashboard needs to load initially without running into performance problems. Conceptually I'm thinking of these as a student's "current" attendance data, but I'm also not wild about this approach.

edavidsonsawyer:
@alexsoble, added some responses. I am definitely open to alternative solutions to the student records problem!

app/models/student.rb
alexsoble: @edavidsonsawyer Is this helpful?

http://guides.rubyonrails.org/active_record_querying.html#specifying-conditions-on-eager-loaded-associations

Rails guides example:

```
Article.includes(:comments).where(comments: { visible: true })
```

app/assets/javascripts/school_administrator_dashboard/dashboard_bar_chart.jsx
alexsoble: Ooooh that's really interesting. The way the school overview page works, if you click one filter, **everything** on the page responds to that change. The percentages for each of the slice table categories, the mini bar charts, and the students table — they all update. I wonder if users will expect other pages like the schoolwide attendance dashboard to perform the same way? In that case, would it make sense to redraw the monthly charts when a new homeroom is selected?

app/models/student.rb
edavidsonsawyer: @alexsoble Yeah, that's what I tried initially, but it doesn't work the way I thought it did (and consequently the current dashboard is doing this incorrectly). If you look at the query that generates, it doesn't filter the included records, but applies the condition to the joined tables. So in our case it would return only those students with absences in the timeframe. That's no good because we do want the students with no absences w/in the last year, we just don't want their absences. Basically, it excludes students with perfect attendance within the timeframe.

app/assets/javascripts/school_administrator_dashboard/dashboard_bar_chart.jsx
edavidsonsawyer: That's a good point, @alexsoble. Here's why I'm inclined to think these are different cases. In the school overview the component, even though everything redraws, that feels natural because every part of the page has the potential to be affected by the filter you select. It's almost reassuring to see the app is recalculating each point to be sure it's correct with the new filter. 

Here, though, the user isn't doing anything to the data displayed by the chart. They just want more information about one of the data points it displays. If the chart redraws at that point, they might assume they accidentally changed the chart parameters. When I see a chart redraw I assume that's because the app thought the chart might need to change because of something I did. I found the redrawing to be jarring even though I know what's happening. 

So that was my rationale. Thoughts?

edavidsonsawyer: Hm. Now that I think about this a little more I think I might want to add an intermediate component since it's actually doing the work to come up with the chart data even though the chart doesn't use it. So even if we want to keep the conditional update in, hold off on this PR so I can make sure we aren't bothering to do more work than we need to.

edavidsonsawyer: Actually, scratch that. The changes didn't affect performance as much as I'd hoped so I'll see if I can handle improving the student chart speed separately.

app/models/student.rb
alexsoble: @edavidsonsawyer Ah, OK, that's some good tech learning! My only suggestion is to think about using a more specific variable name if possible. But I don't think that should be a blocker to merging in this work.

app/assets/javascripts/school_administrator_dashboard/dashboard_bar_chart.jsx
alexsoble: Interesting! User testing will definitely help us make decisions here. As far as the code, I think either let's add a code comment to this lifecycle function explaining why it's here, or implement an intermediate component. First one would let us ship faster but long run I think this `shouldComponentUpdate` override is telling us there should be another component here.

edavidsonsawyer: Alright, @alexsoble. For now I'll add the comment since I think I'd have to overwrite the method even with the other changes I was considering. I'll be interested to see how users respond. It's easy enough to change if redrawing the charts makes for a better experience!

edavidsonsawyer:
Oops! Sorry about the sordid commit history. Mixed up my branches while I was trying alternatives. Everything should be clean now.

alexsoble:
@edavidsonsawyer Did some local QA, looked great to me! Good to merge?

edavidsonsawyer:
@alexsoble let's merge it!

alexsoble:
@edavidsonsawyer Deploying now!

alexsoble:
@edavidsonsawyer Deploy looks good in IE in production!

PR #1274, opened 2017-11-22
Move authorization methods into Authorizer, simplify Educator model

kevinrobinson:
# Who is this PR for?
- Developers, mostly

- It's step towards https://github.com/studentinsights/studentinsights/issues/1080, as discussed in https://github.com/studentinsights/studentinsights/pull/1237.  It's intended to isolate and simplify the authorization rules, so that it's easier to do authorization checks in controller and model code when writing new features.

# What does this PR do?
- Adds `TestPals` class that creates a set of users, students, schools and homerooms that have valid and meaningful relationships, so that we can test authorization rules with them.  This helps since authorization rules are complex and to know if an educator has access to a student often involves traversing a bunch of relations.  The thought here is that this class won't change often, and when it does it will be an expensive change.
- Adds an `Authorizer` class that owns authorization logic.  It adds a helper into `ApplicationController` which is sugar for making this accessible within controller code.  It also uses a collaborator, `AuthorizedDispatcher`, which handles dispatching calls for the block-based API (which isn't used anywhere yet other than in tests).
- Updates the `Educator` model to use `Authorizer`, and refactors to clarify and simplify.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -48,6 +48,20 @@ class ApplicationController < ActionController::Base
     no_sections_path
   end
 
+  # Wrap all database queries with this to enforce authorization
+  def authorized(&block)
+    authorizer.authorized(&block)
+  end
+
+  # Enforce authorization and raise if no authorized models
+  def authorized_or_raise!(&block)
+    return_value = authorizer.authorized(&block)
+    if return_value == nil || return_value == []
+      raise Exceptions::EducatorNotAuthorized
+    end
+    return_value
+  end
+
   # Sugar for filters checking authorization
   def redirect_unauthorized!
     redirect_to not_authorized_path
@@ -78,5 +92,10 @@ class ApplicationController < ActionController::Base
     logger.info "log_timing:end [#{message}] #{timing_ms.round}ms"
 
     return_value
+  end
+
+  private
+  def authorizer
+    @authorizer ||= Authorizer.new(current_educator)
   end
 end
--- a/app/lib/authorizer.rb
+++ b/app/lib/authorizer.rb
@@ -1,0 +1,141 @@
+# This class defines authorization rules.
+# If you want student data, ask through this class.
+# Ideally, start with using `authorized`.
+# Drop down to calling individual methods if you need to.
+#
+# This class should be fully tested, and all changes should be reviewed.
+#
+# Controller usage:
+#   student = authorized { Student.find(params[:id]) }
+#   homeroom = authorized { Homeroom.(params[:id]) }
+#   students = authorized { homeroom.students }
+#
+# Standalone usage:
+#   authorizer = Authorizer.new(educator)
+#   student = authorizer.authorized { Student.find(params[:id]) }
+#   homeroom = authorizer.authorized { Homeroom.(params[:id]) }
+#   students = authorizer.authorized { homeroom.students }
+#
+# Instead of:
+#   Homeroom.find(id).students # no authorization check
+#   Section.find(id).students # no authorization check
+#   current_educator.homeroom.students # doesn't work for all roles
+#   current_educator.school.students # doesn't work for all roles
+#
+class Authorizer
+  # These fields are required on the `Student` model for
+  # authorization checks
+  def self.student_fields_for_authorization
+    [
+      :id,
+      :program_assigned,
+      :limited_english_proficiency,
+      :school_id,
+      :grade
+    ]
+  end
+
+  def initialize(educator)
+    @educator = educator
+    @authorized_dispatcher = AuthorizedDispatcher.new(self)
+  end
+
+  # Usage:
+  #  student = authorized { Student.find(params[:id]) }
+  def authorized(&block)
+    @authorized_dispatcher.dispatch(&block)
+  end
+
+  # This method is the source of truth for whether an educator is authorized to view information about a particular
+  # student.
+  def is_authorized_for_student?(student)
+    begin
+      return true if @educator.districtwide_access?
+
+      return false if @educator.restricted_to_sped_students && !(student.program_assigned.in? ['Sp Ed', 'SEIP'])
+      return false if @educator.restricted_to_english_language_learners && student.limited_english_proficiency == 'Fluent'
+      return false if @educator.school_id.present? && student.school_id.present? && @educator.school_id != student.school_id
+
+      return true if @educator.schoolwide_access? || @educator.admin? # Schoolwide admin
+      return true if @educator.has_access_to_grade_levels? && student.grade.in?(@educator.grade_level_access) # Grade level access
+      return true if student.in?(@educator.students) # Homeroom level access
+      return true if student.in?(@educator.section_students) # Section level access
+    rescue ActiveModel::MissingAttributeError => err
+      # We can't do authorization checks on models with `#select` that are missing
+      # fields.  If this happens, it's probably because the developer is trying to
+      # to optimize a query.  The authorization layer could only recover by making more
+      # queries, so instead we raise and force the developer to figure out how to resolve.
+      #
+      # See `Authorizer.student_fields_for_authorization`
+      raise err
+    end
+    false
+  end
+
+  def is_authorized_for_school?(school)
+    return true if @educator.districtwide_access?
+    return true if @educator.schoolwide_access? && @educator.school == school
+    return true if @educator.has_access_to_grade_levels? && @educator.school == school
+    false
+  end
+
+  def is_authorized_for_section?(section)
+    return true if @educator.districtwide_access?
+
+    return false if @educator.school.present? && @educator.school != section.course.school
+
+    return true if @educator.schoolwide_access? || @educator.admin?
+    return true if section.in?(@educator.sections)
+    false
+  end
+
+  def is_authorized_for_note?(event_note)
+    return false unless is_authorized_for_student?(event_note.student)
+    return false if event_note.is_restricted && !@educator.can_view_restricted_notes
+    true
+  end
+
+  # TODO(kr) remove implementation
+  def students_for_school_overview
+    return [] unless @educator.school.present?
+
+    if @educator.schoolwide_access?
+      @educator.school.students.active
+    elsif @educator.has_access_to_grade_levels?
+      @educator.school.students
+        .active
+        .where(grade: @educator.grade_level_access)
+    else
+      []
+    end
+  end
+
+  # TODO(kr) remove implementation
+  def homerooms
+    # Educator can visit roster view for these homerooms
+    return [] if @educator.school.nil?
+
+    if @educator.districtwide_access?
+      Homeroom.all
+    elsif @educator.schoolwide_access?
+      @educator.school.homerooms.all
+    elsif @educator.homeroom
+      @educator.school.homerooms.where(grade: @educator.homeroom.grade)
+    elsif @educator.grade_level_access.present?
+      @educator.school.homerooms.where(grade: @educator.grade_level_access)
+    else
+      []
+    end
+  end
+
+  # TODO(kr) remove implementation
+  def sections
+    if @educator.districtwide_access?
+      Section.all
+    elsif @educator.schoolwide_access?
+      Section.joins(:course).where('courses.school_id = ?', @educator.school.id)
+    else
+      @educator.sections
+    end
+  end
+end
--- a/app/models/educator.rb
+++ b/app/models/educator.rb
@@ -17,14 +17,14 @@ class Educator < ActiveRecord::Base
 
   validates :email, presence: true, uniqueness: true
 
-  validate :has_school_unless_districtwide,
-           :admin_gets_access_to_all_students,
-           :grade_level_access_is_array_of_strings,
-           :grade_level_strings_are_valid,
-           :grade_level_strings_are_uniq,
-           :grade_level_access_is_not_nil
+  validate :validate_has_school_unless_districtwide,
+           :validate_admin_gets_access_to_all_students,
+           :validate_grade_level_access_is_array_of_strings,
+           :validate_grade_level_strings_are_valid,
+           :validate_grade_level_strings_are_uniq,
+           :validate_grade_level_access_is_not_nil
 
-  VALID_GRADES = [ 'PK', 'KF', '1', '2', '3', '4', '5', '6', '7', '8' ].freeze
+  VALID_GRADES = [ 'PK', 'KF', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12' ].freeze
 
   # override
   # The `student_searchbar_json` field can be really heavy (~500kb), and
@@ -34,85 +34,26 @@ class Educator < ActiveRecord::Base
     super(options.merge({ except: [:student_searchbar_json] }))
   end
 
-  def has_school_unless_districtwide
-    if school.blank?
-      errors.add(:school_id, "must be assigned a school") unless districtwide_access?
-    end
-  end
-
-  def admin_gets_access_to_all_students
-    errors.add(:admin, "needs access to all students") if admin? && !has_access_to_all_students?
-  end
-
-  def grade_level_access_is_array_of_strings
-    unless grade_level_access.all? { |grade| grade.instance_of? String }
-      errors.add(:grade_level_access, "should be an array of strings")
-    end
-  end
-
-  def grade_level_strings_are_valid
-    if grade_level_access.any? { |grade| !(VALID_GRADES.include?(grade)) }
-      errors.add(:grade_level_access, "invalid grade")
-    end
-  end
-
-  def grade_level_strings_are_uniq
-    unless grade_level_access.uniq.size == grade_level_access.size
-      errors.add(:grade_level_access, "duplicate values")
-    end
-  end
-
-  def grade_level_access_is_not_nil
-    errors.add(:grade_level_access, "cannot be nil") if grade_level_access.nil?
-  end
-
-  # This method is the source of truth for whether an educator is authorized to view information about a particular
-  # student.
   def is_authorized_for_student(student)
-    return true if self.districtwide_access?
-
-    return false if self.restricted_to_sped_students && !(student.program_assigned.in? ['Sp Ed', 'SEIP'])
-    return false if self.restricted_to_english_language_learners && student.limited_english_proficiency == 'Fluent'
-    return false if self.school_id.present? && self.school_id != student.school_id
-
-    return true if self.schoolwide_access? || self.admin? # Schoolwide admin
-    return true if self.has_access_to_grade_levels? && student.grade.in?(self.grade_level_access) # Grade level access
-    return true if student.in?(self.students) # Homeroom level access
-    return true if student.in?(self.section_students) # Section level access
-    false
+    Authorizer.new(self).is_authorized_for_student?(student)
   end
 
-  def is_authorized_for_school(currentSchool)
-    self.districtwide_access? ||
-    (self.schoolwide_access? && self.school == currentSchool) ||
-    (self.has_access_to_grade_levels? && self.school == currentSchool)
+  def is_authorized_for_school(current_school)
+    Authorizer.new(self).is_authorized_for_school?(current_school)
   end
 
   def is_authorized_for_section(section)
-    return true if self.districtwide_access?
-
-    return false if self.school.present? && self.school != section.course.school
-
-    return true if self.schoolwide_access? || self.admin?
-    return true if section.in?(self.sections)
-    false
+    Authorizer.new(self).is_authorized_for_section?(section)
   end
 
   def students_for_school_overview(*additional_includes)
-    return [] unless school.present?
-
-    if schoolwide_access?
-      school.students
-            .active
-            .includes(additional_includes || [])
-    elsif has_access_to_grade_levels?
-      school.students
-            .active
-            .where(grade: self.grade_level_access)
-            .includes(additional_includes || [])
-    else
+    students = Authorizer.new(self).students_for_school_overview
+    if students.respond_to?(:includes)
+      students.includes(additional_includes || [])
+    elsif students.size == 0
       logger.warn("Fell through to empty array in #students_for_school_overview for educator_id: #{self.id}")
-      []
+    else
+      students
     end
   end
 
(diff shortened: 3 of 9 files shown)


app/models/educator.rb
```diff
@@ -133,34 +74,11 @@ def has_access_to_grade_levels?
...
-    if districtwide_access?
-      Section.all
-    elsif schoolwide_access?
-      Section.joins(:course).where('courses.school_id = ?', school.id)
-    else
-      sections
-    end
-  end
-
-  def allowed_homerooms_by_name
```
kevinrobinson: No need for this as a method so removed it and updated the one call site.

app/models/educator.rb
```diff
@@ -34,85 +34,26 @@ def as_json(options = {})
...
-      school.students
-            .active
-            .includes(additional_includes || [])
-    elsif has_access_to_grade_levels?
-      school.students
-            .active
-            .where(grade: self.grade_level_access)
-            .includes(additional_includes || [])
-    else
+    students = Authorizer.new(self).students_for_school_overview
```
kevinrobinson: This is the only substantive change - it removes the if branch here and lets `Authorizer` handle that.

app/models/educator.rb
```diff
@@ -17,14 +17,14 @@ class Educator < ActiveRecord::Base
...
-           :grade_level_access_is_not_nil
+  validate :validate_has_school_unless_districtwide,
+           :validate_admin_gets_access_to_all_students,
+           :validate_grade_level_access_is_array_of_strings,
+           :validate_grade_level_strings_are_valid,
+           :validate_grade_level_strings_are_uniq,
+           :validate_grade_level_access_is_not_nil
 
-  VALID_GRADES = [ 'PK', 'KF', '1', '2', '3', '4', '5', '6', '7', '8' ].freeze
+  VALID_GRADES = [ 'PK', 'KF', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12' ].freeze
```
kevinrobinson: `TestPals` revealed that this was a bug and wouldn't allow grade-level access at the high school.

app/models/educator.rb
```diff
@@ -17,14 +17,14 @@ class Educator < ActiveRecord::Base
```
kevinrobinson: This file is almost all renaming or removing code that was moved into `Authorizer`.

app/models/educator.rb
```diff
@@ -198,18 +105,39 @@ def save_student_searchbar_json
...
-    schoolwide_access == false && grade_level_access == [] && homeroom.blank?
+  def validate_admin_gets_access_to_all_students
+    has_acces_to_all_students = (
+      restricted_to_sped_students == false &&
+      restricted_to_english_language_learners == false
+    )
+    errors.add(:admin, "needs access to all students") if admin? && !has_acces_to_all_students
   end
 
-  def has_grade_level_access?
```
kevinrobinson: not called anywhere

app/models/educator.rb
```diff
@@ -198,18 +105,39 @@ def save_student_searchbar_json
   end
 
   private
+  def validate_has_school_unless_districtwide
+    if school.blank?
+      errors.add(:school_id, "must be assigned a school") unless districtwide_access?
+    end
+  end
 
-  def has_access_to_no_students?
```
kevinrobinson: not called anywhere

spec/support/test_pals.rb
```diff
@@ -0,0 +1,108 @@
...
+  attr_reader :shs_ninth_grade_counselor
+  attr_reader :shs_jodi_homeroom
+  attr_reader :shs_biology_course
+  attr_reader :shs_tuesday_biology_section
+  attr_reader :shs_thursday_biology_section
+
+  def create!
+    School.seed_somerville_schools
+
+    # Uri works in the central office, and is the admin for the entire
```
kevinrobinson: Sort of like little stories.  There's lots of details here, so thought that would help us make sense of this, and also let us talk about it like real rules.

kevinrobinson:
Not sure what this test failure is about, it might be test pollution and I have to look more.  This is good to review otherwise though.

alexsoble:
@kevinrobinson Reviewing now!

app/models/educator.rb
alexsoble: Ayyyy great catch. 👍

spec/support/test_pals.rb
alexsoble: @kevinrobinson Great! I love these.

alexsoble:
@kevinrobinson I gave the code a read-through and opened those two patches that you just merged in. Thanks! 👍 

I'd love to do some local QA against this branch too. 

I checked the CI, and in addition to the rubocop failures (which should be fixed now) there's one spec failure. The spec failed locally for me as well. It relates to deleting student section assignments during import, an area which has caused some outage in our app before. So let's investigate before we assume test pollution. Here's the spec failure output:

```
~/somerville-teacher-tool(feature/authorization-simple) » rspec spec/importers/file_importers            alexsoble@Alexs-Computer
.............................................................F..............................

Failures:

  1) StudentSectionAssignmentsImporter#delete_rows happy path deletes all student section assignments except the recently imported one
     Failure/Error: expect(StudentSectionAssignment.count).to eq(1)
     
       expected: 1
            got: 21
     
       (compared using ==)
     # ./spec/importers/file_importers/student_section_assignments_importer_spec.rb:126:in `block (4 levels) in <top (required)>'

Finished in 10.28 seconds (files took 16.45 seconds to load)
92 examples, 1 failure

Failed examples:

rspec ./spec/importers/file_importers/student_section_assignments_importer_spec.rb:125 # StudentSectionAssignmentsImporter#delete_rows happy path deletes all student section assignments except the recently imported one

Coverage report generated for RSpec to /Users/alexsoble/somerville-teacher-tool/coverage. 1786 / 2129 LOC (83.89%) covered.
```

alexsoble:
@kevinrobinson Ooooh, now that I think about it, this could be test pollution related to student section assignments created in TestPals?

alexsoble:
OK, started to make some progress on this final test failure. 

Looks like it has to do with different sets of records in the `schools` table. 

I added some code to print out data on the schools as part of the test suite, and ran the test suite on master and on a patched branch of `feature/authorization-simple` (0018c27e3fb36e9107f873fb6fc7785c2a58dba7) .

```
# master

rspec
..........................................................................................................................................................................................................................................School.count
4

School.pluck(:local_id)
SHS




School.where(local_id: nil).count
3

...........................................................................................................................................................................................................
```

```
# patch/fix-up-final-failing-spec-for-authorization-simple

rspec
..........................................................................................................................................................................................................................................School.count
4

School.pluck(:local_id)
SHS
LOCAL_ID_496
LOCAL_ID_497
LOCAL_ID_498

School.where(local_id: nil).count
0

F.......................................................................................................................................................................................................................
```

alexsoble:
The LOCAL_ID_123 pattern is coming from the schools factory:

```
# spec/factories/schools.rb

factory :school do
  sequence(:state_id) {|n| n }
  sequence(:slug) {|n| "slug#{n}" }
  sequence(:local_id) {|n| "LOCAL_ID_#{n}" }
end
```

We see these factory-made schools in the database in `feature/authorization-simple`, but not `master`.

alexsoble:
What's interesting is that as far as I could tell from some CRTL+F searching, this PR doesn't actually create any schools using the FactoryGirl. So maybe we could safely remove that line and fix the CI error.

alexsoble:
Heyyyyyy CI is passing now. 🎉  cc @kevinrobinson

alexsoble:
Looks like `educator#students_for_school_overview` is broken in some cases on this PR. When I run the seed task with some logging, here's the output I get:

```
Creating demo schools, homerooms, interventions...
Seeding database constants for Somerville...
Creating demo educators...
Creating homerooms..
Creating course and sections..
Creating students for homeroom HEA 500...
Creating students for homeroom WSNS 500...
Creating students for section ART-302A...
Creating students for section ART-302B...
Creating students for section SCI-201A...
Creating students for section SCI-201B...
Querying for jobs...
educator.students_for_school_overview
#<Student:0x007fe27ffb9af0>
#<Student:0x007fe27ffb9280>
#<Student:0x007fe27ffb8380>
#<Student:0x007fe28134f8f8>
#<Student:0x007fe28134eef8>
#<Student:0x007fe28134dbc0>
#<Student:0x007fe28134d300>
#<Student:0x007fe28134cd38>
#<Student:0x007fe28134c798>
#<Student:0x007fe28134c040>
#<Student:0x007fe28134b4d8>
#<Student:0x007fe28134aab0>
#<Student:0x007fe28134a498>
#<Student:0x007fe281349f20>
#<Student:0x007fe2813492c8>
#<Student:0x007fe27ffcfad0>
#<Student:0x007fe27ffcf558>
#<Student:0x007fe27ffcec48>
#<Student:0x007fe27ffce6d0>
educator.students_for_school_overview
true
rake aborted!
NoMethodError: undefined method `map' for true:TrueClass
Did you mean?  tap
/somerville-teacher-tool/app/jobs/precompute_student_hashes_job.rb:41:in `educator_to_jobs'
//somerville-teacher-tool/app/jobs/precompute_student_hashes_job.rb:23:in `block in school_overview_precompute_jobs'
```

alexsoble:
Here's the educator record that throws the error when we try to precomptue school overview for her:

```
#<Educator id: 3, email: "fake-fifth-grade@example.com", created_at: "2017-12-01 16:02:15", updated_at: "2017-12-01 16:02:15", admin: false, phone: nil, full_name: "Teacher, Sarah", state_id: nil, local_id: "450", staff_type: nil, school_id: 2, schoolwide_access: false, grade_level_access: [], restricted_to_sped_students: false, restricted_to_english_language_learners: false, can_view_restricted_notes: false, districtwide_access: false, can_set_districtwide_access: false, student_searchbar_json: nil>
educator.students_for_school_overview
true
```

alexsoble:
@kevinrobinson Opened another patch PR!: #1295

alexsoble:
Now that specs are passing and seed task is running, the next QA thing in my brain is to set up the database with all the TestPals data and make GIFs of trying to login to various things so that I can see the testpals test results with my own 👀 s and make the scenarios a little more concrete. Does that sound like overkill @kevinrobinson?

alexsoble:
OK, working on the strategy I described above! Going to open some comments in the specs / test pals sections as I notice things.

spec/lib/authorizer_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+        thin_student = Student.select(:id, :local_id).find(pals.shs_freshman_mari.id)
+        expect(authorized(pals.uri) { thin_student }.attributes).to eq({
+          'id' => pals.shs_freshman_mari.id,
+          'local_id' => pals.shs_freshman_mari.local_id
+        })
+        expect { authorized(pals.healey_teacher) { thin_student } }.to raise_error(ActiveModel::MissingAttributeError)
+        expect { authorized(pals.shs_bill_nye) { thin_student } }.to raise_error(ActiveModel::MissingAttributeError)
+      end
+
+      it 'if `select` was used on an ActiveRecord::Relation for students, fields needed for authorization are added in' do
```
alexsoble: The test code here is identical to the test code above. But the context block description is different. So we should either change the test code to test a different thing or remove the duplication. I can take a crack at updating the test code based on what you described @kevinrobinson.

app/lib/authorizer.rb
```diff
@@ -0,0 +1,144 @@
...
+
+      return false if @educator.restricted_to_sped_students && !(student.program_assigned.in? ['Sp Ed', 'SEIP'])
+      return false if @educator.restricted_to_english_language_learners && student.limited_english_proficiency == 'Fluent'
+      return false if @educator.school_id.present? && student.school_id.present? && @educator.school_id != student.school_id
+
+      return true if @educator.schoolwide_access? || @educator.admin? # Schoolwide admin
+      return true if @educator.has_access_to_grade_levels? && student.grade.in?(@educator.grade_level_access) # Grade level access
+      return true if student.in?(@educator.students) # Homeroom level access
+      return true if student.in?(@educator.section_students) # Section level access
+    rescue ActiveModel::MissingAttributeError => err
```
alexsoble: I have a question about the `MissingAttributeError` rescue here!  

When I look at the code above, I don't see how any of it could trigger a `MissingAttributeError` error. Here's how I'd imagine a line like `@educator.restricted_to_sped_students` would work:

```
# @educator exists
@educator.restricted_to_sped_students
 => false 
 => true

# does not exist
nil.restricted_to_sped_students
=> NoMethodError: undefined method `restricted_to_sped_students' for nil:NilClass
```

I imagine the same for the `student` variable. 

So where might we see `MissingAttributeError` being raised in the code?

app/lib/authorizer.rb
```diff
@@ -0,0 +1,144 @@
...
+      # We can't do authorization checks on models with `#select` that are missing
+      # fields.  If this happens, it's probably because the developer is trying to
+      # to optimize a query.  The authorization layer could only recover by making more
+      # queries, so instead we raise and force the developer to figure out how to resolve.
+      #
+      # See `Authorizer.student_fields_for_authorization`
+      raise err
```
alexsoble: I have a question about the raise-and-rescue pattern here. We're not doing any logging or any side effect creation with the error message, just re-raising. And we're not in `map` or `each` loop. So is the intent of this raise-and-rescue mainly to catch the eye of developers who might raise this error, and provide a context for the code comments? Or is it meant to do something else that I'm not understanding yet?

spec/lib/authorizer_spec.rb
kevinrobinson: Good catch!  I think based on the text description, intention was to change line 102 to something like `Student.select(:id, :local_id).where(id: pals.shs_freshman_mair.id)` or something that returns an `ActiveRecord::Relation`.

app/lib/authorizer.rb
kevinrobinson: Yeah - this problem here (with using `select`) is an edge case, but it came up in the `lasids` endpoint, which is how I realized it.

So yeah, I was thinking about catching the eye of developers, so added in a comment, and added in the `begin/raise` to make explicit what's happening (we could also add a log message in the raise, I think that's what I did initially in trying to understand what was happening).

If this seems weird, we could remove it for just comments, or we could improve the comments to make it clearer why it's like this or what's happening.

spec/lib/authorizer_spec.rb
alexsoble: Does what I sketched out in #1303 test what you were thinking?

app/lib/authorizer.rb
alexsoble: OK, I see! So right now the `lasids` endpoint looks like this:

```
def lasids
  render json: Student.pluck(:local_id)
end
```

There's no explicit authorization check, and that's because authorization is handled at the controller level like this:

```
before_action :authorize_for_districtwide_access_admin, only: [:lasids]
```

So basically, part of this PR is moving us to a more standardized way of handling things here, where we'd want to call `authorized` in our `lasids` endpoint code, even though there's controller authorization, because that's a more fail-proof and centralized way of doing things. Does that sound about right? 

And as we keep implementing this work, we'd move from `:pluck` (which only works when we assume the educator is a districtwide admin and can access everything) to some kind of `select` statement.

app/lib/authorizer.rb
alexsoble: OK so now thanks to our other thread of conversation about this method I'm looking at https://apidock.com/rails/v4.0.2/ActiveRecord/QueryMethods/select:

```
Accessing attributes of an object that do not have fields retrieved by a select will throw ActiveModel::MissingAttributeError:
```

alexsoble: But what's confusing me here is that `is_authorized_for_student?` is down at the individual student level, whereas `select` is called on an entire class or an active record relation, like this:

```
2.3.0 :001 > Student.select(:local_id)
2.3.0 :002 > Student.all.select(:local_id)
```

So it would seem like we're rescuing at the wrong level here.

kevinrobinson:
Directionally, here's some steps:
- [ ] ship this
- [ ] move Educator specs related to authorization into authorizer specs
- [ ] move specs for `homerooms`, `sections`, etc. into authorizer and improve them
- [ ] update call sites in each controller to call into `Authorizer` methods or to move the authorization logic in the controller into `Authorizer`.  Update corresponding tests
...later...
- [ ] consolidate code in authorizer so logic isn't redundant (eg, `is_authorized_forstudent` vs. `students_for_school_overview`)

app/lib/authorizer.rb
kevinrobinson: @alexsoble I think we talked about this the other day, but when `select` is used on the class or relation like that, it returns active record models that only have some of the attributes of the regular ActiveRecord model.  This is an edge case, but came in trying to update the `students_controller#lasids` method to use the `authorized` API here.  That code was trying to grab all students but only a field or two, and so this would raise to say "hey you have to add some more fields so that the authorization checks can work."

kevinrobinson:
@alexsoble I think we talked about this in person but I didn't track it well enough.  Here's the next steps: https://github.com/studentinsights/studentinsights/issues/1080#issuecomment-350572396

If this seems good, let's ship this PR and then https://github.com/studentinsights/studentinsights/pull/1315 is on top of this work, and updates local development mode to use `TestPals` so that we can do manual testing and automating testing on the substance of the authorization rules.

I think we also need a step to check "do the roles and combinations in `TestPals` actually match the production data?"  The admin UI page in https://github.com/studentinsights/studentinsights/pull/1315 is a step towards looking at that too.

So if this seems good to you, let's 🚢 this, verify it, then aim to ship https://github.com/studentinsights/studentinsights/pull/1315?

alexsoble:
@kevinrobinson I need to finish up that patch which changes the duplicate spec to a different spec for the `select` edge case. Sorry if that's becoming a blocker. As I look through the PR I also have some questions on some other areas. I don't want to drag this work down but I do want to give it thorough review. Please be frank with me about how I can best strike that balance!

app/controllers/application_controller.rb
```diff
@@ -48,6 +48,20 @@ def default_section_path(educator)
     no_sections_path
   end
 
+  # Wrap all database queries with this to enforce authorization
+  def authorized(&block)
+    authorizer.authorized(&block)
+  end
+
+  # Enforce authorization and raise if no authorized models
+  def authorized_or_raise!(&block)
```
alexsoble: I can't see anywhere in the code that calls this method.

kevinrobinson: @alexsoble Yep, it's not used here.  The idea is that it's for controller code doing lookups like:

```
def show
  student = authorized_or_raise! { Student.find(params[:id]) }
  ...
```
 vs.
```
def index
  students = authorized { Student.all }
  ...
```

We could get rid of it if we wanted to simplify, but maybe worth sharing more about why it's here and some other pieces of next steps.

So after we talked about this `authorized` API I made it and then tried to use it with the `students_controller`.  That work is over here: https://github.com/studentinsights/studentinsights/pull/1237/files#diff-11751d644b56a1cce6d8dfc73e331b91R21 and here is a smaller example that shows better what's involved: https://github.com/studentinsights/studentinsights/commit/febdb0883399814a94d64a217efc5575eb640d13.

That involved several other changes: either refactoring the controller code, updating the test setup, or adding other models like Section to the authorizer code.  It also seemed like it'd be good to revise the `students_controller_spec` to use the same test setup code in `TestPals` that is used in the `authorizer_spec` here.  All that seemed like it was getting to be a big change, so I pulled off this piece of work to merge it first.  So that's why I left this in here, and probably should have explained more.

alexsoble: OK that makes a lot of sense! So basically `authorized_or_raise!` would be used for an individual record like a Student or an EventNote.

alexsoble:
@kevinrobinson Thanks for bearing with me and patiently walking me through a bunch of the different tradeoffs and edge cases in this PR! I think we're in a good place to 🚢 . All the existing specs are passing; a bunch of new, more realistic authorization specs are passing; and most of the changes to feature-side code involve moving logic from `educator.rb` to `authorizer.rb`. 

I worked on fixing the spec for that one `#select` edge case we were talking about earlier, but couldn't get it to work or really get my brain around it. That said, as far as I understand it's testing a feature that's not implemented into the codebase yet. I updated #1303 to `skip` the test so that we'd have a reminder to address this eventually, but still let the CI pass. I think that's at least better than leaving a spec that looks legitimate but isn't really testing what it says it's testing. 

@kevinrobinson Let me know if you think #1303 is an acceptable solution for now. If so then let's merge it as a patch and 🚢 .

kevinrobinson:
@alexsoble Thank you for working through this too!  I had a hard time figuring out how to do this in discrete steps, so it's really helpful to talk through it.

I fixed up the issue you found in https://github.com/studentinsights/studentinsights/pull/1303 and discovered the `authorizer` code had a bug that the new test caught.  I pushed it as a commit https://github.com/studentinsights/studentinsights/commit/20f54407c5ae3cc95377d99dc1250bac0b697710.  So I think this is good to go!

alexsoble:
@kevinrobinson Look like the suite fails now tho

kevinrobinson:
@alexsoble Yep, my bad - the failing test in https://travis-ci.org/studentinsights/studentinsights/builds/315615404?utm_source=github_status&utm_medium=notification is from the fix to recover if "thin relations" are passed.  So that test case doesn't fail anymore since `authorized` can recover that.  I updated it to past an array of thin models, which `authorized` can't recover from, and now that test case raises as expected and it passes.

alexsoble:
I can deploy this in a couple of hours after some menorah 🕯 s have been lit! or tomorrow AM.

kevinrobinson:
@alexsoble Cool no rush.  Mazel tov! 🕎

alexsoble:
🚢 ing!

alexsoble:
We don't really have the tools to thoroughly triple-check this in production efficiently? — that would involve changing a bunch of settings on a test user account and logging in each time. But I logged in as test user and clicked around to verify the sheer minimum, plus we've got our test suite and now test pals to verify all these changes and improvements!

kevinrobinson:
@alexsoble 👍 Yeah this is hard to verify - for this and subsequent PRs, I think manual testing locally, improving test coverage for different cases, and adding audit tools for production will help kick the tires and shake things down more as well as we work through https://github.com/studentinsights/studentinsights/issues/1080#issuecomment-350572396 over time.

Thanks for helping ship this!  And if there are exceptions or issues that do come up related to this and I miss that, please feel free to ping me whenever too.

PR #1278, opened 2017-11-27
Patch: add a little more explanation to code comments in authorizer.rb

alexsoble:
# Notes 

+ This PR is a patch onto the `feature/authorization-simple` branch. 
+ @kevinrobinson: Adding a little more explanation to some of the code comments in `authorizer.rb` to make them more explicit about why code like `current_educator.homeroom.students` doesn't work for all roles.


kevinrobinson:
👍

PR #1285, opened 2017-11-28
Bump moment js version to 2.11.2

alexsoble:
# Notes 

+ GitHub is warning about a security vuln in Moment.js 2.11. I took a look and it's a DDOS vulnerability, probably not something that's likely to apply to us since (a) our site's not a DDOS target and (b) someone would have to get behind authentication to exploit it anyway.
+ That said, why not bump the version to resolve the warning.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "jquery": "^3.2.1",
     "js-cookie": "^2.1.4",
     "lodash": "3.10.1",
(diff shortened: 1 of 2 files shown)


alexsoble:
Yeah, first time I’ve seen it as well! Looks like they’re creating
Gemnasium-esque functionality in house.

PR #1286, opened 2017-11-28
Fix missing uploaded_by_educator_id column on Service Uploads table

### LINKED ISSUE #1282: Fix uploaded_by_educator_id assignment in ServiceUpload
See https://github.com/studentinsights/studentinsights/pull/1275#issuecomment-347276598.

alexsoble:

+ Right now back-end isn't storing any values for the `uploaded_by_educator_id` column in the `service_uploads` table. This isn't a huge problem since Uri is the only educator creating service upload records using the UI. 
+ Fix #1282.

# Checklist 

+ [x] Test migration locally
+ [x] Verify uploaded_by_educator_id gets set correctly in local development mode
+ [ ] Code review 
+ [ ] Deploy
+ [ ] Run `heroku run rake data_migration:backfill_service_upload_ids`

### DIFF
--- a/app/controllers/service_uploads_controller.rb
+++ b/app/controllers/service_uploads_controller.rb
@@ -55,7 +55,7 @@ class ServiceUploadsController < ApplicationController
     end
 
     render json: { service_upload: service_upload.as_json(
-      only: [:created_at, :file_name, :id, :uploaded_by_educator_id],
+      only: [:created_at, :file_name, :id],
       include: {
         services: {
           only: [],
--- a/app/controllers/service_uploads_controller.rb
+++ b/app/controllers/service_uploads_controller.rb
@@ -12,7 +12,7 @@ class ServiceUploadsController < ApplicationController
   def create
     service_upload = ServiceUpload.new(
       file_name: params['file_name'],
-      uploaded_by_educator_id: params['uploaded_by_educator_id']
+      uploaded_by_educator_id: current_educator.id
     )
 
     if service_upload.invalid?
--- a/app/controllers/service_uploads_controller.rb
+++ b/app/controllers/service_uploads_controller.rb
@@ -10,7 +10,10 @@ class ServiceUploadsController < ApplicationController
   end
 
   def create
-    service_upload = ServiceUpload.new(file_name: params['file_name'])
+    service_upload = ServiceUpload.new(
+      file_name: params['file_name'],
+      uploaded_by_educator_id: params['uploaded_by_educator_id']
+    )
 
     if service_upload.invalid?
       return render json: {
@@ -52,7 +55,7 @@ class ServiceUploadsController < ApplicationController
     end
 
     render json: { service_upload: service_upload.as_json(
-      only: [:created_at, :file_name, :id],
+      only: [:created_at, :file_name, :id, :uploaded_by_educator_id],
       include: {
         services: {
           only: [],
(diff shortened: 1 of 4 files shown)


app/controllers/service_uploads_controller.rb
```diff
@@ -10,7 +10,10 @@ def authorize_for_districtwide_access_admin
   end
 
   def create
-    service_upload = ServiceUpload.new(file_name: params['file_name'])
+    service_upload = ServiceUpload.new(
+      file_name: params['file_name'],
+      uploaded_by_educator_id: params['uploaded_by_educator_id']
```
kevinrobinson: This is a security bug - it lets users tell us who they are in an untrusted way.  So if I'm logged in as myself, I can upload some services and tell the server "I'm Alex!" and it will say "okay, Alex uploaded this."  For tracking who is doing something, use `current_educator` in the controller instead of trusting the client.

lib/tasks/data_migrations/backfill_service_upload_educator_ids.rake
```diff
@@ -0,0 +1,19 @@
+namespace :data_migration do
+  desc "Backfill all service upload IDs"
+  task backfill_service_upload_ids: :environment do
+    raise "Need an Uri ID!" unless ENV['URI_ID']
+
+    puts "Updating service upload records..."; puts;
+
+    ActiveRecord::Base.transaction do
+      ServiceUpload.all.each do |service_upload|
+        service_upload.uploaded_by_educator_id = ENV.fetch('URI_ID')
```
kevinrobinson: Do we want to do all, or just the nil ones?

app/controllers/service_uploads_controller.rb
alexsoble: Ooooooh, good catch. Thanks, will fix.

alexsoble: Darn untrustworthly client!!!

lib/tasks/data_migrations/backfill_service_upload_educator_ids.rake
alexsoble: Right now they're all `nil` and this is a one-time-only rake task so it's the same thing.

kevinrobinson:
If you accidentally run this take task a year from now it’d lose some
data.  If it only updates nils then it’s always safe to run.  Not a big
deal, but in general code that blindly goes and changes all records in a
table is peligroso!

alexsoble:
Hey @kevinrobinson, I really appreciate your astute observations and feedback on this! I've implemented fixes, updated specs, and got the CI to pass. I'd like to do another round of Internet Explorer QA, then we can re-review, and then hopefully ship the code and run the migration. 👍

kevinrobinson:
@alexsoble no problem! 👍

alexsoble:
This one fell off the burner around seven weeks ago, probably because that's around when we started focusing on shipping a minimal New Bedford Insights instance . . .

But I want to get this fixed before New Bedford is fully operational so that their service uploading works correctly! 🔧 

I fixed the red spec that was holding it up and did a manual verification in my local environment that `uploaded_by_educator_id` gets set correctly when a new ServiceUpload is created. 

@kevinrobinson, would you mind re-reviewing when you have a moment? I think this should be ready to 🚢 .

kevinrobinson:
This looks good in production!

before:
``
$  bundle exec rake missing_foreign_key_relations:scan
...
Missing Educator IDs on ServiceUpload
13 records missing an object. Missing IDs: [nil]
...
```

after shows they're good now!

alexsoble:
@kevinrobinson Yay, thank you so much for merging/deploying/migrating this! 🌟

PR #1293, opened 2017-11-30
Patch: Fix up final failing spec in feature/authorization-simple

alexsoble:
# Notes 

+ This should fix the CI error. I'll wrote more here about why, it's kind of interesting/unexpected. But bottom line it's a spec problem, not a feature-side problem.

### DIFF
--- a/spec/factories/schools.rb
+++ b/spec/factories/schools.rb
@@ -3,7 +3,6 @@ FactoryGirl.define do
   factory :school do
     sequence(:state_id) {|n| n }
     sequence(:slug) {|n| "slug#{n}" }
-    sequence(:local_id) {|n| "LOCAL_ID_#{n}" }
   end
 
   factory :healey, class: School do


alexsoble:
Here's the feature-side method that was causing the error:

```
# student_section_assignments_importer.rb

def delete_rows
  #Delete all stale rows no longer included in the import
  #For the schools imported during this run of the importer
  StudentSectionAssignment.joins(:section => {:course => :school})
                          .where.not(id: @imported_assignments)
                          .where(:schools => {:local_id => school_scope}).delete_all
end
```

Here's what I imagine this code saying: 

"Take all the section assignments. Filter down to the ones associated with the schools we're importing for. Filter out the assignments we imported during this job. Then delete all the ones we have left."

The last line is the one we want to look at. It assumes that `school_scope` is an array. But in the context of the spec we're looking at in `student_section_assignments_import_spec.rb`, `school_scope` is `nil` because it was never set. At the same time, all the schools generated by our factories also had local_ids of `nil` (because there was no sequence set up in FactoryGirl to generate them). Since both values default to nil, in the test context this happens to pass. 

I think a better solution would be to have the test context assume an array-like school scope, like we have in the live, production importer classes. That would make the specs more precise and a better model of our app. We should do that work in a separate branch though so that we don't make it a blocker for this authorization refactor work, which is high priority.

alexsoble:
Gonna merge this in because I'm excited, I wanna see if it'll fix the CI situation, the CI doesn't run unless the branch is against master, and I'm merging into another branch so it's not shipping to prod yet. 👍

alexsoble:
selfie!

PR #1295, opened 2017-12-01
Patch: Send down empty array if educator is authorized for no students

alexsoble:
# Notes 

+ Otherwise, Rails returns `true` which isn't mappable, the precompute job needs a mappable array to work.
+ Tested this patch with a local `rake db:reset`, it worked.


alexsoble:
selife patch! :D

kevinrobinson:
👍  thanks!

PR #1296, opened 2017-12-01
Add another test case to PrecomputeStudentHashesJob spec

alexsoble:
# Notes 

+ This test case would have caught this issue in #1274: https://github.com/studentinsights/studentinsights/pull/1274#issuecomment-348531704
+ Adds a test for how the precomputing job handles classroom teachers who don't get any school overview precomputing; if the API sends down a non-mappable value for one of these teachers, the precomputing job will break; so we should test that all the values being sent down by the `students_for_school_overview` API are mappable (like an array, AR relation, or another enum)


alexsoble:
Merging but noting that we're moving towards a different style of testing authorization for educators, defining more specific roles like in TestPals on @kevinrobinson's authorization refactor branch.

PR #1297, opened 2017-12-01
Experiment – Get JS test coverage reports

alexsoble:
Inspired by @Jngai's comment about test coverage! This is just an experiment I started hacking on while eating lunch. Let's get a badge that shows our JS test coverage stats and motivates us to bring them higher and higher!


alexsoble:
To be explicit, this is low priority and just a messing-around PR.

PR #1300, opened 2017-12-01
Remove incorrect isRequired statements

### LINKED ISSUE #1219: Update development data or UI code to handle no next-gen MCAS data
This is what comes up locally clicking into an academic details page on the student profile page:
screen shot 2017-10-28 at 9 27 05 am

Not sure whether best to handle this in: data generators for local development, querying in Rails, or in the UI.

Related to https://github.com/studentinsights/studentinsights/pull/1178.

alexsoble:
# Issue 

+ Fix #1219. 

# Notes 

+ it's not the case that every student needs to have both next-gen and classic MCAS scores in ELA and Math. Some students will have only one type, some will have both, and some will have neither. See https://github.com/studentinsights/studentinsights/issues/1219#issuecomment-340575572 for more detail.

# Checklists

+ I don't think this needs tooooo much JS IE review because it doesn't touch any functionality, just prop type checking.

### DIFF
--- a/app/assets/javascripts/student_profile/ela_details.jsx
+++ b/app/assets/javascripts/student_profile/ela_details.jsx
@@ -49,7 +49,7 @@ import {merge} from '../helpers/react_helpers.jsx';
 
     propTypes: {
       chartData: React.PropTypes.shape({
-        star_series_reading_percentile: React.PropTypes.array,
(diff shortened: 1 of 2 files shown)


alexsoble:
selfie with good input from @kevinrobinson

alexsoble:
so maybe not a selfie at all

PR #1303, opened 2017-12-04
Patch: Remove duplicate test code for #select edge case on feature/authorization-simple

alexsoble:
# What does this PR do?

+ Removes duplicate test code for `#select` edge case

### DIFF
--- a/spec/lib/authorizer_spec.rb
+++ b/spec/lib/authorizer_spec.rb
@@ -99,13 +99,7 @@ RSpec.describe Authorizer do
       end
 
       it 'if `select` was used on an ActiveRecord::Relation for students, fields needed for authorization are added in' do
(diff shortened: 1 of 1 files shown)


alexsoble:
Note to self: you can finish up this patch, you don't need to test this under a lot of contexts because this is an edge case you're testing, also it doesn't apply to district admins because they bail out early in the authorization tree.

alexsoble:
Also don't call JSON in this test.

kevinrobinson:
@alexsoble I pushed a commit fixing this up here: https://github.com/studentinsights/studentinsights/pull/1274/commits/20f54407c5ae3cc95377d99dc1250bac0b697710

alexsoble:
@kevinrobinson Oh great. Yep, just saw it. So it looks like this spec was being difficult because there was a real bug in the feature side code. Exactly what we want our specs to do!

Issue #1264, opened 2017-11-17
User testing feedback from Absences Dashboard [Winter Hill Community School]


alexsoble:
@edavidsonsawyer — Today we got the chance to user test the absences dashboard with the principal at Winter Hill Community School. @snoopyuri introduced him to the dashboard and asked him to explore, narrate out loud, and explain what he would do with the data he saw. 👍 

The data loaded up great and the principal thought the charts were really useful! He said he would use this to identify homerooms that could improve their attendance and to create a "call list" of students with attendance issues so that he could sit down and call parents to urge better attendance.

We also spotted a few areas for improvement during the user testing session, which I'll document here. We can break these out into separate issues, but for now I'll put all the feedback / areas for refinement that we saw here for discussion and prioritization: 

+ [ ] **Principal thought that the date boxes on either side of the range were editable and tried to type into them to set the dates.** He didn't see them as labels. Lots of ways to address this issue. Personally, this made me think maybe users would find it more intuitive to use two datepickers instead of a sliding range here since we use datepickers in other places in Student Insights?
+ [x] **When principal clicked on the bar for each homeroom, he expected the students table to populate with the students in that homeroom.** This is essentially the same paradigm as the School Overview Page. In some ways the School Overview Page is training users to expect this! 
+ [ ] **Principal wanted to see bar charts default to start at beginning of school year instead of 1 calendar year back.** @edavidsonsawyer, you intuited this in your initial version of this feature! Based on what we've learned from our last iteration, I think we should implement this on the client side rather than the server side. 
+ [x] **Principal would use the students table as a "call list" to call parents about attendance issues.** To me this suggested hyperlinking the names in the students table to student profile pages, so that users can easily find contact info for students and make calls home when attendance issues bubble up.

edavidsonsawyer:
Great stuff, @alexsoble and @snoopyuri! I'll try to follow up with one or both of you by phone to talk through a couple of these points.

alexsoble:
@edavidsonsawyer We can also talk about it on the call today if you can join!

edavidsonsawyer:
Hey @alexsoble and @snoopyuri , I haven't had a lot of time to work on this but #1272 will change the student list to reflect the selected homeroom. I'm still thinking over the appropriate way to switch back to the school view. Right now that's done by clicking on the monthly attendance chart, or clicking on background space in the homeroom attendance chart. It might be better to add a button to make that switch explicit.

alexsoble:
[...]

PR #1305, opened 2017-12-04
Add tardies dashboard

edavidsonsawyer:
# Who is this PR for?
School Principals and Vice Principals who want to track tardies schoolwide and by homeroom

Lack of visibility into trends in student attendance 

# What does this PR do?
Implements a bare bones router so a user can navigate between the two dashboards
Creates a dashboard similar to the absence dashboard which shows the tardies for the past three months.

tardies_dashboard_screenshot

# Checklists

## Javascript QA

*Which features or pages does this PR touch? Make a list here so that you can double-check each affected page in a Internet Explorer before merging.*

School Absence Dashboard
School Tardies Dashboard

+ [x] Author checked latest in IE - School Absence Dashboard
+ [x] Author checked latest in IE - School Tardies Dashboard
+ [x] Reviewer checked latest in IE - School Absence Dashboard
+ [x] Reviewer checked latest in IE - School Tardies Dashboard


alexsoble:
Hey @edavidsonsawyer! Looks like we still have two sub-issues open in #1264. We should fix those first before moving on to the next dashboard. It looks like both of them will apply to the tardies dashboard as well, so would be easier to implement once an re-use rather than implement twice. Let me know what you think!

edavidsonsawyer:
Hey @alexsoble, I agree about the merge order. This is here because I wanted to work through some of these changes before adding the school year restriction in #1264. It seemed possible that would need to change again once these components were modified to accommodate more dashboards. It turns out that won't be a problem, so I should be able to merge this after finishing #1264. The tardy dashboard doesn't deal with date ranges, so the two open issues in #1264 wouldn't affect this feature.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/school_administrator_dashboards.jsx
```diff
@@ -0,0 +1,23 @@
...
+import SchoolwideTardies from './tardies_dashboard/schoolwide_tardies.jsx';
+
+export default React.createClass({
+  displayName: 'SchoolAdministratorDashboards',
+
+  propTypes: {
+    serializedData: React.PropTypes.object.isRequired
+  },
+
+  render: function() {
```
kevinrobinson: I'd suggest finishing off this PR and not solving this right now in this PR, but over time let's converge towards having routing all in one place.  This adds a second "layer" of routing beyond the main one in https://github.com/studentinsights/studentinsights/blob/master/ui/index.js#L27 and https://github.com/studentinsights/studentinsights/blob/master/ui/route.js.

Maybe let's keep moving and come back to "Convert JS routing to use React Router" in a separate issue later?

edavidsonsawyer: @kevinrobinson I'm also open to using the existing router, but I got a little stuck. The problem I'm trying to solve is that I don't want to reload the data for the dashboards every time a user switches the view. If there's a straightforward way to do that without something like react-router, I'm happy to go that route. If you have some bright ideas and a few minutes for a call, I'd be glad to talk it through!

kevinrobinson: @edavidsonsawyer Yeah I get you, and that seems great!  I think doing what you're doing here is a fine immediate-term solution, and no need to block and solve that differently now.

Another immediate-term solution could be having the component here hold a bit of state for which page should be shown, and then pass down a callback to the different dashboard components that let them change to another dashboard.  But I think what you're doing here is a good solution for now - my vote would be to just keep moving and ship it while calling out it's a thing to refactor later.

edavidsonsawyer:
Hey, @alexsoble, @kevinrobinson or whoever else wants to review. This should be ready for a look!

alexsoble:
Reviewing this now with @edavidsonsawyer! 

Eric noticed some funkiness in the absences categories on the x-axis: 

screen shot 2017-12-15 at 12 27 10 pm

We also decided to merge in datepicker branch first and then rebase this on top of it.

edavidsonsawyer:
Alright, @alexsoble, this takes care of the absence dashboard bug and should be ready for review.

alexsoble:
@edavidsonsawyer 👍 👍  — reviewing now

alexsoble:
Hey there @edavidsonsawyer! I gave this a review in Internet Explorer, looks good! Let's merge, keeping in mind @kevinrobinson's longer-term notes about centralizing on one routing solution.

For future refactors/PRs in this area, I think it would be good to start moving these dashboard components over to the same React/ES6 style we're using for other components: using `import React from 'react'` and `class AwesomeClass extends React.Component`. When we add new components like SchoolTardiesDashboard, let's add them this way so that we don't have to refactor later down the line. 

For examples of the React ES6 module style I'm talking about, check out these PRs:

— https://github.com/studentinsights/studentinsights/pull/1346
— https://github.com/studentinsights/studentinsights/pull/1345
— https://github.com/studentinsights/studentinsights/pull/1344

Cheers and congrats on shipping this next dashboard @edavidsonsawyer! 🎉

alexsoble:
@edavidsonsawyer Looks good in prod! 🎉

PR #1306, opened 2017-12-05
Dashboard date picker

edavidsonsawyer:
# Who is this PR for?
Users of the school absence dashboard

There's no native datepicker for date fields in IE. That makes the date fields in the dashboard awkward and buggy.

# What does this PR do?
Adds the jquery datepicker used elsewhere in the app so users can select specific dates

datepicker

# Checklists

## Javascript QA

+ [X] Author checked latest in IE - Absence Dashboard
+ [X] Author checked latest in IE - Student Profile - record service Dashboard
+ [X] Author checked latest in IE - Student Profile - generate report
+ [X] Author checked latest in IE - Service Uploads

+ [x] Author checked latest in IE - Student Profile - record service Dashboard
+ [x] Author checked latest in IE - Student Profile - generate report
+ [x] Author checked latest in IE - Service Uploads

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/date_slider.jsx
+++ b/app/assets/javascripts/school_administrator_dashboard/date_slider.jsx
@@ -38,13 +38,11 @@ export default React.createClass({
 
     return (
       <div>
-          <div className='DashboardDatePicker'>
+        <div className='DashboardDatePicker'>
+          <div>
             <label>From date:</label>
             <Datepicker
               styles={{
-                datepicker: {
-                  padding: 5
-                },
                 input: {
                   fontSize: 12
                 }
@@ -54,16 +52,14 @@ export default React.createClass({
               datepickerOptions={{
                 showOn: 'both',
                 dateFormat: 'yy-mm-dd',
-                minDate: undefined
+                minDate: moment.unix(this.props.rangeStart).format("YYYY-MM-DD"),
+                maxDate: moment.unix(this.state.value[1]).format("YYYY-MM-DD")
               }} />
           </div>
-          <div >
+          <div>
             <label>To date:</label>
             <Datepicker
               styles={{
-                datepicker: {
-                  padding: 5
-                },
                 input: {
                   fontSize: 12
                 }
@@ -73,10 +69,12 @@ export default React.createClass({
               datepickerOptions={{
                 showOn: 'both',
                 dateFormat: 'yy-mm-dd',
-                minDate: undefined
+                minDate: moment.unix(this.state.value[0]).format("YYYY-MM-DD"),
+                maxDate: moment.unix(this.props.rangeEnd).format("YYYY-MM-DD")
               }} />
           </div>
-       <Range
+        </div>
+        <Range
           allowCross={false}
           min = {this.props.rangeStart}
           max = {this.props.rangeEnd}
@@ -85,7 +83,7 @@ export default React.createClass({
           onChange = {this.onSliderChange}
           tipFormatter={(value) => moment.unix(value).format("MM-DD-YYYY")}
           onAfterChange={this.props.setDate}/>
-      </div>
+        </div>
     );
   }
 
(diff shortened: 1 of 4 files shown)


edavidsonsawyer:
Hey @snoopyuri and @alexsoble! This adds a datepicker to the absence dashboard so a user can get to specific dates more easily. 

I didn't remove the date slider here. They both seem like useful features, and since they're synchronized I don't think this should be confusing for users. If you guys feel strongly that the functionality should be either datepicker or slider we can revisit, of course.

alexsoble:
@edavidsonsawyer Great! Could you add some screenshots to the PR description up top when you have a chance?

edavidsonsawyer:
@alexsoble sure thing! Fixed the travis error too...

alexsoble:
@edavidsonsawyer Looks like an improvement! One issue I see is that there's no validation stopping the "To" date from coming before the "From" date, which seems like a good edge case to handle. We might also want to use the same calendar icon we use for datepickers in the rest of the app, like in the student services UI. Last thing, we could left align the "from" and right align the "to." Let me know what you think about these ideas!

edavidsonsawyer:
All sound good, @alexsoble, and all are in now.

Validation on the to and from dates might be worth a double check here. The cleanest way to do it in the UI is to dynamically change the min/max dates on the datepicker, which requires tearing down and rebuilding the datepicker when the wrapper gets new props. 

I don't think this should be a problem, and I didn't find any issue when I tested the other components that use datepicker (I updated the PR to add those tests, though). In fact, it lets us do the same validation in other places we use the datepicker, which isn't done now as far as I can tell. Still, it touches a component that's used in a bunch of places so I wanted to call it out explicitly.

alexsoble:
@edavidsonsawyer This looks awesome visually Eric, great job here! And thanks for calling out the datepicker changes as meriting extra review and checking, I agree with that! One thing — you mentioned adding more tests in your comment, I don't see any JS test files changed in this PR. Are there maybe some test changes hanging around on your local machine that you forgot to push up?

alexsoble:
@edavidsonsawyer One thought on the datepicker change: could we do it non-destructively, like this: 82c5309bca9273253f1994169b519657f206f84f? (jQuery docs: http://api.jqueryui.com/datepicker/#method-option)

edavidsonsawyer:
Nice, @alexsoble! That's much better

edavidsonsawyer:
@alexsoble regarding the tests, I just meant the checklist in the PR description. I do want to add js tests for the dashboard, but I'd prefer to separate that from this PR since the dashboard needs quite a bit of coverage

alexsoble:
@edavidsonsawyer Oh I see! Yeah usually when I say tests I mean "automated tests," sorry for not being more explicit about that. In my brain I think of the checklist items as "manual checks or "QA checks."

alexsoble:
@edavidsonsawyer I checked the React.js documentation for `componentWillReceiveProps` (https://reactjs.org/docs/react-component.html#componentwillreceiveprops) because I was curious what it would say, noticed this:

```
Note that React may call this method even if the props have not changed, so make sure to compare the current and next values if you only want to handle changes. 
This may occur when the parent component causes your component to re-render.
```

Let's do what the documentation suggests and add that comparison. Take a look at dd2f380f244e78fa7d053261eed26d33699d0eee, I sketched out a few guard clauses before we change jQuery options in `componentWillReceiveProps`.

edavidsonsawyer:
Hey @alexsoble, I left some comments on your commit. If I'm missing something basic, just tell me to go do more homework.

alexsoble:
Just talked this through with @edavidsonsawyer, we decided to pass a special prop down that will tell the date picker whether it has to set min/max dynamically or not to avoid a bunch of extra calls to jQuery.

edavidsonsawyer:
Hey, @alexsoble, this is good to go. I added some dummy logs locally to confirm this does indeed prevent the datepicker from loading new options unless this prop is set.

alexsoble:
@edavidsonsawyer Woohoo, reviewing now!

alexsoble:
👍

Issue #1307, opened 2017-12-05
Fix bug: Handle data entry with invalid service end date


alexsoble:
Rollbar: https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/69/

Code:

```
       if service.save
>>>      if estimated_end_date.present? && estimated_end_date.to_time < Time.now
           service.update_attributes(:discontinued_at => estimated_end_date.to_time, :discontinued_by_educator_id => current_educator.id)
         end
         render json: serializer.serialize_service
       else
```

Thoughts: debug and then move this into the serializer object above.

alexsoble:
Tried to reproduce this locally. Nothing seems broken about service recording, was able to record a few fake services locally with the datepicker UI.

Then I tried entering an invalid date, and was able to reproduce the error: 

## UI (inputs invalid service)

![invalid-service-input](https://user-images.githubusercontent.com/3209501/33627458-3382f6d6-d9c3-11e7-87e2-01e964c0a9b8.gif)

## Server logs

![screen shot 2017-12-05 at 1 49 00 pm](https://user-images.githubusercontent.com/3209501/33627461-35ad5636-d9c3-11e7-91de-dfc0ed526e4a.png)

What was unexpected for me was that even though the UI shows "error" at the end and doesn't update the list of services, when the user refreshes the service appears in the list, just without any expected end date: 

## UI (result)

![screen shot 2017-12-05 at 1 50 36 pm](https://user-images.githubusercontent.com/3209501/33627562-845a57a2-d9c3-11e7-9678-4afb214c1f75.png)

So this is my working theory for what cause this bug right now — a user (educator) entering invalid service end date.

alexsoble:
Couple ideas here...

1. We should validate data entered into the service datetime inputs more strictly. 
2. We should push some of the code in students controller that's triggering this error down to the service model. It's ultimately the service model's job to determine is a service is valid or not, not the students controller.

alexsoble:
## Client-side validation asymmetry (GIF)

Looks like we have client-side validation that would catch this for `date_started`, but no such validation for `estimated_end_date`.

![validation-asymmetry](https://user-images.githubusercontent.com/3209501/33628763-66da7b90-d9c7-11e7-8eb8-ae1ae680b0d4.gif)

kevinrobinson:
[...]

Issue #152, opened 2016-03-02
Design - Add red star for required fields in Notes and Services UI


kevinrobinson:
From @johnrobbjr and Kemal's great feedback [here](https://docs.google.com/presentation/d/1t7bXVaFCDU_1d0vxxnqnQ7lMBnXa8oRyYRyMLtwx0hU/edit#slide=id.g11c1c42e02_3_0).

Add a red star to indicate this field is required, so users know how to activate the “save notes” button
<img width="573" alt="screen shot 2016-03-02 at 12 45 48 pm" src="https://cloud.githubusercontent.com/assets/1056957/13469285/b64357c2-e074-11e5-992e-9c32d7431d81.png">

kevinrobinson:
This form only has two elements and both are required, so I think this would be noise for 85% of users experiencing this.  It could help with first-time users, but not sure it's a worthwhile tradeoff or worth adding the visual or implementation complexity.

Perhaps an improvement would be to highlight the section that needs to be filled out when the user hovers over the "Save notes" button, to explain only at the point right before they try to save?

kevinrobinson:
stale

PR #1308, opened 2017-12-05
Show which input fields for recording a service are required in the UI

alexsoble:
# Who is this PR for?

+ Educators who record services using our UI and aren't sure which fields are optional and which are required!

# What does this PR do?

+ Adds bold red asterisks to the required fields and a note explaining what the asterisks mean. 

show-required-services

# Checklists

## Javascript QA

+ [x] Author checked latest in IE - Record Service area of Student Profile
+ [ ] Reviewer checked latest in IE - Record Service area of Student Profile

# Why adding it now?

+ Working on #1307, I started writing code to show users if the end dates they enter are valid or not. As I wrote that code, I realized that according to our server-side code, end dates aren't required, whereas start dates are. The client-side validation should reflect that. If the client-side validation reflects that, we should also have a clear, easy-to-find indications so the user doesn't have to enter invalid data to find out.

# GitHub issue

+ This implements one half of #152. The other half is doing the same thing for the Event Note side of the page. This has only one required input, "What are these notes from?", assuming we don't count the textarea itself as an input. 
+ We've added more fields during the past year, which make this comment (totally valid at the time!) a little dated now: https://github.com/studentinsights/studentinsights/issues/152#issuecomment-191351682

### DIFF
--- a/app/assets/javascripts/student_profile/record_service.jsx
+++ b/app/assets/javascripts/student_profile/record_service.jsx
@@ -137,7 +137,7 @@ import {merge} from '../helpers/react_helpers.jsx';
       return (
         <div>
           <div style={{ marginBottom: 5, display: 'inline' }}>
-            Which service?
+            Which service? <span style={{color: 'red', fontWeight: 'bold'}}>*</span>
           </div>
           <div style={{ display: 'flex', justifyContent: 'flex-start' }}>
             <div style={styles.buttonWidth}>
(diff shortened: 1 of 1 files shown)


kevinrobinson:
@alexsoble This adds red noise to the page that's there all the time, to handle the less common case when someone doesn't add a required field.  Could we invert it, so that if someone clicks on the save button when disable, then the red asterisks appear?

alexsoble:
Sounds like a good idea!

alexsoble:
Closing this for now because we have more immediate opportunities to improve the experience for users. 😎

PR #1315, opened 2017-12-10
Use TestPals in development, add admin/authorization overview

kevinrobinson:
❗️ **This updates the local development process - see the updated README for how to login in development mode** ❗️ 

# Who is this PR for?
Developers and project admin

# What problem does this PR fix?
This updates the local development data to use `TestPals` so that the development data allows manually testing the set of educator roles that are also tested in automated tests.  These should have semantically correct relationships so that authorization rules all make sense.

It also adds an `/admin/authorization` page to audit which users have sensitive access, and to see the homeroom pages for each educator.

# What does this PR do?
It reworks `seeds.rb` to merge it into `TestPals`.

It moves the homepage code out of `ApplicationController` and into `Authorizer` for determining what kind of homepage to show, and `PathsForEducator` for computing the actual path for standalone testing.  In the process, this removes some unnecessary code and pages that were handling cases when the database had no `Section` or `Homeroom` models, and collapses that into a `no_default_page` view.

It adds `/admin/authorization`.

# Screenshot (if adding a client-side feature)
Seeding:
```

Starting seeds.rb...
Checking for existing data in the database...
Seeding database constants for Somerville...
Destroying all educators...
Creating TestPals...
Uploading services...
Creating more students for each homeroom...
  Creating for homeroom: HEA 500...
  Creating for homeroom: HEA 003...
  Creating for homeroom: WSNS 501...
  Creating for homeroom: SHS ALL...
  Creating for homeroom: SHS 917...
  Creating for homeroom: SHS 942...
Creating additional students for the Healey with no homeroom...
Creating more sophomore students...
  Creating students for section SECTION-1...
  Creating students for section SECTION-2...
  Creating students for section ART-302A...
  Creating students for section ART-302B...
  Creating students for section SCI-201A...
  Creating students for section SCI-201B...
Updating indexes...
Querying for jobs...
Found 5 jobs.
Starting precomputing...
Wrote document 1...
Wrote document 2...
Wrote document 3...
Wrote document 4...
Wrote document 5...
Done.
Done seeds.rb.
```

Authorization overview admin page (showing new development data created from `TestPals`):
screen shot 2017-12-10 at 1 53 23 pm
screen shot 2017-12-10 at 2 02 26 pm

PR #1318, opened 2017-12-12
[Blocked] Update `students_controller` to use `authorized`

kevinrobinson:
❗️  This work is on top of https://github.com/studentinsights/studentinsights/pull/1315, which needs to be merged first. ❗️ 

# Who is this PR for?
Developers, students

It uses the `authorized` API to simplify authorization code in `students_controller`.

# What does this PR do?
It updates `student_controller` to use `authorized` to control access to student, event_notes, and section models.  In the process, it adds in support to `authorized` for passing `Section` models, refactors code and tests as needed.  The controller also calls `SearchbarHelper`, so this PR changes it from a helper module to a standalone class, moves it into `lib`, updated to use `authorized` and adds a test.  This also moves the `lasids` endpoint in the `students_controller` into the `service_uploads` controller since it's part of that feature.  With the change to catch `EducatorNotAuthorized` exceptions in `application_controller`, the `IepDocument` tests started failing and so that controller is updated to use the new `authorized` API as well.

This is mostly broken up into commits, and each commit has a message describing what's happening, so a good way to review might be to go through them incrementally.

This doesn't yet rework all of `students_controller_spec` tests to use TestPals, it just does this for tests that had to be updated as part of this work.  That can be a separate step in https://github.com/studentinsights/studentinsights/issues/1080#issuecomment-350572396.


app/controllers/students_controller.rb
```diff
@@ -96,9 +78,9 @@ def service
       :provided_by_educator_name,
       :estimated_end_date
     ])
+    authorized_or_raise! { Student.find(clean_params[:student_id]) }
```
kevinrobinson: This is a real bug - it adds a check that the educator is authorized to read/write data for this student.

app/controllers/students_controller.rb
```diff
@@ -169,34 +144,16 @@ def student_feed(student, restricted_notes: false)
     }
   end
 
-  def school_year_id_range(from_date, to_date)
```
kevinrobinson: unused it seems

kevinrobinson:
Also, this finishes off all the work that was in the branch https://github.com/studentinsights/studentinsights/compare/refactor/authorization.

app/controllers/iep_documents_controller.rb
```diff
@@ -1,37 +1,27 @@
...
-    @bucket_name = ENV['AWS_S3_IEP_BUCKET']
-
-    object = s3.get_object({
-      bucket: @bucket_name,
-      key: @file_name
-    })
-
-    send_data object.body.read, filename: @file_name, type: :pdf
+    iep_document = IepDocument.find_by_id!(params[:id])
+    authorized_or_raise! { iep_document.student }
```
kevinrobinson: This also unrolls the control flow so it's explicit, rather than using `before_action` and split across several methods sharing instance variables.

app/controllers/service_uploads_controller.rb
```diff
@@ -1,12 +1,6 @@
...
-  # Authentication by default inherited from ApplicationController.
-
-  before_action :authorize_for_districtwide_access_admin # Extra authentication layer
-
-  def authorize_for_districtwide_access_admin
-    unless current_educator.admin? && current_educator.districtwide_access?
-      render json: { error: "You don't have the correct authorization." }
-    end
+  before_action do
+    raise Exceptions::EducatorNotAuthorized unless authorizer.can_upload_bulk_services?(current_educator)
```
kevinrobinson: Moving the definition of the "can the educator do X" into `Authorizer`.

app/controllers/students_controller.rb
```diff
@@ -1,36 +1,17 @@
...
-  def authorize!
-    student = Student.find(params[:id])
-    raise Exceptions::EducatorNotAuthorized unless current_educator.is_authorized_for_student(student)
-  end
-
-  def authorize_for_districtwide_access_admin
-    unless current_educator.admin? && current_educator.districtwide_access?
-      render json: { error: "You don't have the correct authorization." }
-    end
-  end
```
kevinrobinson: So this is one of the simplification benefits - this code gets replaced with wrapping `authorized` around data being loaded in the controller methods.

spec/controllers/service_uploads_controller_spec.rb
```diff
@@ -152,7 +153,7 @@ def make_request
         let(:educator) { FactoryGirl.create(:educator) }
         it 'cannot access the page; gets redirected' do
           make_request
-          expect(JSON.parse(response.body)).to eq({ "error" => "You don't have the correct authorization." })
+          expect(response).to redirect_to(not_authorized_path)
```
kevinrobinson: This endpoint returns HTML, so it should have been redirecting, not returning JSON.

spec/controllers/students_controller_spec.rb
```diff
@@ -279,7 +278,51 @@ def make_post_request(student, service_params = {})
...
+    context 'enforces authorization rules' do
+      let(:students) do
+        [
+          pals.healey_kindergarten_student,
+          pals.healey_meredith_student,
+          pals.shs_freshman_mari
+        ]
+      end
+
+      it { expect(can_post_service_for(pals.uri, students)).to eq students }
```
kevinrobinson: Adding the method to try to make this succinct across a bunch of test cases.

spec/controllers/students_controller_spec.rb
```diff
@@ -513,94 +558,58 @@ def make_request(student)
...
+          sign_in(educator)
+          make_request(student.id)
           expect(assigns(:services)).not_to include(old_service)
           expect(assigns(:services)).to include(recent_service)
         end
 
         it 'assigns the student\'s notes correctly excluding restricted notes' do
           restricted_note = FactoryGirl.create(:event_note, :restricted, student: student, educator: educator)
           note = FactoryGirl.create(:event_note, student: student, educator: educator)
+          sign_in(educator)
```
kevinrobinson: This was failing and not sure how it ever worked - the `before` was making the request, but then inside the `it` it was adding the models needed for the test.  This makes the order explicit.

spec/lib/searchbar_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      expect(Searchbar.names_for(pals.healey_vivian_teacher)).to eq [garfield]
+      expect(Searchbar.names_for(pals.healey_ell_teacher)).to eq []
+      expect(Searchbar.names_for(pals.healey_sped_teacher)).to eq []
+      expect(Searchbar.names_for(pals.healey_laura_principal)).to eq [garfield, meredith]
+      expect(Searchbar.names_for(pals.healey_sarah_teacher)).to eq [meredith]
+      expect(Searchbar.names_for(pals.west_marcus_teacher)).to eq []
+      expect(Searchbar.names_for(pals.shs_jodi)).to eq [mari]
+      expect(Searchbar.names_for(pals.shs_bill_nye)).to eq [mari]
+      expect(Searchbar.names_for(pals.shs_hugo_art_teacher)).to eq []
+      expect(Searchbar.names_for(pals.shs_fatima_science_teacher)).to eq []
```
kevinrobinson: This is essentially duplicating the tests of authorization code, but this seems good to do if this is simple and direct to express - here are the names that should come up in the searchbar for each of the educators in `TestPals`.

spec/support/test_pals.rb
```diff
@@ -0,0 +1,252 @@
...
+      school: @shs,
+      homeroom: @shs_jodi_homeroom,
+      grade: '9'
+    })
+    FactoryGirl.create(:student_section_assignment, student: @shs_freshman_mari, section: @shs_tuesday_biology_section)
+
+    self
+  end
+
+  def index!
```
kevinrobinson: This came up in the `sign_in` test, that it was relying on `student_risk_level` and that wasn't being created here, so this adds it.

kevinrobinson:
hmm, merging introduced a few test bugs, need to fix those up

kevinrobinson:
Going to pause this.  This has a lot of changes, and so I started trying to break it apart.  But several of these touch endpoints where we'd like to be able to do better measurement of the impact on response time.  We'd benefit from first having better visibility into this as described in https://github.com/studentinsights/studentinsights/issues/1392.

kevinrobinson:
Closing.  This got stale, and we're not going to ship in now, and the branch isn't going anywhere.  There's still a lot we can and should do to improve the root issues of 1) moving authorization code over to using `Authorizer` and 2) making it safer for developers to write code querying for things like `EventNotes`, particularly when querying through associations.

But the particular steps will probably need to be more targeted, and are interlocked with other feature work (eg, removing homerooms and using sections instead).  All of this is probably better tackled piecewise as we do the related feature work.  We've been able to do that with item 1 above, using `Authorizer` but not with item 2, where in recent new feature development we almost always want to be able to query through associations.  The draft API here doesn't support that, and so a specific change to make querying for `event_notes` safer by default is the next higher impact change we could make to make this kind of code safer to write.

PR #1319, opened 2017-12-13
Seed data for New Bedford school names and school codes

alexsoble:
# Who is this PR for?

+ New Bedford Public Schools! 

# What does this PR do?

+ Adds list of school names and local IDs ("codes") from New Bedford Public Schools. We need this seed data to start importing other tables like students which have associations with schools.
+ Use bang version of `create` because we want this to blow up if there are any errors.


alexsoble:
gonna selfie this

PR #1320, opened 2017-12-13
Remove ELEM shortcode option for import process

alexsoble:
# Why?

+ It adds complexity that makes `import.thor` harder to read, but is rarely if ever used.
  + In my next couple of PRs, I'll be changing `import.thor` to clear a path for importing New Bedford data and this will be a stumbling block. The ELEM code creates another set of configuration for each new district we add. 
  + I can't remember the last time I used the ELEM shortcut for Somerville, if ever.
  + It doesn't look like it would even work: the `school_local_ids` method that calls is is only used in `#validate_schools`, but not in the code that runs the import itself.

### DIFF
--- a/lib/tasks/import.thor
+++ b/lib/tasks/import.thor
@@ -21,10 +21,6 @@ class Import
       'SHS',
     ]
 
-    SCHOOL_SHORTCODE_EXPANSIONS = {
-      "ELEM" => %w[BRN HEA KDY AFAS ESCS WSNS WHCS]
-    }
-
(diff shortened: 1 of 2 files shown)


alexsoble:
QA'ed this locally under the 3 main conditions I could think of:

+ No `--school` argument passed
+ `--school` argument passed with one school
+ `--school` argument passed with multiple schools

Responded correctly under all three conditions! 👌 

Going to selfie this one since it's changing an option that's not currently in use.

PR #1321, opened 2017-12-13
Remove Somerville hard-coding from nightly import code; move import code from Thor DSL to Ruby

alexsoble:
# Who is this PR for?

Developers who want to set up Student Insights instances for other school districts. 

Right now the nightly data import process is hard-coded to seed Somerville school data if it's missing and validate that the school filtering options passed in by developers are valid.

# What does this PR do?

This PR uses the new `LoadDistrictConfig` API for both of the nightly data import features described above: seeding missing school data and validating school filtering options.

It also moves 90% of the code governing nightly data import out of the Thor DSL and into a pure Ruby class. This was a refactor @kevinrobinson and I discussed a while back. It makes it easy for a developer to kick off an import process from the console, and to change the import code without having to wrap their mind around a new DSL. 

# Checklist 

+ [ ] Update README
+ [x] Verify that the import process succeeds for Somerville:

screen shot 2017-12-28 at 7 22 18 pm


alexsoble:
Heyyyyy, this is now working locally for both Somerville and New Bedford imports! 🌈 ⭐️ 💫 

Of course the import tests are all failing, but that shouldn't be too hard to clean up especially since we have more plain Ruby now and less Thor.

alexsoble:
### TODO (ARS)

+ [ ] Restore spec coverage to at least where it was before
+ [ ] Update README
+ [ ] Add New Bedford filenames to `district_new_bedford.yml`
+ [ ] Verify import for both districts

alexsoble:
Woohoo! Specs are passing, and test coverage is on par with `master`. 🎉 

In hindsight, this PR is more "refactor" than "drive forward second district deployment", although it does unblock one obstacle to second district deploy.

I went into the PR with the idea that we'd add a `--district` flag to the nightly import process, to let developers specify which district to import for. But on reflection, I realized that creates a second source of truth for answering the question, "What school district is this?" The only source of truth for that question should be `ENV.fetch('DISTRICT_KEY')`.

So the main change that unblocks a second deployment is around school seeding / school validation. On `master`, this is hard-coded to seed Somerville school data if it's missing and validate that the school filtering options passed in by developers are valid. On this branch, I use the new `LoadDistrictConfig` API for both of those tasks. 

At the same time, I moved 90% of the code governing nightly data import out of the Thor DSL and into a pure Ruby class. This was a refactor @kevinrobinson and I discussed a while back. It makes it easy for a developer to kick off an import process from the console, and to change the import code without having to wrap their mind around a new DSL. 

In terms of a deploy path, I'd like to verify that the import process for Somerville succeeds on this branch, then merge and deploy in the spirit of merge-early-and-often. I see at least two separate PRs on the horizon before the data import process will work for New Bedford: one that adds `remote_filenames` to the New Bedford YAML config, and one that fixes #1325. So I'd rather get this merged in and working for Somerville than wait til it unblocks both districts.

alexsoble:
Re-verified that data import for Somerville works good after the latest round of changes (spec fixes, refactors, etc). 😎

alexsoble:
Did a little more manual verifying that passing in specific importer options works as intended, since that was probably the part of the import process that changed the most . . .

```
» rake db:migrate:reset && thor import:start --source students
```

screen shot 2017-12-29 at 11 21 14 am

alexsoble:
🚢

PR #1322, opened 2017-12-13
Extract remote filenames into YAML configuration

alexsoble:
# Who is this PR for?

+ Districts that want to adopt Student Insights!

+ Districts are going to choose different names for their export files based on their own naming conventions. This is something I didn't anticipate but in hindsight it makes sense! 
+ We should make each Insights instance flexible so that districts can choose any export filename they want, instead of having to conform to our standard.

# What does this PR do?

+ Creates a `district-config.yml` file that stores high-level config values for each district
  + Extract file names of remote files from importer code into `district-config.yml` 
  + Extract info on schools (school name, school local ID) into `district-config.yml` 

# Checklist 

+ [x] Test data import process locally before deploying

screen shot 2017-12-26 at 10 19 50 am

+ [x] Verify school seeding process by re-creating a demo site on this branch 

screen shot 2017-12-26 at 10 14 57 am


alexsoble:
@kevinrobinson — Any comments or feedback on this strategy before I go ahead and set a bunch of ENV vars on the Somerville instance?

app/importers/file_importers/attendance_importer.rb
```diff
@@ -12,7 +12,7 @@ def import
   end
 
   def remote_file_name
-    'attendance_export.txt'
+    ENV.fetch('FILENAME_FOR_ATTENDANCE_IMPORT')
```
kevinrobinson: So one thing I think we'll want is "what do I need to change per-district"?  This makes this hard to do - you have to look through all the env variables.

I'd suggest pulling all this config out to a standalone class in lib, something like `DistrictConfig` that has various methods for these things?  That class can then pull those values from environment variables, from database tables, just hard code them in Ruby based on the deployment, whatever.  And we have all that written in one place and the app code doesn't need to change.

The other thing it seems like it might be useful for these kinds of things is an environment variable that's a single key saying "what Student Insights instance am i?" and is set to `somerville` or `new-bedford` or whatever.

alexsoble: @kevinrobinson Really good feedback and ideas here! Right now we have `ENV['DISTRICT_NAME]` which is used in the login page and the "create new Heroku instance for new district" script.

alexsoble: @kevinrobinson Maybe we want `ENV['DISTRICT_NAME]` for presentation and `ENV['DISTRICT_KEY]` for use in code.

kevinrobinson: 👍 Yeah I think the key is the most important (eg, `DISTRICT_KEY`).  And then there are a pile of other things that branch and are different based on that - some right now are just written in code specific to Somerville (like this PR), others are in ENV variables, etc.

alexsoble:
@kevinrobinson when you've got the chance, I have new code to review here!

I pulled all the config into a `DistrictConfig` class like you suggested, and added a bunch of documentation for devs around how and what to set in Heroku to make this work for a new district.

Downside to the way I set this up is that I made it difficult to inject `ENV` and test `DistrictConfig`, since it relies on assigning values a bunch of constants on app initialization. Pulling the values from the database instead of ENV would be easier to test but might be overkill at this stage (just two districts). What do you think — any suggestions for refactoring `DistrictConfig` to make it more easily testable?

alexsoble:
OK, did some more work on this. Check out #1323, refactors `DistrictConfig` to allow injecting ENV; adds a test.

alexsoble:
Hey @kevinrobinson, give this one another look whenever you have the chance? 

I changed two main things:

+ Moved from using Heroku ENV to using YML (`district-config.yml`)
+ Pulled school data into the district config as well because I saw that storing that in code created some pain when I started working on import process refactor in #1321

alexsoble:
One idea I had as I was working on this: we want each district's config to be held in a separate YML file. That kind of restriction blocks developers from making mistakes where they load the wrong district's data somehow. Trying to get the best parts of using ENV (strict per-instance segregation) and the best parts of using YML (nesting, legibility).

alexsoble:
TK: Add the code comments back in, fix up specs

alexsoble:
Once the specs are green again, next TODO is updating README.

app/config_objects/district_config.rb
```diff
@@ -0,0 +1,69 @@
+class DistrictConfig
+
+  cattr_accessor :remote_filenames
```
kevinrobinson: Here's a style suggestion, that you might be into or not.  No need to hold up shipping this, and up to you whether this seems worthwhile or not.

This is written with only class methods, which makes it hard to see what's a public method and what's not. 
 It also keeps state, so the order you call these in matters, but the method names describe what they do, not how they should be used.

One option might be to write this as a plain class, and then expose only one class method `remote_filenames`. 
 That would read the yaml file and everything is stateless, or there could be a second class method like `load!` or something that has to be called in the Rails initializer and `remote_filenames` raises if not.

app/config_objects/district_config.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  # To configure a Student Insights instance to read data from remote SFTP,
+  # we need to know what filenames to look for.
+
+  # Set filenames for the expected remote files in a .yml file under /config.
+  # See config/district_somerville.yml as an example.
+
+  # The file name should reference an ENV['DISTRICT_KEY'] value, like this:
+  #
+  # => "config/district_#{ENV['DISTRICT_KEY']}.yml"
```
kevinrobinson: I'd suggest making this mapping manual and explicit like `{ somerville: district_somerville.yml }`.  There's no benefit to reducing duplication here, and it avoids using variables to form paths on the filesystem.

config/district_somerville.yml
```diff
@@ -0,0 +1,74 @@
+config:
+  remote_filenames:
+    FILENAME_FOR_STUDENTS_IMPORT: >-
+      students_export.txt
```
kevinrobinson: What does the `>-` do here?  I was expecting a plain

```
FILENAME_FOR_STUDENTS_IMPORT: students_export.txt
```

Is it about the spaces or quotes?

config/initializers/district_config.rb
```diff
@@ -0,0 +1 @@
+DistrictConfig.set_remote_filenames
```
kevinrobinson: This seems like it should be more generic and describe why it's doing this, maybe `load!` or something.

alexsoble: Good suggestion!

app/config_objects/district_config.rb
alexsoble: Good suggestion, didn't realize using variables to form paths on the filesystem was a bad practice but now that I think about it that makes sense.

config/district_somerville.yml
alexsoble: This is some YML syntax that lets us put the value on a newline instead of directly after the key, I did that because some of the values are reaaaaally long and much easier to read if they're on their own line.

app/config_objects/district_config.rb
alexsoble: I like the stateless idea, that reduces some complexity because it lets us get out of the whole initialization game. Let me work on that and add some more commits to this PR.

alexsoble:
@kevinrobinson Made `DistrictConfig` stateless yesterday and added a bunch of documentation to the README. 😎

kevinrobinson:
👍 nice documentation too!

alexsoble:
@kevinrobinson 😎

alexsoble:
Re ran the QA steps and added little screenshots to the issue, time to 🚢 !

Issue #1257, opened 2017-11-13
Test logic in Student Report PDF 


alexsoble:

alexsoble:
Hey @gabivoicu! Carrying our conversation from #1284 over here. 

Alright, so I did some digging through the history of this issue and the overall state of our Student Report testing to crystallize this issue down into three specific improvements:

1. [ ] **Review `try` statements in `student_report.pdf.erb` and remove `try` statements that don't fit with our model logic.** The original bug that motivated this issue is described here: https://github.com/studentinsights/studentinsights/pull/1165#issuecomment-343225723. student_report.pdf.erb is too generous with `try`, which can lead to bugs like the one linked to. We need to review every `try` statement and ask, "is this responding to a piece of data that could be vaild if nil? Or should the data sent down from the server always be valid, in which case we should remove  `try` because we want errors and developer alerts when we have invalid data to address?" Another related improvement is to think about ways we can explicitly handle nil values instead of letting them disappear silently, for example, by printing out "N/A" or "—" on the Student Report so we can make it more explicit in reporting nil values.

2. [ ] **Test more instance variable assignments in `students_controller_spec.rb`.** The existing specs only cover services and event notes, we should add specs for `@student_school_years`, 
`@discipline_incidents`, and `@student_assessments`, all of which have some complex logic going on in the controller code.

3. [ ] **student_profile_pdf.test.js?** I searched through the specs and didn't find any specs for the Student Report JS. This would be great to have even if it's only a "smoke test". Because student_profile_pdf.js is loaded into the app a little differently from the rest of the JS (as we've seen!), it may require a bit more work to test. As a first step, we could pull JS logic out of this file and into helper files that we can test.

@gabivoicu Let me know what you think! I tried to arrange these three in order from most straightforward to most complicated. Happy to jump on a videocall as well. You've already gained a lot of context on the Student Report from your last issue so I'm thinking you should be in a strong position to tackle this one!

gabivoicu:
@alexsoble I'll look into this thoroughly on Wednesday evening, see how much I can figure out on my own and then come back with questions. Thanks for finding another ticket for me. Yay, React 🎉

gabivoicu:
@alexsoble An update on my end, I did not have time to work on this tonight but I am planning to tomorrow night.

gabivoicu:
@alexsoble Started working on this, for the first two parts I believe I can figure out everything I need to complete, but the JS tests part I might need your help, I will let you know.

[...]

PR #1309, opened 2017-12-07
School overview page to es6

Jngai:
# Who is this PR for?
Everybody working on this code base
# What problem does this PR fix?
Continued on with next section of js refactoring.
# What does this PR do?

- Switch school_overview_page.jsx to SchoolOverViewPage.js. Rewrite it into a js module then import it everywhere it happens. Remove window.shared code.
- Refactored its associated test

# Screenshot (if adding a client-side feature)

# Checklists

## Javascript QA

*If the PR touches Javascript, follow this checklist. If the PR doesn't touch Javascript, you can delete this section.*

*Which features or pages does this PR touch? Make a list here so that you can double-check each affected page in a Internet Explorer before merging.*

/schools/hea

*Make one list for the author and another for the reviewer.*

+ [x] Author - Do a rigorous code review. 
+ [x] Author - Know where the js is getting loaded, click around in IE - School Overview 
+ [x] Reviewer - Do a rigorous code review. 
+ [x] Reviewer checked latest in IE - School Overview

PR #1340, opened 2017-12-22
Test logic in Student Report PDF #1257

gabivoicu:
[Issue 1257](https://github.com/studentinsights/studentinsights/issues/1257)
# Who is this PR for?
Educators using student reporting
Improve test coverage and improve code in student report
# What does this PR do?
- add a default value to the places where nil is a valid value
- remove all try that should not be there(we expect the data to be valid)
- add more tests for student report: student assessments, discipline incidents, school years

# Checklists

## Automated Testing

*Does the PR touch an area of the codebase that needs more specs? More examples within the specs? More scenarios covered by specs?*

### DIFF
--- a/app/views/students/student_report.pdf.erb
+++ b/app/views/students/student_report.pdf.erb
@@ -66,8 +66,8 @@
         <ul>
           <li>DOB: <%= format_date(@student.date_of_birth) %></li>
           <li><%= "Grade #{@student.grade}" %></li>
-          <li><%= @student.try(:school).try(:name) %></li>
-          <li><%= @student.try(:homeroom).try(:name) %></li>
+          <li><%= @student.school.name || "N/A" %></li>
+          <li><%= @student.homeroom.name || "N/A" %></li>
           <li><%= @student.program_assigned %></li>
         </ul>
       </div>
@@ -95,7 +95,7 @@
           <div style="float: left; width:370px; padding:5px">
             <ul>
               <li><span style="font-weight:bold;"><%= format_date(event_note.recorded_at) %></span></li>
-              <li><span style="font-style:italic;"><%= event_note.try(:event_note_type).try(:name) %></span></li>
+              <li><span style="font-style:italic;"><%= event_note.event_note_type.name || "N/A" %></span></li>
               <li><%= event_note.text %></li>
             </ul>
           </div>
@@ -124,13 +124,13 @@
                 <span style="font-weight:bold;">
                   <%= format_date(service.date_started) %> -
                   <% if service.discontinued_at.present? %>
-                    <%= "#{format_date(service.try(:discontinued_at))}" %>
+                    <%= "#{format_date(service.discontinued_at)}" %>
                   <% else %>
                     Present
                   <% end %>
                 </span>
               </li>
-              <li><span style="font-style:italic;"><%= service.try(:service_type).try(:name) %></span></li>
+              <li><span style="font-style:italic;"><%= service.service_type.name || "N/A" %></span></li>
               <li><%= service.provided_by_educator_name %></li>
             </ul>
           </div>
--- a/app/views/students/student_report.pdf.erb
+++ b/app/views/students/student_report.pdf.erb
@@ -66,7 +66,7 @@
         <ul>
           <li>DOB: <%= format_date(@student.date_of_birth) %></li>
           <li><%= "Grade #{@student.grade}" %></li>
-          <li><%= @student.school.name || "N/A" %></li>
+          <li><%= @student.school.try(:name) || "No School Assigned" %></li>
           <li><%= @student.homeroom.name || "N/A" %></li>
           <li><%= @student.program_assigned %></li>
         </ul>
@@ -95,7 +95,7 @@
           <div style="float: left; width:370px; padding:5px">
             <ul>
               <li><span style="font-weight:bold;"><%= format_date(event_note.recorded_at) %></span></li>
-              <li><span style="font-style:italic;"><%= event_note.event_note_type.name || "N/A" %></span></li>
+              <li><span style="font-style:italic;"><%= event_note.event_note_type.name %></span></li>
               <li><%= event_note.text %></li>
             </ul>
           </div>
@@ -130,7 +130,7 @@
                   <% end %>
                 </span>
               </li>
-              <li><span style="font-style:italic;"><%= service.service_type.name || "N/A" %></span></li>
+              <li><span style="font-style:italic;"><%= service.service_type.name %></span></li>
               <li><%= service.provided_by_educator_name %></li>
             </ul>
           </div>
--- a/app/views/students/student_report.pdf.erb
+++ b/app/views/students/student_report.pdf.erb
@@ -67,7 +67,7 @@
           <li>DOB: <%= format_date(@student.date_of_birth) %></li>
           <li><%= "Grade #{@student.grade}" %></li>
           <li><%= @student.school.try(:name) || "No School Assigned" %></li>
-          <li><%= @student.homeroom.name || "N/A" %></li>
+          <li><%= @student.homeroom.try(:name) || "No Homeroom Assigned" %></li>
           <li><%= @student.program_assigned %></li>
         </ul>
       </div>
--- a/spec/controllers/students_controller_spec.rb
+++ b/spec/controllers/students_controller_spec.rb
@@ -606,8 +606,16 @@ describe StudentsController, :type => :controller do
         end
 
         it 'assigns the student\'s assesments correctly' do
-          assesment = FactoryGirl.create(:student_assessment, student: student, date_taken: '2016-08-16')
-          expect(assigns(:student_assessments)).to include(assesment)
+          assessment = FactoryGirl.create(:assessment, :access)
+          student_assessment = FactoryGirl.create(:access, student: student, assessment: assessment, date_taken: '2016-08-16')
+          assessment = FactoryGirl.create(:assessment, :math)
+          student_assessment = FactoryGirl.create(:star_math_assessment, student: student, assessment: assessment, date_taken: '2016-08-16')
+          make_request({ student_id: student.id, format: :pdf, from_date: '08/15/2015', to_date: '03/16/2017' })
+
+          expect(assigns(:student_assessments)).to include("ACCESS Composite")
+          expect(assigns(:student_assessments)["ACCESS Composite"]).to be_kind_of(Array)
+          expect(assigns(:student_assessments)).to include(" Mathematics")
+          expect(assigns(:student_assessments)[" Mathematics"]).to be_kind_of(Array)
         end
       end
     end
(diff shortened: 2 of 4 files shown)


gabivoicu:
@alexsoble I'm still working on this, and wanted to run 2 things by you:
1. I've removed `nil` for all fields that are connected to the object via `belongs_to` because we expect that field to always be there, right? Locally I'm missing `homeroom` on 10 students so I wanted to double check.
2. I didn't add an `N/A` for `"#{format_date(service.discontinued_at)}"` because `format_date` returns an empty string if the argument passed in does not respond to `strftime`. That seems a little opaque to me, but I don't have another idea at the moment.

alexsoble:
@gabivoicu Initial PR looks great, let me leave some line-by-line comments!

app/views/students/student_report.pdf.erb
```diff
@@ -124,13 +124,13 @@
                 <span style="font-weight:bold;">
                   <%= format_date(service.date_started) %> -
                   <% if service.discontinued_at.present? %>
-                    <%= "#{format_date(service.try(:discontinued_at))}" %>
+                    <%= "#{format_date(service.discontinued_at)}" %>
```
alexsoble: This is a 👍  change, there's already a `present?` call up top so `try` here is duplicative.

spec/controllers/students_controller_spec.rb
```diff
@@ -604,6 +604,11 @@ def make_request(options = { student_id: nil, format: :pdf, from_date: '08/15/20
           expect(assigns(:event_notes)).to include(note)
           expect(assigns(:event_notes)).not_to include(restricted_note)
         end
+
+        it 'assigns the student\'s assesments correctly' do
```
alexsoble: This is a good start! There's a bunch of logic that goes into the assignment of the `@student_assessments` variable:

```
    @student_assessments = student_assessments_by_date.each_with_object({}) do |student_assessment, hash|
      test = student_assessment.assessment
      test_name = case test.family
        when "STAR" then "#{test.family} #{test.subject} Percentile"
        else "#{test.family} #{test.subject}"
      end

      hash[test_name] ||= []

      result = case test.family
        when "MCAS" then student_assessment.scale_score
        when "Next Gen MCAS" then student_assessment.scale_score
        when "STAR" then student_assessment.percentile_rank
        when "DIBELS" then student_assessment.performance_level
        else student_assessment.scale_score
      end

      hash[test_name].push([student_assessment.date_taken,result])
    end.sort.to_h
```

Would be good to have test code that exercises more of this logic.

app/views/students/student_report.pdf.erb
```diff
@@ -124,13 +124,13 @@
...
                   <% else %>
                     Present
                   <% end %>
                 </span>
               </li>
-              <li><span style="font-style:italic;"><%= service.try(:service_type).try(:name) %></span></li>
+              <li><span style="font-style:italic;"><%= service.service_type.name || "N/A" %></span></li>
```
alexsoble: Every service should have a service type, otherwise we wouldn't know what kind of service it is and the database record would be mostly meaningless. And every service type should have a name. So in this case I think removing the `try` statements is a very good move, but we don't need to add "N/A" as a fallback because we should never have a service without a service type or a service type without a name.

app/views/students/student_report.pdf.erb
```diff
@@ -95,7 +95,7 @@
           <div style="float: left; width:370px; padding:5px">
             <ul>
               <li><span style="font-weight:bold;"><%= format_date(event_note.recorded_at) %></span></li>
-              <li><span style="font-style:italic;"><%= event_note.try(:event_note_type).try(:name) %></span></li>
+              <li><span style="font-style:italic;"><%= event_note.event_note_type.name || "N/A" %></span></li>
```
alexsoble: Same as https://github.com/studentinsights/studentinsights/pull/1340/files#r158587198.

app/views/students/student_report.pdf.erb
```diff
@@ -66,8 +66,8 @@
         <ul>
           <li>DOB: <%= format_date(@student.date_of_birth) %></li>
           <li><%= "Grade #{@student.grade}" %></li>
-          <li><%= @student.try(:school).try(:name) %></li>
-          <li><%= @student.try(:homeroom).try(:name) %></li>
+          <li><%= @student.school.name || "N/A" %></li>
```
alexsoble: If the student isn't assigned to a school, this will blow up with:

```
NoMethodError: undefined method `name' for nil:NilClass
```

We still want to be able to generate the report if the student isn't assigned to a school, so I think some sort of `try` here is useful. Instead of "N/A" maybe we want "No School Assigned" since there's no other way for the reader to tell what attribute of the student is "N/A".

gabivoicu: That makes sense, should I make that change for homeroom as well?

alexsoble: @gabivoicu Yep that sounds right to me!

spec/controllers/students_controller_spec.rb
gabivoicu: @alexsoble What your thoughts on this version of the test, is there more I could be testing?

Another thing I wanted to mention is line `613`, I had to make the call again to be able to see these assessments for the student, and the test above (starting on line `601`) didn't need to call it again so that is puzzling to me.

alexsoble: @gabivoicu I think you've made a good addition, the test covers more surface area now! 

I agree it's puzzling, you shouldn't have to call `make_request` again. One hypothesis is that some of the database relationships between the assessments and student assessments you've created may need a `reload` call to work properly. Take a look at this code sample from https://github.com/rspec/rspec-rails under "Model specs":

```
RSpec.describe Post, :type => :model do
  context "with 2 or more comments" do
    it "orders them in reverse chronologically" do
      post = Post.create!
      comment1 = post.comments.create!(:body => "first comment")
      comment2 = post.comments.create!(:body => "second comment")
      expect(post.reload.comments).to eq([comment2, comment1])
    end
  end
end
```

They use reload to make sure that the `post` object has access to its new `comments`.

That said, I don't think you should spend more than an hour or so on it — broader test coverage is the goal and if some of that coverage has quirkiness in it I think that's alright.

gabivoicu:
Hey @alexsoble, this PR is ready for review & merge, as I have completed the changes in the student report `.erb` and added the tests. A few notes:

- I tried `reload` on a couple different objects (`student`, `student_assessment`), but none of it worked
- once this is merged, I will make another PR for moving and renaming `format_date` (of course, no rush on the merging, it's a holiday weekend after all)
- I had to make some changes to the factory files to get them to work (they don't seem to be used elsewhere so I figured it would be ok)
- I tried using the `school_years.rb` factory for a bit and got very confused by it...it does not appear to be what I needed, but I couldn't figure out what it does? There doesn't seem to be anything using it.
- tomorrow I will work on the JS smoke test(s) for student report and open another PR for it

gabivoicu:
@alexsoble Update on the JS test for student profile: 

I worked a bit on it today and ran into a roadblock. The PR you suggested I look at ([1346](https://github.com/studentinsights/studentinsights/pull/1346)) shows a test pattern I couldn't follow because it seems `AttendanceDetails.js` is written in the new ES6 way, where `student_profile_pdf.js` seems to be the old way (where everything is enclosed in an IIFE, if I am not mistaken).

Assuming I am correct and that is why I am am stuck, then the solution would be to move `student_profile_pdf.js` to the ES6 format and then I can write that test in that manner.

Do you think I am going down the right path here? If so, let me know and I can continue and update that file + add the test. I can use [PR 1309](https://github.com/studentinsights/studentinsights/pull/1309) as a template for what changes to make to the student report js file.
If not, let's pair on this test when you have some time.

spec/controllers/students_controller_spec.rb
```diff
@@ -604,6 +604,41 @@ def make_request(options = { student_id: nil, format: :pdf, from_date: '08/15/20
...
+          assessment = FactoryGirl.create(:assessment, :access)
+          student_assessment = FactoryGirl.create(:access, student: student, assessment: assessment, date_taken: '2016-08-16')
+          assessment = FactoryGirl.create(:assessment, :math)
+          student_assessment = FactoryGirl.create(:star_math_assessment, student: student, assessment: assessment, date_taken: '2017-02-16')
+          make_request({ student_id: student.id, format: :pdf, from_date: '08/15/2015', to_date: '03/16/2017' })
+
+          expect(assigns(:student_assessments)).to include("ACCESS Composite")
+          expect(assigns(:student_assessments)["ACCESS Composite"]).to be_kind_of(Array)
+          expect(assigns(:student_assessments)["ACCESS Composite"]).to eq([["2016-08-16 00:00:00 UTC", nil]])
+          expect(assigns(:student_assessments)).to include(" Mathematics")
```
alexsoble: @gabivoicu Why is there a leading space in `" Mathematics"`? That seems like an easy thing for a developer to miss . . . is the code set up to expect a leading space somehow or is that an artifact of the spec code?

alexsoble:
## Ruby changes + questions from @gabivoicu 

+ Great catch with the `school_years.rb` factory, Gabi. That factory was tied to a model, `school_year.rb`, which was deleted from the codebase a little while ago. Fixing that with #1354.
+ I left one in-line code question here: https://github.com/studentinsights/studentinsights/pull/1340/files#r159294676

alexsoble:
## JS questions from @gabivoicu 

+ I think it would be great to pull as much logic as possible from `student_profile_pdf.js` into ES6 modules! 
+ I'm not certain it will be possible to pull the entire thing into an ES6 module because if the way it needs to interact with Rails to make the PDF.

spec/controllers/students_controller_spec.rb
gabivoicu: @alexsoble The leading space was there because there was no family name for that test type. Looking at the factories again, I realized I should have used
`assessment = FactoryGirl.create(:assessment, :star)`
so now those tests will have a family, and it will be `STAR`.

Thanks for pointing that out!

gabivoicu:
@alexsoble 
Re: the ruby changes and questions: was your comment about the `" Mathematics"` issue? Want to make sure I addressed it, that link isn't showing me any any comments.

Re: JS
I will work on pulling as much of that logic into ES6 as I can. In terms of timeline, this week I don't have a ton of time left, but I will try to at least get started, try to gauge if I will need help.

alexsoble:
Hey @gabivoicu, thanks for your patience with my reviewing here! Looks like you dove into the structure of the Assessment class and fixed the issue with `" Mathematics"`, which is great. Overall this PR looks like it's in really good shape, I'm going to do some manual QA and then 🚢  if all looks good on the Student Report PDF front.

alexsoble:
@gabivoicu OK, so I **did** find a bug in the Student Report while QA'ing, but it **did not** have anything to do with this branch and I verified that it exists on `master`. 

Merging this in now and then I'll go document that bug.

PR #1342, opened 2017-12-26
Improve loading and parsing of district config YAML 

alexsoble:
# Who is this PR for?

Developers who need to read districtwide configuration from YAML.

A couple of inconsistent methods and code style practices. After I merged in #1322 I realized I was unsatisfied with a couple of things: 

+ The method `School.seed_somerville_schools` was a lie. This method read from `ENV` to determine district, so if ENV told it to seed New Bedford schools, it would do that. 
+ The `School` class used an ENV variable to read a path in the filesystem, which isn't a good practice.
+ The YAML config files I wrote had an unnecessary `"config"` key at the top level which added extra lines of code to unwrap. 

# What does this PR do?

+ Fixes the `School.seed_somerville_schools` method by making it pass in a fixed string as district key, so that it always does what it says it will.
+ Creates two separate classes, `LoadDistrictConfig` and `DistrictConfig`. `LoadDistrictConfig` uses district keys to select the right config files from the filesystem and parses them from YAML;  `DistrictConfig` applies validation and other logic. 
  + I separated them out because searching for YAML on the filesystem seemed a like a very different concern from exposing validated config information to the rest of the app. 
  + `School` also asks for a consistent way to find school config, but doesn't need any validation or other logic once it gets the config it needs.
+ Removes the unnecessary top-level key from the district config YAML files.

# Checklist 

+ [x] Test data import process locally before deploying

screen shot 2017-12-26 at 11 10 43 am

+ [x] Verify school seeding process by re-creating a demo site on this branch 

screen shot 2017-12-26 at 11 06 12 am

### DIFF
--- a/app/config_objects/district_config.rb
+++ b/app/config_objects/district_config.rb
@@ -1,3 +1,6 @@
+# This class exposes an API for district config that runs validation or other
+# logic on the parsed YAML config.
+
 class DistrictConfig
 
   def remote_filenames
@@ -5,9 +8,7 @@ class DistrictConfig
 
     remote_filenames = yml_config.fetch("remote_filenames")
 
-    if Rails.env.production?
-      validate_remote_filenames(remote_filenames)
-    end
+    validate_remote_filenames(remote_filenames) if Rails.env.production?
 
     return remote_filenames
   end
--- a/app/config_objects/district_config.rb
+++ b/app/config_objects/district_config.rb
@@ -1,7 +1,9 @@
 class DistrictConfig
 
   def remote_filenames
-    remote_filenames = load_remote_filenames_from_yml
+    yml_config = LoadDistrictConfig.new.load
+
+    remote_filenames = yml_config.fetch("remote_filenames")
 
     if Rails.env.production?
       validate_remote_filenames(remote_filenames)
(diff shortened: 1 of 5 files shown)


alexsoble:
Did some manual command line testing of school seeding on this branch:

```
School.seed_schools_for_district('new_bedford') ; School.count
=> 25

School.destroy_all ; nil
=> nil

School.seed_schools_for_district('somerville') ; School.count
=> 13
```

alexsoble:
@kevinrobinson I'm going to merge this in as a selfie for two reasons: 

+ I think making `School.seed_somerville_schools` a truthfully named method is an important improvement.
+ I'm excited to rebase #1321 on top of this to more consistently access school config data during the import process.

But I'm super open to any feedback or suggestions on these changes! What do you think? What am I missing here? 

I'm especially curious for feedback on splitting `LoadDistrictConfig` from `DistrictConfig`. This seems like a good step to me because the two classes do very different things. Sometimes a dev will just want to read in YAML config, and she should have a pre-built API for doing that. Sometimes a dev will want to add validation or logic on top of the config, and she should have a class to put that in too. 

Is this a good pattern? How can it be improved on? How about the naming of these classes? As a first step I added code comments that make the purpose of each class explicit, which I think is a good start.

alexsoble:
@kevinrobinson Also, obviously we're in holidays mode so you should answer in a super super async way, these questions don't seem urgent to me at all. 😎 🎄 ☃️

PR #1348, opened 2017-12-29
Remove persons_export.sql

alexsoble:
# Notes 

+ All the data from the X2 `persons` table is being joined to the students file and imported via `students_export.sql`
+ There is no `PersonImporter`, `PersonRow`, etc. in the import code, it all goes through `StudentsImporter`
+ Best to delete this since it's misleading to other districts setting up import -- they don't need a separate `persons_export`


alexsoble:
John Breslin removed this SQL export from the Somerville SIS export sequence, merging in deletion.

PR #1349, opened 2017-12-29
Remove student_schedule_export.sql

alexsoble:
# Notes

+ The outfile for this export, `student_schedule_export.txt`, isn't imported anywhere in the importer code
+ Best to delete this since it's misleading to other districts setting up a data import process -- they don't need to export student schedule


alexsoble:
John Breslin removed this SQL export from the Somerville SIS export sequence, merging in deletion.

PR #1353, opened 2018-01-02
Load past Service Uploads async

### LINKED ISSUE #1352: Service Upload page tries to load too much data inline
The Service Upload page has two functionalities — it lets admins upload a new bulk Service Upload, and it shows all past Service Uploads. 

The latter feature loads up all past Service Upload data as JSON, like this:

```
# service_uploads_controller.rb

service_uploads: ServiceUpload.order(created_at: :desc).as_json(
  only: [:created_at, :file_name, :id],
  include: {
    services: {
      only: [],
      include: {
        student: {
          only: [:first_name, :last_name, :id]
        },
        service_type: {
          only: [:name]
        }
      }
    }
  }
)
```

This worked fine when the number of Service Uploads was very low, but now that there's much more data in there, it's a very slow operation. The page is not loading at all for @snoopyuri and the page takes about ~22 seconds to load on my machine. 

At first I thought about caching this, but then I wondered if we could just fetch this data async after the page loads. That seems like a better option to try first. Database caching jobs require some ops setup on Heroku and you need to check the database to debug them, whereas a client-side-code-only solution would be lighter to deploy, QA and maintain.

alexsoble:
# Who is this PR for?

District admins who need to upload bulk service data.

Page is getting to heavy to load inline as amount of service upload data grows (should fix #1352).

# What does this PR do?

Moves the service upload data loading to async so that the page can load first and the past service upload data can load after. 

Adds a spinny loader while the past service upload data loads.

# Notes
+ This is the first spinny loader in the project! I chose a CSS approach in order to have the most lightweight code footprint possible; no JS dependencies. CSS from https://projects.lukehaas.me/css-loaders/. Thank you for the great resource @lukehaas!

# GIF (demo data; Internet Explorer VM)

async-past-services-ie

# Checklists

## Javascript QA

+ [x] Author checked latest in IE - Service Upload and Past Service Upload

## Automated Testing

+ [x] Author included specs for `PastServiceUpload.js`

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -10,7 +10,8 @@
     "React": true,
     "ReactDOM": true,
     "d3": true,
-    "spyOn": true
+    "spyOn": true,
+    "global": true
   },
   "parserOptions": {
     "ecmaVersion": 6,
--- a/app/assets/javascripts/service_uploads/PastServiceUploads.js
+++ b/app/assets/javascripts/service_uploads/PastServiceUploads.js
@@ -3,7 +3,17 @@ import React from 'react';
 class PastServiceUploads extends React.Component {
 
   render() {
+    const { serviceUploads } = this.props;
+
+    if (serviceUploads === null) return this.renderSpinner();
+
     return null;
+  }
+
+  renderSpinner() {
+    return (
+      <div className="loader">Loading...</div>
+    );
   }
 
 }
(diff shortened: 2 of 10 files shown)


alexsoble:
Going to 🚢  this as a selfie since it's unbreaking a page that's currently broken (#1352). Added specs, specs are green, ran through checklist, checklist is good.

PR #1356, opened 2018-01-02
Make target directory on SFTP site configurable

alexsoble:
# Who is this PR for?

School districts who want to set up a Student Insights instance. 

The data files might live in different directories on their remote SFTP sites; this PR makes the directory name configurable. 

# What does this PR do?

Add `target_directory` as a key to the district YML file.

If `target_directory` is unset, read from the home directory.

# Checks

+ [x] Tested locally for New Bedford
+ [x] Tested locally for Somerville

### DIFF
--- a/app/importers/clients/sftp_client.rb
+++ b/app/importers/clients/sftp_client.rb
@@ -27,11 +27,7 @@ class SftpClient < Struct.new :override_env, :env_host, :env_user, :env_password
     local_filename = File.join('tmp/data_download/', remote_file_name)
     local_file = File.open(local_filename, 'w')
 
-    target_file = target_directory
-                  ? "../#{target_directory}/#{remote_file_name}"
-                  : remote_file_name
-
-    sftp_session.download!(target_file, local_file.path)
+    sftp_session.download!(target_file(remote_file_name), local_file.path)
 
     local_file
   end
@@ -41,6 +37,14 @@ class SftpClient < Struct.new :override_env, :env_host, :env_user, :env_password
   end
 
   private
+
+  def target_file(remote_file_name)
+    if target_directory.present?
+      "../#{target_directory}/#{remote_file_name}"
+    else
+      remote_file_name
+    end
+  end
 
   # This is the folder on the remote SFTP site where the data files live
   def target_directory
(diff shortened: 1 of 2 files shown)


config/district_new_bedford.yml
```diff
@@ -10,6 +10,9 @@ remote_filenames:
 
   # istudentschedule.csv -- looks like this is based off of student_schedule_export.sql,
   #   but that isn't a data source we're using
+
+target_directory: newbedford
```
kevinrobinson: Is this a relative path?

app/importers/clients/sftp_client.rb
```diff
@@ -36,6 +37,20 @@ def dir_entries(*args)
...
+    if target_directory.present?
+      "../#{target_directory}/#{remote_file_name}"
+    else
+      remote_file_name
+    end
+  end
+
+  # This is the folder on the remote SFTP site where the data files live
+  def target_directory
+    LoadDistrictConfig.new.load.fetch("target_directory", nil)
```
kevinrobinson: Should this use the `DistrictConfig` API?

app/importers/clients/sftp_client.rb
```diff
@@ -36,6 +37,20 @@ def dir_entries(*args)
   end
 
   private
+
+  def target_file(remote_file_name)
+    if target_directory.present?
+      "../#{target_directory}/#{remote_file_name}"
```
kevinrobinson: This seems like it adds more bits that go into making the file path:
- `tmp/data_download/` in the download file method
- the `..` in this method
- `remote_filename` in the config
- and the actual filename

I'm wondering if there's benefits to this as compared to writing full paths into each remote filenames?  If there are differences between districts, seems good to fully decouple them into config.  Alternately, we could have the "target directory" in config be an absolute path, and remove the `..` and path `/tmp/data_download` paths that are here in code so that it's just `target_directory/remote_filename`.

alexsoble: Oooooh I really like the idea of writing full paths into the remote filenames!

alexsoble:
Closing in favor of write-full-path strategy in #1358!

PR #1357, opened 2018-01-02
Configure Net::SFTP to work with Quotaguard static IPs

alexsoble:
# Who is this PR for?

Districts with IP filtering on their Insights SFTP box. 

For IP filtering and whitelisting to work successfully, the nightly data import process needs to run on the static IP provided by Quotaguard.

# What does this PR do?

Sets up Quotaguard as a proxy for the data import process. 
  

# Checklists

+ [x] Tested with Somerville data import process
+ [x] Tested with New Bedford data import process

### DIFF
--- a/app/importers/clients/sftp_client.rb
+++ b/app/importers/clients/sftp_client.rb
@@ -48,7 +48,7 @@ class SftpClient < Struct.new :override_env, :env_host, :env_user, :env_password
   end
 
   def start_sftp_session
-    if ENV.fetch('QUOTAGUARDSTATIC_URL', false)
+    if ENV.fetch('USE_QUOTAGUARD', false)
       start_sftp_session_with_proxy
     else
       start_sftp_session_no_proxy
--- a/app/importers/clients/sftp_client.rb
+++ b/app/importers/clients/sftp_client.rb
@@ -60,7 +60,10 @@ class SftpClient < Struct.new :override_env, :env_host, :env_user, :env_password
   def start_sftp_session_with_proxy
     quotaguard = URI(ENV.fetch("QUOTAGUARDSTATIC_URL"))
 
-    proxy = Net::SSH::Proxy::HTTP.new(quotaguard.host, quotaguard.port)
+    proxy = Net::SSH::Proxy::HTTP.new(
+      quotaguard.host, quotaguard.port,
+      user: quotaguard.user, password: quotaguard.password
+    )
 
     @sftp_session = Net::SFTP.start(host, user, auth_mechanism({ proxy: proxy }))
   end
--- a/app/importers/clients/sftp_client.rb
+++ b/app/importers/clients/sftp_client.rb
@@ -62,7 +62,7 @@ class SftpClient < Struct.new :override_env, :env_host, :env_user, :env_password
       'quotaguard_proxy_host', ENV.fetch('QUOTAGUARDSTATIC_URL')
     )
 
-    @sftp_session = Net::SFTP.start(host, user, auth_mechanism, proxy: proxy)
+    @sftp_session = Net::SFTP.start(host, user, auth_mechanism({proxy: proxy}))
   end
 
   def start_sftp_session_no_proxy
@@ -74,9 +74,16 @@ class SftpClient < Struct.new :override_env, :env_host, :env_user, :env_password
   end
 
   # secrets
-  def auth_mechanism
+  def auth_mechanism(extra_info = {})
+    base_auth_mechanism.merge(extra_info)
+  end
+
+  def base_auth_mechanism
     return { password: password } if password.present?
-    { key_data: key_data } if key_data.present?
+
+    return { key_data: key_data } if key_data.present?
+
+    raise "Need either a password or a key!"
   end
 
   # avoid storing sensitive data from ENV as an instance variable, read it on-demand
--- a/app/importers/clients/sftp_client.rb
+++ b/app/importers/clients/sftp_client.rb
@@ -1,3 +1,5 @@
+require 'net/ssh/proxy/http'
+
 # These are keys into ENV
 class SftpClient < Struct.new :override_env, :env_host, :env_user, :env_password, :env_key_data
 
--- a/app/importers/clients/sftp_client.rb
+++ b/app/importers/clients/sftp_client.rb
@@ -41,14 +41,30 @@ class SftpClient < Struct.new :override_env, :env_host, :env_user, :env_password
     raise "SFTP information missing" unless sftp_info_present?
 
     if @sftp_session.nil?
-      proxy = Net::SSH::Proxy::HTTP.new(
-        'quotaguard_proxy_host', ENV.fetch('QUOTAGUARDSTATIC_URL')
-      )
-
-      @sftp_session = Net::SFTP.start(host, user, auth_mechanism, proxy: proxy)
+      start_sftp_session
     end
 
     @sftp_session
+  end
+
+  def start_sftp_session
+    if ENV.fetch('QUOTAGUARDSTATIC_URL', false)
+      start_sftp_session_with_proxy
+    else
+      start_sftp_session_no_proxy
+    end
+  end
+
+  def start_sftp_session_with_proxy
+    proxy = Net::SSH::Proxy::HTTP.new(
+      'quotaguard_proxy_host', ENV.fetch('QUOTAGUARDSTATIC_URL')
+    )
+
+    @sftp_session = Net::SFTP.start(host, user, auth_mechanism, proxy: proxy)
+  end
+
+  def start_sftp_session_no_proxy
+    @sftp_session = Net::SFTP.start(host, user, auth_mechanism)
   end
 
   def sftp_info_present?
--- a/app/importers/clients/sftp_client.rb
+++ b/app/importers/clients/sftp_client.rb
@@ -58,11 +58,11 @@ class SftpClient < Struct.new :override_env, :env_host, :env_user, :env_password
   end
 
   def start_sftp_session_with_proxy
-    proxy = Net::SSH::Proxy::HTTP.new(
-      'quotaguard_proxy_host', ENV.fetch('QUOTAGUARDSTATIC_URL')
-    )
+    quotaguard = URI(ENV.fetch("QUOTAGUARDSTATIC_URL"))
 
-    @sftp_session = Net::SFTP.start(host, user, auth_mechanism({proxy: proxy}))
+    proxy = Net::SSH::Proxy::HTTP.new(quotaguard.host, quotaguard.port)
+
+    @sftp_session = Net::SFTP.start(host, user, auth_mechanism({ proxy: proxy }))
   end
 
   def start_sftp_session_no_proxy
--- a/app/importers/clients/sftp_client.rb
+++ b/app/importers/clients/sftp_client.rb
@@ -39,9 +39,15 @@ class SftpClient < Struct.new :override_env, :env_host, :env_user, :env_password
   # This returns an object that will reveal secure data if printed
   def sftp_session
     raise "SFTP information missing" unless sftp_info_present?
+
     if @sftp_session.nil?
-      @sftp_session = Net::SFTP.start(host, user, auth_mechanism)
+      proxy = Net::SSH::Proxy::HTTP.new(
+        'quotaguard_proxy_host', ENV.fetch('QUOTAGUARDSTATIC_URL')
+      )
+
+      @sftp_session = Net::SFTP.start(host, user, auth_mechanism, proxy: proxy)
     end
+
     @sftp_session
   end
 
--- a/.gitignore
+++ b/.gitignore
@@ -13,9 +13,12 @@ node_modules
 yarn-debug.log*
 yarn-error.log*
 
-# This holds configuration for testing production import scripts locally, so it shouldn't be checked into
-# source control.
+# This holds configuration for testing production import scripts locally,
+# so it shouldn't be checked into source control.
 config/local_env.yml
+
+# This is for use with `heroku local`.
+.env
 
 ## Environment normalisation:
 /.bundle


alexsoble:
Thinking about security/privacy tradeoffs here: 

The upside to this approach is that it means we can whitelist our static IPs on our SFTP boxes and use an IP filtering strategy. The downside is that nightly data import traffic will pass through an extra third party cloud provider, QuotaGuard. 

All web traffic is already passing through QuotaGuard's static IP proxies. In my mind, that makes routing nightly data import through QuotaGuard less of a concern — we're not adding a new provider to the mix, just routing more processes through an existing provider. 

As I've been talking with John Breslin, I'm picking up that IP filtering is central to how he thinks about security. For example, only whitelisted IPs can contact Somerville's LDAP (which is why we need Quotaguard to provide static IPs). Setting up the same kind of filtering on our SFTP box would bring our data infrastructure security practices in line with that strategy.

Overall, the considerations above make me think that enforcing IP filtering on our SFTP boxes is a good idea. It adds an extra layer of security, makes our setup look more like Somerville's, and involves sending more traffic through an existing service provider. 

Interested to hear other folks' perspectives!

alexsoble:
Notes to self on how to test this branch locally: 

https://devcenter.heroku.com/articles/quotaguardstatic#local-setup

Looks like this will involve `heroku local`:

https://devcenter.heroku.com/articles/heroku-local

alexsoble:
Progress update: New error!

I took the following steps to simulate a real Heroku nightly import on my local machine: 

1. Added an `.env` file locally with the SIS SFTP details for New Bedford and the QuotaGuard details for Heroku QuotaGuard instance.
2. Whitelisted our QuotaGuard IPs on the SFTP box.
3. Ran `heroku local:run thor import:start`.

Results:

```
Downloading ../newbedford/istaff.csv...🚨  🚨  🚨  🚨  🚨  Error! service name too long (78)
Downloading ../newbedford/istudent.csv...🚨  🚨  🚨  🚨  🚨  Error! service name too long (78)
Downloading ../newbedford/iattendance.csv...🚨  🚨  🚨  🚨  🚨  Error! service name too long (78)
Downloading ../newbedford/iconduct.csv...🚨  🚨  🚨  🚨  🚨  Error! service name too long (78)
Downloading SomervillePublicSchools - Generic SM Extract.csv...🚨  🚨  🚨  🚨  🚨  Error! SFTP information missing
Downloading SomervillePublicSchools - Generic SR Extract.csv...🚨  🚨  🚨  🚨  🚨  Error! SFTP information missing
Downloading ../newbedford/iassessments.csv...🚨  🚨  🚨  🚨  🚨  Error! service name too long (78)
```

"`service name too long`" is a new and interesting error for me, need to dig into it a little more to see where it's coming from and what it means.

alexsoble:
The error is coming from line 65:

```
@sftp_session = Net::SFTP.start(host, user, auth_mechanism({proxy: proxy}))
```

Backtrace:

```
ArgumentError: service name too long (78)
	from app/importers/clients/sftp_client.rb:65:in `start_sftp_session_with_proxy'
	from app/importers/clients/sftp_client.rb:54:in `start_sftp_session'
	from app/importers/clients/sftp_client.rb:46:in `sftp_session'
	from app/importers/clients/sftp_client.rb:31:in `download_file'
	from (irb):5
```

alexsoble:
I looked in Net::SSH and Net::SFTP and couldn't find anything. Looks like this error is coming from Ruby itself. https://searchcode.com/codesearch/view/94774073/ line 456.

alexsoble:
I wonder if the underlying HTTP request with {{PRIVATE KEY}} + {{PROXY INFO}} exceeds some kind of limit. It's certainly possible that our existing requests which just pass a private key are under the limit, and adding more info about a proxy pushes them over somehow. 

But I've never heard of the idea that an HTTP request (or certainly a URL) could be "too long."

alexsoble:
This recipe might be helpful:

http://support.quotaguard.com/support/solutions/articles/5000013943-getting-started-with-ruby-sftp-quotaguard-static

alexsoble:
Progress! In f801cade1696e5d35df699374e07df6832de6d77. The issue was that I wasn't properly unwrapping the values for `quotaguard.host` and `quotaguard.port`. I didn't realize that `"quotaguard_host"` was just a dummy value that needed to be replaced. 😄 

New error message:

```
Downloading educators_export.txt...🚨  🚨  🚨  🚨  🚨  Error! {:version=>"HTTP/1.1", :code=>407, :reason=>"Proxy Authentication Required", :headers=>{"Content-Length"=>"33", "Proxy-Authenticate"=>"Basic realm=Basic"}, :body=>"407 Proxy Authentication Required"}
Downloading courses_sections_export.txt...🚨  🚨  🚨  🚨  🚨  Error! {:version=>"HTTP/1.1", :code=>407, :reason=>"Proxy Authentication Required", :headers=>{"Content-Length"=>"33", "Proxy-Authenticate"=>"Basic realm=Basic"}, :body=>"407 Proxy Authentication Required"}
Downloading educator_section_assignment_export.txt...🚨  🚨  🚨  🚨  🚨  Error! {:version=>"HTTP/1.1", :code=>407, :reason=>"Proxy Authentication Required", :headers=>{"Content-Length"=>"33", "Proxy-Authenticate"=>"Basic realm=Basic"}, :body=>"407 Proxy Authentication Required"}
Downloading students_export.txt...🚨  🚨  🚨  🚨  🚨  Error! {:version=>"HTTP/1.1", :code=>407, :reason=>"Proxy Authentication Required", :headers=>{"Content-Length"=>"33", "Proxy-Authenticate"=>"Basic realm=Basic"}, :body=>"407 Proxy Authentication Required"}
Downloading student_section_assignment_export.txt...🚨  🚨  🚨  🚨  🚨  Error! {:version=>"HTTP/1.1", :code=>407, :reason=>"Proxy Authentication Required", :headers=>{"Content-Length"=>"33", "Proxy-Authenticate"=>"Basic realm=Basic"}, :body=>"407 Proxy Authentication Required"}
Downloading attendance_export.txt...🚨  🚨  🚨  🚨  🚨  Error! {:version=>"HTTP/1.1", :code=>407, :reason=>"Proxy Authentication Required", :headers=>{"Content-Length"=>"33", "Proxy-Authenticate"=>"Basic realm=Basic"}, :body=>"407 Proxy Authentication Required"}
Downloading behavior_export.txt...🚨  🚨  🚨  🚨  🚨  Error! {:version=>"HTTP/1.1", :code=>407, :reason=>"Proxy Authentication Required", :headers=>{"Content-Length"=>"33", "Proxy-Authenticate"=>"Basic realm=Basic"}, :body=>"407 Proxy Authentication Required"}
Downloading student_averages_export.txt...🚨  🚨  🚨  🚨  🚨  Error! {:version=>"HTTP/1.1", :code=>407, :reason=>"Proxy Authentication Required", :headers=>{"Content-Length"=>"33", "Proxy-Authenticate"=>"Basic realm=Basic"}, :body=>"407 Proxy Authentication Required"}
```

Looks like we need Proxy Authentication set up!

alexsoble:
Tested this with both a Somerville and a New Bedford import process, both were 👍 🌈 .

@kevinrobinson since this is an important change do a core process I hope you don't mind if I ask for a review. 💻 🔍

app/importers/clients/sftp_client.rb
```diff
@@ -39,20 +41,52 @@ def dir_entries(*args)
   # This returns an object that will reveal secure data if printed
   def sftp_session
     raise "SFTP information missing" unless sftp_info_present?
+
     if @sftp_session.nil?
-      @sftp_session = Net::SFTP.start(host, user, auth_mechanism)
+      start_sftp_session
```
kevinrobinson: Does this need to assign `@sftp_session` too?  Or could we move the assignment here rather than doing it in each method?

app/importers/clients/sftp_client.rb
```diff
@@ -39,20 +41,52 @@ def dir_entries(*args)
...
+    end
+  end
+
+  def start_sftp_session_with_proxy
+    quotaguard = URI(ENV.fetch("QUOTAGUARDSTATIC_URL"))
+
+    proxy = Net::SSH::Proxy::HTTP.new(
+      quotaguard.host, quotaguard.port,
+      user: quotaguard.user, password: quotaguard.password
+    )
```
kevinrobinson: This makes an HTTP connection with the quotaguard username and password.  Should this be HTTPS too?

alexsoble:
Hey @kevinrobinson, thanks for the thoughtful review! 😄 

A bit earlier in the thread I wrote up some initial thoughts/ideas on security, you may have seen and read this already but just in case: https://github.com/studentinsights/studentinsights/pull/1357#issuecomment-355130330

Thoughts/questions/comments on a bunch of the security ideas your brought up in your review! 

## Private keys over passwords 

I agree 💯 % that this is the best-case scenario. SSH keys with no IP filtering is what we currently have set up with Somerville. With New Bedford, it looks like this isn't possible because they're using a hosted Aspen/X2 UI for imports which only supports password-based authentication. Do you think an SFTP box with password-based auth and no IP filtering is acceptable security-wise if we rotate the password? 

Another thing we could do would be use New Bedford's current, password-based auth to get an initial Heroku instance set up, and then do a deep dive with them to find some alternative to the hosted Aspen UI that would support private keys before we open the instance up to real users.

## Key rotation

This sounds like a fantastic idea. We've definitely rotated the keys on the existing SFTP box a few times when prompted by external circumstances, but we don't have a set schedule. We should set one up! What seems reasonable to you — every 6 months, every 3 months? Once we set up a schedule, how to we enforce it? Add to documentation, maybe send automated email reminders, some kind of ENV var that helps force a reset to make sure we stick to schedule? 

## Alternatives to QuotaGuard

Oooh, good topic to explore! I read up on Heroku Private Spaces (https://devcenter.heroku.com/articles/private-spaces#stable-outbound-ip-addresses) yesterday. Unfortunately they're part of Heroku Enterprise, which I imagine is $$$$ because it doesn't have a price listed on the Heroku website, just a "get in touch!" form. Maybe it will become available to more regular Heroku user folk as time goes on though? 

Rolling our own proxy is an interesting option I hadn't thought of. 😄  

## Encrypting data at rest on SFTP

This is a fantastic idea. I really see the logic here: adding IP filtering and encrypting data at rest are two places we could invest technical work to improve security. If we do IP filtering with a third-party, that creates a new form of risk, whereas encryption-at-rest doesn't come with that risk. So why not invest the time and energy in encryption-at-rest? 👍

kevinrobinson:
Hi @alexsoble!  Yeah this is great stuff to talk out and work on improving :)

One clarification first - I thought this pull request was for adding an outbound static proxy so that the New Bedford SFTP box could whitelist the IP of the Student Insights process that sshs in to imported the CSV files (ie, the [line in this diagram](https://github.com/studentinsights/studentinsights/blob/1b18fffeb1863c208d56303993b31502a8b571b7/README.md#getting-data-into-insights)).  But from reading your latest comment: "With New Bedford, it looks like this isn't possible because they're using a hosted Aspen/X2 UI for imports which only supports password-based authentication" I think maybe I'm misunderstanding.

So to clarify, with New Bedford, are we copying the CSV files directly from a hosted Aspen instance without the SFTP server in between?  Or is it that there are obstacles to setting up ssh keys on the New Bedford SFTP site?  Or maybe here's something else I'm misunderstanding, computers are confusing :)  Thanks!

alexsoble:
@kevinrobinson I re-wrote the sentence to be (hopefully!) clearer: 

"With New Bedford, it looks like using SSH keys to authenticate to the SFTP box isn't possible because they are using a hosted Aspen/X2 UI for imports which only supports password-based authentication."

kevinrobinson:
@alexsoble So here's my mental model of how the data flows now between three computers:

```
|Hosted Aspen X2 box| --pushed to--> |SFTP box| --pulled by--> |Insights Heroku job|
```

and what this PR does:
```
|Hosted Aspen X2 box| --pushed to--> |SFTP box| --pulled through--> |QuotaGuard proxy| --pulled by--> |Insights Heroku job|
```

It sounds like you're saying that the X2 box only supports password-based authentication, which seems like it's the first arrow.  I'm thinking that this pull request is about the second and third arrows, the flow from the SFTP box to Insights (and possibly through QuotaGuard).

Have I got the diagram mixed up here?  It seems like maybe the first arrow is different than I'm thinking it is.

alexsoble:
@kevinrobinson Yes, your mental model is correct! 😄 

I must not have explained the motivation for my question clearly, let me try again with your arrows as help. 😄 

Business reasons are forcing us to use password auth in the first arrow. Since that's a less secure mechanism than using SSH keys would be for the first arrow, are we comfortable with having no IP filtering on our box? Adding IP filtering would be the motivation for QuotaGuard proxy in the second arrow.

kevinrobinson:
@alexsoble Thanks for talking this out! 👍 

So I hear that because of limitations on how the Hosted Aspen X2 setup works, it only supports exporting data to the SFTP site via password auth (the first arrow).  I'm trying to figure out why that's connected to how the SFTP --> Insights flow works (the second arrow).

One thought is "Maybe I've misunderstood the intent of this PR, and it's aiming to add a static IP to the SFTP box so that the Hosted Aspen X2 instance will only export to that particular ip."  This isn't what I thought this pull request was doing though, so maybe I'm mixed up.

Another thought is "Maybe since the first arrow uses password auth, the idea is that the second arrow would also use the same password auth so there's only one way to connect to the SFTP box that both sides use.  And since password auth is less secure than SSH keys, adding a static IP to insights would be good because it would let us whitelist that IP on the SFTP box."  This would make sense to me, but then I'm wondering if we could use the more secure SSH key method for the second arrow (even if we have to use the less secure password auth for the first arrow).

Talking is hard!  We can move to chat too if that's easier :)

alexsoble:
@kevinrobinson OK great I'm calling you now. ☎️

alexsoble:
Had a good ☎️  discussion about this with @kevinrobinson! tl;dr is that we agree the extra risk of sending more student data to QuotaGuard isn't worth enabling the extra benefit of IP filtering. Plus lots of other good discussion and learning along the way. 

+ Look at enabling Fail2ban down the line
+ Do as much as we can to secure connections between computers we control
+ In this case it looks like password auth from Aspen/X2 is a given, not really something we can control and not a totally insecure default 
+ Let's ship a minimal thing this week with password auth and incrementally improve security on parts we control as we go ✨

Issue #1248, opened 2017-11-08
Best way to view Student Report PDF in VM?


alexsoble:
I'm working on verifying the Student Report PDF in an Internet Explorer system. I'm having a problem with the VirtualBox VM (IE 11, Windows 7) where I get stuck in a circular loop and the VM refuses to display the PDF: 

![stuck](https://user-images.githubusercontent.com/3209501/32579349-25e280d2-c4a6-11e7-83ca-7d884147c91a.png)

Are other folks getting stuck here as well? Or are you able to see PDFs in your VM?

jhilde:
Hmmm, I use VMWare Fusion with Windows 10. IE11 prompts me to open or save the download. Both work properly. Unfortunately VMWare Fusion isn't free.

alexsoble:
Answered this one with @kevinrobinson!

1. Launch a Windows VM with the targeted OS/IE version 
2. Go to get.adobe.com/reader in the VM
3. Install Adobe Acrobat in the VM
4. Try opening some PDFs in your VM!
5. If PDFs are openable, save a "Snapshot" so that you can access this VM state easily
6. Make sure you are logged out of Insights in your Snapshot state and that no student report PDFs are lying around your Downloads folder 

TOOD: Document in README

alexsoble:
closed by #1364!

PR #1362, opened 2018-01-04
Install secure_headers gem

alexsoble:
# Why?

+ This gem improves how our app handles cookies. It marks all http cookies with the Secure, HttpOnly and SameSite attributes. 
+ These are good security measures to take given the recent disclosure of the Meltdown and Spectre vulnerabilities: https://spectreattack.com/.
  
# Suggested deploy strategy

1. Push to demo site first
2. Manually verify that the new settings don't impede any functionality or user experience
3. Deploy to production 
  
# Checklist 

+ [x] Tested latest commit in Internet Explorer


alexsoble:
This breaks JS locally with the error:

kevinrobinson:
Nice!  👍

We might want to allow `X-Download-Options` for PDF downloads (skimming https://github.com/twitter/secureheaders).  The Content-Security-Policy might also not "just work" since I think there's some inline JS stubs in various places.  Those would be great things to improve too, but maybe if we can add this gem and most of its defaults that seems great!

kevinrobinson:
For the local JS error, we might be able to only enable the gem in development mode (since nothing is https).

alexsoble:
@kevinrobinson You mean only enable in production right?

alexsoble:
Woah! GitHub leaves notes about deployments now! I'm pushing this up to the demo site so we can see what "just works" and what doesn't.

alexsoble:
Okay! So most things seem to "just work" on the demo site with this patch applied. 

✅  School Overview Page
✅  Student Profile 
✅  Student Profile — Record Service 
✅  Student Profile — Add/Edit Notes

@kevinrobinson As you predicted, Student Report PDF is broken, will tweak that.

alexsoble:
Already seeing some good results here from https://securityheaders.io/.

# Production site (`master`)

c grade

# Demo site (this branch)

screen shot 2018-01-04 at 5 36 17 pm

alexsoble:
Did some manual testing of the demo site in an Internet Explorer VM. 

Looks like Student Report PDF opening might be broken on this branch in Internet Explorer, although its working in Chrome.

config/initializers/secure_headers.rb
```diff
@@ -0,0 +1,23 @@
+SecureHeaders::Configuration.default do |config|
+
+  # Unblock PDF downloading for student report
+  config.x_download_options = nil
```
kevinrobinson: We could also set this in the controller action so that it's only allowed for that the actual download request: https://github.com/twitter/secureheaders/blob/master/docs/per_action_configuration.md

alexsoble:
It's hard to tell based on the VM how/if PDF download is affected in IE. The VM shows that student reports are being successfully downloaded to the filesystem:

screen shot 2018-01-05 at 10 38 54 am

But when I open them in IE, I don't see anything. Not clear if this is somehow due to Insights settings or to limitations of the VM.

kevinrobinson:
Also re: CSP, the browser will report violations if you tell it to.  This is a separate app I've used before to log them to chat or email or something.  This is definitely not critical to do at all, but helped me when trying this out to make sure there wasn't something I mistakenly blocked: https://github.com/mit-teaching-systems-lab/csp-logger.

Also, adding CSP makes this more complicated so we could also just keep that turned off, ship the cookie improvements and then add CSP separately.

alexsoble:
@kevinrobinson OK, I poked around in the production Heroku site, which doesn't have this patch applied, and the PDF download was exactly the same as on the demo site. This is an issue I logged earlier but haven't solved yet, "Best way to view Student Report PDF in VM?" #1248.

My preference right now would be to ship this PR as is. It introduces the cookies security upgrades we wanted, it introduces some bonus CSP safeguards, and to the best of my VM's abilities to detect, it doesn't degrade the user experience at all. We could ping Uri and ask him to try opening a Student Report PDF to verify this.

Once this is shipped and the security improvements are in place, we can go and fine tune if needed. Let me know what you think about that strategy.

kevinrobinson:
@alexsoble oops sorry, didn't hit submit on the PR comments.

My understanding is the current CSP settings in this PR would prevent loading the Mixpanel code for analytics and Rollbar code for error reporting.  If that's the case y vote would be to just cut out CSP and ship to get the other great improvements out (and seems like downloads are good), and then add in CSP separately later.

alexsoble:
@kevinrobinson Yup that sounds like a good plan! Didn't notice the Mixpanel/Rollbar blocking, great catch there. 😊

alexsoble:
@kevinrobinson Does the commit look good? If you're busy now we could also review/deploy after the team meeting this afternoon. 👍 😊

alexsoble:
https://securityheaders.io/ for somerville-teacher-tool-demo with latest commit (`config.csp = SecureHeaders::OPT_OUT`).

screen shot 2018-01-05 at 1 44 36 pm

config/initializers/secure_headers.rb
alexsoble: Took a closer look at this with @kevinrobinson and we realized that `x_download_options` isn't available as a per-controller configuration in their DSL. 😞

alexsoble:
🚢 ing it!

PR #1363, opened 2018-01-05
Updating nokogiri to patch version with fix

kevinrobinson:
Fixing a bug in nokogiri 1.8.0.  It's pulled in transitively by a few different dependencies.


alexsoble:
🚢  optimistically!  BECAUSE we are optimistic!

PR #1364, opened 2018-01-05
Add note about PDF in VM and snapshots

kevinrobinson:
For developers!  From https://github.com/studentinsights/studentinsights/issues/1248#issuecomment-355667686


kevinrobinson:
selfie

PR #1367, opened 2018-01-09
Rename and refactor date formatter

gabivoicu:
# Who is this PR for?
Developers
Code quality
# What does this PR do?
Renames format_date to string_format_date for specificity and moves it into the student controller because it is only used there

# Checklists

## Javascript QA

*Which features or pages does this PR touch? Make a list here so that you can double-check each affected page in a Internet Explorer before merging.*

*Make one list for the author and another for the reviewer.*

*For more info on how to download and use a Virtual Machine to test in IE, see the README.*

*Example:*

+ [x] Author checked latest in IE - Student Report PDF
+ [x] Reviewer checked latest in IE - Student Report PDF

### DIFF
--- a/app/controllers/students_controller.rb
+++ b/app/controllers/students_controller.rb
@@ -11,10 +11,6 @@ class StudentsController < ApplicationController
 
   before_action :authorize_for_districtwide_access_admin, only: [:lasids]
 
-  def string_format_date(date)
-    date.strftime("%m/%d/%Y")
-  end
-
   def authorize!
     student = Student.find(params[:id])
     raise Exceptions::EducatorNotAuthorized unless current_educator.is_authorized_for_student(student)
@@ -80,7 +76,7 @@ class StudentsController < ApplicationController
 
     respond_to do |format|
       format.pdf do
-        footer = "Somerville Public Schools Student Report -- Generated #{string_format_date(todays_date)} by #{current_educator.full_name} -- Page [page] of [topage]"
+        footer = "Somerville Public Schools Student Report -- Generated #{format_date_for_student_report(todays_date)} by #{current_educator.full_name} -- Page [page] of [topage]"
         render(wait_for_js.merge({
           pdf: 'student_report',
           title: 'Student Report',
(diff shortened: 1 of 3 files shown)


gabivoicu:
@alexsoble Had to move the method back into the Application helper because it wasn't working when I tested in IE, not sure how it worked before. I renamed it to something we discussed before, I think that name makes sense now.

In terms of IE testing, I tested on IE11-Win 8.1 so far and I can get the PDF to download, but I can't open it because of resolution issues I haven't been able to sort out, I will try on my 27" monitor when I get the chance as well, see if that helps.

gabivoicu:
@alexsoble Update: downloaded Adobe PDF Reader and was able to verify the PDF on both IEs on my MacBook. They do look a bit pixelated but it's probably because of my tiny screen. Unless you have other comments, this is ready.

alexsoble:
@gabivoicu Hey, glad you got the Adobe PDF Reader set up on your VM! That's a tricky move, I just figured that one out the other week with an assist from Kevin.

Now that the Acrobat-in-VM issue is sorted out, are we still sure that the method breaks when we move it into the students controller? The PR description mentions moving the method into the students controller, so we should either do that or update the PR description.

gabivoicu:
@alexsoble Yup, it still breaks even after getting pdf reading to work 💔 

My suspicion is that's why it is in the `ApplicationController` right now, though I could be wrong. Another thing I was considering was moving all date formatting inside the `.rb` controller file, so formatting all the dates before they were passed to the `.erb` file, but that is definitely a bigger undertaking/a design change we should discuss at length before implementing.

alexsoble:
@gabivoicu Ah, my bad here! The `format_date_for_student_report` method is in `ApplicationHelper`. It won't work in `StudentsController` because that's a controller, not a helper. Controller private methods can be called in controller code, but not view code. The formatting method is called in the `student_report.pdf.erb` view code.

If we wanted to isolate the method in its own helper file, we would do this: http://api.rubyonrails.org/classes/ActionController/Helpers.html

I think that's not necessary for this PR, which uses the method name to indicate where it's used (improvement!) and refactors the method to remove unnecessary error suppression.

alexsoble:
Checked out a PDF in Internet Explorer — all looks good! 👍 

Ready to 🚢 .

alexsoble:
Here's a screenshot from the Internet Explorer VM:

pdf

PR #1369, opened 2018-01-09
Fix bug with TestPals data seeding on demo Heroku site

alexsoble:

TestPals won't build on the demo Heroku site because on `master` it has FactoryGirl as a dependency, which is only loaded in test and development environments. 

# What does this PR do?

Removes FactoryGirl as a dependency for TestPals so that we can rebuild the Heroku demo site; replaces FactoryGirl with straight Ruby, e.g. `Educator.create!`, `Student.create!`.

# See also

https://github.com/studentinsights/studentinsights/pull/1315#issuecomment-356338961

### DIFF
--- a/spec/support/test_pals.rb
+++ b/spec/support/test_pals.rb
@@ -180,7 +180,7 @@ class TestPals
       school: @shs,
       homeroom: @shs_bill_nye_homeroom
     )
-    @shs_biology_course = Course.create!(school: @shs)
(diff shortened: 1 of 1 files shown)


kevinrobinson:
@alexsoble Thanks!  This seems good, I didn't realize that all the educator factory bits are essentially not used here anymore. 👍

alexsoble:
Running the rebuild-demo-site script!

PR #1370, opened 2018-01-10
Limit seeding data to stay under Heroku demo limit

kevinrobinson:
# Who is this PR for?
Developers

Heroku's database row limits on the demo site

# What does this PR do?
Reduce the data generated in the database seed task so that it's under 10k.

```
irb(main):012:0> ActiveRecord::Base.subclasses.map {|klass| klass.all.size }.sum
=> 8265
```

We could improve this overall (eg, updating the number of students per section is probably important for developing UI features for that page), but the goal here is just to unblock the demo site after shipping https://github.com/studentinsights/studentinsights/pull/1315.

Charts are fine.  Demo data isn't all entirely semantically meaningful for all schools, students and grade levels anyway.

screen shot 2018-01-10 at 8 48 54 am

### DIFF
--- a/app/demo_data/fake_student.rb
+++ b/app/demo_data/fake_student.rb
@@ -160,7 +160,7 @@ class FakeStudent
   def add_student_assessments_from_x2
     create_x2_assessment_generators(@student).each do |assessment_generator|
       unless @newstudent
-        5.times do
+        4.times do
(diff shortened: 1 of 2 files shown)


alexsoble:
👍

kevinrobinson:
thanks!

PR #1375, opened 2018-01-11
Allow districtwide admin with no school assigned to visit homeroom pages

### LINKED ISSUE #1304: Allow demo site districtwide admin to access homeroom view
Playing around with the demo data, I noticed that when I'm logged in as the districtwide admin role, I can't see any of the homeroom views. That doesn't seem right! Districtwide admins should have access to all data. 

I traced the issue back to this method:

```
# educator.rb

def allowed_homerooms
  # Educator can visit roster view for these homerooms
  return [] if school.nil?

  if districtwide_access?
    Homeroom.all
  elsif schoolwide_access?
    school.homerooms.all
  elsif homeroom
    school.homerooms.where(grade: homeroom.grade)
  elsif grade_level_access.present?
    school.homerooms.where(grade: grade_level_access)
  else
    []
  end
end
```

We should remove the `return [] if school.nil?` line up at the top. According to the validation logic we've written, districtwide admins can have no school assigned and still be valid (`Educator#has_school_unless_districtwide`). Those educators should still be allowed to access homeroom views. The logic in the if/else statement below is sufficient by itself. 

I ran a database check in production to see if this was affecting any districtwide educators:

```
irb(main):002:0> Educator.where(districtwide_access: true, school: nil).count
=> 0
```

The answer is no, it's not affecting any districtwide educators in production. 

Why do we allow districtwide admins to not be assigned to schools if all of them are assigned to schools? In the early stages of the project, I hand-rolled accounts for districtwide admins, and didn't assign them to any particular school. Now that we're further along, we've imported more data from the Aspen/X2 SIS system. In Aspen/X2, most districtwide educators are assigned to the Parent Information Center as a placeholder school. So they wouldn't be affected by this line of code. 

Because this issue is not affecting any production users, just the demo data (and hypothetical future users), I see it as low priority. I'd like to merge in #1274, which touches `allowed_homerooms` and is far more important, before we tackle this.

alexsoble:
# Who is this PR for?

Folks who want to demo insights on the demo Heroku site (or the fledgeling New Bedford instance).

Admin user can't see homerooms because s/he isn't assigned a school! 

# What does this PR do?

+ Fix #1304

### DIFF
--- a/app/lib/authorizer.rb
+++ b/app/lib/authorizer.rb
@@ -144,9 +144,6 @@ class Authorizer
 
   # TODO(kr) remove implementation
   def homerooms
-    # Educator can visit roster view for these homerooms
-    return [] if @educator.school.nil?
-
     if @educator.districtwide_access?
       Homeroom.all
     elsif @educator.schoolwide_access?


alexsoble:
I reviewed this one and QA'ed it fairly extensively in creating the issue for #1304, so I'll 🚢  this as a selfie.

PR #1376, opened 2018-01-11
Use canonical domain from ENV to set mailer SMTP settings instead of hardcoding

alexsoble:
# Who is this PR for?

Folks who want to set up a new Insights instance. 

The Somerville app's URL is hardcoded into the part of the code that enables sending emails from the app.

# What does this PR do?

Un-hardcode Somerville from mailer SMTP settings and use an ENV variable instead. 

# Notes 

I noticed that for Somerville the CANONICAL_DOMAIN variable does not include an "https://" prefix, while the SMTP settings URL does. 

I figure the simplest thing to do is to follow that pattern in future instances and leave the "https://" hardcoded since we'll always be using SSL.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -80,11 +80,13 @@ Rails.application.configure do
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
 
-
-
   # ---- Student Insights additions ----
   config.action_mailer.raise_delivery_errors = true
-  config.action_mailer.default_url_options = { host: "https://#{ENV.fetch('CANONICAL_DOMAIN')}/" }
+
+  config.action_mailer.default_url_options = {
+    host: "https://#{LoadDistrictConfig.new.canonical_domain}/"
+  }
+
   ActionMailer::Base.smtp_settings = {
     :port           => ENV['MAILGUN_SMTP_PORT'],
     :address        => ENV['MAILGUN_SMTP_SERVER'],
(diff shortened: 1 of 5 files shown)


alexsoble:
@kevinrobinson I figure I should follow @studentinsights-bot's advice on this once since it's touching `production.rb` and ask for a review before merging. 😎

config/environments/production.rb
```diff
@@ -84,13 +84,13 @@
...
   config.action_mailer.raise_delivery_errors = true
-  config.action_mailer.default_url_options = { host: 'https://homeroom-somerville.herokuapp.com/' }
+  config.action_mailer.default_url_options = { host: "https://#{ENV.fetch('CANONICAL_DOMAIN')}/" }
   ActionMailer::Base.smtp_settings = {
     :port           => ENV['MAILGUN_SMTP_PORT'],
     :address        => ENV['MAILGUN_SMTP_SERVER'],
     :user_name      => ENV['MAILGUN_SMTP_LOGIN'],
     :password       => ENV['MAILGUN_SMTP_PASSWORD'],
-    :domain         => 'http://somerville.studentinsights.org/',
+    :domain         => "https://#{ENV.fetch('CANONICAL_DOMAIN')}/",
```
kevinrobinson: https 👍

alexsoble:
@kevinrobinson Naw that's not 🚲 shedding, that's a good suggestion! 👍

kevinrobinson:
@alexsoble 😎 👍

alexsoble:
@kevinrobinson 😎 👍 

I'm going to ship this and then try to manually exercise the two features that depend on canonical URL:

+ Rerouting from the default Heroku URL to the canonical URL 
+ Email generation from the app

alexsoble:
✅  Canonical URL redirection works good

PR #1382, opened 2018-01-16
Adjust STAR data import to handle zipped data

### LINKED ISSUE #1381: Adjust STAR import code to handle a zip file
# What's the issue? 

STAR is an assessment vendor for both Somerville and New Bedford. We integrate STAR data into Insights. 

The SFTP site that STAR set up for Somerville had unzipped flat CSVs in the root directory. 

The SFTP site that STAR up for New Bedford had a single zip file in the root directory. 

I emailed STAR and asked if they could put their New Bedford site in the same format as Somerville, thinking that would save us time since we wouldn't have to write two separate code paths for a single vendor. 

They replied by moving Somerville's SFTP site to zipped format. So now we need to change the code to handle zip files, but once we do that we will be able to import STAR data for New Bedford and Somerville using a single code path.

alexsoble:
# Who is this PR for?

School districts that import STAR data.

New STAR data format (zipped files!) makes the existing code path broken.

# What does this PR do?

+ Fix #1381.

# Checklist 

+ [x] Manually QA STAR data import for Somerville
+ [x] Manually QA STAR data import for New Bedford

### DIFF
--- a/app/importers/file_importers/star_math_importer.rb
+++ b/app/importers/file_importers/star_math_importer.rb
@@ -1,7 +1,7 @@
 class StarMathImporter < Struct.new :school_scope, :client, :log, :progress_bar
 
   def import
-    return unless remote_file_name
+    return unless zip_file_name.present? && remote_file_name.present?
 
     log.write("\nDownloading ZIP file #{zip_file_name}...")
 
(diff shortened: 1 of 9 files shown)


app/importers/file_importers/star_math_importer.rb
```diff
@@ -1,18 +1,34 @@
...
+    log.write("\nDownloading ZIP file #{zip_file_name}...")
 
-    @data.each.each_with_index do |row, index|
-      import_row(row) if filter.include?(row)
-      ProgressBar.new(log, remote_file_name, @data.size, index + 1).print if progress_bar
+    downloaded_zip = client.download_file(zip_file_name)
+
+    Zip::File.open(downloaded_zip) do |zipfile|
+      zipfile.each do |file|
+        if file.name == remote_file_name
```
alexsoble: This is an ugly/hacky way to find the right CSV within the ZIP file, though it worked. A cleaner way would be to use the `#read` API on Zip::File: http://www.rubydoc.info/github/rubyzip/rubyzip/Zip/File#read-instance_method

alexsoble:
Today, progress was made.

Zipped folders were unzipped. 

Files were read. 

Hardcoded remote filenames were removed.

But when I test the process locally, data still isn't being imported yet.

Checklist of steps to try tomorrow: 

+ Visually compare the headers on the Somerville / New Bedford CSVs and make sure they're the same; compare them to the headers we've expected in the past 
+ Try isolating a single CSV row and importing it, see if it throws any errors

alexsoble:
Double checked the headers on the CSV files; they're all consistent.

alexsoble:
Turned out the blocker was that I wasn't using the STAR-specific data transformation classes!

Tested with both Somerville and New Bedford data paths:

Both were ✅  so this should be good to 🚢 !

PR #1390, opened 2018-01-18
Update deploy script to accept N git remotes as arguments

alexsoble:
# Who is this PR for?

Folks who deploy code to Heroku.

One command to deploy and migrate to all the git remotes!

# What does this PR do?

Replace a destroy script that was hardcoded to two remotes with a script that can deploy to N remotes.

Example: 

```
./scripts/deploy/deploy.sh somerville new-bedford heroku-demo
```

### DIFF
--- a/scripts/deploy/deploy.sh
+++ b/scripts/deploy/deploy.sh
@@ -1,5 +1,7 @@
 # Pass in the name of all the git remotes to deploy to as arguments
 
+# Example: ./scripts/deploy/deploy.sh somerville new-bedford heroku-demo
+
 for remote in "$@"
 do
     echo "🚢  🚢  🚢  Deploying code to $remote.";
--- a/scripts/deploy/deploy.sh
+++ b/scripts/deploy/deploy.sh
@@ -1,28 +1,11 @@
-SOMERVILLE_PROD_HEROKU_APP_NAME=$1
-DEMO_HEROKU_APP_NAME=$2
+# Pass in the name of all the git remotes to deploy to as arguments
 
-if [ -z "$SOMERVILLE_PROD_HEROKU_APP_NAME" ]; then
-  echo "🚨  🚨  🚨  Please supply a Somerville production Heroku app name."; exit 0;
-fi
-
-if [ -z "$DEMO_HEROKU_APP_NAME" ]; then
-  echo "🚨  🚨  🚨  Please supply a demo Heroku app name."; exit 0;
-fi
-
-# Deploy to Somerville production app and migrate
-echo "🚢  🚢  🚢  Deploying code to the Somerville production app.";
-git push heroku master
-echo;
-
-echo "⚙  ⚙  ⚙  Migrating the database.";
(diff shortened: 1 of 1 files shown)


scripts/deploy/deploy.sh
```diff
@@ -1,28 +1,13 @@
...
+
+# Example: ./scripts/deploy/deploy.sh somerville new-bedford heroku-demo
+
+for remote in "$@"
+do
+    echo "🚢  🚢  🚢  Deploying code to $remote.";
+    git push $remote master
+    echo;
+    echo "⚙  ⚙  ⚙  Migrating the database for $remote.";
+    heroku run rake db:migrate --remote $remote
```
kevinrobinson: @alexsoble I think my local git and Heroku CLI setup might be a bit different.  This seems great, and a good opportunity to sync these 👍 

To check my understanding, this means you have:
```
$ git remote -v
somerville 	https://git.heroku.com/somerville-teacher-tool.git (fetch)
somerville	https://git.heroku.com/somerville-teacher-tool.git (push)
new-bedford 	https://git.heroku.com/new-bedford-insights.git (fetch)
new-bedford	https://git.heroku.com/new-bedford-insights.git (push)
```

and then since you're using `--remote` instead of `--app`, you use the same `somerville` and `new-bedford` git remote names when running heroku commands.

Am I following?

alexsoble: Yup! So basically we need to do two things, push code to github and run a heroku command. 

The github push requires a remote. The heroku command requires either a `--remote` flag, or `--app` flag. Since we could get them both to do the right thing using the git remote as a single argument, I went with that.

alexsoble: My mental model is that you should be able to set the `<name>` values for your remotes locally to whatever you like, say somerville = `foo` and new bedford = `bar`, and the deploy script should work. Does that match up to what you're thinking?

kevinrobinson: Thanks for teaching me!  This seems great and I updated my setup to match. 👍

alexsoble: @kevinrobinson How did you update your setup?

kevinrobinson:
I updated the Heroku git remotes to be somerville, new-bedford and demo
(they had names like somerville-teacher-tool before).  And then origin for
Github.

And I started using —remote for Heroku commands too, which I didn’t know
about :)

-Kevin

alexsoble:
OK! My impression was that you wouldn't need to update your setup at all, because this script takes remote names as arguments. So I can run:

```
Example: ./scripts/deploy/deploy.sh somerville new-bedford heroku-demo
```

But you could run: 

```
Example: ./scripts/deploy/deploy.sh foo bar foobar
```

. . . if those were your remote names, and the script would run just fine. 

Either way, seems like we're good to 🚢  here!

kevinrobinson:
@alexsoble ah yep, I'm with you 👍

alexsoble:
@kevinrobinson 👍 😄

PR #1393, opened 2018-01-19
Scrub escaped double quotes from data files before importing

### LINKED ISSUE #1391: Fix quotes that throw error in New Bedford iconduct.csv import
Import throws an error on `iconduct.csv` import:

```
Missing or stray quote in line 7
```

Here's the difficult part of line 7:

```
"... yelling \"NO\" to the teacher ..."
```

alexsoble:
# Issue

+ Fix #1391 to unblock conduct import for New Bedford.

# Checklist 

+ [x] QA with Somerville import
+ [x] QA with New Bedford import

### DIFF
--- a/app/importers/data_transformers/csv_transformer.rb
+++ b/app/importers/data_transformers/csv_transformer.rb
@@ -9,9 +9,11 @@ class CsvTransformer
   end
 
   def transform(file)
-    csv = CSV.parse(file, headers: @headers,
-                          header_converters: :symbol,
-                          converters: lambda { |h| nil_converter(h) })
+    cleaned_file = file.gsub("\\\"", "")
+
+    csv = CSV.parse(cleaned_file, headers: @headers,
+                                  header_converters: :symbol,
+                                  converters: lambda { |h| nil_converter(h) })
 
     @pre_cleanup_csv_size = csv.size
 
--- a/spec/importers/data_transformers/csv_transformer_spec.rb
+++ b/spec/importers/data_transformers/csv_transformer_spec.rb
@@ -5,7 +5,7 @@ RSpec.describe CsvTransformer do
   describe '#transform' do
     context 'with good data' do
       context 'headers in csv' do
-        let!(:file) { File.open("#{Rails.root}/spec/fixtures/fake_x2_assessments.csv") }
+        let!(:file) { File.open("#{Rails.root}/spec/fixtures/fake_behavior_export.txt") }
         let(:headers_symbols) {[:local_id,:school_local_id,:assessment_date,:assessment_scale_score,:assessment_performance_level,:assessment_growth,:assessment_name,:assessment_subject,:assessment_test]}
         let(:transformer) { CsvTransformer.new }
         let(:output) { transformer.transform(file) }
--- a/spec/importers/data_transformers/csv_transformer_spec.rb
+++ b/spec/importers/data_transformers/csv_transformer_spec.rb
@@ -8,14 +8,15 @@ RSpec.describe CsvTransformer do
         # Return a File.read string just like the CsvDownloader class does:
         let!(:file) { File.read("#{Rails.root}/spec/fixtures/fake_behavior_export.txt") }
 
-        let(:headers_symbols) {[:local_id,:school_local_id,:assessment_date,:assessment_scale_score,:assessment_performance_level,:assessment_growth,:assessment_name,:assessment_subject,:assessment_test]}
         let(:transformer) { CsvTransformer.new }
         let(:output) { transformer.transform(file) }
         it 'returns a CSV' do
           expect(output).to be_a_kind_of CSV::Table
         end
         it 'has the correct headers' do
-          expect(output.headers).to match_array(headers_symbols)
+          expect(output.headers).to match_array([
+            :event_date, :incident_code, :incident_description, :incident_location, :incident_time, :local_id, :school_local_id
+          ])
         end
       end
       context 'headers not in csv' do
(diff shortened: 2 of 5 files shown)


alexsoble:
This unblocks the old error we were seeing from the conduct file (Missing or stray quote) and leaves us with a much better error, one that means that the file is now readable by our import process: 

```
9310 rows of data in ../newbedford/iconduct.csv pre-cleanup
9310 rows of data in ../newbedford/iconduct.csv post-cleanup

🚨  🚨  🚨  🚨  🚨  Error! Couldn't find Student
```

spec/importers/data_transformers/csv_transformer_spec.rb
```diff
@@ -5,28 +5,18 @@
...
-        let!(:file) { File.open("#{Rails.root}/spec/fixtures/fake_no_headers.csv") }
-        let(:headers) {["section_number","student_local_id","school_local_id","course_number","term_local_id","grade"]}
-        let(:headers_symbols) {[:section_number,:student_local_id,:school_local_id,:course_number,:term_local_id,:grade]}
-        let(:transformer) { CsvTransformer.new(headers:headers) }
-        let(:output) { transformer.transform(file) }
-        it 'returns a CSV' do
-          expect(output).to be_a_kind_of CSV::Table
-        end
-        it 'has the correct headers' do
-          expect(output.headers).to match_array(headers_symbols)
```
kevinrobinson: Can you explain more about why is this removed?

app/importers/data_transformers/csv_transformer.rb
```diff
@@ -9,9 +9,11 @@ def initialize(options={})
   end
 
   def transform(file)
-    csv = CSV.parse(file, headers: @headers,
-                          header_converters: :symbol,
-                          converters: lambda { |h| nil_converter(h) })
+    cleaned_file = file.gsub("\\\"", "")
```
kevinrobinson: The tests are good but seems worth a comment here.

Thinking aloud (feel free to ignore), would the `liberal_parsing` option described in https://github.com/ruby/ruby/pull/1160 work here too?  I see the description is somewhat different, but seems like it might be worth trying (test will confirm if it works or changes any other parsing)).  Alternately https://github.com/intercom/hippie_csv might be worth trying to drop in to see if it works w/o other changes.

alexsoble: @kevinrobinson Cool, thank you, will take a look at those! 👀 😎

alexsoble: @kevinrobinson So HippieCSV looked really interesting and promising. Their description of the use case they focus on sounds really similar to ours. I made a new branch to try it out. One issue I ran into upfront was that I couldn't find any support for headers: https://github.com/intercom/hippie_csv/search?utf8=%E2%9C%93&q=headers&type=

That makes me think HippieCSV could be a good future direction, but isn't a drop-in replacement. To make it work we'd have to stop reading headers from the exported CSVs and pass them in ourselves as hardcoded arrays for each of the files we parse.

alexsoble: Looking at `liberal_parsing` next: https://blog.bigbinary.com/2016/11/22/ruby-2-4-introduces-liberal_parsing-option-for-parsing-bad-csv-data.html, going to download ruby 2.4, do a little reading on what the changes are in 2.4, and see if this works as a drop in replacement.

alexsoble: Alrighty, so #1402 uses Ruby 2.4 and tries `liberal_parsing`. The specs fail, kind of interesting why: 

The blog post above from BigBinary explains that liberal_parsing makes the CSV tolerate errors like "CSV::MalformedCSVError: Illegal quoting in line 1."

In the Travis build (https://travis-ci.org/studentinsights/studentinsights/builds/332371960), we get "CSV::MalformedCSVError: Missing or stray quote in line 3"

Maybe I'm reading too much into the differences between these error messages, but it kind of seems like the message that `liberal_parsing` fixes is about quotes not conforming to the CSV spec ("illegal"), whereas the message we're getting is coming from a deeper level and is saying the CSV library doesn't know how to parse.

alexsoble: Here's the source in `ruby/csv.rb` where the two different error messages come from: https://github.com/ruby/ruby/blob/0593c6a510918ce0b6610d3f0d25748041658a55/lib/csv.rb#L1888

spec/importers/data_transformers/csv_transformer_spec.rb
alexsoble: Ooooh, this is a great question! I assumed that the `headers:` option for CsvTransformer wasn't being used anywhere, but searching through the codebase shows me that was wrong because StudentSectionGradesImporter uses it:

```
#student_section_grades_importer.rb

def data_transformer
  CsvTransformer.new(headers:%w[section_number student_local_id school_local_id course_number term_local_id grade])
end
```

I'm curious why StudentSectionGradesImporter needs to do this, I'll dig into the export file / commit history to find out. In the meantime, you're write, this PR shouldn't remove the spec for this code path so I'll restore.

alexsoble: The StudentSectionGradesImporter imports data from 'student_averages_export.txt', so I assumed we'd have a 'student_averages_export.sql' file under the /x2_exports directory, but I don't see any. I want to find that SQL file and put it under source control so we can see why the importer needs to use the `headers:` option . . . but that's a separate issue.

kevinrobinson:
@alexsoble Thanks for checking out the other libraries, bummer they aren't a good fit.  This seems weird to me to be munging the files manually like that, but let's do it!  I have no better ideas :)

alexsoble:
🚢  and we'll keep an eye out for more elegant solutions in the future!

PR #1394, opened 2018-01-20
Update net-ldap through devise

kevinrobinson:
# Who is this PR for?
everyone!

# What does this PR do?
It update `net-ldap` to a more recent version.  Since the pull request in `devise_ldap_authenticatable`, this forks the project to the studentinsights org to update the dependency.  It also incidentally included a patch version bump to bundler.

This can only be properly tested in production mode, so weekend deploy seems like a good time. :)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,11 @@ gem 'puma'
 gem 'activerecord-import'
 gem 'aws-sdk', '~> 2'
 gem 'devise', '~> 4.3.0'
-gem 'devise_ldap_authenticatable'
+
+# This is forked, waiting on https://github.com/cschiewek/devise_ldap_authenticatable/issues/242 to be merged, at
+# which point we can switch back to the gem upstream.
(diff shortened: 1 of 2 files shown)


kevinrobinson:
selfie to verify during the weekend, cc @alexsoble

kevinrobinson:
This doesn't work and breaks the login.  Reverting and I'll send an email about this @alexsoble.

alexsoble:
Hey @kevinrobinson! Thanks for trying to speed up getting us this fix! Finding a solution that means we don't have to wait for the pull request in `devise_ldap_authenticatable` to merge would be really good, we should pair on this if you have time this week or just discuss async over GitHub email. 

Based on the error you saw, do you think we'll have the same issue once the PR to `devise_ldap_authenticatable` merges?

alexsoble:
Part of my brain is thinking "oooh, maybe we should comment on the PR that bumps up the version to say we're an OS project and we'd like this patch too" but then another part of my brain is thinking "no, that's annoying, whoever maintains the project probably saw the PR already and is aware of this issue, and our extra comment won't help."

PR #1398, opened 2018-01-21
Improving service date saving in RecordService, factor out toMoment

kevinrobinson:
# Who is this PR for?
educators

Fixes https://github.com/studentinsights/studentinsights/issues/1307.

# What does this PR do?
- Educators should be able to leave end date blank. We should change the UI to allow this, and send up null instead of the string "Invalid date". This change would fix the immediate bug that came up last week.
- If there is an end date, validates on the client that it is an actual valid date and not allow saving until that's resolved.
- Makes the default "estimated end date" blank.
- Uses another strategy to parse the dates so that educators can write in natural formats and we assume American date ordering and accept 2/20/18 as validly representing Feb. 20th, 2018 without requiring them to type 02/20/2018.  This function is factored out of the component into its own module.

screen shot 2018-01-21 at 1 11 32 pm
screen shot 2018-01-21 at 1 12 29 pm

+ [x] Reviewer checked latest in IE - Student Profile

## Automated Testing

### DIFF
--- a/app/assets/javascripts/helpers/toMoment.js
+++ b/app/assets/javascripts/helpers/toMoment.js
@@ -1,0 +1,28 @@
+// Allow a few strict formats of dates
+const allowedFormats = [
+  'MM/DD/YYYY',
+  'MM/D/YYYY',
+  'M/DD/YYYY',
+  'M/D/YYYY',
+
+  'MM/DD/YY',
+  'MM/D/YY',
+  'M/DD/YY',
+  'M/D/YY',
+
+  'MM-DD-YYYY',
+  'MM-D-YYYY',
+  'M-DD-YYYY',
+  'M-D-YYYY',
+
+  'MM-DD-YY',
+  'MM-D-YY',
+  'M-DD-YY',
+  'M-D-YY'
+];
+
+// This will always return a `moment` but if`dateText` can't be parsed, that moment will return false
+// when you call `isValid`.
+export function toMoment(dateText) {
+  return moment.utc(dateText, allowedFormats, true); // this does 'strict parsing' on multiple formats
+}
--- a/app/assets/javascripts/student_profile/record_service.jsx
+++ b/app/assets/javascripts/student_profile/record_service.jsx
@@ -6,6 +6,7 @@ import serviceColor from '../student_profile/service_color.js';
 (function() {
   window.shared || (window.shared = {});
   const ProvidedByEducatorDropdown = window.shared.ProvidedByEducatorDropdown;
+  const QuadConverter = window.shared.QuadConverter;
 
   const styles = {
     dialog: {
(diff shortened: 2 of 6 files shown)


alexsoble:
@kevinrobinson Thanks so much for opening up this PR, reviewing now!

app/assets/javascripts/helpers/toMoment.js
```diff
@@ -0,0 +1,28 @@
...
+  'MM-DD-YY',
+  'MM-D-YY',
+  'M-DD-YY',
+  'M-D-YY'
+];
+
+// This will always return a `moment` but if`dateText` can't be parsed, that moment will return false
+// when you call `isValid`.
+export function toMoment(dateText) {
+  return moment.utc(dateText, allowedFormats, true); // this does 'strict parsing' on multiple formats
```
alexsoble: Learning about some great features of `moment` here, if other folks are reviewing and want to follow along here are the moment docs for the features @kevinrobinson is using:

+ https://momentjs.com/guides/#/parsing/strict-mode/
+ https://momentjs.com/guides/#/parsing/multiple-formats/

@kevinrobinson, the multiple formats doc mentions that "the more formats that are used, the longer that parsing takes." We should check out the impact on the client side performance, I don't imagine it will be huge but just something I'll think about when I test out in IE.

alexsoble: Tested this out, adding these format's didn't make service saving slow. 👍

kevinrobinson: @alexsoble Yeah, good to check.  I couldn't notice either  👍

kevinrobinson:
@alexsoble thanks for reviewing and @Jngai if you have any other thoughts please chime in too!  If not I'll ship this sometime later this week.

app/assets/javascripts/student_profile/record_service.jsx
```diff
@@ -66,20 +67,34 @@ import serviceColor from '../student_profile/service_color.js';
       const {nowMoment} = this.props;
       return {
         serviceTypeId: null,
-        providedByEducatorName: "",
+        providedByEducatorName: ''  ,
         dateStartedText: nowMoment.format('MM/DD/YYYY'),
-        estimatedEndDateText: nowMoment.format('MM/DD/YYYY')
+        estimatedEndDateText: ''
```
kevinrobinson: @snoopyuri suggested over email to change this to "the end of the current school year". We should do that, and this PR doesn't reflect that yet.

app/assets/javascripts/student_profile/record_service.jsx
```diff
@@ -64,22 +66,43 @@ import serviceColor from '../student_profile/service_color.js';
...
         dateStartedText: nowMoment.format('MM/DD/YYYY'),
-        estimatedEndDateText: nowMoment.format('MM/DD/YYYY')
+        estimatedEndDateText: this.defaultEstimatedEndDate(nowMoment).format('MM/DD/YYYY')
       };
     },
 
-    dateStartedMoment: function() {
-      const {dateStartedText} = this.state;
-      return moment.utc(dateStartedText, 'MM/DD/YYYY', true); // strict parsing
+    // The default is to assume the service will last until the end of the school year.
```
kevinrobinson: This isn't exact, and may differ by district or by particular programs (eg that are year-round).

kevinrobinson:
mergin'

PR #1399, opened 2018-01-21
Fix test warning for service upload fixture

kevinrobinson:
# Who is this PR for?
developers

test warning

# What does this PR do?
fixes a warning running Jest tests that comes from a bug in the test fixture data

Looking in production, I think the problem is in the format of the fixture data, not the UI code.  This is built on top of `toMoment` factored out in https://github.com/studentinsights/studentinsights/pull/1398, so that pull request needs to be merged first.


kevinrobinson:
Test failure is because this is uses the `toMoment` function factored out in #1398, so that pull request needs to be merged first.

kevinrobinson:
selfie

Issue #1386, opened 2018-01-17
Fix slowness/crashing on Student Profile Page


alexsoble:
Yesterday we received alerts from Logentries about timeouts on the Student Profile Page. Some of them had unusual error messages, like `"ALERT App boot timeout"`. These were unusual because the app wasn't booting during the time we got the errors. 

I saw that some of the URLs in the logs indicated the failures were happening during `_createNote` operations, so we reached out to one of the educators taking notes while this was happening to ask about their experience. 

Here's what they had to say:

> I find that Student Insights is painfully slow most of the time. We try to take notes during SST and there is consistently a lag. By the time you are able to load the student's page, the team has moved on to the next 2 cases and it is difficult to catch up. Yesterday was particularly slow.

@snoopyuri asked:

> Does the slowness happen on the computer in the conference room as well as your own in your office?  I want to come by and try it on some different machines.

The educator replied: 

> Great. Particularly in the conference room during SST. I know Larry and Felix have also expressed similar concerns about slowness at times. It seems to get stuck loading a student's page.

This is a serious problem — an educator finds the app "painfully slow" and it impedes the job they're meant to accomplish using the app. This is surprising because we've had performance issues with the School Overview page in the past, but never the Student Profile page. I've never experienced a slow Student Profile page on my own machine (while I've definitely experienced a slow School Overview before), and haven't heard any reports of Profile slowness from Uri. 

This has me wondering about the interaction between our app and the computer in the conference room that experiences particular slowness: is that computer running an older version of Internet Explorer that we haven't been targeting? Could it be running Internet Explorer 10 instead of Internet Explorer 11 like the rest of the district? That's just a first thought, the slowness could be because of any number of reasons. Whatever it is, we have to debug and fix for our users.

kevinrobinson:
This is great to surface.

If server logs don’t show this slowness but users report it, another thing
we could do is report timing from the client. That would reveal how slow
network and JS render time impact the UX.

alexsoble:
@kevinrobinson Oooh, that's a good suggestion! My instinct was to start with the server first, since the messages we've been getting from Heroku are from the server logs and not the client-side Rollbar monitoring. Just got a new `"ALERT App boot timeout"` error this morning, again associated with a Student Profile page load.

[...]

PR #1400, opened 2018-01-21
Add rake task to show response times from Heroku logs

kevinrobinson:
# Who is this PR for?
developers

# What problem does this PR fix?
visibility into endpoint response times

# What does this PR do?
adds a rake task for analyzing a dump of Heroku logs (eg, from logentries).  This is an initial investigation into https://github.com/studentinsights/studentinsights/issues/1392, just getting aggregate response times over the last week.

# Screenshot (if adding a client-side feature)
This is sorted by p50.

```
$ rake logs:analyze /Users/krobinson/Desktop/somerville-teacher-tool_2018-01-14_192745_2018-01-21_192745.log

count		2xx	3xx	4xx	5xx		p50	p90	p95		key
---	---	---	---	---	---	---	---	---	---	---	---
2	|	50%	0%	50%	0%	|	15051ms	0ms	0ms	|	admin/educators#authorization
7	|	57%	0%	43%	0%	|	10128ms	0ms	0ms	|	students#student_report
3	|	100%	0%	0%	0%	|	9759ms	0ms	0ms	|	service_uploads#past
5	|	80%	0%	20%	0%	|	5332ms	0ms	0ms	|	iep_documents#show
28	|	68%	0%	32%	0%	|	4503ms	20723ms	22956ms	|	students#restricted_notes
179	|	89%	4%	5%	2%	|	3007ms	11081ms	17091ms	|	schools#show
606	|	83%	2%	12%	3%	|	2647ms	11679ms	18803ms	|	students#show
201	|	13%	86%	0%	1%	|	779ms	1624ms	2257ms	|	devise/sessions#create
7	|	100%	0%	0%	0%	|	722ms	0ms	0ms	|	admin/educators#index
2	|	100%	0%	0%	0%	|	394ms	0ms	0ms	|	import_records#index
49	|	84%	16%	0%	0%	|	345ms	1124ms	1821ms	|	sections#show
109	|	58%	42%	0%	0%	|	344ms	1673ms	2447ms	|	homerooms#show
3	|	67%	0%	0%	33%	|	204ms	0ms	0ms	|	students#service
5	|	0%	0%	100%	0%	|	197ms	0ms	0ms	|	[hacking]
34	|	100%	0%	0%	0%	|	116ms	279ms	524ms	|	event_notes#update
3	|	100%	0%	0%	0%	|	100ms	0ms	0ms	|	service_uploads#index
187	|	99%	0%	1%	0%	|	98ms	489ms	810ms	|	event_notes#create
[...]

PR #1401, opened 2018-01-23
WIP – Print more measurements of student profile timing

alexsoble:
# Notes 

+ @kevinrobinson, this branch is my first WIP crack at investigating #1386. My plan is to find the student who triggered Logentry alerts, copy their data down locally, and use this profiling to see which parts of assembling `@serialized_data` take the most time on the server. 
+ What do you think about eventually shipping some version of this branch to production? Then we'd have lots of fine-grained student profile timing data in the logs on an ongoing basis.

### DIFF
--- a/app/controllers/students_controller.rb
+++ b/app/controllers/students_controller.rb
@@ -24,27 +24,70 @@ class StudentsController < ApplicationController
 
   def show
     student = Student.find(params[:id])
-    chart_data = StudentProfileChart.new(student).chart_data
+
+    serialized_student = log_timing('students#serialize_student_for_profile') do
+      serialize_student_for_profile(student)
+    end
+    feed = log_timing('students#feed') do
+      student_feed(student, restricted_notes: false)
+    end
+    chart_data = log_timing('students#chart_data') do
+      StudentProfileChart.new(student).chart_data
+    end
+    dibels = log_timing('students#dibels') do
+      student.student_assessments.by_family('DIBELS')
+    end
+    service_types_index = log_timing('students#service_types_index') do
+      ServiceSerializer.service_types_index
+    end
+    event_note_types_index = log_timing('students#event_note_types_index') do
+      EventNoteSerializer.event_note_types_index
+    end
+    educators_index = log_timing('students#educators_index') do
+      Educator.to_index
+    end
+    access = log_timing('students#access') do
+      student.latest_access_results
+    end
+    iep_document = log_timing('students#iep_document') do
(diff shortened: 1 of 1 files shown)


kevinrobinson:
@alexsoble Yeah, that seems like a good idea!  I think the most important thing that came up from the data in https://github.com/studentinsights/studentinsights/pull/1400#issue-290304976 is the variability.  Some hypotheses about why that might be are that it's from:

1. particular sets of data taking much longer than others (eg, a particular student has lots of data)
2. from changes because of increased load at different times
3. from just inherent variability in our Heroku setup

I added another commit on https://github.com/studentinsights/studentinsights/pull/1400#issuecomment-359641676 to see if there were particular students that were consistently slower.  Looking at individual paths, nothing jumped out at me (eg, paths that were slow at times were fast at others).  But graphing the time vs. latency looked like perhaps the slower requests are happening at similar times (eg, maybe because of concurrent requests).  I also tried running `ab` with a few requests to see if we had large variability at one set of time.  I didn't see anything surprising there except for when there are concurrent requests.

So I think looking at how to make these requests do less work is a great idea.  We could split these into separate requests (eg, don't load STAR math data for charts until the user navigates there, or don't load the "full case history" until they navigate there).  Those might be worth trying as well.

app/controllers/students_controller.rb
```diff
@@ -24,27 +24,70 @@ def authorize_for_districtwide_access_admin
 
   def show
     student = Student.find(params[:id])
-    chart_data = StudentProfileChart.new(student).chart_data
+
+    serialized_student = log_timing('students#serialize_student_for_profile') do
```
kevinrobinson: How about `students#show serialize_student_for_profile`?  That way it's clear it's a step within the `students#show` controller/action.

alexsoble: 👍

alexsoble:
@kevinrobinson Ship this or close it? Slowness in the student profile page doesn't seem to be a big issue anymore and we haven't heard slowness complaints since we shipped the memory improvements, so I could go either way.

kevinrobinson:
It seems great to get more latency data in general.  But I think it depends on whether we think we'll take action on that data in the next week or two.  If so, let's ship it and do that!  And if not, it's easy to add in this great kind of profiling when we do want to do something about it, so nothing much is lost if we close.

alexsoble:
@kevinrobinson OK! I don't think we'll take action in the next week or two so I'll close.

PR #1408, opened 2018-01-24
Validate uniqueness of school local ID

alexsoble:
# Who is this PR for?

Developers who set up new school districts in the future.  

The codebase doesn't throw an error when two schools have the same local ID like in #1407. 

# What does this PR do?

Uses Rails validations to validate school local ID and updates the spec.

### DIFF
--- a/app/importers/file_importers/student_section_assignments_importer.rb
+++ b/app/importers/file_importers/student_section_assignments_importer.rb
@@ -39,11 +39,12 @@ class StudentSectionAssignmentsImporter
   end
 
   def delete_rows
-    # Delete all stale rows no longer included in the import
-    # For the schools imported during this run of the importer
+    #Delete all stale rows no longer included in the import
+    #For the schools imported during this run of the importer
     StudentSectionAssignment.joins(:section => {:course => :school})
                             .where.not(id: @imported_assignments)
                             .where(:schools => {:local_id => @school_scope}).delete_all
+
   end
 
   def import_row(row)
--- a/app/importers/file_importers/student_section_assignments_importer.rb
+++ b/app/importers/file_importers/student_section_assignments_importer.rb
@@ -39,12 +39,11 @@ class StudentSectionAssignmentsImporter
   end
 
   def delete_rows
-    #Delete all stale rows no longer included in the import
-    #For the schools imported during this run of the importer
+    # Delete all stale rows no longer included in the import
+    # For the schools imported during this run of the importer
     StudentSectionAssignment.joins(:section => {:course => :school})
                             .where.not(id: @imported_assignments)
                             .where(:schools => {:local_id => @school_scope}).delete_all
-
   end
 
   def import_row(row)
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -1,6 +1,7 @@
 class School < ActiveRecord::Base
   extend FriendlyId
   friendly_id :local_id, use: :slugged
+  validates :local_id, presence: true, uniqueness: true
   has_many :students
   has_many :educators
   has_many :homerooms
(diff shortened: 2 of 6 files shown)


alexsoble:
@kevinrobinson This ended up touching some feature-side code in StudentSectionAssignmentsImporter when I meant it to only touch the School class + specs, but I think it's a reasonable change. Let me know what you think!

app/importers/file_importers/student_section_assignments_importer.rb
```diff
@@ -38,13 +38,19 @@ def filter
...
+  end
+
+  def rows_to_delete
+    return rows_not_imported if school_scope.nil?
 
+    rows_not_imported.where(:schools => {:local_id => @school_scope})
+  end
+
+  def delete_rows
+    rows_to_delete.delete_all
```
kevinrobinson: @alexsoble I don't really understand why this change is related, but this seems reasonable.  This might be worth explaining at the top of the importer class - this behavior is slightly different than the others and not obvious until you read these particular methods how changing school scope will impact what rows are deleted (not just which ones are imported).  Also probably worth spot checking after this ships.

alexsoble: @kevinrobinson Yep, it's really not related. This change was motivated by the StudentSectionAssignmentsImporter spec. Changing the schools factory broke the spec. I dove in and found the spec was breaking because of its `@school_scope`. This change to the feature-side code was the quickest way I could get it to pass and seemed like a reasonable change . . . I guess that's a bit hacky though huh? 😄 

I'll try to reproduce the spec failure up on Travis and maybe we'll find a better way to address it.

alexsoble: This Travis failure shows the exact same spec error, except in `educator_section_assignments_importer_spec.rb` instead of `student_section_assignments_importer_spec.rb`: 

https://travis-ci.org/studentinsights/studentinsights/builds/338115130?utm_source=github_status&utm_medium=notification

alexsoble: Here's the error: 

```
$ bundle exec rspec spec
..........................................................................................................................................................................................................................F.............................................................................................................................................................................................................................................................
Failures:
  1) EducatorSectionAssignmentsImporter#delete_rows happy path deletes all student section assignments except the recently imported one
     Failure/Error: expect(EducatorSectionAssignment.count).to eq(1)
     
       expected: 1
            got: 21
     
       (compared using ==)
     # ./spec/importers/file_importers/educator_section_assignments_importer_spec.rb:133:in `block (4 levels) in <top (required)>'
```

alexsoble: Ooooh, I dove in deeper and have a hunch that this has to do with the `SchoolFilter` class. The `SchoolFilter` uses school local IDs to do filtering, but on `master`, all schools created by our factories have a `local_id` of `nil`, which isn't a very realistic data setup. My current hypothesis is that giving those factory schools real local IDs changes the way that school filtering interacts with the importer in this spec setup.

alexsoble: @kevinrobinson I was wrong about this actually! Turns out `SchoolFilter` is never called in these specs, because the specs don't test `#import`, they skip it and go straight to testing `#import_row` methods, which assume the rows have already been filtered. 

I found that the issue had to do with the test instances of `EducatorSectionAssignmentsImporter` and `StudentSectionAssignmentsImporter` we were testing: these had `{school_scope: nil}`, which worked when every school created by FactoryGirl had a local_id of `nil`, but started failing once the FactoryGirl schools were assigned local_ids.

alexsoble:
Used git-rebase to force push a new solution onto this branch which adds the validation, touches spec and factory code, but doesn't touch any other feature-side code besides the validation.

alexsoble:
🚢 selfie validations-and-specs improvements!

PR #1411, opened 2018-01-24
Remove unused OJ gem

alexsoble:

Removes an unused gem and reduces our at-boot memory by a small amount. This PR is in the same spirit as #1409.

@kevinrobinson, I believe you experimented with using OJ for JSON serialization at one point, but I don't think we're using the OJ gem anymore. I searched the codebase with Sublime and GitHub (https://github.com/studentinsights/studentinsights/search?utf8=%E2%9C%93&q=oj&type=) and couldn't find any reference to OJ besides installing the gem.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,8 +23,6 @@ gem 'probability'
 gem 'rollbar'
 gem 'rails-erd', require: false, group: :development
 gem 'rubocop', require: false
-gem 'oj'
-gem 'oj_mimic_json'
 gem 'rubystats'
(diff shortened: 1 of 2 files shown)


alexsoble:
Oooooh, thank you for the explanation and the data-gathering @kevinrobinson! I'm going to go ahead and close this PR since clearly OJ is (a) used and (b) improving our JSON serialization. 

Also thanks for highlighting `Oj.optimize_rails`, this sounds like a great improvement to make! What's your sense of how we should prioritize this compared with our other performance improvement work?

PR #1416, opened 2018-01-24
Remove rubystats dependency

alexsoble:

Removes a gem that we're only calling in one site, for a very trivial reason, to reduce our at-boot memory by a small amount. This PR is in the same spirit as #1409.

We are only using Rubystats once, to make the shape of the data for a ⭐️ deprecated⭐️  model in our fake database slightly more realistic. Since overall our demo data is highly unrealistic, trying to make a single model more realistic is not a great reason to download an entire gem.

I replaced the `rubystats` implementation with a simple `sample`.

### DIFF
--- a/app/demo_data/fake_student.rb
+++ b/app/demo_data/fake_student.rb
@@ -246,7 +246,7 @@ class FakeStudent
   def add_deprecated_interventions
     15.in(100) do
       generator = FakeInterventionGenerator.new(@student)
-      intervention_count = Rubystats::NormalDistribution.new(3, 6).rng.round
+      intervention_count = [3, 6].sample
       intervention_count.times do
         intervention = Intervention.new(generator.next)
         intervention.save!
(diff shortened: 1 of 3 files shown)


app/demo_data/fake_student.rb
```diff
@@ -246,7 +246,7 @@ def add_discipline_incidents
   def add_deprecated_interventions
     15.in(100) do
       generator = FakeInterventionGenerator.new(@student)
-      intervention_count = Rubystats::NormalDistribution.new(3, 6).rng.round
+      intervention_count = [3, 6].sample
```
kevinrobinson: maybe `intervention_count = if rand() > 0.5 then 0 else (1..6).sample end` to keep more zero points, and less overall records so we don't all a lot more to the db?

alexsoble:
Implemented good suggestion from @kevinrobinson, tested this by running the `rake db:seed` script locally, all was ✅, time to 🚢 !

PR #1419, opened 2018-01-25
Add script that uses Heroku platform API to restart web dynos

### LINKED ISSUE #1418: Consider changing Heroku setup to restart app more often
Restarting the app resets the memory usage from object creation.

Heroku automatically cycles our app once per day: https://devcenter.heroku.com/articles/dynos

We could write a dump script that forces a restart once an hour, or implement something smarter like https://github.com/arches/whacamole.

alexsoble:

Our Heroku app throws a lot of memory errors. This week those memory errors escalated up to "memory quota vastly exceeded" and the app went down. We can see the memory going up in our metrics dashboard, we think there are memory leaks.

# What does this PR do?

Adds a script that can restart the web dynos of our Heroku apps programmatically, which could mitigate the memory leaks.

Fixes #1418.

# Implementation

We could use the Heroku Scheduler to run this every 1 hour. That would have the same effect as a dev team that shipped code or changed a var every hour, around the clock. 

This approach is soooo much lazier than digging in and trying to fix memory leaks, but part of my lazy developer brain wonders, "if it works, and there are no tradeoffs, why not implement it?"

http://threevirtues.com/

# Tradeoffs?

I wonder if users on the app experience anything when the dyno is being restarted. Heroku automatically restarts dynos whenever we reset ENV vars or ship code, which sometimes happens many times per day, so I can't imagine there's a terribly negative user impact. 

# Improvements to this strategy

This blog post has a write-up that summarizes our situation pretty much exactly: https://stormconsultancy.co.uk/blog/development/ruby-on-rails/automatically-restart-struggling-heroku-dynos-using-logentries/

They offer a more intelligent version of this script that restarts every time they get an R14 memory error. We could ship this first and then consider implementing their version later if needed. I wonder it might trigger too many restarts if we get lots of R14 errors. 

# Checklist

+ [x] Test script locally 
+ [x] Generate OAuth token
+ [ ] Set `HEROKU_OAUTH_TOKEN` and `HEROKU_APP_NAME` env variables
+ [ ] Set up the task in Heroku schedule:

```
rake 'heroku_platform_api:restart_web_dynos[5]'
rake 'heroku_platform_api:restart_web_dynos[17]'
```

### DIFF
--- a/lib/tasks/heroku_platform_api.rake
+++ b/lib/tasks/heroku_platform_api.rake
@@ -15,6 +15,7 @@ namespace :heroku_platform_api do
     now_hour = Time.now.in_time_zone('Eastern Time (US & Canada)').hour
 
     if target_hour == now_hour
+      puts "It feels like a good time to restart our web dynos!"
       heroku = PlatformAPI.connect_oauth(ENV.fetch('HEROKU_OAUTH_TOKEN'))
       dynos = heroku.dyno.list(ENV.fetch('HEROKU_APP_NAME'))
 
@@ -23,13 +24,15 @@ namespace :heroku_platform_api do
           puts "Found web dyno!"; puts
 
           name = dyno.fetch('name')
-          puts "Restaring dyno #{name}..."
+          puts "Restaring dyno #{name}..."; puts
           heroku.dyno.restart(ENV.fetch('HEROKU_APP_NAME'), name)
 
-          puts "Sleeping for 60 seconds so we get a rolling restart ..."
+          puts "Sleeping for 60 seconds so we get a rolling restart ..."; puts
           sleep 60
         end
       end
+    else
+      puts "It doesn't feel like a good time to restart our web dynos right now."
     end
   end
 end
--- a/lib/tasks/heroku_platform_api.rake
+++ b/lib/tasks/heroku_platform_api.rake
@@ -1,0 +1,13 @@
+namespace :heroku_platform_api do
+  desc 'Restart heroku web dyno'
+  task restart_web_dynos: :environment do
+    heroku = PlatformAPI.connect_oauth(ENV.fetch('HEROKU_OAUTH_TOKEN'))
+    dynos = heroku.dyno.list(ENV.fetch('HEROKU_APP_NAME'))
+
+    dynos.each do |dyno|
+      if dyno.fetch("type") == "web"
+        heroku.dyno.restart(ENV.fetch('HEROKU_APP_NAME'), dyno.fetch('name'))
+      end
+    end
+  end
+end
--- a/lib/tasks/heroku_platform_api.rake
+++ b/lib/tasks/heroku_platform_api.rake
@@ -20,7 +20,14 @@ namespace :heroku_platform_api do
 
       dynos.each do |dyno|
         if dyno.fetch("type") == "web"
-          heroku.dyno.restart(ENV.fetch('HEROKU_APP_NAME'), dyno.fetch('name'))
+          puts "Found web dyno!"; puts
+
+          name = dyno.fetch('name')
+          puts "Restaring dyno #{name}..."
+          heroku.dyno.restart(ENV.fetch('HEROKU_APP_NAME'), name)
+
+          puts "Sleeping for 60 seconds so we get a rolling restart ..."
+          sleep 60
         end
       end
     end
--- a/lib/tasks/heroku_platform_api.rake
+++ b/lib/tasks/heroku_platform_api.rake
@@ -1,12 +1,27 @@
 namespace :heroku_platform_api do
   desc 'Restart heroku web dyno'
-  task restart_web_dynos: :environment do
-    heroku = PlatformAPI.connect_oauth(ENV.fetch('HEROKU_OAUTH_TOKEN'))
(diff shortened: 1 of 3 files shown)


kevinrobinson:
@alexsoble Seems worth trying!  I think the main tradeoff is the slower start up time.  If we bump to 1gb memory then perhaps daily restarts are sufficient.  But if not then 6am / 6pm seems like a good points, that way the longest dyno lifetime is 12 hours.

alexsoble:
@kevinrobinson Can you say more about the slower start up time?

alexsoble:
@kevinrobinson Also, I emailed you and Uri about this a few minutes ago, but just so it's up on GitHub: 

Uri and I set the app to use two 2X Standard dynos. 

Let's see what we see this week . . . if there are no memory issues or crashes, we know for sure that the issue was related to memory allocated for us on the Heroku box! If we still have issues, back to the drawing board.

This PR there for us if we decide we need it. In my mind, it makes sense to deploy this PR if the bigger boxes succeed, and we want to find a way to avoid hitting those memory limits without the bigger boxes.

kevinrobinson:
@alexsoble re: startup time, just that the app has some period of time where it's slower to respond during a restart or on the first requests when nothing cached in memory yet.  It's probably not meaningful though, but that's the tradeoff of restarting - any penalties during the tear down/start up phases (which here I think are minimal).

alexsoble:
This latest commit (a913d8a4e45b7fc9fc0bbba00a215a34e7b4adfc) lets the rake task take an argument, so it can be scheduled to run at (say) 5AM and 5PM ET using the built in Heroku Scheduler. 

I think 5AM and 5PM ET are the best times to use for restarts, because there are few users on before 5AM and after 5PM.

Here's what this would look like using two rake tasks in the scheduler: 

```
rake 'heroku_platform_api:restart_web_dynos[5]'
rake 'heroku_platform_api:restart_web_dynos[17]'
```

alexsoble:
@kevinrobinson I think we should ship this before early Monday morning, what say you?

The larger Heroku boxes slow down the time-to-swap-memory by a lot, but the app still hits swap memory within about a day. And I saw earlier today that a manual restart has a huge effect on memory, bringing swap down to zero:

result of restart

So I think this would be a good addition to our "memory error mitigation stack" while we work on fixing the leak(s)!

lib/tasks/heroku_platform_api.rake
```diff
@@ -0,0 +1,28 @@
...
+    now_hour = Time.now.in_time_zone('Eastern Time (US & Canada)').hour
+
+    if target_hour == now_hour
+      heroku = PlatformAPI.connect_oauth(ENV.fetch('HEROKU_OAUTH_TOKEN'))
+      dynos = heroku.dyno.list(ENV.fetch('HEROKU_APP_NAME'))
+
+      dynos.each do |dyno|
+        if dyno.fetch("type") == "web"
+          heroku.dyno.restart(ENV.fetch('HEROKU_APP_NAME'), dyno.fetch('name'))
+        end
```
kevinrobinson: Let's do a rolling restart - sleep for say a minute after each dyno restart, so we don't bring them all down at once.

lib/tasks/heroku_platform_api.rake
```diff
@@ -0,0 +1,28 @@
...
+  #
+  # $ rake 'heroku_platform_api:restart_web_dynos[5]'
+  # $ rake 'heroku_platform_api:restart_web_dynos[17]'
+  #
+  # Good blog post on rake tasks with arguments from Thoughtbot:
+  # https://robots.thoughtbot.com/how-to-use-arguments-in-a-rake-task
+
+  task :restart_web_dynos, [:hour] => :environment do |t, args|
+    target_hour = args[:hour].to_i
+    now_hour = Time.now.in_time_zone('Eastern Time (US & Canada)').hour
```
kevinrobinson: Would it simplify to run the task every hour, and write `if now_hour == 5 || now_hour == 17` in the code?

alexsoble: Yeah, hardcoding in the now_hour values would work too. My thinking here is that allowing `hour` to be set as an argument to the rake task in the Heroku Scheduler would make it simple to adjust the timing of the restarts.

 If we want to change how many times per day we restart, or what times we restart, we just log into the Heroku UI and change the argument on the task. No PR, merge, deploy process needed.

lib/tasks/heroku_platform_api.rake
alexsoble: This is a good idea!

alexsoble:
Here's what the job output looks like with more logging + rolling restarts:

screen shot 2018-01-29 at 4 11 50 pm

I did some more manual testing and verified that each restart pushes memory use way down in the Heroku metrics. 😎 

I think this is good to 🚢 , we have rolling restarts in place and we can always refactor how we set the timing if needed. 💻

alexsoble:
Oooh, when I set the `HEROKU_OAUTH_TOKEN` ENV var in production, I get:

```
 ▸    Warning: The "HEROKU_" namespace is protected and shouldn't be used.
```

alexsoble:
Set new ENV vars and set up the rake tasks in 🕐 Heroku Scheduler🕥  , so this should be up and running! Will keep an eye on the job (and the memory) to confirm. 👍

PR #1421, opened 2018-01-27
Fix the educator_id for the admin user in TestPals

kevinrobinson:
# Who is this PR for?
developers using the demo site

The value for `URI_ID` isn't set in the demo environment, and can't be since `TestPals` doesn't give it a stable id.  This came up trying to test the rake task in https://github.com/studentinsights/studentinsights/pull/1286 on the demo site.

# What does this PR do?
Fixes the id in `TestPals` so we can add a stable `URI_ID` to the demo site.

### DIFF
--- a/spec/support/test_pals.rb
+++ b/spec/support/test_pals.rb
@@ -60,7 +60,7 @@ class TestPals
     # Uri works in the central office, and is the admin for the entire
     # project at the district.
     @uri = Educator.create!(
-      id: 1,
+      id: 999999,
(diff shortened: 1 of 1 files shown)


kevinrobinson:
Updated the id - Rails assumes it owns this column and so it will try to grab it until it conflicts, then it will skip it and go to the next one.  To work around, just set this to a large number.  This code is only for dev/demo/test anyway.

kevinrobinson:
selfie

PR #1275, opened 2017-11-26
Rake task check foreign keys

gabivoicu:
# Who is this PR for?

# What problem does this PR fix?

Checks that foreign keys on an object point to an existing other objects; if not, prints the ID of the missing object.

# What does this PR do?

Add a rake task that does the above. To run the task:
`rake missing_foreign_key_relations:scan`

# Screenshot (if adding a client-side feature)

# Checklist

+ [x] Tested latest version in an Internet Explorer virtual machine? *(If the PR adds Javascript.)*

# Questions / Concerns
- feedback wanted on better names for task, namespace, etc
- I am posting below the output of the rake task, let me know if I can improve it in any way
- locally, the task takes 26 seconds to complete, w/ a total of 447 IDs missing (I deleted some things locally to test)

# Output
```
Missing School IDs on Course
2 records missing an object. Missing IDs: [9]

Missing School IDs on Educator
3 records missing an object. Missing IDs: [nil, 9]

Missing EventNote IDs on EventNoteAttachment
0 records missing an object. Missing IDs: []

Missing EventNote IDs on EventNoteRevision
0 records missing an object. Missing IDs: []

Missing Educator IDs on EventNote
326 records missing an object. Missing IDs: [1]

Missing EventNoteType IDs on EventNote
0 records missing an object. Missing IDs: []

Missing Student IDs on EventNote
0 records missing an object. Missing IDs: []

Missing Educator IDs on Homeroom
1 records missing an object. Missing IDs: [nil]

Missing School IDs on Homeroom
1 records missing an object. Missing IDs: [9]

Missing Student IDs on IepDocument
0 records missing an object. Missing IDs: []

Missing Educator IDs on Intervention
39 records missing an object. Missing IDs: [1]

Missing InterventionType IDs on Intervention
0 records missing an object. Missing IDs: []

Missing Student IDs on Intervention
0 records missing an object. Missing IDs: []

Missing Course IDs on Section
0 records missing an object. Missing IDs: []

Missing Educator IDs on ServiceUpload
4 records missing an object. Missing IDs: [nil]

Missing Educator IDs on Service
6 records missing an object. Missing IDs: [1]

Missing ServiceType IDs on Service
0 records missing an object. Missing IDs: []

Missing ServiceUploads IDs on Service
15 records missing an object. Missing IDs: [nil]

Missing Student IDs on Service
0 records missing an object. Missing IDs: []

Missing Assessment IDs on StudentAssessment
0 records missing an object. Missing IDs: []

[...]

PR #1422, opened 2018-01-27
Update foreign keys to be required by default, opt-out in three cases

kevinrobinson:
# Who is this PR for?
developers

We made an awesome rake task for looking at nil associations when increasing the strictness for foreign keys (not allowing orphans or nils).  Without that historical context, this task implies that we don't want these nil records, but we do as discussed here: https://github.com/studentinsights/studentinsights/pull/1275#issuecomment-347276598.

Also, Rails 5 changed `belongs_to` relations to be "non-nil" by default but we're opted-out because of those three associations where we want to allow nil values.  So there's nowhere in the code right now that expresses what we want.

# What does this PR do?
Updates the rake task to remove the three associations for which `nil` values are normal and expected.  Updates the default `belongs_to` behavior to be stricter and disallow `nil` values by default, and update the three associations where we intentionally allow `nils` with the `optional: true` flag so this is expressed in code and enforced.

### DIFF
--- a/lib/tasks/scan_for_missing_records.rake
+++ b/lib/tasks/scan_for_missing_records.rake
@@ -71,16 +71,6 @@ namespace :missing_foreign_key_relations do
         puts "#{missing.length} records missing an object. Missing IDs: #{missing.uniq}"
         puts ""
 
-        puts "Missing Educator IDs on Homeroom"
-        missing = []
-        Homeroom.find_each do |homeroom|
-            if !homeroom.educator
-                missing << homeroom.educator_id
-            end
-        end
-        puts "#{missing.length} records missing an object. Missing IDs: #{missing.uniq}"
-        puts ""
-
         puts "Missing School IDs on Homeroom"
         missing = []
         Homeroom.find_each do |homeroom|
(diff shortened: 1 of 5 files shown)


kevinrobinson:
cc @gabivoicu @alexsoble this seems reasonable?

lib/tasks/scan_for_missing_records.rake
```diff
@@ -1,244 +1,214 @@
 desc 'Scan for missing foreign key records'
 namespace :missing_foreign_key_relations do
-    task scan: :environment do
```
kevinrobinson: The first commit is the substantive change; second is updating whitespace from 4 spaces to 2.

gabivoicu:
@kevinrobinson This looks great to me, it sounds like we are making more obvious/explicit the fact that we want to allow `nil` for those associations and it's now in a place where a developer can see it when working with that model. 👍🏻👍🏻👍🏻 for me!

lib/tasks/scan_for_missing_records.rake
gabivoicu: Ooops, I had not realized at all I used 4 space. 🐍 habits die hard.

alexsoble:
@kevinrobinson @gabivoicu I like this change as well! Making `belongs_to` validation out-out instead of opt-in is good future-proofing. 

At first I thought that the `school_id` foreign key on the students table should be optional, since I remembered we've had students with `school_id == nil` in the past. But when I went to double check in the production Somerville data, I found:

```
irb(main):001:0> Student.where(school_id: nil).count
=> 0
```

I think stricter is better than looser on these validations, so let's :ship: as-is, and if we find a data modeling situation in the future that's best solved by student records without school_id foreign keys, we can consider changing from there. 👍

PR #1423, opened 2018-01-27
Patch/planet path error pages

kevinrobinson:
# Who is this PR for?
developers mostly

The asset paths in the error pages are relative, so they make 404s when the user sees them (eg, from another 404).  This triggers Rollbar errors.

# What does this PR do?
Fixes them to absolute paths.


kevinrobinson:
selfie

alexsoble:
👍

PR #1424, opened 2018-01-27
Update developer users in Mixpanel filters

kevinrobinson:
# Who is this PR for?
admin and product

Filter out developer users from usage reports; we rotated these ids.  We'll have to rework how we do this for the second district, but this fixes up for now.


kevinrobinson:
selfie

PR #1427, opened 2018-01-27
Add StreamingCsvTransformer and optimize CsvRowCleaner to cut import memory usage

kevinrobinson:
# Who is this PR for?
all users

The import job uses a lot of memory and will be killed by Heroku when it hits the threshold.  This reduces the memory usage.

# What does this PR do?
It adds a `StreamingCsvTransformer` that uses the same interface as `CsvTransformer` but does one iteration on the data, removing two iteration passes for cleaning and transforming data that were done before the iteration to import.

This has a significant impact on memory usage.  The test case uses a production attendance data file and a benchmarking task included in this PR.  The timing method is from https://dalibornasevic.com/posts/68-processing-large-csv-files-with-ruby.  On this benchmark, the change reduces memory usage from 1002MB to 60MB.

```
CsvTransformer
  $ bundle exec rake benchmark:csv_import ~/Desktop/attendance_export.txt
  24530292 bytes
  459407 processed_rows_count
  Time: 101.51
  Memory: 1002.36 MB

StreamingCsvTransformer
  $ bundle exec rake benchmark:csv_import ~/Desktop/attendance_export.txt
  24530292 bytes
  459407 processed_rows_count
  Time: 98.15
  Memory: 60.22 MB
```

The tests are updated to test the the substance of both classes work as expected.

We could do this for each importer, but to start this just swaps it in for `AttendanceImporter`.  This importer is both lower risk (eg, it doesn't involve user login) and higher impact (it's one of the larger files) so a good one to start with.

This also includes the same workaround for escaped quotes from https://github.com/studentinsights/studentinsights/pull/1393.

### DIFF
--- a/app/importers/data_transformers/csv_row_cleaner.rb
+++ b/app/importers/data_transformers/csv_row_cleaner.rb
@@ -1,14 +1,19 @@
+
 class CsvRowCleaner < Struct.new :row
+  DATE_HEADERS = [:event_date, :date_taken, :assessment_date]
+  BOOLEAN_VALUES = [true, false, '1', '0', 1, 0, 'true', 'false']
+  BOOLEAN_HEADER_NAMES = [:absence, :tardy, :has_exact_time]
 
   def dirty_data?
     !clean_date? || !clean_booleans?
   end
 
   def transform_row
-    row[date_header] = parsed_date if has_dates?
+    row[date_header] = parsed_date unless date_header.nil?
     row
   end
 
+  private
   def clean_date?
     return true if date_header.blank?       # <= No dates here, so no more checks to do
     return false if date_from_row.blank?    # <= Column that should be a date is blank
@@ -18,24 +23,19 @@ class CsvRowCleaner < Struct.new :row
   end
 
   def clean_booleans?
-    return true if boolean_headers.blank?
-    boolean_headers.all? do |header|
-      row[header].in? [true, false, '1', '0', 1, 0, 'true', 'false']
+    headers.each do |header|
+      next unless header.in? BOOLEAN_HEADER_NAMES
+      return false unless row[header].in? BOOLEAN_VALUES
     end
+    true
   end
-
-  private
 
   def headers
-    row.headers
-  end
-
-  def date_headers
-    [:event_date, :date_taken, :assessment_date]
+    @headers ||= row.headers
   end
 
   def date_header
-    headers.detect { |header| header.in? date_headers }
+    @date_header ||= headers.detect { |header| header.in? DATE_HEADERS }
   end
 
   def date_from_row
@@ -43,26 +43,15 @@ class CsvRowCleaner < Struct.new :row
   end
 
   def parsed_date
-    return date_from_row if date_from_row.is_a?(DateTime)
-    Date.parse(date_from_row)
-  rescue ArgumentError
-    false
+    @parsed_date ||= begin
+      return date_from_row if date_from_row.is_a?(DateTime)
+      Date.strptime(date_from_row, '%Y-%m-%d')
+    rescue ArgumentError
+      false
+    end
   end
 
   def date_out_of_range
     (parsed_date < (Date.today - 50.years)) || (Date.today < parsed_date)
   end
-
-  def has_dates?
-    headers.any? { |header| header.in? date_headers }
-  end
-
-  def boolean_header_names
-    [:absence, :tardy, :has_exact_time]
-  end
-
-  def boolean_headers
-    headers.select { |header| header.in? boolean_header_names }
-  end
-
 end
--- a/app/importers/data_transformers/csv_row_cleaner.rb
+++ b/app/importers/data_transformers/csv_row_cleaner.rb
@@ -1,4 +1,3 @@
-
 class CsvRowCleaner < Struct.new :row
   DATE_HEADERS = [:event_date, :date_taken, :assessment_date]
   BOOLEAN_VALUES = [true, false, '1', '0', 1, 0, 'true', 'false']
--- a/app/importers/file_importers/csv_downloader.rb
+++ b/app/importers/file_importers/csv_downloader.rb
@@ -23,7 +23,7 @@ class CsvDownloader
 
   def download_file
     @log.write("\nDownloading #{@remote_file_name}...")
-    
+
     downloaded_file = @client.download_file(@remote_file_name)
     File.read(downloaded_file)
   end
--- a/app/importers/file_importers/csv_downloader.rb
+++ b/app/importers/file_importers/csv_downloader.rb
@@ -10,9 +10,9 @@ class CsvDownloader
   end
 
   def get_data
-    file = download_file
+    file_contents = download_file
 
-    data = @transformer.transform(file)
+    data = @transformer.transform(file_contents)
 
     CleanupReport.new(
       @log, @remote_file_name, @transformer.pre_cleanup_csv_size, data.size
@@ -23,12 +23,9 @@ class CsvDownloader
 
   def download_file
     @log.write("\nDownloading #{@remote_file_name}...")
-
+    
     downloaded_file = @client.download_file(@remote_file_name)
-
-    File.read(downloaded_file).encode('UTF-8', 'binary', {
-      invalid: :replace, undef: :replace, replace: ''
-    })
+    File.read(downloaded_file)
   end
 
 end
--- a/app/importers/file_importers/streaming_csv_transformer.rb
+++ b/app/importers/file_importers/streaming_csv_transformer.rb
@@ -1,0 +1,66 @@
+require 'csv'
+
+# This implements:
+#   the Transformer interface {transform, pre_cleanup_csv_size}
+#   our ad-hoc CSV data interface {each_with_index, size}
+class StreamingCsvTransformer
+  def initialize(options = {})
+    @headers = options.key?(:headers) ? options[:headers] : true
+    @total_rows_count = nil
+    @processed_rows_count = nil
+  end
+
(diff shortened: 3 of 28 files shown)


app/importers/file_importers/streaming_csv_transformer.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    CSV.parse(@csv_string, {
+      headers: @headers,
+      header_converters: :symbol,
+      encoding: 'binary:UTF-8',
+      converters: lambda { |h| nil_converter(h) }
+    }) do |row, index|
+      @total_rows_count = @total_rows_count + 1
+      cleaner = CsvRowCleaner.new(row)
+      next if cleaner.dirty_data?
```
kevinrobinson: This is a lot of processing, that from the perspective of the importers is hidden inside the call to `CsvDownload` which sounds like it'd be a pretty simple operation

 I think ideally we'd push this out to importer classes; looking at production logs for "rows of data in" to see how much this is used it only impacts attendance import.  If this PR lands, we could factor this out there, and then it seems there's no need to be running these checks for other importers.  Nothing to do now, just noting it since I didn't remember about this.

kevinrobinson:
fwiw I don't really understand why the impact on memory usage is so significant here.  The difference isn't as big on smaller CSV files (eg 2000 lines) but I ran this a bunch of times and with different variations and could consistently reproduce.

kevinrobinson:
@alexsoble It'd probably also good to bump the dyno size up on the worker too, even if just for this month.

kevinrobinson:
oops, I screwed up the last change to where the UTF8 encoding happens and busted all the tests back again :)  I'll fix this up tomorrow.

kevinrobinson:
I took another pass at this, since the memory numbers were surprising to me.  This involved looking at two other ways to measure memory (`get_process_mem` and `memory_profiler`).  `memory_profile` added significant overhead and couldn't run the whole file so I ran it on the 10% data set, and then also verified similar patterns on a 50% data set (that data isn't included here).

Reading the allocation counts in `memory_profiler` was the most helpful.  `get_process_mem` returned similar values to the previous `print_memory_usage` method and both had noise in repeated runs.  `memory_profiler` gave deterministic allocation counts and more detailed information about where these were occurring.  This led to changes in the `CsvRowCleaner` to hoist constant values and to cache repeated computation to avoid the intermediate allocations.

Here's the comparison using the same measurement as earlier in this PR, on a test case
```
CsvTransformer
  $ bundle exec rake benchmark:csv_import ~/Desktop/attendance_export.txt
  24530292 bytes
  459407 processed_rows_count
  Time: 73.24sec
  Memory: 733.96 MB

StreamingCsvTransformer
  $ bundle exec rake benchmark:csv_import ~/Desktop/attendance_export.txt
  24530292 bytes
  459407 processed_rows_count
  Time: 55.49sec
  Memory: 66.12 MB
```

So for `StreamingCsvTransformer` this shows the tradeoff in the `CsvCleanerRow` caching - it adds 10% memory usage, but reduces execution time by around 50%.

Here's a more detailed comparison of how the changes to `CsvRowCleaner` reduce allocated memory, keeping `StreamingCsvTransfomer` constant:
```
StreamingCsvTransformer on the 10% data set (with CsvRowCleaner optimizations):
  2440757 bytes
  45987 processed_rows_count
  Time: 11.57sec
  Total allocated: 264MB
  streaming_csv_transformer.rb: 5.7MB
  csv_row_cleaner.rb: 5.5MB

StreamingCsvTransformer on the 10% data set (no CsvRowCleaner optimizations):
  2440757 bytes
  45987 processed_rows_count
  Time: 27.28sec
  Total allocated: 662MB
  csv_row_cleaner.rb: 183MB
  streaming_csv_transformer.rb: 4.3MB
```

All this to say, this should improve memory allocations in an observable way for the import job.

app/importers/file_importers/csv_downloader.rb
```diff
@@ -10,9 +10,9 @@ def initialize(log:, client:, remote_file_name:, transformer:)
   end
 
   def get_data
-    file = download_file
+    file_contents = download_file
 
-    data = @transformer.transform(file)
+    data = @transformer.transform(file_contents)
```
kevinrobinson: Clarifying that the previous API to `#transform` was to pass the bytes of the CSV file as as string, not a file handle.

spec/importers/data_transformers/csv_row_cleaner_spec.rb
```diff
@@ -34,7 +34,7 @@
         context 'date is not in the future or the far past' do
           let(:row) { CSV::Row.new([:event_date, :pizza_slices_count], ['2015-10-11', '3']) }
           it 'returns true' do
-            expect(row_cleaner.clean_date?).to eq true
+            expect(row_cleaner.send(:clean_date?)).to eq true
```
kevinrobinson: I made all these methods private since only two are ever called.

spec/importers/file_importers/students_importer_spec.rb
```diff
@@ -5,7 +5,7 @@
   describe '#import_row' do
 
     context 'good data' do
-      let(:file) { File.open("#{Rails.root}/spec/fixtures/fake_students_export.txt") }
+      let(:file) { File.read("#{Rails.root}/spec/fixtures/fake_students_export.txt") }
```
kevinrobinson: This was a test bug before, the `CsvDownloader` code passes a string of the CSV file to the transformer class.  But because of the way `transform` was implemented before, it just so happened that passing a file in the test would work as well.  This isn't the case anymore with these changes, so this fixes the test bug to correctly pass a string.

app/importers/data_transformers/csv_row_cleaner.rb
```diff
@@ -1,14 +1,19 @@
+
 class CsvRowCleaner < Struct.new :row
+  DATE_HEADERS = [:event_date, :date_taken, :assessment_date]
+  BOOLEAN_VALUES = [true, false, '1', '0', 1, 0, 'true', 'false']
+  BOOLEAN_HEADER_NAMES = [:absence, :tardy, :has_exact_time]
```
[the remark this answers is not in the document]
alexsoble: 👍

app/importers/data_transformers/csv_row_cleaner.rb
```diff
@@ -18,51 +23,35 @@ def clean_date?
...
   end
 
   def parsed_date
-    return date_from_row if date_from_row.is_a?(DateTime)
-    Date.parse(date_from_row)
-  rescue ArgumentError
-    false
+    @parsed_date ||= begin
+      return date_from_row if date_from_row.is_a?(DateTime)
+      Date.strptime(date_from_row, '%Y-%m-%d')
```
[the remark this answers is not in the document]
alexsoble: Awesome! Great catch here. 😎

app/importers/file_importers/csv_downloader.rb
alexsoble: Yeah, good name change. 👍

alexsoble:
@kevinrobinson Merged in upstream changes with #1438.

alexsoble:
@kevinrobinson For review I'm going to do a manual import process with Somerville attendance data. I'll run a small import process by narrowing down the filter to only include 1 school. That will check for any things that are different with the production import environment that we might not be thinking of. It'll also produce some performance data.

spec/importers/data_transformers/csv_row_cleaner_spec.rb
alexsoble: This is such a sneaky way to call a private method!! 👀

alexsoble:
Turns out that trying to benchmark a production data import takes a loooong time — too long to make it a useful review activity. This must be what the "this gives detailed information, but incurs significant overhead" code comment is about in `csv_benchmark.rake`. 😄 

Instead of trying to profile a production data import, I'll replicate your data from `csv_benchmark.rake` @kevinrobinson.

alexsoble:
Alrighty! I ran `csv_benchmark.rake` against three different scenarios: StreamingCsvTransformer, New CsvTransformer (no streaming, with your refactors), and Old CsvTransformer (no refactors, on `master`). 

I used the default `profile_memory` setting, which was probably a mistake because it made the benchmarking take a long time, which in turn led me to work with a much smaller file size. 

Here's what I found (with some light editing of the output for readability):

```
# Old CsvTransformer, no streaming:

1327208 bytes
 processed_rows_count: 24995
Time: 17.56sec
Printing memory report ...
Total allocated: 371,323,667 bytes (6,324,958 objects)
```

```
# New CsvTransformer, no streaming:

1327208 bytes
 processed_rows_count: 24994
Time: 6.79sec
Printing memory report ...
Total allocated: 179,759,751 bytes (2,825,625 objects)
```

```
# StreamingCsvTransformer:

1327208 bytes
 processed_rows_count: 24994
Time: 8.44sec
Printing memory report ...
Total allocated: 147,767,572 bytes (2,325,739 objects)
Total retained:  5644 bytes (55 objects)
```

These are fantastic results! ✨ 🎉 ✨ 🎉   

They show that the refactored CsvTransformer uses 51.8% less memory than the existing CsvTransformer on `master`, and the StreamingCsvTransformer uses 17.9% less memory than the refactored CsvTransformer. Interestingly, the StreamingCsvTransformer took a little more time to run than the refactored CsvTransformer in this sample. Some of those differences could have to do with the size of the data being transformed, and since your sample size was much larger it's probably a much better model of the production data.

Awesome refactoring and improvements here @kevinrobinson! And `csv_benchmark.rake` is a fantastic tool, thank you for setting it up and bringing a scientific 🔬  angle to testing our CSV parsing performance! Let's 🚢  this.

PR #1429, opened 2018-01-29
WIP – Tolerant attendance importer

alexsoble:
# Who is this PR for?

New Bedford import process! (Benefits Somerville's too.)

# What does this PR do?

Unblocks attendance import for New Bedford by making attendance importer tolerant of invalid rows.

Instead of blowing up when it hits the first invalid row, the attendance importer skips those rows, and keeps a count of invalid rows plus error messages.


alexsoble:
Closing in favor of #1440, which is rebased on top of the other New Bedford-related attendance import refactor and so in a better position to get merged in smoothly. 😎 🚭

PR #1434, opened 2018-01-29
Make 'tmp' directory if it doesn't exist locally; use Dir.exist?

alexsoble:
# Who is this PR for?

Developers.

The local development cleanup script deletes the `tmp` directory entirely. The import script expects a `tmp` directory. This PR creates a `tmp` directory if it's missing. It also uses more idiomatic Ruby.


alexsoble:
selfie

PR #1439, opened 2018-01-31
Add "only_recent_attendance" import option; refactor passing down import options

alexsoble:
# Who is this PR for?

+ New Bedford, so that we can run an initial import process that imports all New Bedford attendance data, not just recent data.
+ Developers, so that adding new arguments to the file importer classes is much easier by making those classes more flexible with their arguments. 

# What does this PR do?

## Big changes

+ Adds a new argument for the import process thor interface: `:only_recent_attendance`
+ Refactors the file importer options to read from an `:options` hash instead of passing in arguments one-by-one to `Struct.new`

Using `Struct.new` makes the number of options passed inflexible: 

```
# Using Struct.new:

2.3.0 :001 > class Fruit < Struct.new :color, :size
2.3.0 :002?>   end
2.3.0 :003 > Fruit.new('red', 'small', 'tasty')
ArgumentError: struct size differs
	from (irb):3:in `initialize'
	from (irb):3:in `new'
	from (irb):3
```

```
# Using initialize + options hash:

2.3.0 :004 > class Veggie
2.3.0 :005?>   def initialize(options:)
2.3.0 :006?>     end
2.3.0 :007?>   end
2.3.0 :012 > Veggie.new(options: {color: 'red'})
 => #<Veggie:0x007fe31225f678> 
2.3.0 :013 > Veggie.new(options: {
2.3.0 :014 >       color: 'red',
2.3.0 :015 >       size: 'small',
2.3.0 :016 >       taste: 'delicious'})
 => #<Veggie:0x007fe312247d70> 
```

## Smaller changes 

+ Remove `:progress_bar` option, which to the best of my knowledge is totally unused by developers at this point, to slim down the args being passed in
+ Instead of passing in `:client` as an argument, each class should know what their client is; whether an importer needs STAR or X2 is constant

# Checklist 

+ [x] Add spec coverage for `:only_recent_attendance` test cases
+ [x] QA with Somerville data import
+ [ ] After deploy, add ` --only-recent-attendance` flag to Somerville import

Issue #1371, opened 2018-01-10
Deploy Insights New Bedford app


alexsoble:
Using this Issue as a thread to post updates and notes on the deployment process, will close when the app is successfully set up and notes are moved over to proper documentation. 💻 🏫

alexsoble:
Couple of config notes:

```
DISTRICT_KEY: new_bedford
DISTRICT_NAME: New Bedford
SHOULD_USE_LDAP: false    # we're using database auth to start
USE_QUOTAGUARD: false    # quotaguard is only used to talk to somerville ldap
```

alexsoble:
Need to set buildpacks in Heroku (can use UI) before first deploy:

```
heroku/nodejs
heroku/ruby
```

![screen shot 2018-01-10 at 1 16 08 pm](https://user-images.githubusercontent.com/3209501/34790776-74ca7216-f608-11e7-9951-aeb771f26f3e.png)

alexsoble:
```
heroku run rake db:migrate
```

alexsoble:
Running first data import process!

alexsoble:
Students import ran without any errors, attendance import ran but then crashed with:

```
 🚨  🚨  🚨  🚨  🚨  Error! Couldn't find Student 
```

alexsoble:
Precomputing all the overview pages because some of the bigger ones don't load.

alexsoble:
Shipping #1373 to unbreak Educator Permissions Overview page for New Bedford.

alexsoble:
Set up three critical tasks to run nightly in the background using Heroku Scheduler:

+ `rake precompute:school_overview`
+ `rake chores:import_data_in_background`
+ `rake chores:update_searchbar_data_for_educators_who_log_in`

I was wondering why the School Overview page performance seemed so poor on the New Bedford app. Then I looked at the logs and saw precomputed JSON wasn't being loaded. Then I looked at the code and remembered that precomputed school overview JSON is only loaded if the job has been run within the last day. So it's really critical to get these background processes up, because a one-off precomputing job for demo purposes gets stale before too long.

alexsoble:
Next thing to do here in my mind is set up email integration so that we can get emailed error alerts when the nightly data import process fails.

alexsoble:
Opening up #1376 to unblock setting up a mailer resource for New Bedford Insights.

alexsoble:
Lesson learned: all mailer config must be set up properly in order to run the import process all the way through.

[...]

PR #1440, opened 2018-01-31
Tolerant attendance importer for New Bedford (rebased onto #1439)

alexsoble:
# Who is this PR for?

New Bedford Public Schools.

# What does this PR fix?

Blocked attendance import as described in https://github.com/studentinsights/studentinsights/issues/1371#issuecomment-360859776 and https://github.com/studentinsights/studentinsights/issues/1371#issuecomment-360864063.

# What does this PR do?

Makes the attendance importer tolerant of invalid rows. Instead of blowing up when it hits the first invalid row, the attendance importer skips those rows, and keeps a count of invalid rows plus error messages. 

This will also be beneficial to the Somerville app, because it will produce a more detailed report of how the attendance export file is getting parsed and how many invalid rows are in the file.

### DIFF
--- a/app/importers/file_importers/attendance_importer.rb
+++ b/app/importers/file_importers/attendance_importer.rb
@@ -18,14 +18,14 @@ class AttendanceImporter
 
     @data.each_with_index do |row, index|
       import_row(row) if filter.include?(row)
-      log.write(
+      @log.write(
         "\r#{@success_count} valid rows imported, #{@error_list.size} invalid rows skipped"
       )
     end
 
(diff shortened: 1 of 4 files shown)


alexsoble:
Ran a local trial import of attendance for New Bedford, looks like this branch was doing a good job of detecting and reporting invalid rows without blowing up:

new-bedford-attendnace

alexsoble:
Here's the final output from that import job:

screen shot 2018-02-01 at 11 27 01 am

alexsoble:
Closing to point this branch to master now that #1439 is merged.

PR #1441, opened 2018-01-31
Document expected fields in from X2 Assessment export

alexsoble:
# Who is this PR for?

+ External developers who write export scripts for an Insights instance and are using `x2_export/assessment_export.sql` as a model
+ Internal developers who need a reminder of what the different fields in `assessment_export.sql` mean

New Bedford's assessment export had the columns switched around, which means that I didn't properly document/explain them.

# What does this PR do?

Adds code comments to the SQL file to explain what key column names mean and what they should look like.

### DIFF
--- a/x2_export/assessment_export.sql
+++ b/x2_export/assessment_export.sql
@@ -7,9 +7,9 @@ SELECT
   'assessment_scale_score',
   'assessment_performance_level',
   'assessment_growth',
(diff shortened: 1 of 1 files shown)


alexsoble:
New Bedford updated their export files to match this documented format, so this seems like a good time to merge. ✅

PR #1446, opened 2018-02-02
Reduce school overview memory usage further

alexsoble:
# Notes

This PR builds on top of #1445 and follows a similar methodology. 

When I looked at the MemoryProfiler output for the school overview page, I found another suspicious thing: among the "Retained String Report" output were strings like "999-999-9999 C-Mom" and "student@example.com". 

These were suspicious because they are fake generated student contact info. The school overview never surfaces contact info!

Using `#select` to pull only the necessary fields for students brought the retained memory down (measuring warm path to warm path):

```
master (current app state): 
Total allocated: 3785005 bytes (35109 objects)
Total retained:  443397 bytes (3354 objects)

with student fields #selected in students_query_helper:
Total allocated: 3168397 bytes (29097 objects)
Total retained:  376418 bytes (2603 objects)
```

Combined with #1445, this change brings retained memory down from the baseline by about 15%.

## Code quality notes

I searched in vain through the Rails API for an "opposite of `#select`" method for ActiveRecord::Association that would let me specify fields to exclude. Since I couldn't find one, I ended up with a big, ugly `INCLUDE_FOR_STUDENTS` constant at the top of the class. Anyone got any good alternatives to this? 😄 

Similarly, I had hope that `#select` would work on an individual record, but it didn't, and neither did `#pluck`. That left me with:

```
student_fields = Student.where(id: student.id).select(INCLUDE_FOR_STUDENTS)
```

... which is ugly but gets the job done.

Finally, we're still querying the un-`#selected` student from the database to get discipline_incidents_count, absences_count, and tardies_count. I think that's the reason we still have some "student@example.com"s showing up in the memory profiling report (but much fewer than in the baseline report).

## Checklist

+ [x] Manually QA visiting School Overview Page
+ [x] Manually QA running rake school_overview precompute job

### DIFF
--- a/app/helpers/students_query_helper.rb
+++ b/app/helpers/students_query_helper.rb
@@ -1,6 +1,39 @@
 require 'digest'
 
 module StudentsQueryHelper
+  INCLUDE_FOR_STUDENTS = [
+    :id,
+    :grade,
+    :hispanic_latino,
+    :race,
+    :free_reduced_lunch,
+    :homeroom_id,
+    :first_name,
+    :last_name,
+    :home_language,
+    :registration_date,
+    :program_assigned,
+    :sped_placement,
+    :disability,
+    :sped_level_of_need,
+    :plan_504,
+    :limited_english_proficiency,
+    :most_recent_mcas_math_growth,
+    :most_recent_mcas_ela_growth,
+    :most_recent_mcas_math_performance,
+    :most_recent_mcas_ela_performance,
+    :most_recent_mcas_math_scaled,
+    :most_recent_mcas_ela_scaled,
+    :most_recent_star_reading_percentile,
+    :most_recent_star_math_percentile,
+    :enrollment_status,
+    :date_of_birth,
+    :risk_level,
+    :gender,
+    :house,
+    :counselor,
+  ]
+
   INCLUDE_FOR_EVENT_NOTES = [
     :id,
     :student_id,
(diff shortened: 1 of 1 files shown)


app/helpers/students_query_helper.rb
```diff
@@ -1,6 +1,39 @@
 require 'digest'
 
 module StudentsQueryHelper
+  INCLUDE_FOR_STUDENTS = [
```
kevinrobinson: Maybe `Student.column_names.map(&:to_sym) - [:whatever_you_want_to_exclude]`?  Similar to the comment in https://github.com/studentinsights/studentinsights/pull/1445#issuecomment-363300279, I think this depends on whether we want this to be "a `Student` object with one or two things missing" or we want this to be "exactly the partial view of a `Student` that we want to show in the product." 

Based on your PR description, is the idea here "everything but the address and contact info?"

alexsoble: Ooooh, I like the `Student.column_names` idea! 

And yep, "everything but the address and contact info." But mostly because those are the two fields I could find that aren't consumed by the school overview page. Giving it a second look, we might want to exclude enrollment_status also. And if you spot more fields that aren't used in the overview page let me know. 🔎 🔎

alexsoble:
QA looks good, this should be good to 🚢 after our weekly checkin meeting!

PR #1447, opened 2018-02-05
Bug Fix: double counting each student's first absence

edavidsonsawyer:
# Who is this PR for?
Dashboard Users
The first absence for each student is being counted twice. This is only a problem in the absence dashboard. Tardies shouldn't have this issue. 

# What does this PR do?
Counts each absence once!

# Checklists

## Javascript QA

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/absences_dashboard/school_absence_dashboard.jsx
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/absences_dashboard/school_absence_dashboard.jsx
@@ -50,7 +50,7 @@ export default React.createClass({
     let studentAbsenceCounts = {};
     this.state.displayDates.forEach((day) => {
       _.each(this.props.schoolAbsenceEvents[day], (absence) => {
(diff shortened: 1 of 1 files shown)


edavidsonsawyer:
@alexsoble, when you get the chance, this should fix the bug @snoopyuri saw this morning.

alexsoble:
@edavidsonsawyer Can we add/expand spec coverage for this function? Since this is really important code, let's make sure our specs cover it really thoroughly.

edavidsonsawyer:
@alexsoble we definitely do want to spec coverage for this and the other calculation functions for the dashboard. There are quite a few of them, but I had wanted to wait to confirm that there was nothing wrong with the overall approach (something like the dashboard totals not looking anything like the Aspen data).  

Later tonight I'll be able to push a more comprehensive spec for all the things the dashboard is doing here. But we should move this fix in so we don't have broken code in production, I'm so we have a chance to surface any other issues.

alexsoble:
@edavidsonsawyer Works for me, I'll merge this and make a new issue about expanding spec coverage.

alexsoble:
@edavidsonsawyer Deployed!

PR #1450, opened 2018-02-05
Document StudentSectionGradesImporter CSV headers

### LINKED ISSUE #1404: Document where student_averages_export.txt comes from

alexsoble:
# Who is this PR for?

Insights developers.

It's not immediately clear without more context why StudentSectionGradesImporter passes in headers for its CSV parsing instead of reading them from the CSV file. 

# What does this PR do?

Adds documentation based on email conversation with John Breslin from Somerville IT. 

# Issue 

+ Fix #1404.

### DIFF
--- a/app/importers/file_importers/student_section_grades_importer.rb
+++ b/app/importers/file_importers/student_section_grades_importer.rb
@@ -1,5 +1,24 @@
 class StudentSectionGradesImporter
 
+  # Most of our file importer classes match to a SQL file in the /x2_export folder.
+  # Somerville IT uses the SQL to export data  nightly, and Rails imports nightly.
+
+  # The StudentSectionGradesImporter is a bit different. It has no matching SQL export
+  # file, because its CSV gets delivered by Aspen nightly, using a custom routine
+  # they created on their back end.
+
+  # The CSV output of the custom Aspen routine doesn't include headers, so we
+  # specify them explicitly here:
+
+  CSV_HEADERS = [
+    'section_number',
+    'student_local_id',
+    'school_local_id',
+    'course_number',
+    'term_local_id',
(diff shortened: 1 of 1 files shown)


app/importers/file_importers/student_section_grades_importer.rb
```diff
@@ -1,5 +1,24 @@
...
+  # specify them explicitly here:
+
+  CSV_HEADERS = [
+    'section_number',
+    'student_local_id',
+    'school_local_id',
+    'course_number',
+    'term_local_id',
+    'grade',
+  ]
```
kevinrobinson: This is great to add!  I'm wondering if this all should be moved into `DistrictConfig`?  It seems idiosyncratic to Somerville unless this seems like it's a reasonable default to assume.

alexsoble: @kevinrobinson To be honest, I'm not sure what other districts will look like. The next data point we'll get will come when we import grades for New Bedford high schools. My instinct is to 🚢  and avoid moving anything to district config until we have a second data point.

kevinrobinson: 👍 @alexsoble yeah do what you feel is the right balance!  Especially since you're in it you'll have a better sense of that than I do.  It might be useful to collect all the idiosyncratic bits while you go, but that also just might be friction and worth putting off and doing separately after things are up and running.

alexsoble: 👍 👍

alexsoble: @kevinrobinson Thanks for the advice and perspective here!

PR #1452, opened 2018-02-05
Tolerant attendance importer [unblock New Bedford attendance import]

alexsoble:
# Who is this PR for?

New Bedford Public Schools.

# What does this PR fix?

Blocked New Bedford attendance import as described in https://github.com/studentinsights/studentinsights/issues/1371#issuecomment-360859776 and https://github.com/studentinsights/studentinsights/issues/1371#issuecomment-360864063.

# What does this PR do?

Makes the attendance importer tolerant of invalid rows. Instead of blowing up when it hits the first invalid row, the attendance importer skips those rows, and keeps a count of invalid rows plus error messages. 

This will also be beneficial to the Somerville app, because it will produce a more detailed report of how the attendance export file is getting parsed and how many invalid rows are in the file.

# Checklist 

+ [x] QA attendance import process for Somerville Public Schools

# GIF 

New attendance import command line interface: 

somerville-attendance-import

The row counts for the CsvTransformer are a bit broken because the StreamingCsvTransformer doesn't know the total number of rows until it has streamed all the rows. I think that's fine for now, shipping New Bedford is more important than making the command line interface gorgeous. 

Somerville attendance import looks ✅ , I also have GIFs of New Bedford attendance import using this branch in #1440.

### DIFF
--- a/app/importers/file_importers/attendance_importer.rb
+++ b/app/importers/file_importers/attendance_importer.rb
@@ -18,14 +18,14 @@ class AttendanceImporter
 
     @data.each_with_index do |row, index|
       import_row(row) if filter.include?(row)
-      log.write(
+      @log.write(
         "\r#{@success_count} valid rows imported, #{@error_list.size} invalid rows skipped"
       )
     end
 
     @error_summary = @error_list.inject(Hash.new(0)) { |h, e| h[e] += 1 ; h }
-    log.write("\n\nInvalid attendance rows summary: ")
-    log.write(@error_summary)
+    @log.write("\n\nInvalid attendance rows summary: ")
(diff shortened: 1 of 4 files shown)


alexsoble:
selfie 🚢

alexsoble:
Manually kicking off an attendance-only import for New Bedford.

alexsoble:
Summary from attendance import filtered to our two pilot schools in New Bedford:

```
7052 valid rows imported, 1818 invalid rows skipped

Invalid attendance rows summary: {
{:student=>["can't be blank"]}=>168,
"Neither absence nor tardy"=>1650
}
```

PR #1453, opened 2018-02-05
Allow scheduled import task to take arguments 

alexsoble:
# Who is this PR for?

Developers who want to configure data import job.

Currently the only way to schedule an import job in the background is to set:

```
rake chores:import_data_in_background
```

The problem with this is that it's not configured to accept any arguments, so you can't pass in flags like `--only-recent-attendance` to the scheduled job.

# What does this PR do?

This PR removes the `chores:import_data_in_background` rake tasks, which don't accept command line arguments very elegantly, and uses thor instead as the top level API for starting background jobs, because thor has all great command-line interface tooling. 

The thor CLI accepts a `--background` flag that can be used to run the process on a worker (i.e. in production), and can also be run `--no-background` (i.e. in local development).

This PR also removes the `--test-mode` flag, which is better implemented by calling `Rails.env.test?` directly instead of asking the developer who calls thor to supply an argument. 

# QA Checklist 

+ [x] Test the `--background` import path
+ [x] Test the `--no-background` import path

# Post-Deploy Checklist

+ [ ] Update Somerville task to `thor import:start --background --only-recent-attendance`
+ [ ] Update New Bedford task to `thor import:start --background --school '123' '115'`*

*Once we have historical attendance data in the app for every New Bedford school we can add the `--only-recent-attendance` flag

### DIFF
--- a/app/importers/import_task.rb
+++ b/app/importers/import_task.rb
@@ -7,9 +7,6 @@ class ImportTask
     # options["source"] describes which external data sources to import from
     @source = options.fetch("source", ["x2", "star"])
 
-    # options["test_mode"] is for the test suite and supresses log output
-    @test_mode = options.fetch("test_mode", false)
-
     # options["only_recent_attendance"]
     @only_recent_attendance = options.fetch("only_recent_attendance", false)
   end
@@ -58,7 +55,7 @@ class ImportTask
   ## SET UP COMMAND LINE REPORT AND DATABASE RECORD ##
 
   def log
-    @test_mode ? LogHelper::Redirect.instance.file : STDOUT
+    Rails.env.test? ? LogHelper::Redirect.instance.file : STDOUT
   end
 
   def set_up_record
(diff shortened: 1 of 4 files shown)


alexsoble:
This morning I checked some metrics and realized that this issue is higher-priority than it seems. Since I deployed #1439, I haven't been able to pass the  `--only-recent-attendance` flag to the Somerville nightly import job because of this issue with how the rake task and thor task are structured. 

I checked the metrics for the Somerville worker dyno on the Heroku UI, and saw that the dyno is running way past its memory limits every night: 

screen shot 2018-02-12 at 11 27 47 am

Setting the thor task to default to `--only-recent-attendance=true` would get this fixed quicker, but this PR is a more flexible and future way to fix it and will allow us to pass custom arguments to the import job of each Insights instance going forward.

alexsoble:
Ran the QA manually, looks ✅ so going to 🚢 this as a selfie and change the nightly import jobs on our Insights Heroku apps.

Next week I'll look back at the metrics to see if the worker dyno for Somerville is less exhausted/overloaded because it only has to import recent attendance events!

alexsoble:
Updated the scheduler jobs on both instances to use the new thor API with arguments + backgroundizing.

PR #1454, opened 2018-02-05
Tolerant behavior importer [unblock behavior import for New Bedford]

alexsoble:
# Who is this PR for?

New Bedford Public Schools.

# What does this PR fix?

Blocked New Bedford behavior import as described in https://github.com/studentinsights/studentinsights/issues/1371#issuecomment-360859776 and https://github.com/studentinsights/studentinsights/issues/1371#issuecomment-360864063.

# What does this PR do?

Makes the behavior importer tolerant of invalid rows. Instead of blowing up when it hits the first invalid row, the behavior importer skips those rows, and keeps a count of invalid rows plus error messages. 

This will also be beneficial to the Somerville app, because it will produce a more detailed report of how the behavior export file is getting parsed and how many invalid rows are in the file.

# Checklist 

+ [x] QA attendance import process for Somerville Public Schools
+ [x] QA attendance import process for New Bedford Public Schools

### DIFF
--- a/app/importers/file_importers/attendance_importer.rb
+++ b/app/importers/file_importers/attendance_importer.rb
@@ -23,8 +23,10 @@ class AttendanceImporter
       )
     end
 
-    @error_summary = @error_list.inject(Hash.new(0)) { |h, e| h[e] += 1 ; h }
-    @log.write("\n\nInvalid attendance rows summary: ")
+    @error_summary = @error_list.each_with_object(Hash.new(0)) do |error, memo|
+      memo[error] += 1
+    end
+    @log.write("\n\nInvalid behavior rows summary: ")
     @log.write(@error_summary)
   end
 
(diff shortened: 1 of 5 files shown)


alexsoble:
New Bedford behavior import for our pilot schools: 

screen shot 2018-02-05 at 4 18 29 pm

alexsoble:
Somerville behavior import: 

screen shot 2018-02-05 at 4 34 43 pm

alexsoble:
I was curious about the difference between "rows of data" and "valid rows imported" for Somerville. Since I didn't scope Somerville's import by school, I assumed all rows would be imported. I printed out some data about the filtering mechanism:

```
if filter.include?(row)
  import_row(row)
else
  puts "not in filter! row[:school_local_id] = #{row[:school_local_id]}"
end
```

I got back a whole lot of rows that looked like this:

```
not in filter! row[:school_local_id] = SCALE
```

SCALE is Somerville's adult learning program (http://www.somerville.k12.ma.us/adult-learning/scale). 

We aren't importing or surfacing data from SCALE, so we definitely don't want to import these rows. That explains why the output in the screenshot looks the way it does, and confirms the import process is working as it should. 😎 🆒

PR #1456, opened 2018-02-09
Specs for Dashboard

edavidsonsawyer:
# Who is this PR for?
Developers and maintainers 

Dashboard specs are currently just shallow smoke tests

# What does this PR do?
Tests entire component contract for each dashboard component
## Automated Testing

*Does the PR touch an area of the codebase that needs more specs? More examples within the specs? More scenarios covered by specs?*


edavidsonsawyer:
@alexsoble , this should be all set as well. I'm using enzyme to test the render output and functions of each of the dashboard components.

spec/javascripts/school_administrator_dashboard/students_table_spec.jsx
```diff
@@ -0,0 +1,31 @@
...
+    expect(table.instance().totalEvents()).toEqual(12);
+  });
+
+  it('orders the students by total events by default', () => {
+    expect(table.instance().orderedRows()[5].first_name).toEqual("Arlecchino");
+  });
+
+  it('orders students by name when Name is clicked', () => {
+    table.find('.sort-header').first().simulate('click');
+    expect(table.state().sortBy).toEqual("last_name");
```
alexsoble: Would it make sense to have an expectation here about the first value in the table changing?

spec/javascripts/school_administrator_dashboard/dashboard_bar_chart_spec.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { shallow } from 'enzyme';import DashboardBarChart from '../../../app/assets/javascripts/school_administrator_dashboard/dashboard_components/dashboard_bar_chart.jsx';
```
alexsoble: Should these be on separate lines?

spec/javascripts/school_administrator_dashboard/students_table_spec.jsx
edavidsonsawyer: It would, @alexsoble. And it's a good test b/c it may have uncovered a bug in the sorthelpers. I'll check on what's going on there and update this then.

edavidsonsawyer: @alexsoble So looking into this a bit, the function I'm using to sort here is the same used by the school overview. It sorts by last names, but doesn't do a secondary sort on first names. That might be a change worth making at some point, but since I don't know why it was implemented the way it was, I'm inclined to leave the behavior as is and alter the test data so secondary sorts aren't a problem.

edavidsonsawyer:
@alexsoble Thanks for the check. I incorporated your suggestions so this should be ready for another look when you have the time.

spec/javascripts/school_administrator_dashboard/students_table_spec.jsx
alexsoble: @edavidsonsawyer Agreed!

alexsoble:
@edavidsonsawyer 👍 👍 Thanks for adding this test suite, shipping now! 🚢 🚢 

One other code improvement I think we should make before we add new features is bringing the JS module style in line with the direction we're using for our other classes.

The tardies dashboard uses `React.createClass`, but the rest of our JS is moving toward `class ClassName extends React.Component`. We're also changing the format of our filenames to match create-react-app styles: FileName.js and FileName.test.js instead of file_name.jsx and file_name_spec.jsx. Check out #1467 for an example. 

I think you're really well positioned to do this refactoring since you know these classes intimately and have been in the weeds of writing them over the last couple of weeks. Let me know if that sounds like a good plan to you / what questions you have!

PR #1458, opened 2018-02-12
Remove rails_12factor gem

alexsoble:

Extra gem in Gemfile that can be removed.

The `rails_12factor` gem was required to launch our app on Heroku when we were at Rails 4.x. Now that we're at Rails 5.1.1, we don't need the gem anymore.

Our app already has the required `production.rb` setup described in the "Migrating to Rails 5" section of the 12_factor docs: https://github.com/heroku/rails_12factor#rails-5

+ [x] `config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?`
+ [x] `if ENV["RAILS_LOG_TO_STDOUT"].present? . . .`
 
# What does this PR do?

Removes the `rails_12factor` gem from the production group.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,10 +35,6 @@ gem 'uglifier', '>= 1.3.0'
 gem 'wicked_pdf'
 gem 'wkhtmltopdf-binary'
 
-group :production do
-  gem 'rails_12factor'
-end
-
 group :development, :test do
   gem 'bourbon', '~> 4.3.2'
   gem 'capybara'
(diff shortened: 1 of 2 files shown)


alexsoble:
I QA'ed this by force-pushing to the somerville-teacher-tool-demo Heroku app. The deploy succeeded. Then I ran `heroku logs -t -a somerville-teacher-tool-demo` to tail the logs, and clicked around the demo site at the same time to double check that the logging was working right.

alexsoble:
Weekend 🚢

alexsoble:
Clicked around prod, looks ✅, tailed logs, they look ✅ too.

PR #1459, opened 2018-02-12
Delete WeeklyUpdateMailer code

alexsoble:

Insights used to send out a weekly email, subject line "Student Insights Weekly update for educators." The email showed lists of students in SST and MTSS who hadn't had a note created for them in a long time, plus students with the most absences/tardies. 

The weekly update hasn't been sent since May, because of a bug. But no one noticed, which seems like a signal that this weekly email had quickly lost urgency and become spam. We could build these features in the web app instead of an email format, which might be a better form of delivering the information. In fact, the absences/tardies dashboards that @edavidsonsawyer is building include features that replicate the Weekly Update mailer.

In the meantime, this generates a Rollbar error every week, creating white noise. Removing the mailer will cull unused code paths, remove error reporting white noise, and remove an unused feature.

# What does this PR do?

Removes the `WeeklyUpdateMailer` class and the rake task that calls it.


alexsoble:
🚢

PR #1461, opened 2018-02-12
Move service types to district config

alexsoble:
# Who is this PR for?

New Bedford insights.

Service types are currently stored as constants and are all hard-coded to Somerville service types. 

# What does this PR do?

Pulls service types to district config YAML.

TK 

# Checklists

## Javascript QA

+ [ ] Author checked latest in IE - Student Profile

## Automated Testing

### DIFF
--- a/config/district_new_bedford.yml
+++ b/config/district_new_bedford.yml
@@ -82,3 +82,32 @@ schools:
   -
     name: Whaling City Jr./Sr. High
     local_id: '515'
+service_types:
+  -
+    name: Behavior Plan
+    id: 1
+  -
+    name: Safety Plan
+    id: 2
+  -
+    name: Counseling – Individual
+    id: 3
+  -
+    name: Group Counseling
+    id: 4
+  -
+    name: Social Skills Group
+    id: 5
+  -
+    name: Afterschool Teacher Help
+    id: 6
+  -
+    name: Reading Interventionist
+    id: 7
+  -
+    name: Lexia Core 5
+    id: 8
+  -
+    name: MySidewalks Intervention
+    id: 9
+
--- a/app/config_objects/load_district_config.rb
+++ b/app/config_objects/load_district_config.rb
@@ -15,6 +15,10 @@ class LoadDistrictConfig
     load_yml.fetch("schools")
   end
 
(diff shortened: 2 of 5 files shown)


alexsoble:
Exploring how this looks locally with the New Bedford service types configured! 

The School Overview page works great and shows the new types, so it must not have their IDs hardcoded: 

screen shot 2018-02-12 at 1 53 27 pm

The Student Profile page is broken, it looks alright before you try to click around. But clicking the "Record Service" button does nothing for the user and throws a JS error:

```
Uncaught TypeError: Cannot read property 'name' of undefined
    at Object.renderServiceButton (webpack-internal:///389:239)
    at Object.renderWhichService (webpack-internal:///389:218)
    at Object.render (webpack-internal:///389:176)
```

As I recall this area of the JS has service type IDs hardcoded, which will have to change.

config/district_new_bedford.yml
```diff
@@ -82,3 +82,32 @@ schools:
   -
     name: Whaling City Jr./Sr. High
     local_id: '515'
+service_types:
```
kevinrobinson: @alexsoble This is super interesting!  I'm excited to learn more about the adoption path for writing and reading this data in New Bedford.  Without knowing more about that, it's hard to comment more, but one thing that jumped out to me was that we're forking the way that we talk about these services and so they will stay pretty specific about what's happening.

Evaluating this more really depends on expectations and assumptions about the adoption path in New Bedford.  It might be good to capture more about that here if it's come up (or if not we'll maybe learn more shortly!)

app/assets/javascripts/student_profile/record_service.jsx
```diff
@@ -157,33 +158,47 @@ import serviceColor from '../student_profile/service_color.js';
     },
 
     renderWhichService: function() {
+      const {serviceTypesIndex} = this.props;
+      const serviceTypeIds = _.map(_.values(serviceTypesIndex), 'id');
+      const chunks = _.values(_.groupBy(serviceTypeIds, (id) => {
+        return serviceColor(id);
+      }));
```
kevinrobinson: Would this change the button ordering?

PR #1462, opened 2018-02-12
Use latest devise_ldap_authenticatable

alexsoble:
# Who is this PR for?

Updates `devise_ldap_authenticatable` to the most recent version available on GitHub, which includes a patch to `net-ldap`.

Resolves an issue with an outdated version of `net-ldap`.

# ❓Will this deploy successfully❓ 

@kevinrobinson worked on deploying another branch to get us the patched `net-ldap`, #1394. His branch didn't ship because, as he reported, "This doesn't work and breaks the login." 

That's excellent background to have and because of it, I know not to ship this PR during the school day / school week! This is a ship-at-night-or-on-the-weekend-and-test PR.

The open question here is, does the patched `devise_ldap_authenticatable` gem also suffer from whatever issue blocked #1394 from shipping? If so, this branch isn't much farther ahead of #1394, it just points to the gem upstream instead of a fork.


kevinrobinson:
See also https://github.com/studentinsights/studentinsights/pull/1545

alexsoble:
Closing in favor of #1545

PR #1468, opened 2018-02-15
Convert ServiceUploadsPage.js to use ES6 modules; add Api class

alexsoble:
# Who is this PR for?

Developers who want to contribute to the codebase by adding or improving client-side features. 

Not all our JS files use ES6 modules, which can be confusing for new JS devs coming on to the project.

# What does this PR do?

+ Converts `ServiceUploadsPage.js ` to an ES6 module. 
+ Pulls two API calls made by the page into an `Api.js` helper class.

# Checklist 

+ [x] Service Upload — tested latest commit in Internet Explorer

# GIF of everything but the file upload — Internet Explorer

ie-upload

# GIF including file upload — Chrome

upload-file

# GIF of LASID validation — Chrome

lasid-fail


alexsoble:
This PR touches sensitive logic around which students get assigned when a bulk service is uploaded. So I want to use this as an opportunity to add more spec coverage for that functionality. 👍 😎

PR #1469, opened 2018-02-16
Investigate New Bedford MCAS dates

alexsoble:

+ Only one MCAS event per student is coming through in the New Bedford MCAS instance

# What does this PR do?

+ Analyzing the raw iassessment.csv export file shows that there are assessment records from multiple test dates in the raw data, which means that it's an issue with the import process:

```

["2010-05-01",
"2011-05-01",
"2012-05-01",
"2012-11-01",
"2013-02-01",
"2013-03-01",
"2013-05-01",
"2013-11-01",
"2014-02-01",
"2014-03-01",
"2014-05-01",
"2014-11-01",
"2015-02-01",
"2015-03-01",
"2015-05-01",
"2015-11-01",
"2016-02-01",
"2016-03-01",
"2016-05-01",
"2017-02-01",
"2017-03-01",
"2017-05-01",
"2017-11-01"]
```

### DIFF
--- a/lib/analysis/analyze_assessments_table.rb
+++ b/lib/analysis/analyze_assessments_table.rb
@@ -61,4 +61,12 @@ class AnalyzeAssessmentsTable < Struct.new(:path)
         .sort
   end
 
+  def new_bedford_pilot_school_mcas_dates
+    data.select { |row| row[:assessment_test] == 'MCAS' }
+        .select { |row| row[:school_local_id] == '123' || row[:school_local_id] == '115' }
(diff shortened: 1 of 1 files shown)


alexsoble:
Adding a method to look specifically at the pilot schools showed why we're only seeing one MCAS result in our New Bedford Insights instance: 

```
2.3.0 :004 > a.new_bedford_pilot_school_mcas_dates
 => ["2017-05-01"] 
```

Takeaway: there are multiple years of MCAS data in the export file, but only one year's worth for students in our pilot schools.

PR #1470, opened 2018-02-16
Don't show years enrolled on School Overview page unless we have data

alexsoble:
# Who is this PR for?

New Bedford Public Schools users of the School Overview page.

Right now they see all students as having "Years Enrolled" equal 48, because we don't have any registration data for New Bedford in our Insights instance yet. 

# What does this PR do?

Removes the "Years Enrolled" slice panel if we don't have any registration data. There are many slice panels, and showing one more or one fewer doesn't have a huge effect on the functionality of the page, especially "Years Enrolled" which shows a similar kind of data to "Grade." My theory here is that it's better to show a user one less data category than show an obvious data issue.

### DIFF
--- a/spec/javascripts/components/SlicePanels.test.js
+++ b/spec/javascripts/components/SlicePanels.test.js
@@ -68,8 +68,8 @@ SpecSugar.withTestEl('high-level integration tests', (container) => {
        student registration dates`, () => {
     const el = container.testEl;
     const studentsWithRegistration = FixtureStudents.map((student) => {
-      return _.merge(student, {registration_date: '2018-02-13T22:17:30.338Z'})
-    })
+      return _.merge(student, {registration_date: '2018-02-13T22:17:30.338Z'});
+    });
 
     helpers.renderInto(el, {
       students: studentsWithRegistration,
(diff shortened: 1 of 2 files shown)


spec/javascripts/components/SlicePanels.test.js
```diff
@@ -47,10 +47,35 @@ const helpers = {
...
+      [ 'Disability', 'Low Income', 'LEP', 'Race', 'Hispanic/Latino', 'Gender' ],
+      [ 'Grade', 'Risk level' ],
+      [ 'STAR Reading', 'MCAS ELA Score', 'MCAS ELA SGP' ],
+      [ 'STAR Math', 'MCAS Math Score', 'MCAS Math SGP' ],
+      [ 'Discipline incidents', 'Absences', 'Tardies' ],
+      [ 'Services', 'Summer', 'Notes', 'Program', 'Homeroom' ]
+    ]);
+  });
+
+  it(`renders everything on the happy path for elementary school with
```
kevinrobinson: If this is about differences in data systems across schools, let's add that to the comment.

kevinrobinson:
@alexsoble 👍 🚢

PR #1471, opened 2018-02-16
Convert SchoolAdministratorDashboards.jsx to es6 stateless function

edavidsonsawyer:
# Who is this PR for?
Developers who want to contribute to the codebase by adding or improving client-side features.

Not all our JS files use ES6 modules, which can be confusing for new JS devs coming on to the project.

# What does this PR do?
Converts SchoolAdministratorDashboards.jsx to ES6 module

# Checklists

## Javascript QA

+ [x] Author checked latest in IE - School Administrator Dashboards Links
+ [x] Reviewer checked latest in IE - School Administrator Dashboards Links


edavidsonsawyer:
@alexsoble here's the first of the es6 conversions. Since there are no handler methods or state here it's just the function, which seems to be the [typical standard](https://github.com/airbnb/javascript/tree/master/react#class-vs-reactcreateclass-vs-stateless). If we want to just make everything a class instead, let me know.

alexsoble:
@edavidsonsawyer Woah, this is neat! I learned a new thing from you, thank you for sharing that Airbnb React style guide documentation! Let's change the file name too to match the CamelCasing standard we're moving towards.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/school_administrator_dashboards.jsx
```diff
@@ -1,23 +1,21 @@
 import React from 'react';
+import PropTypes from 'prop-types';
```
alexsoble: This looks unused.

edavidsonsawyer: It's used down in the propTypes variable to check the serializedData req. I could just use React.propTypes, but that's being depricated

edavidsonsawyer:
@alexsoble sure thing, all set

edavidsonsawyer:
@alexsoble well _now_ it's all set!

app/assets/javascripts/school_administrator_dashboard/dashboard_components/school_administrator_dashboards.jsx
alexsoble: Is `SchoolAdministratorDashboards` still a React component? It doesn't inherit from React.Component or React.createClass anymore, so it looks like just a plain function to me. That makes me wonder if `serializedData` is still props, or just a plain argument to a function now?

edavidsonsawyer: It is, actually! Anything that takes props as an argument and returns a React component is a valid component as far as React is concerned. https://reactjs.org/docs/components-and-props.html

Though you're right. It's also just a plain function with an argument. If we treat it like a react component we get type checking for the props and probably other stuff I'm just not thinking of atm.

alexsoble: @edavidsonsawyer Ah I am learning so much here! This is cool!

PR #1473, opened 2018-02-18
Live dashboard link for principals

edavidsonsawyer:
# Who is this PR for?
Principals and Assistant Principals

If a principal wants to see the dashboard for their school, they have to manually type the URL

# What does this PR do?
Adds a link to their School Overview which takes them to their dashboard

screen shot 2018-02-28 at 8 39 37 am

# Checklists

## Javascript QA

+ [x] Author checked latest in IE - Dashboard link with principal (school admin)
+ [x] Author checked latest in IE - Dashboard link with non-admin with schoolwide access
+ [ ] Reviewer checked latest in IE - Dashboard link with principal (school admin)
+ [ ] Reviewer checked latest in IE - Dashboard link with non-admin with schoolwide access

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -40,7 +40,7 @@
               &nbsp; &nbsp; &nbsp;
             <% end %>
             <% if current_educator.admin? %>
-              <%= link_to 'School Administrator Dashboard', school_administrator_dashboard_school_path(current_educator.school) %>
+              <%= link_to 'Administrator Dashboard', school_administrator_dashboard_school_path(current_educator.school) %>
               &nbsp; &nbsp; &nbsp;
             <% end %>
             <p class="search-label">Search for student:</p>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -39,6 +39,10 @@
               <%= link_to 'School Overview', school_path(current_educator.school) %>
               &nbsp; &nbsp; &nbsp;
             <% end %>
+            <% if current_educator.admin? %>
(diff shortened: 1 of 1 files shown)


alexsoble:
Hey @edavidsonsawyer! Uri is out of town on vacation this week, so why don't we ship this live link next week when Uri's around to intro the new feature to principals, handle questions or run trainings as needed?

edavidsonsawyer:
Works for me! I'm out of town until next Tuesday myself so I just wanted to be sure this was ready in case anyone needed it in the next week

edavidsonsawyer:
@alexsoble @snoopyuri  now that I'm back I wanted to check in on this. Uri, I remember you saying you wanted these links live either this week or next but couldn't remember if there was a particular date you were shooting for. In any case, this is ready when you guys are and I'm back to field any issues that pop up.

kevinrobinson:
@edavidsonsawyer @alexsoble This is awesome!  Does it seem reasonable to change this to two links: Absences and Tardies so folks can navigate directly to what they most care about?

I think it might also be great to add a "prototype" or "experimental" banner to the pages, which could also include an invitation to share feedback.  Here's one example of what that might look like from an older prototype:

screen shot 2018-03-02 at 12 39 31 pm

That can help both to set expectations ("ah they're sharing something that's still in development, I'm an early collaborator or partner in developing this, it would make sense if things weren't super beautiful or fast and that's cool") as compared to other parts of the product where people are using them every week, and they'll expect things to be used in meetings regularly, have no bugs and be speedy, etc.

kevinrobinson:
oops, I missed that this was here and shipped https://github.com/studentinsights/studentinsights/pull/1523, sorry about that!

PR #1474, opened 2018-02-20
WIP – Make overlapping services invalid

alexsoble:
# Who is this PR for?

Educators who record services in Insights.

Right now it's possible for two educators to create two services of the same type for the same student at the same time. This situation can lead to duplicate, confusing Student Profile data. I've verified there at 10 duplicates in the Somerville Insights database already. Goal of shipping this is to make sure that number shrinks over time and doesn't grow!

# What does this PR do?

+ Adds Rails validation logic so that adding an overlapping active service is invalid.
+ Refactor the service upload controller to surface Rails error messages for the end user.
  + This makes error message display more useful (show more specific data about the error) and more flexible (handle a wider range of errors).

# GIF (service upload: overlapping service)

TK

# GIF (service upload: happy path)

TK

# GIF (service upload: invalid LASIDs)

TK

# GIF (service upload: educator tries recording overlapping service in Student Profile)

TK

### DIFF
--- a/app/assets/javascripts/service_uploads/NewServiceUpload.js
+++ b/app/assets/javascripts/service_uploads/NewServiceUpload.js
@@ -261,13 +261,9 @@ class NewServiceUpload extends React.Component {
           <br />
           <br />
           <ul>
-            {this.props.incorrectLasids.map(function(lasid) {
-              return (
-                <li>
-                  {lasid}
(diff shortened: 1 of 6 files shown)


kevinrobinson:
@alexsoble Should we close this?  I have it tracked on the backlog list.

PR #1326, opened 2017-12-18
Finish and merge in @Jngai Notes + Help Bubble ES6 PR

alexsoble:
# Who is this PR for?

Developers who want to contribute to the codebase by adding or improving client-side features. 

# What problem does this PR fix?

Not all our JS files use ES6 modules, which can be confusing for new JS devs coming on to the project.

# What does this PR do?

Merges in #1311 and adds some changes on top of it. 

Updates the following files to use ES6 modules and follow create-react-app naming conventions:

+ RestrictedNotesPageContainer.js
+ NotesDetails.js
+ Api.js
+ HelpBubble.js

Also adds initial test coverage for:

+ Api.js
+ HelpBubble.js
# Merge Checklist

## Javascript QA

+ [x] Checked latest in IE - Create/Edit Notes
+ [x] Checked latest in IE - Create/Edit Restricted Notes
+ [x] Checked latest in IE - Help Bubble on Academic Detail

## Specs 

+ [x] Included specs for new code.

PR #1443, opened 2018-02-02
NotesFeed Page V1 (shipping dark)

gabivoicu:
# Who is this PR for?
Educators
**Feature will not be link-live yet, navigate to `/educators/notes_feed`**

# What problem does this PR fix?
Educators wanting to see a feed of all of their notes.

# What does this PR do?
Adds a new page, `Notes Feed`, where an educator can see all of the notes they have created, in descending chronological order (most recent ones at the top).

# Screenshot (if adding a client-side feature)
screen shot 2018-02-15 at 9 31 26 pm
screen shot 2018-02-15 at 9 31 31 pm
screen shot 2018-02-23 at 8 48 09 am

# Checklists

## Javascript QA

+ [x] Author checked latest in IE - Notes Feed
+ [x] Author checked latest in IE - Student Dashboard - Interventions - Notes List: display & functionality
+ [x] Reviewer checked latest in IE - Notes Feed
+ [x] Reviewer checked latest in IE - Student Dashboard - Interventions - Notes List: display & functionality

## Automated Testing

*Does the PR introduce a new feature or new section of code?*

+ [x] Author included specs for new code (JS)
+ [x] Author included specs for new code (Ruby)

PR #1475, opened 2018-02-21
Remove duplicate notes list file

alexsoble:

There are two files that export the `NotesList` class: one with `.js` and one with `.jsx`. That's confusing and could potentially introduce bugs down the line: we need only one file per class. This problem was introduced in #1326.

# What does this PR do?

Removes the `.js` file, which isn't being called or used anywhere. 

# Screenshot (demo student)

screen shot 2018-02-21 at 12 03 47 pm


alexsoble:
Taking a look at the two files, I'm seeing that `notes_list.js` is an ES6 module transformation in progress. It doesn't modify `window.shared`, and it ends with `export default NotesList`. The other file exports using `window.shared.NotesList =`. That's good because it means these are on totally separate paths. Searching for "`import NotesList`" turned up nothing, which means that the code on the `notes_list.js` path is totally uncalled and unused.

alexsoble:
There are two ways to tackle this. One would be to finish up the ES6 module conversion, swap that in for the `window.shared.NotesList` API that's currently being used, QA and ship. The other is to just delete the ES6 file and convert it later. For me the deciding factor is that #1443 is currently using the `window.shared` API, and I want to create as little disruption for that PR as possible. So this PR will delete the ES6 file and we can convert `NotesList` later.

alexsoble:
🚢

PR #1476, opened 2018-02-21
Use more general variable names for LDAP config

alexsoble:
# Notes 

+ Right now the LDAP-related ENV config keys are hardcoded with a "SOMERVILLE_" prefix, doesn't make much sense for us to use when setting up New Bedford's LDAP integration

### DIFF
--- a/config/ldap.yml
+++ b/config/ldap.yml
@@ -22,9 +22,9 @@ authorizations: &AUTHORIZATIONS
 ## Environment
 
 production:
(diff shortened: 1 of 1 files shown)


alexsoble:
⚙️ ✅ ⚙️ ✅ . . . configured `DISTRICT_LDAP_BASE`, `DISTRICT_LDAP_HOST`, and `DISTRICT_LDAP_PORT` on the Somerville instance so that this can ship and connect correctly. 

Will remove the old env variables keys once I verify the new ones are working as intended.

PR #1479, opened 2018-02-26
Fix bug with data sent down to homeroom page

kevinrobinson:
# Who is this PR for?
classroom teachers

The homeroom roster page isn't showing students.  I couldn't track down when this was introduced.

# What does this PR do?
Removes a UI-side filter on active students.  This is already done on the server, and the server isn't sending down the field to do the filtering on, which caused the bug.

Before:
screen shot 2018-02-26 at 11 59 54 am

After:
screen shot 2018-02-26 at 11 59 43 am

# Checklists
+ [x] Author checked latest in IE - Homeroom roster page


kevinrobinson:
@alexsoble Nice, thanks!  I was thinking we could still ship this one too then (since the server is already filtering active out anyway), what do you think?

alexsoble:
@kevinrobinson Oh yeah, good call! I didn't realize the server was doing the same thing but you're totally right, I found it in `homerooms_controller.rb`:

```
def eager_students(*additional_includes)
  @homeroom.students.active.includes([
    :interventions,
    :student_risk_level,
    :homeroom,
  ] + additional_includes)
end
```

Let's 🚢!

PR #1483, opened 2018-02-27
Remove fromPair and migrate FeedHelpers to imported ES6 module

kevinrobinson:
# Who is this PR for?
just developers, part of https://github.com/studentinsights/studentinsights/issues/1180

# What does this PR do?
`fromPair` won't be needed with the new build process, but I didn't want to change the complicated `PageContainer` code yet, so I inlined it there and removed the other uses.  `FeedHelpers` is just renamed and removed from the legacy import file.

# Checklists


kevinrobinson:
selfie

PR #1486, opened 2018-02-28
My notes page 1405

gabivoicu:
# Who is this PR for?
Educators

# What problem does this PR fix?
Improve Notes Feed feature based on feedback

# What does this PR do?
- Change notes feed page from loading notes by days to loading notes by count
- Only display load more button if there are more notes to show

# Screenshot (if adding a client-side feature)

# Checklists

## Javascript QA

+ [x] Author checked latest in IE - Notes Feed
+ [x] Author checked latest in IE - Student Dashboard - Interventions - Notes List: display & functionality
+ [x] Reviewer checked latest in IE - Notes Feed
+ [x] Reviewer checked latest in IE - Student Dashboard - Interventions - Notes List: display & functionality

PR #1492, opened 2018-03-01
Factor out IncidentCard from AttendanceDetails as a component

kevinrobinson:
# Who is this PR for?
developers and designers

This is moving towards a home page with a feed of what's happening for students, building on https://github.com/studentinsights/studentinsights/pull/1486.  One piece of information that isn't included now is discipline incidents, which are some of the highest impact events for students.  We have a feed-style UI for this now buried at the bottom of the attendance and behavior tab in the student profile, so this factors it out so we can elevate this kind of "card" as a first-class component.  I'm assuming we'll soon change the layout and visual presentation, but that this element will be useful to have as a design element.

# What does this PR do?
Straight refactor.

screen shot 2018-03-01 at 4 32 04 pm

# Checklists

### DIFF
--- a/app/assets/javascripts/components/IncidentCard.js
+++ b/app/assets/javascripts/components/IncidentCard.js
@@ -50,6 +50,14 @@ class IncidentCard extends React.Component {
     );
   }
 }
+IncidentCard.propTypes = {
+  incident: React.PropTypes.shape({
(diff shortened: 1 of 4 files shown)


kevinrobinson:
selfie

PR #1493, opened 2018-03-02
 Add last SST to absences dashboard

kevinrobinson:
# Who is this PR for?
SST facilitators and school leaders

The prototype absences dashboard shows kids with absences, but folks have to "join" that manually in their head with who isn't already in SST in order to take action.  This does that work for them.

# What does this PR do?
Adds a "Last SST" column to the table in the prototype absences dashboard.  It also simplifies some styling on the table.

screen shot 2018-03-02 at 7 56 02 am

# Checklists
+ [x] Author checked latest in IE - Absences Dashboard

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/StudentsTable.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/StudentsTable.js
@@ -40,6 +40,8 @@ class StudentsTable extends React.Component {
       return rows.sort((a, b) => SortHelpers.sortByString(a, b, sortBy));
     case 'number':
       return rows.sort((a, b) => SortHelpers.sortByNumber(a, b, sortBy));
+    case 'date':
+      return rows.sort((a, b) => SortHelpers.sortByDate(a, b, sortBy));
     default:
       return rows;
     }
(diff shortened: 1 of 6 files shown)


kevinrobinson:
@edavidsonsawyer Check out the styling changes here.  I found that table layout didn't work like I expected when I went to add this new column, and it looked like it was because the cells had `float: left` and the table elements all had `display: block` and were sized explicitly.  I removed those so that the table would layout like a typical table, but maybe there's other stuff I missed there in the process.

Also, it looked to me like this might have been related to getting only the `tbody` to scroll, so I removed that bit for now and just render them all out.  We could render only the top n students here, or use `react-virtualized` to render the whole list (and it does the nice scrolling and optimizes the rendering so you can throw huge lists at it).

Does this seem a worthwhile enough improvement to ship as-is?  There's other feedback from the MTSS meeting yesterday I can share later today too, just trying to chip in :)

edavidsonsawyer:
@kevinrobinson this looks like a nice feature! The CSS scrolling was probably due for a better solution anyway, so I don't see any need to keep it. 

One thing I'm wondering about here is whether we'll want "Last SST" for every dashboard. Right now the column shows for both absences and tardies, but we're only passing the information to populate it for the absence dashboard. I think we should either show the SST date in every case, or only show the column with the absence dashboard. I'd imagine we'll still want that column in the discipline list in the future so it might make sense to just have that information for all three. What do you think?

kevinrobinson:
@edavidsonsawyer re: CSS, cool that sounds good!

re: putting this everywhere, sure that sounds great, but I don't think that should stop us from shipping value incrementally.  Absences are a particular focus of SST meetings in the way that tardies are not, and this need around absences came up in conversation yesterday with folks at two schools.  So I think it's great if we can show them incremental progress and a short cycle time from discovering a need to shipping something that helps with the things educators care about.  These needs are often more complex and nuanced that "consistency everywhere" 

All that said, I agree that it'd be great to add this to the tardies dashboard too!

kevinrobinson:
@edavidsonsawyer So unless you have strong objections, let's ship 🚢 this now and then potentially add to other places as well?  There's a bunch of other early feedback on this that I can share later today too and we can talk about what we should tackle when.

edavidsonsawyer:
Sure, I don't think there's anything wrong with shipping this right now as is. I can create a PR myself to add it to the tardies dashboard since I think it's just two lines.

PR #1494, opened 2018-03-02
Guard against missing educator name in Research Matters exporter

alexsoble:
# Who is this PR for?

Research Matter.

Test test accounts don't have full names associated with them and are assigned to a school, make the script guard against missing educator name when outputting `teacher_file`.


alexsoble:
🚢

PR #1496, opened 2018-03-02
Add App.js and top-level react-router

kevinrobinson:
# Who is this PR for?
developers

Answers "How do I add a new page?" and creates a migration path for removing code in `legacyRouteHandlers`.  It came out of building towards https://github.com/studentinsights/studentinsights/issues/1497.

# What does this PR do?
Changes the top-level `index.js` file so that it checks for a `main` element, runs through `legacyRouteHandler`, and then if not it mounts the new `App.js` on `main`.  This also adds a placeholder route for `/home`, which is for the new home page prototype.

On the server side, also adds a controller action for the server to respond with on client-side routes, `ui#ui`, which sends back just the current educator and then leaves any other data fetching up to the client.  So for example, on a request to `/home`, Rails sends down the same application layout with the nav bar and embeds the `current_educator` data in `serialized-data`.  The JS code then runs, tries out the legacy routes (none match), then mounts `App.js` which reads the minimal generic `serializedData` and then renders `HomePage`.  It's up to `HomePage` to request what other data it needs.

(note that this is placeholder; this PR is about the routing mechanism)
screen shot 2018-03-02 at 12 12 18 pm

# Checklists
## Javascript QA
+ [x] Author checked latest in IE - Import records
+ [x] Author checked latest in IE - Bulk upload
+ [x] Author checked latest in IE - Section page
+ [x] Author checked latest in IE - Homeroom page

## Automated Testing

### DIFF
--- a/ui/index.js
+++ b/ui/index.js
@@ -1,7 +1,11 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
 import datepickerConfig from '../app/assets/javascripts/datepicker_config.js';
 import sessionTimeoutWarning from '../app/assets/javascripts/session_timeout_warning.js';
 import studentSearchbar from '../app/assets/javascripts/student_searchbar.js';
-import route from './route';
+import legacyRouteHandler from './legacyRouteHandler';
+import App from './App';
+
 
 // First, run side effects to inject code into window.shared
 import './legacy.js';
@@ -25,4 +29,12 @@ if ($('.student-searchbar').length > 0) {
 }
 
 // Routing
-route();
+// Some pages are server-rendered and have a different structure
+// other than #main so we ignore those.  Other pages add in class names
+// to the body tag that `legacyRouteHandler` works with.  Newer pages
+// should handle routing with react-router inside the `App` component.
+const mainEl = document.getElementById('main');
+if (mainEl) {
+  const didRoute = legacyRouteHandler(mainEl);
+  if (!didRoute) ReactDOM.render(<App />, mainEl);
+}
--- a/app/assets/javascripts/home/HomePage.js
+++ b/app/assets/javascripts/home/HomePage.js
@@ -1,0 +1,19 @@
+import React from 'react';
+
+
+/*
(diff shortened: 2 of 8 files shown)


kevinrobinson:
The idea here is that this could be how we add new client-side routes for new pages, so I asked everybody to review and make sure this seems reasonable, and to help surface any questions or other ideas about this too.  Thanks y'all!

ui/index.js
```diff
@@ -25,4 +29,12 @@ if ($('.student-searchbar').length > 0) {
 }
 
 // Routing
-route();
+// Some pages are server-rendered and have a different structure
+// other than #main so we ignore those.  Other pages add in class names
+// to the body tag that `legacyRouteHandler` works with.  Newer pages
+// should handle routing with react-router inside the `App` component.
+const mainEl = document.getElementById('main');
+if (mainEl) {
```
alexsoble: Is `index.js` going to run on every single page/page load in the app?

kevinrobinson: Yep, it already does except for a few admin pages that don't load any JS at all.

Ideally we'd run the routing, then run any other code based on that.  But we're still doing a lot of JS execution in the different legacy.js files as part of the migration from Sprockets.

alexsoble:
@kevinrobinson I want to make a flow chart to show how routing will work with this PR! I've read through the code a few times and the core concepts make sense, I think a visual will help clarify the flow.

kevinrobinson:
@alexsoble @edavidsonsawyer @gabivoicu I'm planning on shipping this today, so if you have other comments or feedback please feel free to share them either before or after this is merged.

Reviewing is a good place to ask questions, learn about stuff, and clarify what we're all building together over time, and I'm a big fan of after-the-fact reviews too especially when people are getting other stuff shipped too :)

alexsoble:
@kevinrobinson and I took some time to walk through the JS, diagram of routing coming soon! 

In the meantime, 🚢 🚢 🚢! 😎

PR #1499, opened 2018-03-02
Fix padding factoring in IncidentCard

kevinrobinson:
# Who is this PR for?
educators

Follow-on to fix styling bug from refactoring in https://github.com/studentinsights/studentinsights/pull/1492.

# What does this PR do?
Fixes refactoring bug; I missed this with just one incident but it's quite visible in production with more than one incident :)

(avoiding screenshotting any data, you can see the vertical margin)
screen shot 2018-03-02 at 1 08 51 pm

# Checklists
## Javascript QA

### DIFF
--- a/app/assets/javascripts/components/IncidentCard.js
+++ b/app/assets/javascripts/components/IncidentCard.js
@@ -9,42 +9,40 @@ class IncidentCard extends React.Component {
   render() {
     const {incident} = this.props;
     return (
-      <div style={{paddingTop: 60}}>
-        <div style={styles.box} key={incident.id}>
-          <div style={styles.header}>
-            <div style={styles.item}>
-              <span style={styles.itemHead}>
-                {'Date: '}
-              </span>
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/components/IncidentCard.js
```diff
@@ -9,43 +9,41 @@ class IncidentCard extends React.Component {
   render() {
     const {incident} = this.props;
     return (
-      <div style={{paddingTop: 60}}>
```
kevinrobinson: The diff looks messy, but all this is doing is removing the `<div style={{paddingTop: 60}}>` wrapping everything; that shouldn't have been factored out of the caller (which still has it here: https://github.com/studentinsights/studentinsights/blob/master/app/assets/javascripts/student_profile/AttendanceDetails.js#L141).

kevinrobinson:
selfie

PR #1502, opened 2018-03-04
WIP Discipline dashboard

edavidsonsawyer:
# Who is this PR for?
Principals and Assistant Principals 

School admins would like a breakdown of data on discipline incidents

# What does this PR do?
Creates a dashboard showing tally of discipline incidents broken out by category. 

# Checklists

## Javascript QA

+ [ ] Author checked latest in IE - Discipline Dashboard

## Automated Testing

+ [ ] Author improved specs for code in need of better test coverage

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
@@ -1,0 +1,85 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import _ from 'lodash';
+
+import DashboardHelpers from '../DashboardHelpers';
+import StudentsTable from '../StudentsTable';
+import DashboardBarChart from '../DashboardBarChart';
+import DateSlider from '../DateSlider';
+
+class SchoolDisciplineDashboard extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      startDate: '2012-08-10', //DashboardHelpers.schoolYearStart(),
+      selectedType: null
+    };
+    this.setDate = (range) => {
+      this.setState({
+        startDate: moment.unix(range[0]).format("YYYY-MM-DD")
+      });
+    };
+    this.setStudentList = (highchartsEvent) => {
+      this.setState({selectedType: highchartsEvent.point.category});
+    };
+    this.resetStudentList = () => {
+      this.setState({selectedType: null});
+    };
+  }
+
+  filterIncidentDates(incidents) {
+    return incidents.filter((incident) => {
+      return moment(incident.occurred_at).isSameOrAfter(moment(this.state.startDate));
+    });
+  }
+
+  render() {
+    return(this.renderDisciplineChart(this.props.disciplineIncidentsByLocation));
+  }
+
+  renderDisciplineChart(group) { //turn this into independant function
+    let seriesData = [];
+    let categories = Object.keys(group);
+    categories.forEach((type) => {
+      const incidents = this.filterIncidentDates(group[type]);
+      seriesData.push([type, incidents.length]);
+    });
+
+    return (
+        <DashboardBarChart
+          id = {'string'}
+          categories = {{categories: categories}}
+          seriesData = {seriesData}
+          titleText = {group}
+          measureText = {'Number of Incidents'}
+          tooltip = {{
+            pointFormat: 'Total incidents: <b>{point.y}</b>'}}
+          onColumnClick = {this.resetStudentList}
+          onBackgroundClick = {this.resetStudentList}/>
+    );
+  }
+
+  renderDateRangeSlider() { //maybe its own function too
+    const firstDate = DashboardHelpers.schoolYearStart();
+    const lastDate = moment();
+    return (
+      <DateSlider
+        rangeStart = {parseInt(moment(firstDate).format("X"))}
+        rangeEnd = {parseInt(moment(lastDate).format("X"))}
+        setDate={this.setDate}/>
+    );
+  }
+}
+
+SchoolDisciplineDashboard.propTypes = {
+  disciplineIncidentsByLocation: PropTypes.object.isRequired,
+  disciplineIncidentsByTime: PropTypes.object.isRequired,
+  disciplineIncidentsByClassroomn: PropTypes.object.isRequired,
+  disciplineIncidentsByGrade: PropTypes.object.isRequired,
+  disciplineIncidentsByDay: PropTypes.object.isRequired,
+  disciplineIncidentsByOffense: PropTypes.object.isRequired,
+  disciplineIncidentsByRace: PropTypes.object.isRequired
+};
+
+export default SchoolDisciplineDashboard;
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolWideDisciplineIncidents.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolWideDisciplineIncidents.js
@@ -1,0 +1,52 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import _ from 'lodash';
+
+import SchoolDisciplineDashboard from './SchoolDisciplineDashboard';
+
+class SchoolWideDisciplineIncidents extends React.Component {
+
+  schoolDisciplineEvents() {
+    //structures the discipline incident data for easy grouping
+    const students = this.props.dashboardStudents;
+    let schoolDisciplineEvents = [];
+    students.forEach((student) => {
+      student.discipline_incidents.forEach((incident) => {
+        schoolDisciplineEvents.push({
+          student_id: incident.student_id,
+          location: incident.incident_location,
+          time: incident.has_exact_time,
+          classroom: null,
+          student_grade: student.grade,
+          day: moment(incident.occurred_at).format("DDD"),
+          offense: incident.incident_code,
+          student_race: student.race,
+          occurred_at: incident.occurred_at
+        });
+      });
+    });
+
+    return schoolDisciplineEvents;
+  }
+
+  render() {
+    const schoolDisciplineEvents = this.schoolDisciplineEvents();
+    return (
+      <SchoolDisciplineDashboard
+        disciplineIncidentsByLocation={_.groupBy(schoolDisciplineEvents, 'location')}
+        disciplineIncidentsByTime={_.groupBy(schoolDisciplineEvents, 'time')}
+        disciplineIncidentsByClassroomn={_.groupBy(schoolDisciplineEvents, 'classroom')}
+        disciplineIncidentsByGrade={_.groupBy(schoolDisciplineEvents, 'student_grade')}
+        disciplineIncidentsByDay={_.groupBy(schoolDisciplineEvents, 'day')}
+        disciplineIncidentsByOffense={_.groupBy(schoolDisciplineEvents, 'offense')}
+        disciplineIncidentsByRace={_.groupBy(schoolDisciplineEvents, 'student_race')}
+      />
+    );
+  }
+}
+
+SchoolWideDisciplineIncidents.propTypes = {
+  dashboardStudents: PropTypes.array.isRequired
+};
+
+export default SchoolWideDisciplineIncidents;
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -32,7 +32,7 @@ class SchoolsController < ApplicationController
       redirect_to not_authorized_path and return #TodDo: determine whether there's a more appropriate action here
     end
 
-    dashboard_students = students_for_dashboard(@school).includes(:homeroom, :dashboard_absences, :dashboard_tardies)
+    dashboard_students = students_for_dashboard(@school).includes(:homeroom, :dashboard_absences, :dashboard_tardies, :discipline_incidents)
                                                         .map { |student| individual_student_dashboard_data(student) }
 
     @serialized_data = {students: dashboard_students.to_json}
@@ -131,9 +131,12 @@ class SchoolsController < ApplicationController
       first_name: student.first_name,
       last_name: student.last_name,
       id: student.id,
+      grade: student.grade,
+      race: student.race,
       homeroom: student.try(:homeroom).try(:name),
       absences: student.dashboard_absences,
-      tardies: student.dashboard_tardies
+      tardies: student.dashboard_tardies,
+      discipline_incidents: student.discipline_incidents
       }
     )
   end
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardOverview.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardOverview.js
@@ -9,6 +9,7 @@ export default function DashboardOverview () {
       <ul>
         <li><Link to='/absences_dashboard'>Absences Dashboard</Link></li>
         <li><Link to='/tardies_dashboard'>Tardies Dashboard</Link></li>
+        <li><Link to='/discipline_dashboard'>Discipline Dashboard</Link></li>
       </ul>
     </div>);
 }
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/SchoolAdministratorDashboards.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/SchoolAdministratorDashboards.js
@@ -5,6 +5,7 @@ import {Switch, Route} from 'react-router-dom';
 import DashboardOverview from './DashboardOverview';
 import SchoolwideAbsences from './absences_dashboard/SchoolwideAbsences';
 import SchoolwideTardies from './tardies_dashboard/SchoolwideTardies';
+import SchoolWideDisciplineIncidents from './discipline_dashboard/SchoolWideDisciplineIncidents';
 
 export default function SchoolAdministratorDashboards ( {serializedData} ) {
   const {students} = serializedData;
@@ -13,6 +14,7 @@ export default function SchoolAdministratorDashboards ( {serializedData} ) {
     <Route exact path="/" render={ () => <DashboardOverview />} />
     <Route path="/absences_dashboard" render={ () => <SchoolwideAbsences dashboardStudents={students}/>} />
     <Route path="/tardies_dashboard" render={ () => <SchoolwideTardies dashboardStudents={students}/>} />
+    <Route path="/discipline_dashboard" render={ () => <SchoolWideDisciplineIncidents dashboardStudents={students}/>} />
   </Switch>);
 }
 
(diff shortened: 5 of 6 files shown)


edavidsonsawyer:
@kevinrobinson This may work for a first pass. Instead of throwing a bunch of charts in here, there's just a dropdown that lets you select which one to view. It's simple to omit any grouping we don't want to immediately put in front of test users. 

As is, the branch doesn't include a link to the discipline section so we can control when users see it, though that's also just a moment to put in. The routes for the dashboard are all in place and you can reach it directly: /school_administrator_dashboard#/discipline_dashboard

app/controllers/schools_controller.rb
```diff
@@ -34,6 +34,9 @@ def school_administrator_dashboard
       individual_student_dashboard_data(student)
     end
 
+    dashboard_students = students_for_dashboard(@school).includes([homeroom: :educator], :dashboard_absences, :event_notes, :dashboard_tardies, :discipline_incidents)
+                                                        .map { |student| individual_student_dashboard_data(student) }
```
kevinrobinson: Let's talk about this; it impacts the existing pages load more data that users can't yet see, and those pages are already slow enough we'll need to optimize this.  I can look at this more next week with you and figure out what design makes sense, and then I can take care of the server-side optimizations since it's it's hard to optimize for production data without being able to measure :)

app/demo_data/fake_student.rb
```diff
@@ -218,11 +218,11 @@ def add_attendance_events
...
-      0 => 0.83,
-      1 => 0.10,
-      2 => 0.03,
-      (3..5) => 0.03,
-      (6..15) => 0.01,
+      0 => 0.23,
+      1 => 0.30,
+      2 => 0.23,
+      (3..5) => 0.13,
+      (6..15) => 0.11,
```
kevinrobinson: Can you say more about this?

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolWideDisciplineIncidents.js
```diff
@@ -0,0 +1,60 @@
...
+class SchoolWideDisciplineIncidents extends React.Component {
+
+  schoolDisciplineEvents() {
+    //structures the discipline incident data for easy grouping
+    const students = this.props.dashboardStudents;
+    const startDate = DashboardHelpers.schoolYearStart();
+    let schoolDisciplineEvents = [];
+    students.forEach((student) => {
+      student.discipline_incidents.forEach((incident) => {
+        if (moment(incident.occurred_at).isSameOrAfter(moment(startDate))) {
```
kevinrobinson: for now, `moment.utc` in these?

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolWideDisciplineIncidents.js
```diff
@@ -0,0 +1,60 @@
...
+            location: incident.incident_location,
+            time: incident.has_exact_time ? moment(incident.occurred_at).format("h A") : "Not Logged",
+            classroom: student.homeroom_label,
+            student_grade: student.grade,
+            day: moment(incident.occurred_at).format("ddd"),
+            offense: incident.incident_code,
+            student_race: student.race,
+            occurred_at: incident.occurred_at,
+            last_sst_date_text: latestNoteDateText(300, student.event_notes)
+          });
```
kevinrobinson: Can you say more about why you're mapping these items into a different shape here?  I was expecting this to happen in the rendering and I'm wondering if this is maybe because of how you've factored the charts.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -0,0 +1,184 @@
...
+    super(props);
+    this.state = {
+      startDate: DashboardHelpers.schoolYearStart(),
+      selectedChart: {
+        type: "location",
+        data: this.props.disciplineIncidentsByLocation,
+        title: "Incidents by Location"},
+      selectedCategory: null
+    };
+    this.setDate = (range) => {
```
kevinrobinson: Can you say more about what all these methods in the constructor are doing?  If there's some initial state that the component should have, can we just set that in `this.state = {...}` and let render react to that and render the correct things?

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -0,0 +1,184 @@
...
+    let studentDisciplineIncidentCounts = {};
+    const incidents = incidentCategory ? this.state.selectedChart.data[incidentCategory] : this.props.totalDisciplineIncidents;
+    this.filterIncidentDates(incidents).forEach((incident) => {
+      studentDisciplineIncidentCounts[incident.student_id] = studentDisciplineIncidentCounts[incident.student_id] || 0;
+      studentDisciplineIncidentCounts[incident.student_id]++;
+    });
+    return studentDisciplineIncidentCounts;
+  }
+
+  updateChartSelection(selectedChart) {
```
kevinrobinson: These values are all a function of the value of `selectedChart`, so you don't need to store them as state.  They can be computed in `render` or one of the other methods it calls.  The overall idea is "state should have the minimal bits of actual state, and then compute everything else based on that."

app/demo_data/fake_student.rb
edavidsonsawyer: @kevinrobinson I meant to back this out after seeding my local db, but it might be worth keeping. With the default distribution, we don't have enough fake students to to get any discipline incidents in the dashboard, so the demo site will just look blank.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolWideDisciplineIncidents.js
edavidsonsawyer: @kevinrobinson yup! Will change.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: @kevinrobinson this was just to bind the methods since they're all passed as callbacks. Looking around I guess it's more typical to make a prototype and bind it in the constructor, so I'll switch to that

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: Good point @kevinrobinson.

app/demo_data/fake_student.rb
kevinrobinson: Ah got it.  Reasonably realistic distributions are good, but development data that is actually helpful for development is the most important :)

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
kevinrobinson: Ah gotcha, I misread this.  Yeah, you can define these as methods just like `filterIncidentDates`, and then do the binding in the constructor, which I think is the same thing your'e saying here :)

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolWideDisciplineIncidents.js
edavidsonsawyer: @kevinrobinson It's not necessary to do this here. This component could be combined with the disciplineDashboard component, but it made more sense to me to just get the items once and pass them as a prop to the component that needs to manipulate them. Is that what you meant by having this happen in the render?

app/controllers/schools_controller.rb
edavidsonsawyer: @kevinrobinson I'd love to go over this. I've tried to tighten up the queries, but probably the way to go is to separate these based on the dashboard the user is looking for.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolWideDisciplineIncidents.js
kevinrobinson: Ah yeah I see what you're saying.  This component is transforming the data from the server, and the `SchoolDisciplineDashboard` component is visualizing it in different ways.  So you're putting that transformation here, and then the other component doesn't have to transform further, and can render that data in a few different ways, is that right?

The main thing I'm commenting on here is that those two pieces of code are tightly coupled but not located next each other - the transformation here is only correct if it's doing what the UI code in `SchoolDisciplineDashboard` expects, and the UI code in `SchoolDisciplineDashboard` is only correct if it's working with data in the shape that this component is making.  This might be fine here, but if you keep the distance between them, maybe add in the `propTypes` about what particular fields need to be passed to `SchoolDisciplineDashboard`.  Then that adds a small check that the "seam" between those two components works correctly.

app/controllers/schools_controller.rb
kevinrobinson: Sure, I'll look at this this morning and send you the other notes that came up Friday.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolWideDisciplineIncidents.js
edavidsonsawyer: @kevinrobinson yeah, that's it. AFAIK it won't be such a big deal given the likely number of discipline incidents, but separating out this transform made a big difference in the absence dashboard's performance. 

Good point about making the connection clear. I'll make that more explicit.

kevinrobinson: Gotcha.  Yeah, caching repeated computations is a great way to make this faster!  The comment here suggests that the code is doing this for easy grouping, so it might be good to edit or add a one-line comment explaining how it impacts performance.

kevinrobinson:
Hey @edavidsonsawyer!  From the WIP description, I'm assuming you're still working on this, which is fine.  Ping me if you need anything else here or if you want to review and ship.

edavidsonsawyer:
@kevinrobinson @alexsoble, I think this should be alright to review. The route is separate so we won't see any extra performance trouble on the existing dashboard.

alexsoble:
@edavidsonsawyer I can take a look!

alexsoble:
@edavidsonsawyer I gave this a review in Internet Explorer and the feature looks great, overall I thing this is good to ship! While reading the code I had some thoughts about variable names that confused me in a few points and I'll leave notes about those, let me know if you think those suggestions make sense — we don't have to block shipping for them tho.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolwideDisciplineIncidents.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import SchoolDisciplineDashboard from './SchoolDisciplineDashboard';
+import DashboardHelpers from '../DashboardHelpers';
+import {latestNoteDateText} from '../../../helpers/latestNoteDateText';
+
+class SchoolWideDisciplineIncidents extends React.Component {
+
+  schoolDisciplineEvents() {
```
alexsoble: This data shape is referred to as `totalDisciplineIncidents` in the SchoolDisciplineDashboard. When I was trying to figure out where the data shape came from, I kept searching for `totalDisciplineIncidents`, but it's named `schoolDisciplineEvents()` at the parent level. Do you think it would simplify to use `totalDisciplineIncidents()` up here too?

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolwideDisciplineIncidents.js
```diff
@@ -0,0 +1,51 @@
...
+    //structures the discipline incident data for easy grouping
+    const students = this.props.dashboardStudents;
+    const startDate = DashboardHelpers.schoolYearStart();
+    let schoolDisciplineEvents = [];
+    students.forEach((student) => {
```
alexsoble: Very small thing but I wonder if using `map` would be more JS-idiomatic here.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -0,0 +1,160 @@
...
+
+  filterIncidentDates(incidents) {
+    return incidents.filter((incident) => {
+      return moment.utc(incident.occurred_at).isSameOrAfter(moment.utc(this.state.startDate));
+    });
+  }
+
+  studentDisciplineIncidentCounts(incidentCategory) {
+    let studentDisciplineIncidentCounts = {};
+    const selectedChart = this.getChartData(this.state.selectedChart);
```
alexsoble: On the left side of this assignment `selectedChart` is an object, but on the right side `selectedChart` is a string.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -0,0 +1,160 @@
...
+          {this.renderDateRangeSlider()}
+          {this.renderStudentDisciplineTable()}
+        </div>
+      </div>
+    );
+  }
+
+  renderDisciplineChart(group) {
+    let seriesData = [];
+    const selectedChart = this.getChartData(this.state.selectedChart);
```
alexsoble: I think this a good example of where the variable naming can make a difference. This function takes an argument called `group`, what's getting passed in is `this.getChartData(this.state.selectedChart)`. So it looks like we're computing `this.getChartData(this.state.selectedChart)` a second time when we already have it available as `group`. 

It looks like we could safely delete this line. Instead of `group`, maybe we could call the argument `selectedChartObject`? Or something else that shows it's different from the string value of `selectedChart` stored on `this.state`?

edavidsonsawyer: @alexsoble thanks for catching this! Some of this is left over from an earlier iteration of this component and this piece was moved out without being removed here.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: @alexsoble same as below, this should have been changed with an earlier refactor. Thanks!

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
alexsoble: @edavidsonsawyer Ah, awesome, glad this was helpful!

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolwideDisciplineIncidents.js
edavidsonsawyer: Definitely!

edavidsonsawyer:
@kevinrobinson @alexsoble I hate to keep pinging on this, but I remember this week was going to have a few opportunities to show this to folks in school so if it's possible to ship this, we should try to do it soon.

kevinrobinson:
@edavidsonsawyer Yeah sorry about the slowdown in communication on my part (and for missing the last two Fridays).

There's three things going on - 1) we're working on starting up a separate thread of work, 2) there have been a bunch of in-person conversations with folks where we've learned more about the complexity of the human data entry flows for discipline data, which means we'll need to spend more time talking with people about what they see and digging into how that data is entered, rather than the prototype being immediately usable and 3) many of the same folks we'd have these conversations with about discipline data are the same people who are waiting on improvements to the absences and tardies features, which are still in progress.  So for me, helping out with absences and tardies improvements is what I've been prioritizing, which is not to say it's moving fast enough :)

All that being said, I know it's a bummer when you make progress on something and a few days later it still hasn't shipped or been used by people yet.  Overall I think we should keep striving to reduce the time from something being ready as a prototype to getting feedback, but this is always hard to balance.  Concretely, If you're feeling like you need the next thing to work on, I'd suggest focusing on the improvements to the absences views next (which people are already actively using), and if there's more I can do to help tomorrow I can spend some time on that tomorrow and Friday too.

alexsoble:
Hey there @edavidsonsawyer — no need to apologize for the ping, I think it was helpful! I agree with what @kevinrobinson said above about why we've been a bit slow on this / alternative directions we can be making progress in. 

I had some time this afternoon to review, so I figured I'd try to cut out an iteration by reviewing this feature against real data from the Healey. I noticed a few things that I think would be helpful to fix up on this PR.

## X-Axis: days of the week

Here's how the x-axis looks right now, I think users would expect this to be sorted in MTWThF order:

days of week

## X-Axis: time of day

I think users will expect these to be in order: 

time

## X-Axis: location

Let's handle `null`:

locations

These things may have been hard to notice/discover without production data! Let me know if you need more information about how the data looks in production in order to debug. My assumption is that both days of the week and time of day are coming from the `occurred_at` datetime, so the demo data should be similar enough to the production data to work with.

edavidsonsawyer:
@kevinrobinson thanks for the update! I don't mind myself when things take a while to ship (I'll be opening up a PR soon with some of the other dashboard improvements I've been working on in the meantime), but wanted to be sure we weren't missing a window after Uri was asking about it. 

It's interesting to hear about the discipline data! I'm curious to know what you've discovered. Probably a good topic for Friday.

kevinrobinson:
@edavidsonsawyer oops, some miscommunication here.  @alexsoble or I will help ship this today or tomorrow, so @snoopyuri can show it to folks tomorrow, and we'll let the absences and tardies improvements slip.

alexsoble:
Works smoothly in prod @edavidsonsawyer. One other x-axis fix I noticed is that we should put grade levels in order too:

screen shot 2018-04-12 at 3 31 00 pm

Issue #1500, opened 2018-03-02
JS error reporting analytics on school_administrator_dashboard route


kevinrobinson:
<img width="487" alt="screen shot 2018-03-02 at 1 15 07 pm" src="https://user-images.githubusercontent.com/1056957/36914575-cbe6c33e-1e1b-11e8-8a4c-5a46c6aefae9.png">

I'm assuming this is because `currentEducator` isn't set.  We could either not track visits to this page, which would be fine, or could fix the bug so that value is present for the code trying to track these analytics.  cc @edavidsonsawyer 

This also isn't causing anyone any problems :)

kevinrobinson:
Actually this impacts when you visit either dashboard directly too, so we definitely need to fix this up.  It probably didn't come up locally because analytics logging is disabled.  We could improve the way we do that so that we'd catch things like this locally more easily too.

kevinrobinson:
Fixed by https://github.com/studentinsights/studentinsights/pull/1503

PR #1503, opened 2018-03-05
fixing analytics error in dashboard

edavidsonsawyer:
# Who is this PR for?
Anyone making use of analytics
currentEducator isn't being set in the dashboard, throwing a js error and preventing us from tracking visits cf #1500 
# What does this PR do?
sets the currentEducator

# Checklists

+ [ ] Author checked latest in IE - School Dashboard
+ [ ] Reviewer checked latest in IE - School Dashboard

### DIFF
--- a/spec/javascripts/school_administrator_dashboard/StudentsTable.test.js
+++ b/spec/javascripts/school_administrator_dashboard/StudentsTable.test.js
@@ -21,7 +21,7 @@ describe('Dashboard Students Table', () => {
     expect(cellTexts).toEqual([
       "Pierrot Zanni",
       "3",
-      moment().subtract(3, 'months').format('M/D/YY'),
+      moment.utc().subtract(3, 'months').format('M/D/YY'),
     ]);
   });
 
--- a/spec/javascripts/school_administrator_dashboard/StudentsTable.test.js
+++ b/spec/javascripts/school_administrator_dashboard/StudentsTable.test.js
@@ -21,7 +21,7 @@ describe('Dashboard Students Table', () => {
     expect(cellTexts).toEqual([
       "Pierrot Zanni",
       "3",
-      moment.utc().subtract(3, 'months').format('M/D/YY'),
+      moment().subtract(3, 'months').format('M/D/YY'),
     ]);
   });
 
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -35,7 +35,9 @@ class SchoolsController < ApplicationController
     dashboard_students = students_for_dashboard(@school).includes(:homeroom, :dashboard_absences, :event_notes, :dashboard_tardies)
                                                         .map { |student| individual_student_dashboard_data(student) }
 
-    @serialized_data = {students: dashboard_students.to_json}
+    @serialized_data = {
+      students: dashboard_students.to_json,
+      current_educator: current_educator}
     render 'shared/serialized_data'
   end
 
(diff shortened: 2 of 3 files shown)


edavidsonsawyer:
Hi @kevinrobinson, I think this should address #1500 but I'm not exactly sure how to test it. I can log currentEducator and see that the value is set appropriately, but I don't know how to verify the visits are registered.

edavidsonsawyer:
Or rather it will once I figure out why my spec is showing inconsistent dates.

spec/javascripts/school_administrator_dashboard/StudentsTable.test.js
```diff
@@ -21,7 +21,7 @@ describe('Dashboard Students Table', () => {
     expect(cellTexts).toEqual([
       "Pierrot Zanni",
       "3",
-      moment.utc().subtract(3, 'months').format('M/D/YY'),
+      moment().subtract(3, 'months').format('M/D/YY'),
```
kevinrobinson: @edavidsonsawyer Can you say more about this, shouldn't this change fail the test?

kevinrobinson:
👍

spec/javascripts/school_administrator_dashboard/StudentsTable.test.js
edavidsonsawyer: You know, at the time I didn't think so because I assumed the SST data needed to do something like moment(threemonthsago).utc instead of moment.utc(threemonthsago) but going over the moment docs again I guess that's wrong. So it should fail the test.

Here's the thing. The test only stopped failing after I removed utc here. If you check out the travis error on my initial commit you'll see it was getting inconsistent dates between the spec and the test data. I'll work through this a bit more later tonight and see if I can figure out what's up here. Out of curiosity, why do you parse the SST date as utc?

edavidsonsawyer: @kevinrobinson actually, thinking about this again, I see what's up. The threeMonthsAgo variable creates a local moment and subtracts three months. The SST date parses that string as utc with moment.utc(threeMonthsAgo), but that doesn't change the time, it just sets the utc offset to 0.  

In the spec, moment.utc().subtract(3, 'months) appropriately gives you a date 3 months from the current UTC time since it creates a utc moment and subtracts three months from that. So that won't be the same as the date above and this spec will fail for any code I push after dinner.

edavidsonsawyer: I don't know how important it is to keep the SST date in utc. We could fix it by creating the initial variable in utc. We could use moment(threeMonthsAgo).utc() to get the SST test date and revert this change we're commenting on. It kind of looks like we could also just hard code the date since I don't believe we're testing any code that requires the SST date to happen within a specific timeframe, but I could be wrong here.

edavidsonsawyer: Nevermind all that. That odd parsing behavior seems like an artifact of the way those dates were formatted. If they're changed to ISO 8601 then utc parsing works like you'd expect, so I just did that.

kevinrobinson: @edavidsonsawyer Yeah this is great to bring up.  In general, my default assumption has been "store all times in UTC always," and then if we want to project them to local time, only do that at the very end when we show it to the user.  But we probably don't handle that entirely correctly in a bunch of places.  The new home page will surface this in https://github.com/studentinsights/studentinsights/pull/1505/files#diff-8194c0c736a7a4045e1d0a21d7715997R46, so we can sort it out there.

Separately, in code and in tests I've found it can be helpful to make the "now" time explicit, so that in tests all runs are deterministic and consistent.  The way you tackled that with making all the times relative is great too, but sometimes it's clearer in the test to write concretely "the date is 3/5/18" and so events that happened on "12/19/17" shouldn't show up.

kevinrobinson:
thanks @edavidsonsawyer! Merging now and will verify this in production.

PR #1482, opened 2018-02-27
Add columns for sorting by last MTSS, NGE and 10GE on school overview, homeroom and section

kevinrobinson:
# Who is this PR for?
high school teachers, JR and other NGE teachers in particular

# What problem does this PR fix?
short-term fix for not being able to find which students have recently had NGE notes.

# What does this PR do?
Updates the school overview page, homeroom page and section page to add columns for last MTSS meeting, and in the high school last NGE meeting and 10GE meeting.  Per JR, they don't use MTSS at the high school but do still have SST.  This also reorders the columns to put house and counselor and homeroom on-screen.

In the process, a test bug surfaced where the students created by `TestPals` don't have an `enrollment_status` set, so this fixes that as well.

# Screenshot (if adding a client-side feature)
#### before, HS:
screen shot 2018-02-27 at 9 28 45 am

#### after, HS:
screen shot 2018-02-27 at 5 05 02 pm

#### before, K8:
screen shot 2018-02-27 at 9 28 38 am

#### after, K8:
screen shot 2018-02-27 at 9 25 28 am

#### before, HS:
screen shot 2018-02-27 at 5 08 09 pm

#### after, HS:
screen shot 2018-02-27 at 5 06 16 pm
screen shot 2018-02-27 at 5 07 16 pm

#### before, K8:
screen shot 2018-02-27 at 5 08 06 pm

#### after, K8:
screen shot 2018-02-27 at 5 06 31 pm

#### before, HS:
screen shot 2018-02-27 at 5 08 02 pm

#### after, HS:
screen shot 2018-02-27 at 5 07 36 pm

# Checklists
+ [x] Author checked latest in IE - School Overview
+ [x] Author checked latest in IE - Homeroom Page
+ [x] Author checked latest in IE - Section Page
+ [ ] Reviewer checked latest in IE - School Overview
+ [ ] Reviewer checked latest in IE - Homeroom Page
+ [ ] Reviewer checked latest in IE - Section Page
+ [x] Author improved specs for code in need of better test coverage
+ [x] Author included specs for new code

PR #1505, opened 2018-03-06
Add new home page with feed and initial insight box

kevinrobinson:
this is on top of https://github.com/studentinsights/studentinsights/pull/1496, and merges it in

# Who is this PR for?
all educators, HS NGE teachers are first partners

Teachers can't really see what's happening recently across all their students, even if those students have a lot of information entered in Insights.  They have to look at student profiles individually.  https://github.com/studentinsights/studentinsights/pull/1482 helps this a bit, but only per section for most teachers, and even then they can't see the substance of the notes.

Also, for NGE teachers in particular, teachers and dept. heads can't see which students are failing at-a-glance, so they skim through and do this across several screens now.  This adds a first "insight" box that directly answers this question using the threshold that the NGE team uses.

# What does this PR do?
Adds a new home page, with UI and endpoints.  It also factors out a few UI components and adds some related test helper packages.

This PR doesn't look at optimizing the endpoint response time yet, and my intention was to ship this dark and then look at optimizing that separately.  We'll also have to build some other tools or endpoints to look at this across users.  So the intention with this PR is to ship the first iteration for Friday morning ahead of 10GE, so any page that loads at all is fine, even if it takes 20 seconds.

This PR also is intended to be used just by HS teachers for now, so it doesn't branch on the "insight" box based on whether the user is in a HS or not, or based on the deployment.

It also adds in a commit that uses React `context` to pass through a clock function, and then uses that in tests to freeze the time: https://github.com/studentinsights/studentinsights/pull/1505/commits/238bcb361e0ea5ef4240dff7845f7760f446b00c

screen shot 2018-03-06 at 5 22 18 pm

# Checklists
## Javascript QA

## Automated Testing


app/assets/javascripts/components/Educator.js
```diff
@@ -13,11 +13,14 @@ class Educator extends React.Component {
   }
 
   render() {
-    const educator = this.props.educator;
+    const {educator, style} = this.props;
```
kevinrobinson: Allowing `style` to be passed in, say to change font size.

app/assets/javascripts/components/eventNoteType.js
```diff
@@ -0,0 +1,22 @@
+const textMap =  {
+  300: 'SST Meeting',
+  301: 'MTSS Meeting',
+  302: 'Parent conversation',
+  304: 'Something else',
+  305: '9th Grade Experience',
+  306: '10th Grade Experience'
+};
```
kevinrobinson: Handling this may be related to https://github.com/studentinsights/studentinsights/pull/1461 and https://github.com/studentinsights/studentinsights/pull/1481, but I was trying to punt all that from this PR and deal with it next week.

app/assets/javascripts/components/eventNoteType.js
```diff
@@ -0,0 +1,22 @@
...
+  return textMap[eventNoteTypeId] || 'Other';
+}
+
+const colorMap =  {
+  300: '#31AB39',
+  301: '#EB4B26',
+  302: '#CDD71A',
+  305: '#139DEA',
+  306: '#333333'
+};
```
kevinrobinson: Similarly, I think these colors are experimental at this point, and all subject to change next week.

kevinrobinson:
Hey @alexsoble!  Want to talk through this on Thursday and then review?  I was going to write up a better note for reviewing, but ran out of time today.

For reviewing, my suggestion would be check out the screenshot, then review the Rails code starting in `routes.rb` and `home_controller.rb`.  And then the new page for the UI is `HomePage` which has `HomeFeed` and `HomeInsights` as the main two pieces.  Also let me know what else I can do to make this more reviewer-friendly too!

kevinrobinson:
This should fix up the tests.  https://github.com/studentinsights/studentinsights/pull/1505/commits/238bcb361e0ea5ef4240dff7845f7760f446b00c uses React context to pass through a clock function (`nowFn`), and then uses that in tests to freeze the time.  There was one place where it was still using `moment#fromNow` which implicitly reads the JS wall clock and so the tests started failing today.

kevinrobinson:
https://github.com/studentinsights/studentinsights/pull/1505/commits/ea8904adfe57021d6929c4bb53bd86e3cca4af94 Removes restricted notes from the feed for this first iteration.  I think a better UX could be showing a placeholder card that says there's a restricted note for the student from a particular on a particular date would be good, with an explicit "click to fetch and show the content" might be better, but I want to cut that out of this chunk of work and deal with it separately.

app/lib/feed.rb
```diff
@@ -0,0 +1,69 @@
+class Feed
+  def initialize(educator)
+    @educator = educator
+    @authorizer = Authorizer.new(@educator)
+  end
+
+  # Return a list of `FeedCard`s for the feed based on event notes.
+  def event_note_cards(from_time, limit)
+    event_notes = @authorizer.authorized do
```
alexsoble: Talking this out with @kevinrobinson, maybe we want to pass the event notes as a straight argument?

kevinrobinson:
Talked through some of this with @alexsoble, going to ship this in beta state now, with the main issue being load times from <1sec - ~15sec for `feed_json` depending on the user.  Improving that'll be the focus for tomorrow, but this lets folks look at their real data for this tomorrow morning.

kevinrobinson:
and @alexsoble if you have time tomorrow, please share and review feedback you have if you're able, and if not we can talk through the other bits tomorrow afternoon and pair on improving them 👍

PR #1506, opened 2018-03-07
Changing homeroom name to teacher name in dashboard

edavidsonsawyer:
# Who is this PR for?
Principals and APs

Per MTSS meeting, it's more useful to have teacher names than Homeroom names in the dashboard

# What does this PR do?
Swaps homeroom names for teacher names in dashboard

screen shot 2018-03-07 at 10 21 29 am

# Checklists

## Javascript QA

### DIFF
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -131,7 +131,7 @@ class SchoolsController < ApplicationController
       first_name: student.first_name,
       last_name: student.last_name,
       id: student.id,
-      homeroom: student.try(:homeroom).try(:educator).try(:full_name),
+      homeroom_label: student.try(:homeroom).try(:educator).try(:full_name),
       absences: student.dashboard_absences,
       tardies: student.dashboard_tardies,
       event_notes: student.event_notes
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -32,7 +32,7 @@ class SchoolsController < ApplicationController
       redirect_to not_authorized_path and return #TodDo: determine whether there's a more appropriate action here
     end
 
-    dashboard_students = students_for_dashboard(@school).includes(:homeroom, :dashboard_absences, :event_notes, :dashboard_tardies)
+    dashboard_students = students_for_dashboard(@school).includes([homeroom: :educator], :dashboard_absences, :event_notes, :dashboard_tardies)
                                                         .map { |student| individual_student_dashboard_data(student) }
 
     @serialized_data = {students: dashboard_students.to_json}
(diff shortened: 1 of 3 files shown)


edavidsonsawyer:
@alexsoble @kevinrobinson I can't remember whether this was definitely a change we wanted to make, but it's simple and ready to go if we want to merge it.

app/controllers/schools_controller.rb
```diff
@@ -131,7 +131,7 @@ def individual_student_dashboard_data(student)
       first_name: student.first_name,
       last_name: student.last_name,
       id: student.id,
-      homeroom: student.try(:homeroom).try(:name),
+      homeroom: student.try(:homeroom).try(:educator).try(:full_name),
```
kevinrobinson: This seems like it could easily be misinterpreted.  This seems like a great change, but what do you think about renaming this field to `homeroom_text` to clarify it's not a homeroom, homeroom id, or homeroom name?

kevinrobinson:
@edavidsonsawyer Awesome!  👍 

It'd be great to ship this ahead of a Thursday morning MTSS (assuming we don't get too much ❄️ ).  This might need some other visual/design work, since there are many more homerooms than this in the production data set (say 20 or so in a K8 school).  I think we should ship this as-is and see what happens, but if you have time we could also simulate that with more homerooms locally to see what it looks like.

app/controllers/schools_controller.rb
edavidsonsawyer: @kevinrobinson good point. How about homeroom_label?

kevinrobinson: do it! 👍

edavidsonsawyer:
@kevinrobinson it took a quick look at the test pals script we're using for demo data now and it might not be until tonight before I can add the homerooms and teachers. If you're comfortable shipping as is, that might be faster.

kevinrobinson:
@edavidsonsawyer Yeah I was going to do this for this morning, but it's a snow day so not as important :)  Let's ship this as-is for now.

FWIW, here's what I mean about the number of homerooms, this is a K8 school:

screen shot 2018-03-08 at 9 32 47 am

Here there are 36 homerooms, and so the labels on the bottom don't show all the homeroom names anyway, and definitely won't with the longer educator names.  This doesn't need to stop us from shipping this first version to folks though!

Stepping back a bit, it might be good to talk a bit more holistically about how folks might use this and the newer discipline dashboard as well.  The two things that have come up so far are: how do we help folks interpret what they see in this data, and not overreact to noise?  and how do we help them translate data into the actions or next steps they might take?  That might be a good thing to spend 15m on this Friday or next - essentially trying to use these interfaces ourselves to answer these questions and see what our experience is like.  cc @snoopyuri @alexsoble

kevinrobinson:
🚢ed, and folks can now see teachers names instead of homeroom numbers, making it slightly easier to take action (eg, if there are differences across classrooms, which teachers should I go talk to and see how I can help?) 👍

PR #1508, opened 2018-03-08
Validate discipline incident codes, improve local data

kevinrobinson:
# Who is this PR for?
developers, designers

Document expected values for incident codes, for validate and for building features around this.

# What does this PR do?
Adds a validation which adds in codes that we see, and also improves the local demo data to use real codes and locations.


kevinrobinson:
@edavidsonsawyer This might help a bit with https://github.com/studentinsights/studentinsights/pull/1502 too.

edavidsonsawyer:
@kevinrobinson this is great! Where did these come from? Is it a Somerville specific list?

kevinrobinson:
@edavidsonsawyer Yep, these is the aggregate of `DisciplineIncident.all.map(&:incident_code).uniq.sort` across all data we import in Somerville.  I assume it's a combination of how Somerville defines this, how Aspen supports this, and what DESE requires for reporting, but getting into more details there is probably a conversation for email.  This isn't trying to make realistic distributions or anything, just to make the codes quasi-realistic rather than the placeholder "A", "B", "C" they were before.

kevinrobinson:
Pushed another commit that updates all the specs to now use valid data, and removed the `FakeDisciplineIncident` class since it and the factory were doing the same thing.

PR #1511, opened 2018-03-09
Responsive Dashboard Styles

edavidsonsawyer:
# Who is this PR for?
Dashboard users

There are reports the dashboard doesn't fit well in all screen resolutions.

# What does this PR do?
Gives the dashboard more responsive styles

screen shot 2018-03-09 at 12 01 05 am

# Checklists

## Javascript QA

+ [x] Reviewer checked latest in IE - Tardies Dashboard

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardBarChart.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardBarChart.js
@@ -4,34 +4,6 @@ import _ from 'lodash';
 
 import HighchartsWrapper from '../../student_profile/HighchartsWrapper';
 
-const styles = {
-  title: {
-    color: 'black',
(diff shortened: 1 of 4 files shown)


edavidsonsawyer:
@kevinrobinson I'm hoping this will address some of the screen resolution issues you saw earlier. Even if something else needs to be done there, this at least improves the layout so users don't have to scroll to see all the information.

kevinrobinson:
@edavidsonsawyer This looks good to me!  I think that the actual issue at one of the K8 schools is that the project is only 800x600 resolution.  There's a `min-width` style on the `body` tag that applies to everything, so if that's the case then we'd have to change it there.  I was going to confirm this yesterday but it was a ❄️ day :)

This looks good to be on 1024x768, which is what most other pages are built for, so I'd say let's :ship this improvement!

PR #1512, opened 2018-03-09
Restrict dashboard date range to school year

edavidsonsawyer:
# Who is this PR for?
Dashboard users

The absence dashboard shows all absences for the past calendar year, but school year is more useful to users

# What does this PR do?
Restricts the absence dashboard to the current school year

## Javascript QA

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardHelpers.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardHelpers.js
@@ -44,7 +44,7 @@ export default {
 
   schoolYearStart: function() {
     const today = moment();
-    return today.month() < 8 ? today.subtract(1, 'year').year() + "-08-10" : today.year() + "-08-10";
+    return today.month() < 8 ? today.subtract(1, 'year').year() + "-08-10" : today.year() + "-08-15";
   },
 
   //slightly faster than Array.filter for getting a new date range
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardHelpers.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardHelpers.js
@@ -44,7 +44,7 @@ export default {
 
   schoolYearStart: function() {
     const today = moment();
-    return today.month() < 8 ? today.subtract(1, 'year').year() + "-08-10" : today.year() + "-08-15";
+    return today.month() < 8 ? today.subtract(1, 'year').year() + "-08-15" : today.year() + "-08-15";
   },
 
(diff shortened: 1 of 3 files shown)


kevinrobinson:
@edavidsonsawyer Awesome, this will really help principals looking at this!

Locally, it looks like there are two issues:
screen shot 2018-03-09 at 8 37 26 am

The big one is that the date range in the upper right (starting 8/10) doesn't match the date range in the chart on the left (starting 4/17), so we'll have to fix that before shipping this.

The other one is minor but I'm wondering if 8/10 will be intuitively read as "the start of the school year."  I see we could change that in `DashboardHelper`, and also that there's some repetition between that and the assumptions in `QuadConverter`, which uses 8/15.  We could clean that up at some point later (probably my bad for not noticing before).  But for data like this where we're showing people those dates directly in the UI (versus using them as a cutoff when aggregating), I think `9/1` is what most folks will expect as a default for the start of the school year.  What do you think?

edavidsonsawyer:
@kevinrobinson I'm glad you caught that! I had a local change that never made it into my commit! The initial chart rendering should match the slider date range now. Though now I see I need to change the way the spec handles this case. I should be able to get this passing soon. 

Re: school year start. Looking back at somerville school history, they consistently begin in August, but it's only a couple of days. I'm not sure whether information from the first couple days is useful or throwaway, I could see the argument both ways. I tend to think of August as the start of the school year, but I'm happy to defer to you if September's better. For now I matched August 15, just to be consistent.

edavidsonsawyer:
@kevinrobinson yeah, the way I tried to do that would have been rough on our specs. This is much simpler.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardHelpers.js
```diff
@@ -44,7 +44,7 @@ export default {
 
   schoolYearStart: function() {
     const today = moment();
-    return today.month() < 8 ? today.subtract(1, 'year').year() + "-08-10" : today.year() + "-08-10";
+    return today.month() < 8 ? today.subtract(1, 'year').year() + "-08-10" : today.year() + "-08-15";
```
kevinrobinson: Should these both be the same, whether it's 08-10 or 08-15?

kevinrobinson:
@edavidsonsawyer Cool, sounds good!  Let's go with what you think is best then and :ship and see how people use it next week. 👍

app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardHelpers.js
edavidsonsawyer: @kevinrobinson they certainly should.

PR #1513, opened 2018-03-09
Performance improvements for home page insight box around low grades

kevinrobinson:
this builds on https://github.com/studentinsights/studentinsights/pull/1505

# Who is this PR for?
HS educators, particularly dept. heads or counselors who have access to many students

~25 second load time for the insight box about low grades, for dept. heads

# What does this PR do?
It optimizes the query for students with low grades but who haven't been discussed recently in NGE or 10GE.

No changes:
screen shot 2018-03-09 at 8 14 27 am

# Checklists

### DIFF
--- a/app/assets/javascripts/home/HomeInsights.js
+++ b/app/assets/javascripts/home/HomeInsights.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import qs from 'query-string';
 import Card from '../components/Card';
 import Educator from '../components/Educator';
 import GenericLoader from '../components/GenericLoader';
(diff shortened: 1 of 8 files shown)


kevinrobinson:
This was the script I used to profile this:
```
def measure(tag)
  value = nil
  puts tag + (Benchmark.measure { value = yield }).to_s
  value
end

values = []
times = []
time_now = Time.at(1520529065)
limit = 20
educators = []
educators.each do |educator|
  puts "educator: #{educator.id}, email: #{educator.email}"
  time = Benchmark.measure do
    # cards = Feed.new(educator).event_note_cards(time_now, limit)
    cards = InsightUnsupportedLowGrades.new(educator).assignments(time_now, time_now - 30.days, 69)
    values << cards
  end
  puts time
  times << time
  puts
  puts
  puts '---------'
end;nil
times.map {|time| (time.real * 1000).round(0) }
```

I ran this across a few samples of 20 educators first and found that time ranged from 1sec - 30sec.  Then I focused on three HS educators that had slow query times and then added `measure` blocks around different parts of the code and found that the loop through `low_assignments` calling `has_experience_team_commented?` was taking almost all the time.  So I restructured that to make flat queries for low grades and uncommented students and then join manually, which cut the time down a lot.

The overall latency was pretty variable - ranging from 6sec to 25sec for the same test case at different times.  But here's some example data of one direct comparison showing how long in ms it took to run test case for each of 3 HS educators:
```
before: [6672, 7298, 7117]
after: [641, 649, 823]
```

kevinrobinson:
selfie ahead of the meeting this morning, @alexsoble can you review later today?

PR #1514, opened 2018-03-09
Update copy for birthday card on current day

kevinrobinson:
# Who is this PR for?
educators

for students who's birthday is today, it says their birthday "was" on today.
screen shot 2018-03-09 at 8 23 45 am

This also updates `BirthdayCard` to use the `nowFn` in context, and updates the `withDefaultNow` test helper to return a new `now` each time to prevent accidental test pollution with moment's mutable methods (eg, `now.startOf('day')`).


kevinrobinson:
selfie

PR #1516, opened 2018-03-09
Fix insight copy to show all student count, not limit

kevinrobinson:
# Who is this PR for?
HS educators

insight box copy would show the limit of how many students were returned, not actual real number of students
screen shot 2018-03-09 at 9 01 16 am

# What does this PR do?
fixes that!

# Checklists


kevinrobinson:
selfie

PR #1517, opened 2018-03-09
Fix JS bug in loading homeroom page

kevinrobinson:
bug fix from https://github.com/studentinsights/studentinsights/pull/1505

# Who is this PR for?
K8 teachers primarily

Homeroom page doesn't load.  This doesn't produce any JS errors so monitoring didn't catch it.  It's hard to write automated tests for this, since it's a seam between the Rails HTML and the JS bootstrapping.  We could add production tests for this if we wanted.

# What does this PR do?
fixes a bug introduced in https://github.com/studentinsights/studentinsights/pull/1505 with what HTML the JS code expects to be present

before:
screen shot 2018-03-09 at 12 21 31 pm

after:
screen shot 2018-03-09 at 12 21 39 pm

# Checklists
+ [x] Author checked latest in IE - K8 homeroom

### DIFF
--- a/app/assets/javascripts/homeroom_table/main.jsx
+++ b/app/assets/javascripts/homeroom_table/main.jsx
@@ -31,6 +31,6 @@ export default function homeroomMain() {
       rows={serializedData.rows}
       school={serializedData.school}
     />,
-    document.getElementById('homeroom-table')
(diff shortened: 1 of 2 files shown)


kevinrobinson:
selfie

PR #1519, opened 2018-03-12
Load more home page cards

kevinrobinson:
# Who is this PR for?
educators, HS NGE/10GE to start

The home page only loads a fixed number of cards on page load; there's no way to browse back further.

# What does this PR do?
Adds a "see more" link to load more cards into the feed.

This PR also adds a `time_now` method to `TestPals`, consistent with what's in `NowContainer` in the JS specs.  This is to help make it easier to make correct timestamps, etc. across tests by fixing test time to a particular moment.

screen shot 2018-03-12 at 10 18 09 am

# Checklists


kevinrobinson:
selfie

Issue #1092, opened 2017-09-24
Add notes on Ruby linter within Sublime


kevinrobinson:
https://github.com/SublimeLinter/SublimeLinter-rubocop maybe?

So we can catch whitespace errors in the IDE, or fix them up in a more automated way, rather than later in CI: https://travis-ci.org/studentinsights/studentinsights/builds/279269714#L2405

alexsoble:
@kevinrobinson This is a good idea. Folks can also run `rubocop -a` to autofix most issues instead of having to do that manually. Now I'm thinking that shortcut should go into the README...

jhilde:
How about a git pre-commit hook that runs eslint and rubobop before committing. Some of us are getting old :older_man: and occasionally forget.

kevinrobinson:
FWIW, https://github.com/typicode/husky can do this and https://github.com/okonet/lint-staged optimizes only running those precommit tasks on staged files only.  If we added that with robocop and eslint, it would force developers to strictly run those tasks on each local commit or on each git push.

When there's an error, it shortens the cycle between when developers find out about it from Travis.  When there's not an error, it adds a few seconds of delay on every commit.  That might be slow and annoying (depending on how fast those tasks are), or it might be helpful for folks since then they don't have to think about running those tools.  Adding this on push might be a middle ground.

Either was could be cool, @alexsoble might have thoughts on what would be helpful for folks in general across the project.

kevinrobinson:
We can do this next time we're onboarding!

PR #1520, opened 2018-03-12
Add note to README about rubocop linter, remove node_modules from it

kevinrobinson:
addresses https://github.com/studentinsights/studentinsights/issues/1092

# Who is this PR for?
developers

don't run rubocop on `node_modules`, and add a note for adding Rubocop linting in Sublime


kevinrobinson:
selfie

PR #1521, opened 2018-03-12
Add Ruby linter

kevinrobinson:
# Who is this PR for?
developers

`ruby -wc` has helpful checks like variables that are unused, so let's add these warnings into the IDE and the build

# What does this PR do?
Adds a `rake lint` and calls it in the Travis build


kevinrobinson:
This one seems like an actual migration bug: https://github.com/studentinsights/studentinsights/blob/master/db/migrate/20161222175109_add_summer_boolean_to_service_type.rb#L18

kevinrobinson:
This is what this looks like in Sublime, the most helpful bit for catching actual bugs being that you can see in the IDE when you have unused variables (with the dot in the gutter, just like with eslint):

screen shot 2018-03-26 at 11 32 21 am

And here's the status bar:
screen shot 2018-03-26 at 11 32 29 am

kevinrobinson:
Actually reading more Rubocop can do this better and faster (Lint/UselessAssignment), but it doesn't seem to be included in our config right now.

@alexsoble I see there's a rubocop.yml file, but that many of the items that are "enabled by default" according to the docs aren't caught: http://rubocop.readthedocs.io/en/latest/cops_lint/ because of `DisabledByDefault: true` at the top.  I figure this came about from you tuning this when you first installed it, want to chat more about this at some point?  I'd vote for putting in a bunch of these, especially the ones that are lint cops aimed at catching bugs as compared to style cops (where I don't have any preference).

kevinrobinson:
Closing this.  @alexsoble let's chat more in person or over https://github.com/studentinsights/studentinsights/pull/1521?

PR #1524, opened 2018-03-15
Fix bug with last SST on tardies dashboard, add `nullableWithKey` propType

kevinrobinson:
# Who is this PR for?
K8 principals and asst. principals

Tardies dashboard last SST column is missing data

# What does this PR do?
Fixes that, adds propTypes to the `StudentsTable` component in the dashboards to catch this.

After fix:
screen shot 2018-03-15 at 8 15 18 am

`nullableWithKey` propType failure messages:
screen shot 2018-03-15 at 8 26 51 am
screen shot 2018-03-15 at 8 27 05 am

# Checklists
## Javascript QA
+ [x] Author checked latest in IE - Absence dashboard
+ [x] Author checked latest in IE - Tardies dashboard

### DIFF
--- a/app/assets/javascripts/helpers/prop_types.jsx
+++ b/app/assets/javascripts/helpers/prop_types.jsx
@@ -32,7 +32,21 @@ const PropTypes = {
   history: React.PropTypes.shape({
     pushState: React.PropTypes.func.isRequired,
     replaceState: React.PropTypes.func.isRequired
-  })
+  }),
+
+  // Return a propType for us in `PropTypes.shape`
+  // that requires a key to be passed, but allows
+  // null values (which `.isRequired` doesn't allow).
+  nullableWithKey(propType) {
+    return (props, propName, componentName) => {
+      if (!props.hasOwnProperty(propName)) {
+        return new Error(
+          'Missing prop `' + propName + '` in ' +
+          ' `' + componentName + '`. Nulls are allowed, but must be passed explicitly.'
+        );
+      }
+    };
+  }
 };
 
 export default PropTypes;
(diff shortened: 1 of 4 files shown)


kevinrobinson:
I guess based on https://github.com/studentinsights/studentinsights/pull/1493 I only did this for absences before, but thought it was for both (and the change to `StudentsTable` impacted both).  This fixes up both.

FYI @edavidsonsawyer

kevinrobinson:
selfie

edavidsonsawyer:
Hey @kevinrobinson quick question. What is the benefit to doing this vs not requiring the sst prop?

kevinrobinson:
@edavidsonsawyer yeah good question, I should have explained more!

This requires that the caller passed in the prop, even if the value is null or undefined.  So it's trying to be stricter to say "this prop is required, so callers better express that they know that by passing it, but null values are allowed and make sense."  If you just leave off `.isRequired`, then you can leave out the key altogether (which in this case would be a bug).  If you add `.isRequired`, then you can't pass `last_sst_date: null` like we would like to allow here.

edavidsonsawyer:
@kevinrobinson thanks for that! In other places I've been much less strict about requiring props that can have valid null values. In the future, would you prefer we use this construct in places where the component is going to be interpreting something as a prop with a (valid) null value, as opposed to an optional prop?

kevinrobinson:
@edavidsonsawyer If you think that makes sense, or helps communicate your intent as you're writing, sure!  I've made this kind of mistake a bunch of times, so have found this kind of check helpful when it's about fields in data objects that you think should always be there, even if their value is null.  But in some places props are truly optional (like a `className` or a `limit` to how many things are shown, or a `defaultSortKey`).

PR #1523, opened 2018-03-15
Add links to navbar for absences and tardies

kevinrobinson:
# Who is this PR for?
K8 principals and asst. principals

# What problem does this PR fix?
Let folks see the absences and tardies dashboards on their own.

# What does this PR do?
Updates navbar links.

# Screenshot (if adding a client-side feature)
#### school leader
screen shot 2018-03-15 at 7 08 54 am

#### classroom teacher
screen shot 2018-03-15 at 7 11 12 am

#### districtwide (access for school through admin page)
screen shot 2018-03-15 at 7 12 23 am

PR #1522, opened 2018-03-12
Add educator page for admin to view data as a particular educator

kevinrobinson:
# Who is this PR for?
admin, developers

# What problem does this PR fix?
With features that are customized by particular user roles, it's difficult to get feedback or beta test these without working across a large group of people.  This allows just folks with districtwide admin access to see a new `educators/view/:id` page that shows what that educator has access to or can see.  This PR starts with doing that for the new home page.

# What does this PR do?
It adds in a new `/educators/view/:id` page.  `HomePage` and related components are refactored to require an explicit `educator_id`, and the controller code is updated to respect that only for folks with districtwide access.  This can enable better development, debugging and feedback for features that branch on educator type or data that they have access to.

This also lifts up recovery from authorization failures to `application_controller` so that it now responds correctly for both HTML requests (with a redirect) and JSON requests (still returning JSON).  Relatedly, it factors out UI code for making JSON requests into `apiFetchJson` and updates code to use that.

# Screenshot (if adding a client-side feature)

screen shot 2018-03-12 at 4 13 20 pm

screen shot 2018-03-12 at 4 13 29 pm

# Checklists
+ [x] Author checked latest in IE - Home page
+ [x] Author checked latest in IE - Homeroom
+ [x] Author checked latest in IE - Sections
+ [x] Author checked latest in IE - Section
+ [x] Author checked latest in IE - Student Profile
+ [x] Author checked latest in IE - Student Report PDF
+ [x] Author checked latest in IE - School Overview
+ [x] Author checked latest in IE - Principal dashboards
+ [x] Author improved specs for code in need of better test coverage
+ [x] Author included specs for new code

PR #1525, opened 2018-03-15
Fix bug with navbar routes for absences dashboards

kevinrobinson:
# Who is this PR for?
K8 principals, asst. principals

Bug in https://github.com/studentinsights/studentinsights/pull/1523 with how navbar links are rendered on different pages across roles.

# What does this PR do?
Fixes the bug by explicitly passing the school to these links.

In https://github.com/studentinsights/studentinsights/compare/patch/navbar-across-roles, I also spiked on pulling out the navbar from Rails and rendering in JS so we could add it into https://github.com/studentinsights/studentinsights/pull/1522 and be able to see these easily in production across roles (right now we can't), but this is more involved that I can do this morning and we have other things to ship for tomorrow.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -39,10 +39,10 @@
               <%= link_to 'Roster', school_path(current_educator.school) %>
               &nbsp;
             <% end %>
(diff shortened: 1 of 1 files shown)


kevinrobinson:
selfie

PR #1526, opened 2018-03-15
Update NGE/10GE insight box to focus on students in teachers' courses

kevinrobinson:
# Who is this PR for?
HS classroom teachers

For HS classroom teachers, this often showed more information about students than they could directly take action on.  This included low grades for courses they didn't teach, or for courses that they taught older students.  This isn't solving any problems for dept. heads and other roles who do want to see wider views (but in a different way).

# What does this PR do?
This focuses it to 9/10th grade students in sections that they teach.  It also has some related copy changes, since it focuses on students rather than courses, and it changes numeric grades to letter grades (how teachers talk about this in conversation in these meetings).  For local data, it also adds in `course_description` and `house` fields to objects in `TestPals`.

before, showing messages about low grades in other teacher's classes:
screen shot 2018-03-15 at 4 39 20 pm

after, focused on just grades in the current teacher's courses:
screen shot 2018-03-15 at 4 34 53 pm

# Checklists
## Javascript QA


app/assets/javascripts/educator/EducatorPage.test.js
```diff
@@ -1,24 +1,21 @@
...
-import unsupportedLowGradesJson from '../../../../spec/javascripts/fixtures/home_unsupported_low_grades_json';
+
 
 function testProps() {
   return {
     educatorId: 101
   };
 }
 
+jest.mock('../home/HomePage');
```
kevinrobinson: Using this in place of Enzyme's `shallow` renderer since it lets all the code run like normal and just mocks this one dependency (which requires other test setup because of the HTTP requests, we don't need to test all that here).

app/assets/javascripts/home/HomePage.js
```diff
@@ -37,6 +37,7 @@ const styles = {
   },
   column: {
     flex: 1,
+    width: 0,
```
kevinrobinson: This fixes a related issue with making sure the two-column layout distributes extra space evenly and doesn't actually flex.

kevinrobinson:
selfie, if anyone can review after-the-fact, that'd be awesome!

PR #1528, opened 2018-03-15
Update filter for NGE and 10GE students

kevinrobinson:
# Who is this PR for?
HS classroom teachers

bug in https://github.com/studentinsights/studentinsights/pull/1526, which didn't properly filter down to 9th and 10th grade students

# What does this PR do?
fixes it!

# Checklists

This came up because we don't have enough students and courses generated in local development for this to be visible.  We should fix this but I don't have time to do it now ahead of tomorrow morning.


kevinrobinson:
selfie

PR #1529, opened 2018-03-15
Tune down initial fetch for home feed as short-term performance optimization

kevinrobinson:
# Who is this PR for?
HS classroom teachers

Short-term change to reduce latency loading home page feed

# What does this PR do?
Tunes down initial fetch from 20 (more than fits on screen) to 10.


kevinrobinson:
selfie

PR #1530, opened 2018-03-15
Fix bug with experience team filter in home page insight box

kevinrobinson:
Fixes bug in https://github.com/studentinsights/studentinsights/pull/1528

# Who is this PR for?
HS classroom teachers

Filtering by NGE/10GE students didn't work correctly in https://github.com/studentinsights/studentinsights/pull/1528, because we filtered by this on one part of the query but not the other part when joining. 

# What does this PR do?
fixes it!

Tested in production console for 2 sample teachers.


kevinrobinson:
selfie

PR #1531, opened 2018-03-16
Revise dashboard JS tests to explicitly pass through current time

kevinrobinson:
# Who is this PR for?
developers

Test failures depending on the timezone and time of day, eg: https://travis-ci.org/studentinsights/studentinsights/builds/354091918.  This was blocking https://github.com/studentinsights/studentinsights/pull/1530 and https://github.com/studentinsights/studentinsights/pull/1529, so just resolving the short-term issue now.

# What does this PR do?
Revises the absence/tardies dashboard specs to explicitly pass through all times.

# Checklists

There's two deeper underlying issues here, one is that we're parsing times into the incorrect timezone in `toMoment` (unused here) and in other places we're taking dates and converting them to timestamps in UTC days on the server.  The other is that I keep misinterpreting `moment`'s API even after using it for years (eg, functions mutate the date, `moment.utc(text).format()` tells it to tag formatting as in UTC rather than parse it as UTC).  https://date-fns.org/ might be our jam in a separate piece of work.

### DIFF
--- a/spec/javascripts/school_administrator_dashboard/DashboardHelpers.test.js
+++ b/spec/javascripts/school_administrator_dashboard/DashboardHelpers.test.js
@@ -1,11 +1,12 @@
 import DashboardHelpers from '../../../app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardHelpers.js';
 
-import * as Data from './DashboardTestData.js';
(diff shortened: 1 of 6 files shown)


kevinrobinson:
selfie

PR #1533, opened 2018-03-16
Temporary workaround for NB homeroom names in analysis pages

kevinrobinson:
# Who is this PR for?
NB K8 educators

Absence and tardy dashboards don't show homeroom breakdowns, since there aren't any homeroom/educator relationships being imported.

# What does this PR do?
I didn't look further yet, this just works around the immediate problem so the breakdown is meaningful.


kevinrobinson:
selfie

PR #1534, opened 2018-03-16
Show different disability filters by district

kevinrobinson:
# Who is this PR for?
NB K8 educators

New Bedford uses different values than Somerville to describe levels of disability in special education.  This makes the summary values in the overview page inaccurate, and the filter values are not all meaningful.

# What does this PR do?
Threads the server env value down to `window.shared.Env.districtKey` and then branches on it in the `SlicePanels` UI to show different filters for each district.

before, showing wrong filters:
screen shot 2018-03-15 at 9 56 16 pm

after, showing correct filters (but for demo data using Somerville disability values):
screen shot 2018-03-15 at 9 57 04 pm

# Checklists


kevinrobinson:
selfie

PR #1535, opened 2018-03-16
Add PerDistrict, render admin home page schools differently in New Bedford

kevinrobinson:
# Who is this PR for?
New Bedford districtwide admin

We're only importing data for pilot schools, but the districtwide admin pages shows all schools and there are a lot, with the two pilot schools buried way below the fold.

# What does this PR do?
Adds a `PerDistrict` class Rails code like this and tags `LoadDistrictConfig` as deprecated.  We'll need to distinguish between what the purpose of these two classes are, and my intent here is to tag `PerDistrict` as "the place to put Ruby code that branches on ENV['DISTRICT_KEY']" rather than "the class that reads yml config."  I think we can revise `LoadDistrictConfig` to just do that, and have `PerDistrict` be the Ruby-facing API for "I need to do something different based on the district."

None.  We don't have a good way to work with Rails code that works for features across districts right now, either in development mode or in RSpec tests, and we need to work on this.  I'll deploy and verify manually for now.

### DIFF
--- a/app/config_objects/load_district_config.rb
+++ b/app/config_objects/load_district_config.rb
@@ -1,6 +1,7 @@
 # This class is responsible for pulling YAML district config out of the
 # filesystem and parsing it into a Ruby hash.
 
+# Deprecated, see PerDistrict
 class LoadDistrictConfig
 
   def initialize(district_key = ENV['DISTRICT_KEY'])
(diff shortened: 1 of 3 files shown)


kevinrobinson:
selfie

PR #1538, opened 2018-03-16
Include discipline incidents in the home page feed

kevinrobinson:
# Who is this PR for?
educators

Show more of what we know about what's going on with students, improve zero-day for home page in new districts.

# What does this PR do?
Adds discipline incidents into the home page feed.
screen shot 2018-03-15 at 11 21 51 pm

Also factors out the formatting from `EventNoteCard` into `FeedCardFrame` for re-use in the incident card.  No regressions:

screen shot 2018-03-15 at 11 22 09 pm

# Checklists

### DIFF
--- a/app/assets/javascripts/components/Badge.js
+++ b/app/assets/javascripts/components/Badge.js
@@ -1,0 +1,28 @@
+import React from 'react';
+
+
+// A visual UI element for a badge indicating a particular 
+// type of data (eg, in the home page feed).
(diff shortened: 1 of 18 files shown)


kevinrobinson:
selfie; right now this is disabled everywhere, and can be enabled per-district with an env variable.  My intention is to switch it on just for New Bedford, and ship for Somerville and Demo next week.

PR #1539, opened 2018-03-16
Fix one more non-deterministic test from times

kevinrobinson:
follow-on to https://github.com/studentinsights/studentinsights/pull/1531

# Who is this PR for?
developers

builds that fail at certain times like https://travis-ci.org/studentinsights/studentinsights/builds/354154156

# What does this PR do?


kevinrobinson:
missed one and it's breaking the build; selfie

PR #1541, opened 2018-03-16
Add more values for lunch program

kevinrobinson:
# Who is this PR for?
developers

add values into demo data that aren't in there now, adds validation to model


kevinrobinson:
selfie

PR #1543, opened 2018-03-18
Add socks.conf to gitignore

kevinrobinson:
# Who is this PR for?
developers

This file is generated if you run qgsocksify locally, and writes out IP and port info that we wouldn't ever want to check in.

# What does this PR do?
gitignore it


kevinrobinson:
selfie

PR #1544, opened 2018-03-18
Remove RSpec code about mocking LDAP that doesn't do anything

kevinrobinson:
# Who is this PR for?
developers

These tests make it look like we're testing the LDAP authorization, but we're not doing that right now.  The tests run without setting SHOULD_USE_LDAP, so in the educator model, we're using `database_authenticatable`.  The code that's mocking different Devise LDAP objects is having no effect on the code being executed in the tests.  You can swap the reject/authorize calls, or remove them altogether as this PR does and the tests still pass.

I tried running these specs with `SHOULD_USE_LDAP=true` and they fail, but the message seems related to the test setup not the LDAP calls:
screen shot 2018-03-18 at 9 53 16 am

I'm not sure how to test this, but if we did I think it'd involve adding a separate RSpec run to the build that runs with `SHOULD_USE_LDAP=true`, and then we could fix forward from this bug above.  But for now, this PR removes these tests since they suggest we have coverage that we don't.


alexsoble:
Great find @kevinrobinson! I did some local investigating and verified everything you said. 😄  Should we ship this first since #1545 touches some of the same spec code?

kevinrobinson:
@alexsoble 👍

PR #1545, opened 2018-03-18
Add ldap_authenticatable_tiny to replace devise_ldap_authenticatable and allow SSL config

kevinrobinson:
# Who is this PR for?
security, developers

Upgrading net-ldap, and passing in particular encryption options to improve security.  We couldn't do this directly with devise_ldap_authenticatable, so I submitted a PR 246 over there (no link intentionally).

# What does this PR do?
Remove devise_ldap_authenticatable altogether and make our own ldap_authenticatable_tiny that's <100 lines.  It only does the LDAP bind exactly how we need, and lets us pass whatever config we need to `Net::LDAP`.

I tested this a few ways:
1. To test the Devise/Warden integration I ran it locally with `SHOULD_USE_LDAP=true`, commenting out the call to the LDAP server.
2. I ran the LDAP code in production on a one-off dyno using the updated version of the gem and running through the proxy.  This worked as expected.

There's no automated tests for this (see https://github.com/studentinsights/studentinsights/pull/1544), as I couldn't figure out a way to do this that would actually increase confidence in it working correctly beyond the two steps above.

This requires some additional ENV values to ship.  And this PR has some risk to ship, so ideally we'd do it today while it's not a school day.  But I'm going to let this sit for a bit and either come back later today or get a better review and talk more next week.  @alexsoble I wanted to make sure this path was workable, since this change is interrelated to some of the other changes we're working on for LDAP and user authentication with New Bedford.  We'll definitely want to manually test this quite a bit as soon as we ship.


kevinrobinson:
ha!  just ran into this and realized it's basically this PR: https://github.com/plataformatec/devise/wiki/How-To:-Authenticate-via-LDAP

alexsoble:
@kevinrobinson Oooh great find!

app/lib/ldap_authenticatable_tiny/strategy.rb
```diff
@@ -0,0 +1,73 @@
...
+          fail(:error)
+        end
+        nil
+      end
+
+      private
+      # Create a Net::LDAP instance, `bind` to it and close.
+      # Return true or false if they're authorized.
+      def is_authorized_by_ldap?(login, password)
+        options = create_ldap_options_for(login, password)
```
alexsoble: `ldap_options_for`?

kevinrobinson:
@alexsoble This has some specs now!  Thanks for talking through this together 👍

alexsoble:
🚢ing with @kevinrobinson!

PR #1547, opened 2018-03-19
Improve import logging and PerDistrict branching for New Bedford

kevinrobinson:
# Who is this PR for?
New Bedford educators, developers

There are a few bits in the import process that aren't working correctly.

# What does this PR do?
This adds some more logging to see what's going on, and also addresses two issues.  One is with the validation on `DisciplineInciden#incident_code`, which doesn't apply for New Bedford, so the PR removes it.  The other is moving the mapping of login_name to email to `PerDistrict`.

# Checklists
+ [ ] Author improved specs for code in need of better test coverage


kevinrobinson:
selfie

PR #1549, opened 2018-03-19
Update school dashboard authorization to use Authorizer method

kevinrobinson:
# Who is this PR for?
educators with districtwide or schoolwide access that aren't flagged as admin in Aspen

They can't access the absences and tardies dashboards, but should be able to.

# What does this PR do?
Updates the absences and tardies dashboard controller code to use the `Authorizer#is_authorized_for_school?` method.


kevinrobinson:
selfie, cc @edavidsonsawyer just FYI, this came up for someone today so finishing off that todo :)

PR #1550, opened 2018-03-19
Allow demo district_key

kevinrobinson:
# Who is this PR for?
people using the demo site

the demo site is down

# What does this PR do?
Allows 'demo' as a district_key value; this was made stricter in https://github.com/studentinsights/studentinsights/pull/1547/files#diff-9ba06106d1f1d97c7f2d9ece90d5b81dR11 this fixes it up for the demo site.

It also removes a stray log statement from https://github.com/studentinsights/studentinsights/pull/1549.


kevinrobinson:
selfie

PR #1551, opened 2018-03-20
Move student profile files to ES6 modules

gabivoicu:
# Who is this PR for?
Developers

Makes our JS code more consistent

# What does this PR do?
Move `AcademicSummary` and `SummaryWithoutSparkline` to ES6 modules.

# Checklists

## Javascript QA

+ [x] Reviewer checked latest in IE - Student Profile


gabivoicu:
@alexsoble @kevinrobinson This is my first WIP draft of the ES6 modules. I have a bunch of tests failing that I can't crack, can you take a look when you get a second and tell me if I'm missing something major?

Also, I am starting to wonder if we will need to convert `PageContainer` as part of this PR to get the tests to work 🤔

app/assets/javascripts/student_profile/AcademicSummary.js
```diff
@@ -0,0 +1,548 @@
...
+            </tr>
+          </tbody>
+        </table>
+      </div>
+    );
+  }
+
+  render() {
+    return (
+      <div className="AcademicSummary">
```
kevinrobinson: This is an existing bug, but should be `SummaryWithoutSparkline`, can you rename while you're in here?

app/assets/javascripts/student_profile/AcademicSummary.js
```diff
@@ -0,0 +1,548 @@
...
+    );
+  }
+}
+
+SummaryWithoutSparkline.propTypes = {
+  caption: React.PropTypes.string.isRequired,
+  value: React.PropTypes.string // or null
+};
+
+export {
```
kevinrobinson: Nice!

This is unrelated, but rather than exporting both, let's split these out into two separate files while we're at it.  I'm not sure why they were one file before, they seem pretty separate components.  Sometimes if there's a tiny "private" component that's good to define in the same file, but not export.  Or sometimes if you have two components that are really linked together you might define them in the same file and export both.  But these two seem unrelated, except for sharing a few minor styles like padding (which are fine to define locally in each file).

kevinrobinson:
@gabivoicu Awesome! 👍 I left two small changes I noticed here, which are slightly beyond literally translating to ES6 but in the spirit of "make this code written in component files like others," which would be great if you can do while you're in here too.

From your [comment](https://github.com/studentinsights/studentinsights/pull/1551#issuecomment-374435123) it sounded like you got things figured out, and that you can chip away at the different components one-at-a-time, which is great.  You can also merge these in small separate PRs if you like too (like, merge this one, then come back to another file) rather than a big batch.  Please update the PR description with whatever you choose to do.

Also, the test failure here is unrelated - I thought I had fixed this up, but you can ignore for now.  It's related to the way times are set up in code, with the tests failing at particular times of the day, depending on what timezone you're in.  :)  Feel free to dig in and fix it, and if not I'll take a look later this week.

app/assets/javascripts/student_profile/AcademicSummary.js
gabivoicu: FYI, I changed this in the other file, GitHub is just confused.

gabivoicu:
@kevinrobinson This is ready for a review. Do we have a convention for import order? I changed the import order in one file (mostly to get add another commit because things were weird on GitHub).

I used the Python convention: separate third party imports from our own code and alphabetize it all. Thoughts?

kevinrobinson:
@gabivoicu Awesome, I'll check this out tonight or tomorrow morning!  For finding where components are used in the UI, you can grep for the component name, but these are all within the student profile page.  

re: imports, that sounds good!  I've been putting third-party first then others in whatever order.  Have you found ordering mistakes lead to bugs, or is it just a way to be organized when there are lots of dependencies?

gabivoicu:
@kevinrobinson I've yet to find bugs from import order, I think what could happen would be duplicate imports of the same class/function, but the linter would warn us about it. It's more for code organization and ease of finding where a module is imported from if we know they're ordered alphabetically.

kevinrobinson:
@gabivoicu sounds good!  I guess another benefit to alphabetic ordering (after everything has been ordered), is that git diffs are simpler.  For this one it's hard to see what actually changed, but if everything were always alphabetical that might be simpler

screen shot 2018-03-23 at 11 44 27 am

PR #1553, opened 2018-03-20
Parsing all dashboard times as utc

edavidsonsawyer:
# Who is this PR for?
Really it's for developers

Timestamps are parsed differently in different parts of the dashboard, making it difficult to correctly test behavior

# What does this PR do?
Parses all times in UTC for consistency

# Checklists

## Javascript QA

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardHelpers.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardHelpers.js
@@ -43,7 +43,7 @@ export default {
   },
 
   schoolYearStart: function() {
-    const today = moment();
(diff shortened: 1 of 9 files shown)


edavidsonsawyer:
Hey @kevinrobinson, what do you think of this? It still seems more "correct" to have the dashboard show information in the user's local time, but keeping track of which things should and shouldn't be parsed as utc is proving to be a headache and probably isn't worth the upkeep cost.

@gabivoicu this should also prevent the build errors we were running into last night.

kevinrobinson:
@edavidsonsawyer Yeah, let's fix the flaky build first (this seems good!) and then talk Friday about the underlying issue, since timezones are confusing :)

PR #1554, opened 2018-03-20
Add teacher pageview column to Research Matters export

alexsoble:
# Who is this PR for?

Research Matters.

Adds missing column with # of pageviews in Mixpanel per teacher.

### DIFF
--- a/app/reports/raw_mixpanel_data_downloader.rb
+++ b/app/reports/raw_mixpanel_data_downloader.rb
@@ -21,18 +21,15 @@ class RawMixpanelDataDownloader
 
     raw_output = `#{cmd}`
 
-    parsed_output = raw_output.split("\n").map do |json_line|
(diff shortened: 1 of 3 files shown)


alexsoble:
This PR should be good to go! I added a test case describing the `pageview_count` column in the teacher file, and the specs are passing. ✅ 

The only issue: When I try to run it locally, I either get an error message from Mixpanel ("error completing export request"), or it just hangs. I don't see anything that would explain this on the Mixpanel Status page (https://status.mixpanel.com/, "Data Export API: Operational").

Might just ship this PR and see if the Mixpanel query works in prod. Or wait a couple of hours, try again, and see if the Data Export API is more responsive.

alexsoble:
The Mixpanel Data Export API calls are working now; no more error messages. Added more console output during the download phase to give developers more feedback, so they can see the difference between "downloading data..." and "hanging there..."

PR #1556, opened 2018-03-21
Filter out event note text from logs

alexsoble:
# Who is this PR for?

Developers who use the logs and get email alerts from our logging service. 

Event note text is currently printed in the logs as plaintext. This isn't the best security practice because it makes the surface larger for accessing sensitive event note data. We want to shrink that surface and make it as small as possible. If the logging service sends an email alert that includes a log trace of an event note being created/updated, without filtering in place we'll have copies of sensitive event note text in developers' inboxes.

# What does this PR do?

Filters out any field named "text" from the logs.

## Event note creation log (demo data, master branch)
screen shot 2018-03-21 at 10 29 43 am

## Event note creation log (demo data, this branch)
screen shot 2018-03-21 at 10 31 00 am

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -1,4 +1,4 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
+Rails.application.config.filter_parameters += [:password, :text]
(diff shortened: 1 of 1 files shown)


alexsoble:
@kevinrobinson Good idea, just added a code comment. 👍

alexsoble:
@kevinrobinson Don't see anything about indicating particular models using this config. 😞 

Here are what the Rails guides have to say about it: http://guides.rubyonrails.org/security.html#logging

PR #1559, opened 2018-03-21
Update New Bedford domain suffix to be correct

kevinrobinson:
# Who is this PR for?
NB educators, JV in particular.

Folks can't sign in.

# What does this PR do?
Updates the domain suffix for New Bedford to actually be correct :)  This will require a manual migration too.


kevinrobinson:
Ran the migration: `Educator.all.each {|e| e.email= "#{e.email.split('@')[0]}@newbedfordschools.org"; e.save! };nil`

kevinrobinson:
selfie

PR #1560, opened 2018-03-22
Fix note rendering on home page for multi-line notes

kevinrobinson:
# Who is this PR for?
all educators

The text of event notes were being rendered without respecting whitespace like newlines.  This is especially problematic for longer notes like from MTSS meetings.
# What does this PR do?
Factors out the rendering of static and editable notes from `note_card` in the student profile into `NoteText` and `EditableNoteText` components that can be used elsewhere.  Uses `NoteText` in the home feed's `EventNoteCard`.

before:
screen shot 2018-03-22 at 12 52 33 pm

after:
screen shot 2018-03-22 at 12 53 14 pm

no regressions:
screen shot 2018-03-22 at 12 36 11 pm

# Checklists


kevinrobinson:
selfie

PR #1565, opened 2018-03-26
Add in timing measurements for absences dashboard

kevinrobinson:
# Who is this PR for?
developers, @edavidsonsawyer in particular

We don't have a way to get measurements on client-side performance in IE right now

# What does this PR do?
Adds a `measurePageLoad` function for grabbing page load timing data, and adds `MountTimer` which measures the time to mount a React component tree.

# Checklists
+ [x] Author checked latest in IE - Absences dashboard as smoke test

### DIFF
--- a/app/assets/javascripts/components/MountTimer.js
+++ b/app/assets/javascripts/components/MountTimer.js
@@ -1,0 +1,40 @@
+import React from 'react';
+
+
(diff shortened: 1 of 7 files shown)


kevinrobinson:
@edavidsonsawyer I'm going to ship this so we can get some data for production, but if you can review after-the-fact that'd be awesome!

PR #1566, opened 2018-03-26
scripts/deploy/all

kevinrobinson:
# Who is this PR for?
developers

Adds a new `script/deploy/all.sh` that uses `concurrently` to run all deploys concurrently and color-code their output.  This is a small convenience for deploying across all sites without running each in parallel windows.

screen shot 2018-03-26 at 11 06 29 am

### DIFF
--- a/scripts/deploy/all.sh
+++ b/scripts/deploy/all.sh
@@ -4,7 +4,6 @@ git fetch origin
 
 echo "🚢  Deploying...";
 yarn run concurrently \
-  --kill-others \
   --names "demo,somerville,new-bedford" \
   -c "yellow.bold,blue.bold,magenta.bold" \
   'scripts/deploy/deploy.sh demo' \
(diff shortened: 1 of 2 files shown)


scripts/deploy/all.sh
```diff
@@ -0,0 +1,13 @@
+# Deploy to all sites in parallel
+echo "🚢  Fetching from GitHub...";
+git fetch origin
+
+echo "🚢  Deploying...";
+yarn run concurrently \
```
alexsoble: I didn't realize that `yarn run` could take shell scripts as arguments!

alexsoble: This is super cool and useful, I'm surprised that yarn doesn't seem to document it very clearly: https://yarnpkg.com/lang/en/docs/cli/run/

kevinrobinson: Yeah!  What's happening here is that `yarn run [script]` is running `concurrently` within the package.json env (like `bundle exec` in Ruby).  And that package, concurrently, happens to take a set of command line arguments that it should run concurrently.  This is how the `yarn start` command works too, to run the dev server and Rails concurrently.

alexsoble: Ooooooooooooh 👍

PR #1567, opened 2018-03-26
Refactor/quad converter es6

kevinrobinson:
# Who is this PR for?
developers

simplifying the way JS is written so there's less to learn for new folks

# What does this PR do?
Refactor `QuadConverter` and `ProvidedByEducatorDropdown`.  ProvidedByEducatorDropdown also has an implicit external dependency on jQuery autocomplete, so I added that into a note in https://github.com/studentinsights/studentinsights/issues/1180#issuecomment-374814129 to do later.

# Checklists


kevinrobinson:
selfie

app/assets/javascripts/student_profile/QuadConverter.js
```diff
@@ -0,0 +1,103 @@
...
+  return [toMoment(quad).valueOf(), toValue(quad)];
+}
+
+// The "Star Object" adds an additional data point (4-indexed) to a quad.
+// This allows the rendering of gradeLevelEquivalent data in highcharts tooltip
+export function toStarObject(quad){
+  return {
+    x: toMoment(quad).valueOf(),
+    y: toValue(quad),
+    gradeLevelEquivalent: toGradeLevelEquivalent(quad)
```
alexsoble: @kevinrobinson I'm learning something here . . . so when we have an JS file that's a list of `export`'ed functions, the functions can reference each other — like `toStarObject` calling `toGradeLevelEquivalent` — and they don't have to be in any particular order, is that correct?

kevinrobinson: Yeah, the way I think about it is these are just plain `function` like normal JS.  And so code can call functions just like normal.  In JS function declarations are all "hoisted" to the top of the function scope so that you can write them in any order you want.  This is different than if you define anonymous functions and assign them to variables like this: `const foo = function() { return 'wat'; }`

And then separately, you can tag `export` in front of functions to make them publicly available to other code that imports the module.

alexsoble: Oh that's super helpful, thanks for explaining @kevinrobinson!

PR #1572, opened 2018-03-27
Patch section authorization bug for HS teachers

kevinrobinson:
# Who is this PR for?
HS educators
follow-on to https://github.com/studentinsights/studentinsights/pull/1570


kevinrobinson:
with @alexsoble

PR #1573, opened 2018-03-27
Make login case insensitive 

alexsoble:
# Who is this PR for?

Educators who log into Insights and don't want to have to worry about if their email is typed uppercase or lowercase.

# What does this PR do?

Make the login code check for emails case-insensitively when comparing user-entered emails with emails in our database. 

We still pass the original, non-downcased email over to the LDAP server for authentication.


alexsoble:
specs pass locally! ✅

PR #1574, opened 2018-03-27
Widen scope of Research Matters exporter to all elementary schools

alexsoble:
# Notes 

Include all seven elementary schools in the query.

# Who is this PR for?

Research Matters.


alexsoble:
selfie 🤖

PR #1575, opened 2018-03-28
Separate dashboard routes

edavidsonsawyer:
# Who is this PR for?
Dashboard users

Dashboards can have long load times without any visible ui

# What does this PR do?
Renders a UI immediately so users can see the app while the data loads.
Only requests the data necessary to display the current dashboard to reduce load time.

## Javascript QA

+ [x] Author checked latest in IE - Absence Dashboard (from /school_administrator_dashboard)
+ [x] Author checked latest in IE - Absence Dashboard (from /absences)
+ [x] Author checked latest in IE - Tardies Dashboard (from /school_administrator_dashboard)
+ [x] Author checked latest in IE - Tardies Dashboard (from /tardies)
+ [x] Reviewer checked latest in IE - Absence Dashboard (from /school_administrator_dashboard)
+ [x] Reviewer checked latest in IE - Absence Dashboard (from /absences)
+ [x] Reviewer checked latest in IE - Tardies Dashboard (from /school_administrator_dashboard)
+ [x] Reviewer checked latest in IE - Tardies Dashboard (from /tardies)

## Automated Testing


app/assets/javascripts/school_administrator_dashboard/dashboard_components/tardies_dashboard/TardiesDataLoader.js
```diff
@@ -0,0 +1,48 @@
...
+    };
+  }
+
+  componentWillMount() {
+    //if the parent component already has the dashboard data, we don't need to duplicate it
+    if(this.props.dashboardStudents.length) return;
+    const url = `/schools/${this.props.schoolId}/tardies_dashboard_data`;
+    this.setState({isLoading: true});
+
+    apiFetchJson(url)
```
kevinrobinson: There's a `<GenericLoader>` component if you want ([example](https://github.com/studentinsights/studentinsights/blob/ccba754239820bac5371275b25f56be0ed0537b7/app/assets/javascripts/school_courses/SchoolCoursesPage.js#L31)), not sure if it's actually useful or an improvement over the explicit code here, probably the only thing you get is a generic error message (not handled here) and a default loading message.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/absences_dashboard/SchoolwideAbsences.js
```diff
@@ -3,24 +3,48 @@ import PropTypes from 'prop-types';
...
   }
 }
 
 SchoolwideAbsences.propTypes = {
-  dashboardStudents: PropTypes.array.isRequired
+  dashboardStudents: PropTypes.array.isRequired,
+  schoolId: PropTypes.number
 };
 
+measurePageLoad(info => console.log(JSON.stringify(info, null, 2))); // eslint-disable-line no-console
```
kevinrobinson: This code looks like it will run when this file is imported, not when this component is rendered.  Should this be moved to `App#componentDidMount` to run on startup for now?  Same for the other usage.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/absences_dashboard/SchoolwideAbsences.js
```diff
@@ -3,24 +3,48 @@ import PropTypes from 'prop-types';
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      dashboardStudents: this.props.dashboardStudents,
+      isLoading: false
+    };
+  }
+
+  componentWillMount() {
+    //if the parent component already has the dashboard data, we don't need to duplicate it
```
kevinrobinson: If I follow - this is because the existing `SchoolAdministratorDashboards` is calling this component (with the data read from `serializedData`), and now we're trying to call it from `App` but this component needs to get data on its own now.

You might consider either factoring this into a loader component that then calls `SchoolwideAbsences`, or moving the `DashboardHelpers.absenceEventsByDay` work into `SchoolwideAttendance`.  But also we'll remove the original caller shortly, so no use in overthinking this and just head towards whatever direction seems best :)

config/routes.rb
```diff
@@ -57,12 +57,16 @@
...
+  get '/schools/:school_id/absences' => 'ui#ui'
+  get '/schools/:school_id/tardies' => 'ui#ui'
   resources :schools, only: [:show] do
     member do
       get :overview
       get :school_administrator_dashboard
       get :overview_json
       get :csv
+      get :absence_dashboard_data
+      get :tardies_dashboard_data
```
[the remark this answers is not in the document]
edavidsonsawyer: Yeah, I like that! tbh I've always been a little unsure about when it's better to list these explicitly vs having them as a member on an existing route. The way you've done it with the courses feels right for a data endpoint.

ui/App.js
```diff
@@ -64,7 +71,15 @@ class App extends React.Component {
     return <EducatorPage educatorId={educatorId} />;
   }
 
-  // Ignore this, since we're hybrid client/server and perhaps the 
+  renderAbsenceDashboard() {
+    return <SchoolWideAbsences dashboardStudents={[]} schoolId={this.props.currentEducator.school_id} />;
```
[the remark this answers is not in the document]
edavidsonsawyer: Ah, I forgot about districtwide access here!

app/assets/javascripts/school_administrator_dashboard/dashboard_components/tardies_dashboard/TardiesDataLoader.js
```diff
@@ -0,0 +1,48 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      dashboardStudents: this.props.dashboardStudents,
+      isLoading: false
+    };
+  }
+
+  componentWillMount() {
+    //if the parent component already has the dashboard data, we don't need to duplicate it
```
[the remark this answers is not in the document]
edavidsonsawyer: Yeah, this doesn't need to be here for the tardies component, it was only an issue with the absence piece, and we can remove that as well once we're only calling these in one way.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/tardies_dashboard/TardiesDataLoader.js
edavidsonsawyer: Neat, I didn't see that!

app/assets/javascripts/school_administrator_dashboard/dashboard_components/absences_dashboard/SchoolwideAbsences.js
edavidsonsawyer: I might end up using a less specific loader component for both pieces here or at least keeping the structure for the tardies loader and not doing more than one thing in this piece.

app/controllers/schools_controller.rb
```diff
@@ -143,6 +164,7 @@ def authorize_for_school
 
   def set_school
     @school = School.find_by_slug(params[:id]) || School.find_by_id(params[:id])
+    puts @school
```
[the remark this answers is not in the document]
edavidsonsawyer: The shame...

kevinrobinson:
@edavidsonsawyer ah sorry, I bet that's from the node version.  If I download fresh from your repo and `yarn install && yarn test` I see the same thing when running:

```
$ node -v
v6.11.3
```

This is on a different computer than usual, and if I install the latest long-term support version, 8.11.0, things work great.

```
$ node -v
v8.11.0
```

We could add the node engine version to the `package.json` to communicate this for the project and catch things like this in the future, but for now you can also just install the updated version which only takes a minute or two: https://nodejs.org/en/

kevinrobinson:
Hey @edavidsonsawyer!  From the description and conversation here, my read of this PR is that it's still WIP, which is fine.  Let me know if you need anything else here or just ping when this feels ready to ship.

edavidsonsawyer:
@kevinrobinson I think this one should now be ready to go since we're shipping dark and using this to get some information on performance in production. Let me know if there's anything about the current PR you've got concerns with!

kevinrobinson:
@edavidsonsawyer awesome, I'll take a final look Thursday and then let's ship this!  @alexsoble if you are around and have time to 🚢 before then that's great too. 👍

kevinrobinson:
@edavidsonsawyer I apologize, I didn't get to this this morning and am out tomorrow for a family wedding.  You should definitely keep working on top of this on what's next!

I'll ship this next week or @alexsoble if you have time and are able to ship that's great too, no worries either way.

alexsoble:
@edavidsonsawyer @kevinrobinson Gave both sets of endpoints a manual QA in Internet Explorer and gave the code another read-through . . . you two had awesome discussions here and I learned a whole bunch!  👍 

Shipping now, will deploy and share out the page load timing measurements from production + Internet Explorer.

alexsoble:
@edavidsonsawyer @kevinrobinson Tested out the new routes in Internet Explorer and Chrome. 

The first thing I noticed is that the timing logging in the console doesn't reflect how long it takes to see data in `/schools/:id/absences` and `/schools/:id/tardies` for the user. For those pages, the MountTimer appears almost immediately in the console, while the user sees "Loading..." for a while afterwards, which isn't accounted for in the timing log. 

I used the Google Chrome Stopwatch to time how long I saw "Loading..." for the Healey school absences page using the new JSON endpoint, it came out to between 10 seconds and 12 seconds over a couple of tries. 

So, with that caveat, here are the timing logs I saw for Chrome: 

## Chrome: schools/hea/school_administrator_dashboard

screen shot 2018-04-05 at 3 22 24 pm

## Chrome: schools/hea/absences (again, does not include time spent "Loading...")

screen shot 2018-04-05 at 3 22 28 pm

Here's what I saw for Internet Explorer, using my Virtual Machine:

## IE: schools/hea/school_administrator_dashboard
screen shot 2018-04-05 at 3 27 21 pm

## IE: schools/hea/absences 

For some reason, I couldn't see any output here in the Developer tools, even after trying a couple of times. I used a stopwatch to time the amount of time between requesting the page and seeing data and got 17 seconds. 

## Subjectively...

The old/existing dashboard pages felt faster to me than the new ones. This could be because looking at the word "Loading..." for 10 seconds made me much more aware that I was waiting for data to load.

alexsoble:
Woah! It says @studentinsights-bot was the author of that last comment but I'm guessing the real author was @kevinrobinson or @edavidsonsawyer... unless the @studentinsights-bot AI is seriously next level...

For what it's worth I think it would be good to profile and optimize the Ruby code first. Watching the JSON-powered, separate-endpoint pages hang on "Loading..." for ~10 seconds gave me the sense that we could gain a lot from speeding up the server-side queries. I also think improving the Ruby code is easier to profile and measure since we don't have to worry about the effect of things like Internet Explorer dev tools.

PR #1577, opened 2018-03-28
Fixup last login to be more readable

kevinrobinson:
# Who is this PR for?
project team, school leads

# What problem does this PR fix?
readability in https://github.com/studentinsights/studentinsights/pull/1576

# What does this PR do?
screen shot 2018-03-28 at 7 45 53 am

PR #1578, opened 2018-03-28
Fix bug with sorting by nil school name

kevinrobinson:
# Who is this PR for?
NB project leads

bug in https://github.com/studentinsights/studentinsights/pull/1577, can't sort arrays that contain different types (eg, string vs. nil)


kevinrobinson:
selfie

PR #1579, opened 2018-03-29
Add node version to package json

alexsoble:
# Who is this PR for?

Developers. 

The JS specs will fail unless the dev environment is using NodeJS >= 8.50, which introduced an initial implementation of `perf_hooks`: https://github.com/gabrielschulhof/node/commit/38520245fe63baf58fc69aa493ed73c6b15479b9.

Performance hooks are called in `setupTests.js`, line 13:

```
import {performance} from 'perf_hooks';
```

# What does this PR do?

Lists the minimum required version of NodeJS in `package.json`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,6 @@ language: ruby
 rvm:
   - ruby-2.3.0
 
-env:
-  - TRAVIS_NODE_VERSION="8.11.0"
-
 sudo: required
 
 services:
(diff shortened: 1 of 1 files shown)


alexsoble:
@kevinrobinson Travis passes on this branch, which is great. The main question in my mind is if this will require any changes to our Heroku setup.

alexsoble:
@kevinrobinson Looks like Heroku will read the desired Node version from our `package.json` and install/use it:

https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version

PR #1580, opened 2018-03-30
Fix bug with student section assignment import

alexsoble:

Bug with assigning students to sections in the nightly import process. 

The current import process is linking students to sections based on just section names, but section names aren't unique across all schools and terms. 

# What does this PR do?

Fixes the import bug by linking students to sections based on section name, section school (determined by section => course => school), and section term. 

# Checklist 

+ [x] Run student section assignment process locally and visually check results at `/schools/:id/courses` end point

### DIFF
--- a/app/importers/rows/student_section_assignment_row.rb
+++ b/app/importers/rows/student_section_assignment_row.rb
@@ -38,7 +38,11 @@ class StudentSectionAssignmentRow < Struct.new(:row, :school_ids_dictionary)
   def section
     return unless row[:section_number]
 
-    Section.find_by(section_number: row[:section_number], course: course)
+    Section.find_by(
+      section_number: row[:section_number],
+      course: course,
+      term_local_id: row[:term_local_id]
+    )
   end
 
 end
--- a/app/importers/rows/student_section_assignment_row.rb
+++ b/app/importers/rows/student_section_assignment_row.rb
@@ -14,19 +14,27 @@ class StudentSectionAssignmentRow < Struct.new(:row, :school_ids_dictionary)
   end
 
   def build
-    if student and section
-      student_section_assignment = StudentSectionAssignment.find_or_initialize_by(student: student,
-                                                                                section: section)
-      return student_section_assignment
-    end
+    return if course.nil? || student.nil? || section.nil?
+
+    StudentSectionAssignment.find_or_initialize_by(student: student, section: section)
   end
 
   def student
-    return Student.find_by_local_id(row[:local_id]) if row[:local_id]
+    return unless row[:local_id]
+
+    Student.find_by_local_id(row[:local_id])
+  end
+
+  def course
+    return unless row[:course_number]
+
+    Course.find_by_course_number(row[:course_number])
   end
 
   def section
-    return Section.find_by_section_number(row[:section_number]) if row[:section_number]
+    return unless row[:section_number]
+
+    Section.find_by(section_number: row[:section_number], course: course)
   end
 
 end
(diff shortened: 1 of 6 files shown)


alexsoble:
Ran a local import job for students and student sections assignments and used the `/schools/:id/courses/` view to inspect each school. Saw dramatic improvement from the current state on master. There was one instance of a course with students from two schools, AFAS (Argenziano School) and PIC (Parent Information Center). This was for a 1st Grade level Art course at AFAS, which seems like it could plausibly have a PIC-assigned student, especially since PIC can be used as a "placeholder" school assignment for new students who are entering the system.

app/importers/rows/student_section_assignment_row.rb
```diff
@@ -14,19 +14,34 @@ def self.build(row)
   end
 
   def build
-    if student and section
-      student_section_assignment = StudentSectionAssignment.find_or_initialize_by(student: student,
-                                                                                section: section)
-      return student_section_assignment
-    end
+    return if course.nil? || student.nil? || section.nil?
```
kevinrobinson: Can we make the flow more explicit?  If I read it right, `build` executes eight database queries for each row of data, and this isn't super visible right now.  The logging also throws away the information about why theses rows can't be imported (eg, there's no student that matches, there's no course that matches, there's no section that matches).

alexsoble: Sure, I can make those changes so that the importer logs more information about why a row can't be imported.

kevinrobinson: 👍 I don't know what the baseline is here (whether we're importing 100% of the rows or 63%) so might be good to see that while we're doing this, but that might be creeping too so whatever you think is best :)

alexsoble:
@kevinrobinson: Responding to review comments above:

1 — Yep, that's the meaning. The local database is empty though, not newly seeded, when I run the import job.
2 — My mental model is that the first time the import job runs after this PR is merged, `StudentSectionAssignmentsImporter#delete_rows` will delete all bad data. 

As the job runs, all the valid assignment IDs will be pushed into the `@imported_assignments` array.

The invalid IDs will not be gathered into the array, because the row importer model looks for `    StudentSectionAssignment.find_or_initialize_by(student: student, section: section)`. That code will call `initialize` instead of `find` if the record is bad, because the bad row will have the wrong section. Then at the end of that job, all the bad rows not part of `@imported_assignments` get deleted.

app/importers/rows/student_section_assignment_row.rb
alexsoble: Hmmm, yes, perhaps this would be creep since our immediate goal is to fix the bad data. I'm also not too worried about 8 database queries from a performance point of view since this runs in the background.

kevinrobinson: 👍 Yeah I don't care about database queries here either, more to learn which query failed to see why the row is invalid.

kevinrobinson:
@alexsoble got it, sounds good!  thanks for talking through 👍

app/importers/rows/student_section_assignment_row.rb
alexsoble: @kevinrobinson Oooh, gotcha!

alexsoble:
@kevinrobinson No problem! I'll merge and deploy this either tonight or tomorrow evening.

alexsoble:
Did some more manual checking here before merging — triple-checked that this PR has no impact on the SHS sections/courses import. Looked good! 

🚢 ing now

kevinrobinson:
@alexsoble Are you going to run the import job manually now?  Just trying to coordinate so I don't ship https://github.com/studentinsights/studentinsights/pull/1586 and https://github.com/studentinsights/studentinsights/pull/1585 at the same time you're running the awesome update task.  I'm fine with either order, just probably good to do one set at a time :)

PR #1581, opened 2018-04-02
Support branching for different "insight boxes" on the home page

kevinrobinson:
# Who is this PR for?
K8 educators, HS counselors and SST staff

In the original iteration of the home page, we showed an "insights box" for NGE and 10GE teachers about students with low grades.  This doesn't make sense for all users, so this branches that part of the UI to show it only if they're labeled as being in the experience team.  The next step will be to add in other insights boxes for other roles (the first one being students with absences not being talked about in SST, aimed at K8 SST staff and classroom teachers).

# What does this PR do?
To do the branching, this PR adds a `EducatorLabel` database table, that tags educators with different labels.  After merging and running the migration, I'll add in the list of NGE and 10GE educators here.  This will be console-only for now, but we can add admin tools for this, or can surface this on the educator page as we go.  Upon shipping, the effect branching will be visible in the educator view page.

I'm not so sure about the factoring of how the data flows (eg, between `ui#ui` and `PerDistrict`) but I think this is fine for now and we don't need to solve any more general problem here.  This is mainly aiming at improving over just storing this list in an env variable.

This PR then factors out `CheckStudentWithLowGradesView` from `HomeInsights` and only shows it to experience team folks.

For Bill in TestPals, who's now tagged with this label:
screen shot 2018-04-02 at 10 49 08 am

For Uri in TestPals, who's not tagged:
screen shot 2018-04-02 at 10 49 20 am

# Checklists


kevinrobinson:
selfie

kevinrobinson:
Ran this in production:
```
emails = [...]
educators = Educator.where(email: emails)
educators.each do |educator|
  EducatorLabel.create!(educator: educator, label_key: 'shs_experience_team')
end
```

PR #1585, opened 2018-04-03
Add high absence "insights box" to home page, refactor educator labels

kevinrobinson:
# Who is this PR for?
SST educators primarily, classroom teachers secondarily.  Follow-on to https://github.com/studentinsights/studentinsights/pull/1581.

# What problem does this PR fix?
Shows SST educators students with potential gaps in services that might be appropriate for SST.

# What does this PR do?
This PR makes three main changes.  First it, updates  the way the server and UI communicate about which staff are in the experience team.  In https://github.com/studentinsights/studentinsights/pull/1581 we decided this on the server, but this PR changes that to send down the list of `educator_labels` and let's the UI decide based on that (in `PerDistrict#inExperienceTeam`).  This impacts the API for `App`, `HomePage` and `HomeInsights`, which are updated.

Second, it adds `CheckStudentsWithHighAbsences` to the UI and `InsightStudentsWithHighAbsences ` to the server, which are shown for all educators (see screenshot).  These are aimed at surfacing gaps in service, where students have a high number of absences but aren't yet being discussed in SST.  SST staff can use this to directly set the agenda for the meeting, or classroom teachers can ask the SST team to do this.  There's a performance test for this as well, here's production measurements on a 10% sample:
```
ALL_FOR_EDUCATOR, sample size: 97
  all_for_educator         	median: 489ms		p95: 1615ms
```

Third, it updates the "insights box" about low student grades to used 45 days as the time period, keeping the time period consistent at one quarter for both insights boxes.

Incidentally, it changes a few other things that were mostly driven by changes to tests:
1) It moves `HelpBubble` from the student profile page into `components` so it can be re-used in this feature.  In the process it revises the `teaser` to accept a React node instead of just a string.
2) Moves the home page endpoints all under `/api/` since they were split with some under that path and some not.
3) It adds a `GENERIC_LOADER_THROW_ON_REJECT_IN_TEST` flag in test that makes `GenericLoader` throw instead of logging rejections, which gives informative stack traces on test failures.
4) Updates revised tests to remove `SpecSugar` and `jest.mock` in favor of Enzyme's `shallow` and `mount` renderers, which allow checking that props to children components are passed correctly (this was revealed since initially the changes in API to components around `educatorLabels` didn't fail tests).
5) Updates matcher in `authorizer_spec.rb` to use `contain_exactly`, since depending on what tests are run, this might not match the exact order.

# Screenshot (if adding a client-side feature)
### For NGE/10GE, with some students:
[...]

PR #1586, opened 2018-04-03
Update navbar links by role, in preparation for cutting over to /home for homepage

kevinrobinson:
# Who is this PR for?
This is related to https://github.com/studentinsights/studentinsights/pull/1585, and is a shorter-term alternative to https://github.com/studentinsights/studentinsights/compare/refactor/navbar (which aimed to replace the server-side navbar HTML and CSS with UI code, but is a more involved change).

It updates the navbar links for different users based on their roles, in preparation for changing the "home page" which is currently different by role.  This essentially takes the current home page values and translates them to links in the navbar, so that users still have links to access the pages they've previously seen when they first sign in.

There's a bunch of complexity in the authorization rules, and the users in `TestPals` don't exactly match production.  There's no change here to actual authorization code, but the default home page is changed so that if the teacher has a homeroom at the HS they won't be shown the homeroom page.  In production data, no HS teachers have a homeroom set so this has no functional difference, but expresses in code that right now the Homeroom view is only for K8 teachers (and avoiding expanding the surface area of the product).

### Classroom teacher
screen shot 2018-04-03 at 4 56 11 pm

### HS teacher
screen shot 2018-04-03 at 4 55 49 pm

### Schoolwide
screen shot 2018-04-03 at 4 55 59 pm

### Districtwide
screen shot 2018-04-03 at 4 55 35 pm

### Adds in links to the admin educator page for simpler debugging
screen shot 2018-04-03 at 4 53 33 pm

# Checklists

I tested these manually locally across roles as well.

This script verified in production that the change to `homepage_type` has no impact:
```
# (pasting in homepage_type)
output = []
educators = Educator.all
educators.each do |educator|
  old_homepage_type = Authorizer.new(educator).homepage_type
  @educator = educator
  new_homepage_type = homepage_type
  output << {
    old_homepage_type: old_homepage_type,
    new_homepage_type: new_homepage_type,
    match: old_homepage_type == new_homepage_type
  }
end
pp output.select {|o| !o[:match] };nil
```

### DIFF
--- a/app/lib/authorizer.rb
+++ b/app/lib/authorizer.rb
@@ -130,20 +130,6 @@ class Authorizer
     true
   end
 
-  # There are five types of entry experiences, depending on levels
-  # of access.
-  def homepage_type
-    begin
-      return :districtwide if @educator.districtwide_access?
-      return :school if @educator.schoolwide_access? || @educator.has_access_to_grade_levels?
-      return :section if @educator.school.school_type == 'HS' && @educator.default_section
-      return :homeroom if @educator.school.school_type != 'HS' && @educator.default_homeroom
-      return :nothing
-    rescue Exceptions::NoAssignedHomeroom, Exceptions::NoAssignedSections => err
-      :nothing
-    end
-  end
-
   # TODO(kr) remove implementation
   def students_for_school_overview
     return [] unless @educator.school.present?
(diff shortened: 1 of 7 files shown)


kevinrobinson:
selfie

PR #1587, opened 2018-04-03
Add Storybook

kevinrobinson:
This adds [Storybook](https://github.com/storybooks/storybook) for making UIs.  You write tiny "stories" about individual components, like you might for tests, and then run `yarn storybook` and you can look at them and interact with them at `http://localhost:6006`.  This is useful as a component library, and especially useful when you have components that can be in several different states that are hard to see all at once in the actual product.

This doesn't add in global styles yet since those are built by Rails (I spiked on this a while back on [feature/sass](https://github.com/studentinsights/studentinsights/tree/feature/sass) but that's more involved so cutting it for now).  Because of this, the storybook view isn't an exact match for what you see in development.  But for most components, this should be a minimal difference and we should be able to make these match as we remove the last bits in `legacy.js` and in Rails SASS files.

Here's an example:
screen shot 2018-04-03 at 5 13 27 pm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,8 @@ before_install:
 
 install:
   - . $HOME/.nvm/nvm.sh
-  - nvm install 8.11.0
-  - nvm use 8.11.0
+  - nvm install --lts
+  - nvm use --lts
   - bundle install --retry=3
   - yarn install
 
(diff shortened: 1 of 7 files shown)


kevinrobinson:
selfie cc @alexsoble @edavidsonsawyer this might be useful for new features but also feel free to ignore if it doesn't seem worthwhile.  I've found it super helpful for small component things, but less helpful for whole pages (where there's more setup to get the data needed for the component, similar to when writing tests).  But your mileage may vary :)

PR #1588, opened 2018-04-03
Revert "Fix bug with student section assignment import"

alexsoble:
Reverting because trying to trigger a one-off import job from the command line fails. Unclear if that has to do with this PR or with some other problem. 

Reverts studentinsights/studentinsights#1580.


alexsoble:
Deploying reversion . . .

PR #1590, opened 2018-04-04
Remove flash notice to keep navbar layout fixed

kevinrobinson:
# Who is this PR for?
educators, developers

The flash messages about signing in and out aren't super informative, and right now they change the layout of the page based on whether they're present or not.  This removes notices, and moves error messages so they stay within the containing navbar.

### before
screen shot 2018-04-03 at 8 13 03 pm
screen shot 2018-04-03 at 8 12 56 pm
screen shot 2018-04-03 at 8 12 48 pm

### after
screen shot 2018-04-03 at 8 19 49 pm
screen shot 2018-04-03 at 8 19 30 pm
screen shot 2018-04-03 at 8 19 39 pm

### DIFF
--- a/app/assets/stylesheets/nav.scss
+++ b/app/assets/stylesheets/nav.scss
@@ -84,8 +84,8 @@
 
   .flash {
     position: absolute;
-    right: 390px;
-    top: 61px;
+    right: 40px;
+    top: 62px;
   }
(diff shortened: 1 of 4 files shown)


kevinrobinson:
selfie

PR #1591, opened 2018-04-04
Fix typo in high absences copy

kevinrobinson:
# Who is this PR for?
educators

# What problem does this PR fix?
the mistake visible in snapshots and screenshots in https://github.com/studentinsights/studentinsights/pull/1589 :\

PR #1593, opened 2018-04-04
Fix copy typo in absences box, again

kevinrobinson:
one more from https://github.com/studentinsights/studentinsights/pull/1591/files


kevinrobinson:
selfie

PR #1594, opened 2018-04-04
Fix date formatting in absences box to use nth

kevinrobinson:
From https://github.com/studentinsights/studentinsights/pull/1593, fix the date formatting.  The bug here using `Qo` for "quarter" looked correctly but wasn't actually the right date.


kevinrobinson:
selfie

PR #1597, opened 2018-04-04
Add newlines to make import task logging more readable

alexsoble:
# Who is this PR for?

Developers who work with the data import process. 

We lost newlines when we moved from using `puts` to `write`, so the log output for the import task got difficult to read: 

screen shot 2018-04-04 at 9 15 10 am

Adding in newlines automatically makes the logs more legible:  

screen shot 2018-04-04 at 9 19 23 am

### DIFF
--- a/app/importers/import_task.rb
+++ b/app/importers/import_task.rb
@@ -151,6 +151,6 @@ class ImportTask
   end
 
   def log(msg)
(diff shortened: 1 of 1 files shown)


alexsoble:
selfie

Issue #1596, opened 2018-04-04
Fix bug with one-off data import jobs succeeding silently


alexsoble:
# The problem 

Nighty data import jobs for Somerville are running smoothly, but when I try to run a one-off job from the command line, it fails without any error logging or meaningful output.

# What I see so far

Running the import job against production with the `--background` flag activated results in silent failing. 

Running with `--no-background` results in a successful import job. (I tried this late last night so as to not affect any live users.)

So the issue must involve backgroundizing the job. 

I checked `Delayed::Job` in the console in production and found that every time I ran a backgroundized import job, another job record was being added to the database table:

```
irb(main):002:0> Delayed::Job.all.count
=> 2

# try to run another import job from the background

irb(main):008:0> Delayed::Job.all.count
=> 3

irb(main):004:0> Delayed::Job.first
=> #<Delayed::Backend::ActiveRecord::Job 
id: 502, 
priority: 0, 
attempts: 0, 
handler: "--- !ruby/object:ImportJob\noptions: !ruby/hash:Tho...", 
last_error: nil, 
run_at: "2018-04-03 21:48:22", 
locked_at: "2018-04-03 21:48:26",
failed_at: nil,
queue: nil, 
created_at: "2018-04-03 21:48:22", 
updated_at: "2018-04-03 21:48:22">
```

alexsoble:
According to the DelayedJob docs, `locked_at` is "set when a client is working on this object." So it looks like these jobs are being started, but aren't finishing or failing.

alexsoble:
Alright, here's a little more context/info from #1598.

I ran the following one-off import job from the console:

```
$ heroku run thor import:start --background --source students --school HEA
```

Here's the result on the Import Records page:

```
Import Record #1512
Job completed in 1 minute.
Started: 04/04/2018 11:56 AM.
Ended: 04/04/2018 11:57 AM.
Down-to-the-file details:
[
  {
    "importer": "StudentsImporter",
    "start_time": "2018-04-04T15:56:43.897Z",
    "end_time": "2018-04-04T15:57:58.362Z"
  }
]
```

This is telling me that the job is running, in the sense that it's looping through the file importers and reaching this line in the code that updates "time_ended":

```
@record.update(
  importer_timing_json: timing_log.to_json,
  time_ended: DateTime.current,
)
```

But it's NOT creating any log output and it's NOT clearing the Delayed::Job record from the database as expected.

alexsoble:
[...]

PR #1598, opened 2018-04-04
Add more detailed info to import records page

alexsoble:
# Who is this PR for?

Anyone who has to debug the data import process. 😄 

We're storing a bunch of detailed information about the data import process in the database, but we're not surfacing much of it in the "Import records" view in the app. Surfacing more of that data will paint a more complete picture of what's going on with data import and simplify debugging.

# Screenshot (local data)

screen shot 2018-04-04 at 10 03 11 am

### DIFF
--- a/app/models/import_record.rb
+++ b/app/models/import_record.rb
@@ -8,4 +8,22 @@ class ImportRecord < ActiveRecord::Base
     time_ended - time_started
   end
 
+  def time_started_display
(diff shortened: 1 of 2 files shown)


alexsoble:
selfie! excited to see how this looks in prod!

alexsoble:
Should also add that this is motivated by debugging #1596.

PR #1599, opened 2018-04-04
Store import options in database and display on "Import Records" page

alexsoble:
# Who is this PR for?

Developers who need to debug the data import process. 😄 

Right now there's no visibility into what options were passed down to the import task. Historical import records don't tell us too much if we don't know what options were passed in.

# What does this PR do?

Store import task options in the database as JSON and display on "Import Records" page. 

# Screenshot (local data)

screen shot 2018-04-04 at 10 39 23 am

### DIFF
--- a/app/importers/import_task.rb
+++ b/app/importers/import_task.rb
@@ -1,14 +1,16 @@
 class ImportTask
 
   def initialize(options:)
+    @options = options
+
     # options["school"] is an optional filter; imports all schools if nil
-    @school = options.fetch("school", nil)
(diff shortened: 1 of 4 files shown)


alexsoble:
selfie. I expect shipping this will have no effect on the existing "Import Record" page but new import jobs will show the `options` hash going forward.

alexsoble:
🎉 👍  This is looking good in prod

kevinrobinson:
this is a great idea! 🌮

alexsoble:
Thanks @kevinrobinson! :D

PR #1601, opened 2018-04-05
Fix CSS on import records page

alexsoble:

Fix misaligned CSS on import records page. 

# Screenshots
screen shot 2018-04-05 at 9 54 51 am
screen shot 2018-04-05 at 9 54 12 am
screen shot 2018-04-05 at 9 52 06 am


alexsoble:
selfie

PR #1604, opened 2018-04-06
Update gem versions

alexsoble:
# Notes 

+ `bundle update`
+ Fix a non-high-risk security warning that GitHub is showing for a dependenciy of a dependency


alexsoble:
The dependencies I see that have major version bumps are autoprefixer-rails, uglifier, capybara, and xpath.

Gemfile.lock
```diff
@@ -377,4 +374,4 @@ RUBY VERSION
    ruby 2.3.0p0
 
 BUNDLED WITH
-   1.15.2
+   1.16.1
```
kevinrobinson: @alexsoble Let's keep this matching Heroku - check out https://github.com/studentinsights/studentinsights/pull/1542 (most of those comments aren't relevant but the links to the Heroku docs explain why they fix this, and it'd probably be good to follow their advice unless we have reasons to do otherwise).

alexsoble: Ooooh, good catch, thanks for pointing me to #1542. 👍

alexsoble:
Want to get this one shipped so that it's not just hanging around . . . I used an older version of bundler but the BUNDLED WITH wouldn't decrement, apparently it only increases: https://github.com/bundler/bundler/issues/3697#issuecomment-107223143. So I'll just manually back out that change.

alexsoble:
Did some clicking around and note-taking on my local branch with demo data to check for anything that the tests might have missed, didn't see anything. Going to force-push this branch to the demo site and then, if all goes well there, do a real deploy.

alexsoble:
Deploy to demo site looks good, demo site looks good . . . 🚢! selfie

PR #1606, opened 2018-04-06
Show import task logs on ImportRecord page; move ImportRecord page to React

alexsoble:
# Who is this PR for?

Developers working on data import.

Finding the import logs takes a whole bunch of digging, especially if you're looking for logs from the not-as-recent past.

# What does this PR do?

Shows the log output for each import task on the Import Records page. 

Because the logs are long, I wanted to be able to add a show/hide toggle so that the logs don't take up all the space on the page.

That led me to convert the entire Import Records page from Rails ERB HTML to React, which may have been a little bit of 🚲 🚲 _bikeshedding_ 🚲 🚲.

Overall though I'm feeling confident this will be useful new tooling for understanding/debugging the import process.

# GIF (completed import)

import-records-react

# Screenshot (import that did not complete)

screen shot 2018-04-09 at 11 56 33 am


app/assets/javascripts/import_records/ImportRecordsPage.js
```diff
@@ -0,0 +1,76 @@
...
+    return (
+      <div className="ImportRecordsPage">
+        <GenericLoader
+          promiseFn={this.fetchImportRecords}
+          render={this.renderPage} />
+      </div>
+    );
+  }
+  renderImportRecords(importRecords) {
+    const divStyle = {margin: '15px 0'};
```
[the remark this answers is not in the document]
kevinrobinson: And in general, does it seem okay to factor out the static styles up to the top or bottom of the file, outside of the component itself (eg, [one possible example](https://github.com/studentinsights/studentinsights/blob/master/app/assets/javascripts/school_courses/SchoolCoursesPage.js#L141))?  It may not particularly matter for what we're doing here, but seems good to keep them somewhat separate (but still in the same file for easy editing and since they're tightly coupled).

app/importers/import_task.rb
```diff
@@ -156,6 +156,8 @@ def run_update_tasks
   end
 
   def log(msg)
-    @log.write("\n\n💾  ImportTask: #{msg}")
+    full_msg = "\n\n💾  ImportTask: #{msg}"
+    @log.write full_msg
+    @record.log += full_msg
```
kevinrobinson: This line seems like the most meaningful change.  We'll keep these logs indefinitely, in contrast to the rolling expiration with other logs.  And adding the log output to the record increases the record size and grows the `ImportRecords` table at that rate each day.  These sound fine, but both are slightly different than how I usually think about logs.

This also introduces a new choice to developers - if someone wants to see the logs for what happened in an import task, should they look here in the UI or in the logs in logentries?  Both "products" for looking at logs have different features and affordances (eg, sorting, grepping, seeing historical comparisons, etc).  It's good how you flagged this as maybe 🚴 shedding and when it starts to feel like you're re-creating another product it's good to ask if it's worthwhile (and sometime it definitely is!).  Another intermediate option might be to have this UI link directly to the relevant bits in logentries (maybe possible, not sure).

Anyway, to avoid more 🚴s, I'd say from here you should comment and do whatever you think is best and  🚢it!  👍

app/assets/javascripts/import_records/ImportRecordToggleableSection.js
```diff
@@ -0,0 +1,66 @@
...
+
+    const preStyle = {
+      fontSize: 14,
+      color: '#3d3d3d',
+      marginTop: 10,
+      marginBottom: 20,
+    };
+
+    return (
+      <pre style={preStyle}>{children}</pre>
```
kevinrobinson: If I read right, the point of this component is to render preformatted text and hide/show it.  So this doesn't seem like `children` is the right API - would it ever make sense to put other React elements inside here other than plain text?

app/assets/javascripts/import_records/ImportRecordsPage.js
```diff
@@ -0,0 +1,76 @@
...
+    const titleStyle = {borderBottom: '1px solid #333'};
+
+    return (
+      <div style={outerWrapperStyle}>
+        <div style={columnStyle}>
+          <h4 style={titleStyle}>Import Records</h4>
+          {this.renderImportRecords(json.import_records)}
+        </div>
+        <div style={columnStyle}>
+          <h4 style={titleStyle}>Queued Jobs</h4>
```
kevinrobinson: These might be places for `<SectionHeading />`.

app/assets/javascripts/import_records/QueuedJobCard.js
```diff
@@ -0,0 +1,39 @@
...
+      locked_at,
+      failed_at,
+      attempts,
+      last_error,
+    } = this.props;
+    const divStyle = {border: '1px solid #eee', padding: 15, margin: '15px 0'};
+    const cardTitleStyle = {marginBottom: 15, borderBottom: '1px solid #eee', fontSize: 18};
+
+    return (
+      <div style={divStyle}>
```
kevinrobinson: Maybe a place for `<Card />`?

app/assets/javascripts/import_records/ImportRecordCard.js
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  render() {
+    const {id} = this.props;
+    const divStyle = {border: '1px solid #eee', padding: 15, margin: '15px 0'};
+    const cardTitleStyle = {marginBottom: 15, borderBottom: '1px solid #eee', fontSize: 18};
+
+    return (
+      <div style={divStyle}>
+        <div style={cardTitleStyle}>Import #{id}</div>
```
kevinrobinson: maybe `<Card />` and `<SectionHeading />` would remove some of the styling duplication?

app/importers/import_task.rb
alexsoble: Yeah, agreed this is the most meaningful change! 

I want to create a "one-stop shop" for debugging and monitoring the nightly import jobs. Right now it's a mix of queries in the console, Logentries, maybe error alert emails from the job, and the Import Record page in the app. By adding a copy of the logs to the Import Record page my goal is to make one page that can answer the question, "what's going on with the import job?" 

When I mentioned 🚲shedding, I was just referring to the HTML => React part of the work in particular.

app/assets/javascripts/import_records/ImportRecordsPage.js
alexsoble: Yes!

app/assets/javascripts/import_records/ImportRecordToggleableSection.js
alexsoble: Ooooh good catch!

app/assets/javascripts/import_records/ImportRecordCard.js
```diff
@@ -0,0 +1,97 @@
...
+      showOptions: false,
+      showFileByFile: false,
+      showLog: false,
+    };
+  }
+
+  onToggleState(stateKey) {
+    return function () {
+      this.setState({ [stateKey]: !this.state[stateKey] });
+    }.bind(this);
```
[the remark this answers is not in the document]
alexsoble: You're right, that's much simpler  . . . the power of binding at the point of passing in the func instead of up in the constructor!

app/assets/javascripts/import_records/ImportRecordsPage.js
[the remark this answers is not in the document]
alexsoble: Yeah, that sounds like a good practice and allowing for re-use within the file is a good idea. Kind of quasi-CSS-like in the sense that we're separating styles from elements, albeit on the same page. 

I like putting the styles at the bottom of the file as opposed to the top, that way when you open the file for the first time you still see the component name and constructors first.

config/routes.rb
```diff
@@ -57,9 +58,10 @@
   end
   resources :homerooms, only: [:show]
   resources :sections, only: [:index, :show]
-  resources :import_records, only: [:index]
   resources :iep_documents, only: [:show]
 
+  get 'import_records' => 'ui#ui'
```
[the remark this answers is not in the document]
alexsoble: I like that idea! Follows what you did with scoping `get '/authorization' => 'educators#authorization'` under the `/admin` namespace.

app/importers/import_task.rb
kevinrobinson: 👍

app/controllers/import_records_controller.rb
```diff
@@ -1,17 +1,49 @@
...
+      }
+    else
+      {
+        id: import_record.id,
+        completed: false,
+        time_started_display: import_record.time_started_display,
+        importer_timing_json: import_record.importer_timing_json,
+        task_options_json: import_record.task_options_json,
+        log: import_record.log,
+      }
```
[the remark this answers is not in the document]
alexsoble: Yup, exactly. ImportRecordTimingSection tests for `completed` in `render()`. And you're exactly right, we're leaning on the server for `distance_of_time_in_words`.

alexsoble:
🚢 ing

PR #1607, opened 2018-04-09
Add page to see district enrollment patterns

kevinrobinson:
# Who is this PR for?
districtwide admin, Insights team

In looking at classroom assignments data across schools and grades, there's some differences in enrollment that impact how this would would across schools and grades.  This shows this at-a-glance (the same data could be pieced together from the `/equity/principal` view or `/schools/:id` view but this makes it more readable.

# What does this PR do?
Adds a new districtwide only endpoint.

It also adds some test setup code to make tests a bit stricter, which came up in this work.

screen shot 2018-04-09 at 10 03 06 am

# Checklists


app/assets/javascripts/home/HomeInsights.test.js
```diff
@@ -13,6 +13,9 @@ function testProps(props = {}) {
   };
 }
 
+jest.mock('./CheckStudentsWithLowGrades');
+jest.mock('./CheckStudentsWithHighAbsences');
```
kevinrobinson: After updating `setupTests` to fail on unhandled promise rejections, these were failing.  Mocking these components loses the propTypes checking, but then we can do that in the test code below.  Ideally we'd get all this automatically, but not sure how to do this right now.

ui/config/setupTests.js
```diff
@@ -21,4 +21,12 @@ console.error = jest.fn(error => { throw new Error(error); }); //eslint-disable-
 console.warn = jest.fn(warn => { throw new Error(warn); }); //eslint-disable-line no-console
 
 // flag for GenericLoader
-global.GENERIC_LOADER_THROW_ON_REJECT_IN_TEST = true;
\ No newline at end of file
+global.GENERIC_LOADER_THROW_ON_REJECT_IN_TEST = true;
+
+// Unhandled promise rejections should fail tests
```
kevinrobinson: This makes tests stricter - it fails if there's an unhandled promise rejection (which in practice happens when a component makes a request to the server that's not mocked).

kevinrobinson:
selfie

PR #1612, opened 2018-04-10
Fix path helper in districtwide admin page

kevinrobinson:
# Who is this PR for?
districtwide users

districtwide page broken in https://github.com/studentinsights/studentinsights/pull/1606


kevinrobinson:
selfie cc @alexsoble

PR #1613, opened 2018-04-12
Update Delayed::Job failures modes

kevinrobinson:
# Who is this PR for?
NB educators

NB import jobs have been failing, and repeatedly retrying and failing.  We haven't been notified of these, and they haven't healed themselves because the root problem is still active.

# What does this PR do?
Changes the tuning for Delayed::Job to get different behavior in failure modes (see comments in code) and adds in email notifications on import job failures (although these aren't currently working, which we need to look at separately).

I'll verify this manually - deploy, run the import, clear the job queue, and then try killings some import tasks to see this behaves as expected.

After that we should be stable and can look at the actual issue with the import and why it's running out of memory.

### DIFF
--- a/app/jobs/import_job.rb
+++ b/app/jobs/import_job.rb
@@ -1,3 +1,4 @@
+# These methods are all hooks for https://github.com/collectiveidea/delayed_job
 class ImportJob
 
   def initialize(options:)
@@ -16,4 +17,14 @@ class ImportJob
     6.hours
   end
 
+  def error(job, exception)
+    extra_info =  { "hook" => "error", "job" => job.as_json }
+    ErrorMailer.error_report(exception, extra_info).deliver_now if Rails.env.production?
+  end
+
+  def failure(job)
+    extra_info =  { "hook" => "failure", "job" => job.as_json }
+    exception = StandardError.new('ImportJob#failure')
+    ErrorMailer.error_report(exception, extra_info).deliver_now if Rails.env.production?
+  end
 end
--- a/app/jobs/import_job.rb
+++ b/app/jobs/import_job.rb
@@ -18,12 +18,12 @@ class ImportJob
   end
 
   def error(job, exception)
-    extra_info =  { "hook" => "error", "job" => job.as_json }
+    extra_info =  { "hook" => "error", "job" => job.as_json(except: :log) }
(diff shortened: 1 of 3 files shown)


alexsoble:
While I'm 🚢ing PRs tonight . . .

alexsoble:
Sorry @kevinrobinson, I didn't take any explicit steps towards verifying this last night. I knew the New Bedford job would run overnight, and since hasn't been reaching the X2AssessmentImport because of an error it has been swallowing silently, I expected that we'd get an email alert with an error message. 

I didn't clear out the queue or trigger any import process manually, and I got the No. 111 SignalException: SIGTERM email from Rollbar a few hours later. I take this as confirmation that `Delayed::Worker.raise_signal_exceptions = true` is working now and being noisy about jobs that end because they run out of memory, which is a great improvement. 

Two other things I noticed that I'm wondering about:

+ My assumption is that the SIGTERM email we got is for the New Bedford app. It looks like right now there's no way to tell which instance a Rollbar email is coming from. Both Heroku instances share the same `ROLLBAR_ACCESS_TOKEN` variable. And we're not passing in any extra data about which instance is triggering the error. I think it would be good to pass that information to Rollbar, I'm going to go ahead and add that.

+ We got the Rollbar notification, which is good. I didn't get any notification from Student Insights ErrorMailer, which should have been triggered by the `error` Delayed::Job hook if I'm reading things right. I don't think this is a big problem but wanted to note and share what I'm seeing.

kevinrobinson:
@alexsoble got it, thanks for explaining!  Let's focus first on: restoring the Delayed::Job queue to a healthy state, and getting the import job to run successfully.

I emailed you separately the other day about `ErrorMailer` since I think it requires you to set up the DNS in MailGun, and Rollbar is on the list of things that need to be done for New Bedford in the timeline doc already.

kevinrobinson:
@alexsoble Looking at this now, ping me if that is a bad idea :)

alexsoble:
@kevinrobinson Go for it, I'm looking at setting up MailGun for New Bedford right now. Our domain name provider's UI doesn't seem to afford a way to set up the TXT records we need for Mailgun so I'm chatting with their customer support right now.

alexsoble:
@kevinrobinson OK, made some progress on the Mailgun setup . . . waiting for the records to propagate . . . ⏲

PR #1615, opened 2018-04-13
Fix sorting for times and days in discipline dashboard

edavidsonsawyer:
# Who is this PR for?
Dashboard Users

Times and days of the week are not sorted in the discipline dashboard charts

# What does this PR do?
Imposes order on the time and day charts

# Checklists

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
@@ -61,6 +61,34 @@ class SchoolDisciplineDashboard extends React.Component {
       title: "Incidents by " + selectedChart};
   }
 
+  sortChartData(chart) {
+    switch(chart.type) {
+    case 'time': return this.sortByTime(chart.disciplineIncidents);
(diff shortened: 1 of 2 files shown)


edavidsonsawyer:
@kevinrobinson @alexsoble, this should fix the sorting issues for Day and Time discipline charts. 

However, this is a quick fix for tomorrow. To do this properly I need to rewrite the component to use maps instead of objects, which I won't have the chance to do tonight.

alexsoble:
@edavidsonsawyer The fix looks like good progress when I test it locally, shipping now. Here's what the days of the week chart looks with my local demo data setup, maybe one more change we can make is setting the count to zero by default for each day of the week so all the days of the week show up even if there are no events that day: 

screen shot 2018-04-13 at 7 20 59 am

alexsoble:
@edavidsonsawyer Ah, I see why this may have been difficult to spot:

```
# /spec/factories/discipline_incidents.rb

factory :discipline_incident do
  ...
  has_exact_time { if Random::rand(1.0) > 0.91 then true else false end }
  ...
end
```

Only 9% of the demo incidents have an exact time. At first I assumed that was realistic to the production data, but when I look at the Healey school in the discipline dashboard, it seems that the clear majority of incidents have an exact time. This may be different across different schools.

In any case, having only 9% of incidents with exact time means that debugging the "incident time" chart with demo data would be hard to do.

PR #1616, opened 2018-04-13
Time sorting in Disc Dashboard

edavidsonsawyer:
Improves sorting in discipline chart, and also sorts grade.

+ [ ] Author checked latest in IE - discipline dashboard
+ [ ] Author checked latest in IE - discipline dashboard

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
@@ -64,14 +64,14 @@ class SchoolDisciplineDashboard extends React.Component {
   sortChartKeys(chartKeys) {
     switch(this.state.selectedChart) {
     case 'time': return this.sortedTimes(chartKeys);
-    case 'day': return this.sortedDays();
+    case 'day': return this.sortedDays(chartKeys);
     case 'grade': return this.sortedGrades(chartKeys);
     default: return chartKeys;
     }
   }
 
-  sortedDays() {
-    return ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'];
+  sortedDays(chartKeys) {
+    return ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'].filter((day) => chartKeys.includes(day));
   }
 
   sortedTimes(chartKeys) {
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
@@ -61,32 +61,27 @@ class SchoolDisciplineDashboard extends React.Component {
       title: "Incidents by " + selectedChart};
   }
 
-  sortChartData(chart) {
-    switch(chart.type) {
-    case 'time': return this.sortByTime(chart.disciplineIncidents);
-    case 'day': return this.sortByDay(chart.disciplineIncidents);
-    default: return chart.disciplineIncidents;
+  sortChartKeys(chartKeys) {
+    switch(this.state.selectedChart) {
+    case 'time': return this.sortedTimes(chartKeys);
+    case 'day': return this.sortedDays();
+    case 'grade': return this.sortedGrades(chartKeys);
+    default: return chartKeys;
     }
   }
 
-  sortByDay(incidentsGroupedByDay) {
-    const dayOrder = ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'];
-    let sortedHash = {};
-    dayOrder.forEach((day) => {
-      if (incidentsGroupedByDay[day]) sortedHash[day] = incidentsGroupedByDay[day];
-    });
-    return sortedHash;
+  sortedDays() {
+    return ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'];
   }
 
-  sortByTime(incidentsGroupedByTime) {
-    const sortedTimes = Object.keys(incidentsGroupedByTime).sort((a, b) => {
+  sortedTimes(chartKeys) {
+    return chartKeys.sort((a, b) => {
       return new Date('1970/01/01 ' + a) - new Date('1970/01/01 ' + b);
     });
-    let sortedHash = {};
-    sortedTimes.forEach((hour) => {
-      sortedHash[hour] = incidentsGroupedByTime[hour];
-    });
-    return sortedHash;
+  }
+
+  sortedGrades(chartKeys) {
+    return ['PK', 'KF', '1','2','3','4','5','6','7','8','9','10','11','12'].filter((grade) => chartKeys.includes(grade));
   }
 
   render() {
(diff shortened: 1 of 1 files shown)


alexsoble:
Hey @edavidsonsawyer! I texted Uri to ask what time he was planning on showing the discipline dashboards, and he said noon ET which is in ~30 minutes, so I got super focused on trying to fix the time sorting issue before then. Just shipped #1617 and then saw your PR — sorry about that! It looks like this PR has a couple of great additional changes beyond the time chart sorting, so if you don't mind rebasing this on top of master, I'd love to get this one shipped too! It does look like our time chart changes conflict, I just changed the sorting whereas it looks like you changed the data type that gets passed down to the chart.

edavidsonsawyer:
Oh man, I didn't realize you were working on this! I was trying to get this in before noon. I'll rebase this when I get back to my computer.

alexsoble:
@edavidsonsawyer My solution in #1617 shipped a mistake with how "12 PM" is handled (see below). 

screen shot 2018-04-13 at 10 41 51 am

Looks like your PR solves this in a more thorough way by improving how the time data is being passed down to the chart, which is 👍. Excited to get this shipped, give me a shout if I can help here.

alexsoble:
@edavidsonsawyer Feel free to back out my change! It's totally hacky and definitely not the best way to approach this, I just wanted to get a short-term fix up there.

edavidsonsawyer:
@alexsoble alright, this should be okay to merge in when you have the chance.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -61,48 +61,30 @@ class SchoolDisciplineDashboard extends React.Component {
...
 
-  sortByDay(incidentsGroupedByDay) {
-    const dayOrder = ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'];
-    let sortedHash = {};
-    dayOrder.forEach((day) => {
-      if (incidentsGroupedByDay[day]) sortedHash[day] = incidentsGroupedByDay[day];
-    });
-    return sortedHash;
+  sortedDays(chartKeys) {
+    return ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'].filter((day) => chartKeys.indexOf(day) >= 0);
```
alexsoble: I'm not sure filtering out days with zero discipline incidents is what we want for the x-axis. If I'm looking at a days of the week chart, I'd expect to see MTWThFSSun on the x-axis, even if there are (for example) no incidents on Tuesday, Wednesday, or Sunday. In the context of this chart I think days with no incidents could be especially interesting . . . "why are there no incidents on Mondays but lots on Fridays?," for example.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -61,48 +61,30 @@ class SchoolDisciplineDashboard extends React.Component {
...
-      let b_AmPm = b.split(' ')[1];
-
-      if (a_AmPm === 'AM' && b_AmPm == 'PM') return -1;
-
-      if (a_AmPm === 'PM' && b_AmPm == 'AM') return 1
-
-      return a_Hour - b_Hour;
+  sortedTimes(chartKeys) {
+    return chartKeys.sort((a, b) => {
+      if (a == "Not Logged") return -1;
```
alexsoble: I think a code comment about what shape we expect `a` and `b` to be passed down as would be really helpful here, this was something I had to spend some time hunting down in #1617. Also maybe a couple of words about why you chose `format('h:mm a')` — was that in part so that they formatted values would "play nice" with new Date?

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -61,48 +61,30 @@ class SchoolDisciplineDashboard extends React.Component {
...
+      if (b == "Not Logged") return 1;
+      return new Date('1970/01/01 ' + a) - new Date('1970/01/01 ' + b);
```
alexsoble: If I understand this right, 1970/01/01 is an arbitrary value and is just here so we can compare two times as `new Date(...) - new Date(...)`, is that right?

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -61,48 +61,30 @@ class SchoolDisciplineDashboard extends React.Component {
...
     });
+  }
+
+  sortedGrades(chartKeys) {
+    return ['PK', 'KF', '1','2','3','4','5','6','7','8','9','10','11','12'].filter((grade) => chartKeys.indexOf(grade) >= 0);
```
alexsoble: Same as with days of the week above, I don't think we want to be filtering out grades with no incidents. If 3rd grade has no incidents, that could be a really meaningful thing for an educator to see. Of course some schools will have no students in certain grades, like an elementary school that has no students in 9 through 12. But I think we should handle that by looking directly at the student data, not by asking if there are any incidents at a grade level, since zero incidents could occur at a valid grade level.

alexsoble:
@edavidsonsawyer Left some functionality notes and comments for you to review, let me know if these are helpful!

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: @alexsoble makes sense. 

Yes, the format was changed just so Date could parse the time, which seemed like the simplest solution. We could strip the minutes out again later. I can't decide whether I think having the minutes here is distracting.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: @alexsoble it does seem to work with just new Date(a) and it's probably fine to do that. My thinking was I'm not totally sure how it handles edge cases where a - b crosses into another day. Adding the time to a specific date removes ambiguity about which time is bigger.

edavidsonsawyer:
@alexsoble thanks for the comments! I pretty much agree, except that I'd prefer to keep the Date sorting as-is unless you feel strongly. Otherwise I've pushed some changes to address the other bits you found.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -61,48 +61,31 @@ class SchoolDisciplineDashboard extends React.Component {
...
+    //chartKeys will either contain a time like "4:00 pm", "10:00 am", or "Not Logged"
+    return chartKeys.sort((a, b) => {
+      if (a == "Not Logged") return -1;
+      if (b == "Not Logged") return 1;
+      return new Date('1970/01/01 ' + a) - new Date('1970/01/01 ' + b);
     });
+  }
+
+  sortedGrades(chartKeys) {
+    return ['PK', 'KF', '1','2','3','4','5','6','7','8','9','10','11','12'];
```
kevinrobinson: Check it out, hot off the press this week! https://github.com/studentinsights/studentinsights/blob/master/app/assets/javascripts/helpers/SortHelpers.js#L83

alexsoble: This is great @kevinrobinson, we should use the SortHelper instead of rolling our own here.

PR #1617, opened 2018-04-13
Fixup x-axis ordering for "incident time" chart

alexsoble:
# Who is this PR for?

Discipline dashboard chart users.

X-axis of "incident time" chart isn't sorting correctly (demo data): 

screen shot 2018-04-13 at 10 11 25 am

# What does this PR do?

Change custom sort function to parse times in the format they're being passed down as: `'5 AM', '4 PM', 'Not Logged'`.

screen shot 2018-04-13 at 10 24 24 am


alexsoble:
selfie, not waiting for the specs to pass because this looks good locally, is only changing a sort function in one place, and uri is going to do user research on this feature soon (12 et)!

alexsoble:
Darn. Looking at this in production, it's not **totally** accurate because it puts 12PM in the wrong place. So an incremental improvement because we went from not sorted to sorted with one mistake . . . still needs more fixing. 

screen shot 2018-04-13 at 10 41 51 am

kevinrobinson:
Hey @alexsoble!  This looks like it's breaking master: https://travis-ci.org/studentinsights/studentinsights/builds/366165877.

alexsoble:
@kevinrobinson Yeah, I saw that . . . I pushed the fix in a hurry and missed that linter output. #1616
 will fix when it lands.

PR #1618, opened 2018-04-13
Fix distribution of demo discipline incidents has_exact_time

alexsoble:
# Who is this PR for?

Devs working with the demo data.

Demo data doesn't match production! 

```
# production:

DisciplineIncident.count
=> 16884

DisciplineIncident.where(has_exact_time: true).count
=> 15450

DisciplineIncident.where(has_exact_time: false).count
=> 1434
```

Looks like the demo data generator has it backwards, instead of ~90% of discipline incidents **not** having an exact time, ~90% **do** have an exact time in the database. 

# What does this PR do?

Make the demo data generator match what we see in production, 91.6% of discipline incidents have an exact time.


kevinrobinson:
oops, thanks for fixing up! 🌮 👍

alexsoble:
spec failure is unrelated

PR #1622, opened 2018-04-13
Send district key context to Rollbar errors: server-side

alexsoble:
# Who is this PR for?

Developers who get error emails from Rollbar.

We don't know which school district / Heroku instance the error is coming from! 

# What does this PR do?

Sends the district key as an extra bit of information along with the error report.


alexsoble:
Going to ship this one first, test, and then add the same bit of district data to the client side error reporting.

PR #1624, opened 2018-04-13
Fix bug with importer logging mod line

kevinrobinson:
continuing https://github.com/studentinsights/studentinsights/pull/1621

I messed up the check in reducing logging.


kevinrobinson:
selfie

PR #1625, opened 2018-04-13
log.puts instead of .write, hash invalid rows

kevinrobinson:
more from https://github.com/studentinsights/studentinsights/pull/1613#issuecomment-381155196.

This changes `log.write` back to `logs.puts` (changed in https://github.com/studentinsights/studentinsights/pull/1547 for no particular reason; I think I thought they were equivalent).  Since `write` doesn't put a newline at the end, in the import task when there's a lot of logging, there's a lot of data written without ever flushing.  This was particularly influential before https://github.com/studentinsights/studentinsights/pull/1621, when we logged a full line of text status for every row in the attendance and behavior files.  This switches everything back to puts.  I believe this also explains why jobs would appear not to run - I just ran into this now where a job looked like it wasn't running, and soon as I cycle the dyno there's many many log lines that must have been running for a while but never flushing.

Separately, this hashes the output of invalid rows in logging; these contain data we don't want in logs.

### DIFF
--- a/app/importers/file_importers/attendance_importer.rb
+++ b/app/importers/file_importers/attendance_importer.rb
@@ -18,15 +18,15 @@ class AttendanceImporter
 
     @data.each_with_index do |row, index|
       import_row(row) if filter.include?(row)
-      @log.write("processed #{index} rows.") if index % 10000 == 0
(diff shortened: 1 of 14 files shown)


kevinrobinson:
selfie

PR #1629, opened 2018-04-16
Ensure absences and tardies are being eager loaded

alexsoble:
# Who is this PR for?

Users of the absences/tardies dashboards.

Slow to load: 12-15 seconds to see data in production. 

# Notes

I loaded the schools/hea/absences/ page in my local environment. In the logs I saw an N+1 pattern, meaning that each absence record was being loaded one by one instead of all at once:

screen shot 2018-04-16 at 1 41 30 pm

I saw that the query code for this endpoint JSON was using `includes`, but it was passing a named scoped association `:dashboard_absences` to the includes clause instead of just `:absences`. I tried passing down `:absences` and saw that the N+1 pattern went away:

screen shot 2018-04-16 at 1 50 33 pm

I wanted to add back in the "only load absences for the past year" feature provided by `:dashboard_absences`, so I used a `where` clause as described in the docs: http://guides.rubyonrails.org/active_record_querying.html#specifying-conditions-on-eager-loaded-associations.

Here's what the logs looked like now:

screen shot 2018-04-16 at 2 09 25 pm

Along the way I saw significant performance improvements. Here is the timing I saw from the from the GET request to the JSON endpoint at the beginning, before any refactors:

```
[0] Completed 200 OK in 5985ms (Views: 4358.3ms | ActiveRecord: 399.0ms)
```

Here's what I see now:

```
[0] Completed 200 OK in 1639ms (Views: 1136.4ms | ActiveRecord: 118.2ms)
```

So it looks like making sure absences and tardies are being eager loaded makes the endpoint 3 times faster. @edavidsonsawyer, do you see a similar improvement on your end? (I'm using the "more demo data" setting by the way.) I changed only the dark endpoints here so we could ship this PR and it wouldn't affect any of the user-facing dashboard pages.

### DIFF
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -44,7 +44,9 @@ class SchoolsController < ApplicationController
   #Individual dashboard data endpoints will replace school_administrator_dashboard once they're stable
   def absence_dashboard_data
     student_absence_data = students_for_dashboard(@school)
-      .includes([homeroom: :educator], :dashboard_absences, :event_notes)
+      .includes([homeroom: :educator], :absences, :event_notes)
+      .where("absences.occurred_at >= ?", 1.year.ago).references(:absences)
+
     student_absence_data_json = student_absence_data.map do |student|
       individual_student_absence_data(student)
     end
@@ -54,7 +56,9 @@ class SchoolsController < ApplicationController
 
   def tardies_dashboard_data
     student_tardies_data = students_for_dashboard(@school)
-      .includes([homeroom: :educator], :dashboard_tardies, :event_notes)
+      .includes([homeroom: :educator], :tardies, :event_notes)
+      .where("tardies.occurred_at >= ?", 1.year.ago).references(:tardies)
+
     student_tardies_data_json = student_tardies_data.map do |student|
       individual_student_tardies_data(student)
     end


app/controllers/schools_controller.rb
```diff
@@ -44,7 +44,9 @@ def school_administrator_dashboard
   #Individual dashboard data endpoints will replace school_administrator_dashboard once they're stable
   def absence_dashboard_data
     student_absence_data = students_for_dashboard(@school)
-      .includes([homeroom: :educator], :dashboard_absences, :event_notes)
+      .includes([homeroom: :educator], :absences, :event_notes)
+      .where("absences.occurred_at >= ?", 1.year.ago).references(:absences)
```
edavidsonsawyer: When I look at the absence dashboard on this branch, it looks like this: 
screen shot 2018-04-16 at 5 49 55 pm

But on master, it looks like this: 
screen shot 2018-04-16 at 5 50 09 pm

Note the percentages are significantly higher. I think adding the where clause here applies the where filter to the entire includes query, not just the absences. That filters out all students with no absences within the past year.

edavidsonsawyer:
@alexsoble That is really interesting! I also see an n+1 problem in the separate endpoints, but not the combined endpoint which uses the same pre-trimmed absences. 

I think using a where clause is a much cleaner approach, but I left some comments on this particular way of doing it. It would definitely be a better solution than having the cutoff date in the model!

edavidsonsawyer:
@alexsoble oh! I see why the query is misbehaving. The /absence and /tardies endpoints are looking for student.absences not student.dashboard_absences. I'll submit a PR to fix that, but I think it's still worth looking for a way to make this approach work!

app/controllers/schools_controller.rb
alexsoble: @edavidsonsawyer Ooooh, good catch here!

alexsoble:
Closing in favor of #1630.

PR #1630, opened 2018-04-16
fixing n+1 on individual dashboard endpoints

edavidsonsawyer:
# Who is this PR for?
Dashboard users

The individual absence dashboard endpoints had a bug causing n+1 conditions in the queries

# What does this PR do?
Fixes it!

### DIFF
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -206,7 +206,7 @@ class SchoolsController < ApplicationController
       last_name: student.last_name,
       id: student.id,
       homeroom_label: homeroom_label,
-      absences: student.absences,
+      absences: student.dashboard_absences,
       event_notes: student.event_notes
     })
   end
@@ -219,7 +219,7 @@ class SchoolsController < ApplicationController
       last_name: student.last_name,
       id: student.id,
       homeroom_label: homeroom_label,
-      tardies: student.tardies,
+      tardies: student.dashboard_tardies,
       event_notes: student.event_notes
     })
   end


edavidsonsawyer:
@alexsoble this should fix the n+1 issues you saw in #1629

alexsoble:
@edavidsonsawyer Hey, this is a great catch! I didn't realize that the N+1 behavior was different on the individual endpoints versus the combined endpoint. Really excited to see how much this drives down the load time for the individual endpoints. 

Did you look in the rails server logs in your development environment to see how long the individual endpoints took to load the JSON before and after this change? I think it's important to have that in this PR so that we have a sense of how much impact we think this change might have, and just to double check that we're making the endpoints load faster. If it's easy for you to add that data now, awesome, if not I can add it when I get to the office.

alexsoble:
@edavidsonsawyer Just remembered you have more free coding time in the evenings/at night. I have more coding time during the day, so we can work on these PRs together async. 👍 

OK, so here's what I see in the logs. I'm looking at the logs locally with demo data in "more demo students" mode.

**Individual absences endpoint, master branch:**

```
[0] Completed 200 OK in 5435ms (Views: 4710.5ms | ActiveRecord: 446.3ms)
```


**Individual absences endpoint, this branch:**

```
[0] Completed 200 OK in 2280ms (Views: 1760.4ms | ActiveRecord: 62.7ms)
```

**Combined school_administrator_dashboard endpoint:**

```
[0] Completed 200 OK in 3310ms (Views: 2469.8ms | ActiveRecord: 104.4ms)
```

Here's what this tells me:

+ Fixing this query makes the individual endpoints respond ~2.4 times as fast. 
+ The "combined endpoint" page is faster than the individual endpoints right now because of this N+1 issue. That explains why I found the combined endpoints were faster when I tried them out even though they loaded more data: https://github.com/studentinsights/studentinsights/pull/1575#issuecomment-379069904.

If we ship this fix, and then substitute the individual endpoints in production instead of the combined endpoints, we'll make the data load ~45% faster (2280ms versus 3310ms). I'll 🚢 this now and then take a crack at substituting in the individual endpoints. 🎉

PR #1631, opened 2018-04-17
Switching from slider to range buttons in dashboard

edavidsonsawyer:
# Who is this PR for?
Dashboard users

Feedback from users indicated they tend to have definite date ranges and locating them with the slider is frustrating

# What does this PR do?
Replaces the slider with preselected date ranges

rangebuttons

# Checklists

## Javascript QA

+ [x] Author checked latest in IE - Absence Dashbaord
+ [ ] Reviewer checked latest in IE - Absence Dashbaord

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashButton.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashButton.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+
+//Custom all purpose dashboard button
+class DashButton extends React.Component {
+
+  render() {
+    return (
+      <div className="DashButton">
+        <a
+          style={{
+            backgroundColor: this.props.isSelected? 'rgba(49, 119, 201, 0.75)' : ''
+          }}
+          onClick={this.props.onClick}>
+          {this.props.buttonText}
+        </a>
+      </div>
+    );
+  }
+}
+DashButton.propTypes = {
+  buttonText: PropTypes.string.isRequired,
+  onClick: PropTypes.func.isRequired,
+  isSelected: PropTypes.bool.isRequired
+};
+
+export default DashButton;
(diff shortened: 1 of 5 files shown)


edavidsonsawyer:
@alexsoble This is a very quick way to get this feature working. If it turns out we don't need to range slider at all, these dashboard components should be refactored and simplified. But since it sounds like users were playing with the slider on the discipline dashboard I think it's best to keep that functionality available in the absence dashboard behind the scenes for now.

alexsoble:
@edavidsonsawyer What do you think about matching the same set of styles we used for the "Reset Students" button here?

edavidsonsawyer:
@alexsoble Yup, I'm working on that now!

alexsoble:
Hey there @edavidsonsawyer! Gave this a look locally. 

labels

As a user, when I navigate to the page, my first question is, "What does 'School Year' / '90 days' / '45 days' mean?" It's not immediately clear to me that these are filters. They appear directly above the students table, but when I click them, the entire page updates. 

A couple ideas on how to make some of this information hierarchy clearer to the user: maybe we could add some more direct labeling saying "Filter By: This School Year | Past 90 Days | Past 45 Days." I'd also suggest moving the labels above both the graphs and the students table, to make it more visually obvious that it controls both of them. Also, clicking each filter changes the context of each charts, so I think it would be helpful to reflect that in the chart title: "Average Attendance By Month (This School Year)", "Average Attendance By Month (Past 90 Days)", "Average Attendance By Month (Past 45 Days)". These are just three ideas for how to communicate what's going on in the page more clearly, you may have more ideas and better ideas for how to do this. 

I'm also wondering if it would be a good idea to use the filters across both the Absences and Tardies dashboard. Absences and Tardies are two measures of attendance, so I think it could be confusing to have two totally separate UIs for the two different dashboard.

alexsoble:
@edavidsonsawyer I can merge in this PR and then make some of the changes described above and then we'll see how it looks. 😄 👍

PR #1633, opened 2018-04-17
Ship individual absences/tardies endpoints

alexsoble:
# Who is this PR for?

School and district administrators who use the absences and tardies dashboards to gain insights into attendance issues. 

# What problem does this PR fix?

Dashboards slow to load.

# What does this PR do?

Speeds up dashboard loading by ~45% (based on what we've seen locally) by subbing in individual endpoints that load less data. More detail on measurements here: https://github.com/studentinsights/studentinsights/pull/1630#issuecomment-382035893.

# GIF (principal user role, Internet Explorer)

schoolwide

# GIF (districtwide admin user role, Internet Explorer)

districtwide

PR #1634, opened 2018-04-18
Only send SST note data to dashboard pages

alexsoble:
# Who is this PR for?

Users of the absences/tardies dashboards.

Slow load times: ~4 seconds to see data in production. (Note that this was a major improvement from a couple days ago when load times were at 12+ seconds! See #1629, #1630.)

# What does this PR do?

Since we're only displaying Last SST Note on the dashboard views, only send down SST Notes data to the front-end by using a scoped query. If we need to use more event note data in the dashboards, we can always add to the query.

I tested the effect of these changes on load times (local environment, "more students mode" demo data, absences dashboard). I loaded each page 5 times to account for caching effects.

**loading the dashboard on `master`:**

```
1. [0] Completed 200 OK in 2486ms (Views: 1696.9ms | ActiveRecord: 78.8ms)
2. [0] Completed 200 OK in 2253ms (Views: 1638.0ms | ActiveRecord: 58.5ms)
3. [0] Completed 200 OK in 2228ms (Views: 1601.4ms | ActiveRecord: 82.4ms)
4. [0] Completed 200 OK in 2169ms (Views: 1630.0ms | ActiveRecord: 69.7ms)
5. [0] Completed 200 OK in 2195ms (Views: 1625.7ms | ActiveRecord: 58.5ms)
```

**loading the dashboard on this brach:**

```
1. [0] Completed 200 OK in 1401ms (Views: 635.2ms | ActiveRecord: 59.5ms)
2. [0] Completed 200 OK in 1011ms (Views: 570.9ms | ActiveRecord: 36.7ms)
3. [0] Completed 200 OK in 1259ms (Views: 732.3ms | ActiveRecord: 30.4ms)
4. [0] Completed 200 OK in 1143ms (Views: 657.9ms | ActiveRecord: 37.5ms)
5. [0] Completed 200 OK in 1093ms (Views: 585.7ms | ActiveRecord: 48.3ms)
```

Overall it looks like this change cuts load times by 50%. 

This PR also makes some other changes to the dashboard Ruby code which I'll add individual GitHub comments on.

### DIFF
--- a/app/assets/javascripts/helpers/latestNoteDateText.js
+++ b/app/assets/javascripts/helpers/latestNoteDateText.js
@@ -2,6 +2,12 @@ import _ from 'lodash';
 
 // Find the latest event note of a particular type, and return a string
 // of that date.
+
+// TODO (ARS): Sometimes we want to use this function on a list of notes that
+// all have the same event_note_type. In that case we just want to use the sorting
+// and text formatting functions here. We could switch the order of the arguments
+// passed in and make eventNoteTypeId an optional second argument.
+
 export function latestNoteDateText(eventNoteTypeId, eventNotes) {
   const sortedEventNotes = _.sortBy(eventNotes, note => new Date(note.recorded_at));
   const latestNoteOfType = _.findLast(sortedEventNotes, { event_note_type_id: eventNoteTypeId });
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardHelpers.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardHelpers.js
@@ -5,12 +5,7 @@ import _ from 'lodash';
 export default {
 
   groupByHomeroom: function(studentRecords) {
-    const studentsByHomeroom = _.groupBy(studentRecords, 'homeroom_label');
-    if (studentsByHomeroom[null]) {
-      studentsByHomeroom["No Homeroom"] = studentsByHomeroom[null];
-      delete studentsByHomeroom[null];
-    }
-    return studentsByHomeroom;
+    return _.groupBy(studentRecords, 'homeroom_label');
   },
 
   averageDailyAttendance: function(absenceEventsByDay, size) {
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -165,66 +165,48 @@ class SchoolsController < ApplicationController
     end
   end
 
-  # Methods for Dashboard
-
-  def individual_student_dashboard_data(student)
-    # This is a temporary workaround for New Bedford
-    homeroom_label = student.try(:homeroom).try(:educator).try(:full_name) || student.try(:homeroom).try(:name)
-    HashWithIndifferentAccess.new({
-      first_name: student.first_name,
-      last_name: student.last_name,
-      id: student.id,
-      grade: student.grade,
-      race: student.race,
-      homeroom_label: homeroom_label,
-      absences: student.dashboard_absences,
-      tardies: student.dashboard_tardies,
-      event_notes: student.event_notes
-    })
-  end
+  # Methods for Dashboards #
 
   def individual_student_absence_data(student)
-    # Gathers only the information needed for a specific dashboard view.
-    homeroom_label = student.try(:homeroom).try(:educator).try(:full_name) || student.try(:homeroom).try(:name)
-    HashWithIndifferentAccess.new({
+    {
       first_name: student.first_name,
       last_name: student.last_name,
       id: student.id,
-      homeroom_label: homeroom_label,
+      homeroom_label: homeroom_label(student.homeroom),
       absences: student.dashboard_absences,
       event_notes: student.event_notes
-    })
+    }
   end
 
   def individual_student_tardies_data(student)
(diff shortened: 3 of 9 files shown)


app/assets/javascripts/helpers/latestNoteDateText.js
```diff
@@ -2,9 +2,15 @@ import _ from 'lodash';
 
 // Find the latest event note of a particular type, and return a string
 // of that date.
+
+// TODO (ARS): Sometimes we want to use this function on a list of notes that
+// all have the same event_note_type. In that case we just want to use the sorting
+// and text formatting functions here. We could switch the order of the arguments
+// passed in and make eventNoteTypeId an optional second argument.
```
alexsoble: I think this would be a good improvement to make, but after this PR lands.

app/controllers/schools_controller.rb
```diff
@@ -28,8 +28,8 @@ def csv
   end
 
   def absence_dashboard_data
-    student_absence_data = students_for_dashboard(@school)
-      .includes([homeroom: :educator], :dashboard_absences, :event_notes)
+    student_absence_data = active_students(@school)
```
alexsoble: I renamed this method because it added an extra step to reading/writing code in this class. The first time I saw the `students_for_dashboard` method, it made me wonder, "ooooh, I wonder what queries we're running to get the right students for the dashboard!" and went and looked at the method. But it turned out we were just querying for active students. I changed the method name to be a little more upfront about what it's doing.

app/controllers/schools_controller.rb
```diff
@@ -165,66 +165,48 @@ def authorized_students_for_overview(school)
...
-      event_notes: student.event_notes
-    })
-  end
+  # Methods for Dashboards #
 
   def individual_student_absence_data(student)
-    # Gathers only the information needed for a specific dashboard view.
-    homeroom_label = student.try(:homeroom).try(:educator).try(:full_name) || student.try(:homeroom).try(:name)
-    HashWithIndifferentAccess.new({
+    {
```
alexsoble: I believe `HashWithIndifferentAccess` is slower than just a plain hash and uses more memory. (Here's one person's benchmark data: https://gist.github.com/tiagoamaro/c82a27aceedfc901b081#file-results_with_4-2-5-txt.) Since swapping it out for a plain hash didn't cause any feature regression that I could see, I figured why not.

app/controllers/schools_controller.rb
```diff
@@ -165,66 +165,48 @@ def authorized_students_for_overview(school)
     end
   end
 
-  # Methods for Dashboard
-
-  def individual_student_dashboard_data(student)
```
alexsoble: This was for the combined page that was removed in #1633.

app/controllers/schools_controller.rb
```diff
@@ -165,66 +165,48 @@ def authorized_students_for_overview(school)
...
-      absences: student.dashboard_absences,
-      tardies: student.dashboard_tardies,
-      event_notes: student.event_notes
-    })
-  end
+  # Methods for Dashboards #
 
   def individual_student_absence_data(student)
-    # Gathers only the information needed for a specific dashboard view.
-    homeroom_label = student.try(:homeroom).try(:educator).try(:full_name) || student.try(:homeroom).try(:name)
```
alexsoble: Since this logic looked complicated and was repeated three times I factored it into a method.

app/controllers/schools_controller.rb
```diff
@@ -165,66 +165,48 @@ def authorized_students_for_overview(school)
...
+    }
   end
 
-  def students_for_dashboard(school)
+  def active_students(school)
     school.students.active
   end
+
+  def homeroom_label(homeroom)
+    homeroom.try(:educator).try(:full_name) || homeroom.try(:name) || "No Homeroom"
```
alexsoble: This is probably scope creep for this PR, but . . . I figured as long as we're testing for whether or not the student has a homeroom label here, why not return the string value for when there's no homeroom? That lets us remove JS code in a couple of places in the dashboards testing for no-homeroom.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashboardHelpers.js
```diff
@@ -5,12 +5,7 @@ import _ from 'lodash';
 export default {
 
   groupByHomeroom: function(studentRecords) {
-    const studentsByHomeroom = _.groupBy(studentRecords, 'homeroom_label');
```
alexsoble: See https://github.com/studentinsights/studentinsights/pull/1634/files#r182439306.

alexsoble:
@edavidsonsawyer Agreed, I'm curious to find out too! Let me know if you have any questions or comments, I'll ship this now but we can keep chatting in this PR if there's more you want to explore. 

Since this started out as a Ruby-code-only PR, I deleted the sections in the pull request template about manually checking the UI in Internet Explorer. But then I added a few commits that did touch JS code, specifically around handling "No Homeroom."

I checked the absences, discipline, and tardies dashboards manually in my local environment and paid attention to how they handled "No Homeroom", they looked ✅to me. But if you spot anything, let me know! I really appreciated the way to spotted a bug I missed in #1629.

PR #1636, opened 2018-04-18
Limit dashboard students table height and make scrollable

alexsoble:
# Who is this PR for?

Dashboard users. 

Height of the table is huge, awkward to navigate. Here's what it look like with an amount of demo data equal to the number of students at Healey School:

big-list

# What does this PR do?

Caps the dashboard students table height at 480px and sets overflow-Y to scroll to make a scrollbar appear by default in Internet Explorer. 

# GIF (absences)
absences-fixed-table

# GIF (tardies)
tardies-fixed-table

# GIF (discipline)
discipline-fixed-table

# Checklists

## Javascript QA

+ [x] Author checked latest in IE - Absences Dashboard
+ [ ] Reviewer checked latest in IE - Absences Dashboard

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/StudentsTable.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/StudentsTable.js
@@ -78,8 +78,8 @@ class StudentsTable extends React.Component {
   }
 
   render() {
-    return(
-      <div className= 'StudentsList'>
+    return (
+      <div className='StudentsList' style={style.root}>
         <div style={{display: 'flex', justifyContent: 'space-between'}}>
           {this.renderCaption()}
           <DashResetButton clearSelection={this.props.resetFn} selectedCategory={this.props.selectedCategory}/>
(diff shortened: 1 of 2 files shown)


alexsoble:
@edavidsonsawyer That's a good question! We definitely have a mix of both right now, some styling in CSS and some in JS. I don't know if we've explicitly talked about which direction we want to go here. But I believe most of the new components Kevin and I are shipping are going the JS route. 

Some things I like about styling in the JS . . . it lets me stay in the flow of {edit JS => code re-compiles => see your changes in the browser} without forcing me to switch mental models over to CSS; it keeps all the code in the same file which reduces toggling between files; and it keeps the styles super-localized to each page which reduces the global-scope problems that happen with CSS. 

There's lots \ written about the tradeoffs online, here's one article that talks about some of them: https://blog.jumpcut.com/how-should-you-style-your-react-app-89ced099643. I bet @kevinrobinson has some really good thoughts on this too! 

When the whole team is around I think this could be a good thing to decide on together, but for now I'd say either way is fine but we should try not to mix-and-match in a single file.

PR #1638, opened 2018-04-18
Handle SignalException in import task; don't rely on Delayed::Job for restarts

alexsoble:
# Who is this PR for?
Nightly import job.

Two problems.

First problem: the nightly import job often gets interrupted by Heroku deploys or restarts. This causes it to end. We'd like to gracefully restart. As @kevinrobinson writes in `delayed_job_config.rb`:

  > The failure cases we've seen are running out of memory or a dyno restart from a config change, when Heroku sends a SIGERM and then SIGKILL.  In that case, we want want the job to be retried one more time and then be marked as "failed" and not retried anymore if it fails the second time. 

Second problem: Delayed::Job isn't restarting jobs that fail, even after @kevinrobinson changed the config in #1613.

Quoting @kevinrobinson again: 

> It doesn't look like "destroyed_failed_jobs = false" worked. Logs show this line: Job ImportJob (id=73) REMOVED permanently because of 1 consecutive failures. That's the opposite of what I thought that config line would do :\ So I think we are still in a state where if there's a deploy, the job just gets interrupted and dropped.
> ...
> And in trying to debug this now, I can't seem to get the workers to pick up jobs in the queue. It may be there's some "backoff on failure" config that I can't understand. The README in their repo doesn't describe this though so I got stuck and couldn't debug further because I couldn't get any jobs to run. Before stopping, I cleared the queue again so it was empty when I stopped.

# What does this PR do?
Add code to handle restarting the job ourselves. This logic is very simple to add, whereas the Delayed::Job config doesn't seem to work properly and is very difficult to read and debug. I'd rather skip dealing with Delayed::Job and just roll this ourselves.

# GIF

retry

This shows how interrupting the worker task causes the job to log the signal interruption, re-try exactly once, and then give up if interrupted again leaving now Delayed::Job behind. 

# Next steps

+ Trigger a manual import job and a manual `heroku restart` in production to verify that this works as expected
+ Verify that removing the Delayed::Job config for `destroy_failed_jobs` and `max_attempts` has no effect

### DIFF
--- a/app/importers/import_task.rb
+++ b/app/importers/import_task.rb
@@ -18,7 +18,6 @@ class ImportTask
     begin
       @record = create_import_record
       @report = create_report
-
       log('Starting validation...')
       validate_district_option
       seed_schools_if_needed
@@ -36,10 +35,15 @@ class ImportTask
       log('Done.')
     rescue SignalException => e
       log("Encountered a SignalException!: #{e}")
-      log('Putting a new job into the queue...')
-      Delayed::Job.enqueue ImportJob.new(
-        options: @options.merge({ attempt: @options.attempt + 1 })
-      )
+
+      if (@options.attempt == 0)
+        log('Putting a new job into the queue...')
+        Delayed::Job.enqueue ImportJob.new(
+          options: @options.merge({ attempt: @options.attempt + 1 })
+        )
+      else
+        log('Already re-tried this once, not going to retry again...')
+      end
     end
   end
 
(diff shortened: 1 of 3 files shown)


alexsoble:
selfie 👍

alexsoble:
Initial results from testing with a manual restart on Heroku were . . . fair-to-good? 

The good is that a restart job was successfully queued and started.

The bad is that that second job ended pretty quickly and it's not clear from Logentries or the Import Records page why.

I added some more logging in #1639 and #1640, hoping that these make it easier to see more about the jobs.

PR #1639, opened 2018-04-19
Save more import process logs

alexsoble:
# Who is this PR for?

Developers working on debugging the import process.

Logs aren't being saved and sent to the Import Records page except after a successful import; we want to capture logs even when there's a SIGTERM sent to the job (see #1638).

# What does this PR do?

Saves the logs of an Import Process job every time they're added to (about ~20 times per import task).


alexsoble:
selfie

PR #1641, opened 2018-04-20
Adding range captions to dashboard charts

edavidsonsawyer:
# Who is this PR for?
Dashboard Users

The charts show data within the selected range, but don't communicate that to users

# What does this PR do?
Adds a caption to the chart so it explains the range. See discussion in #1631 

chartcaptions

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/absences_dashboard/SchoolAbsenceDashboard.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/absences_dashboard/SchoolAbsenceDashboard.js
@@ -15,7 +15,8 @@ class SchoolAbsenceDashboard extends React.Component {
     super(props);
     this.state = {
       displayDates: this.props.dateRange,
-      selectedHomeroom: null
+      selectedHomeroom: null,
+      selectedRange: 'School Year'
     };
     this.setStudentList = (highchartsEvent) => {
       this.setState({selectedHomeroom: highchartsEvent.point.category});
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/school_administrator_dashboard/dashboard_components/absences_dashboard/SchoolAbsenceDashboard.js
```diff
@@ -87,7 +88,7 @@ class SchoolAbsenceDashboard extends React.Component {
           seriesData = {filteredAttendanceSeries}
           yAxisMin = {80}
           yAxisMax = {100}
-          titleText = {'Average Attendance By Month'}
+          titleText = {'Average Attendance By Month (' + this.state.selectedRange + ')'}
```
alexsoble: @edavidsonsawyer Could be a good spot for interpolation! 

```
`Average Attendance By Month (${this.state.selectedRange})`
```

edavidsonsawyer: @alexsoble that's slick, let's do that

edavidsonsawyer:
@alexsoble here's a version with the buttons on top. I want to find some way to distinguish them because to my eyes they get lost up there. Maybe some kind of box? 
buttonstop

alexsoble:
@edavidsonsawyer I think a box is a good idea! I agree, my eyes get lost too — maybe moving the text to the upper-left instead of upper-centered would help? My eyes look to the upper-left corner first, probably because English text goes left-right.

edavidsonsawyer:
@alexsoble the only reason I might prefer the upper right here is that it strikes me that users are mostly using this page for the student's table. If that's the main focus then it seems like the filters should stay near it. Though come to think of it, perhaps the table should be on the right and the charts pushed. Thoughts?

alexsoble:
@edavidsonsawyer Yeah I think if the table is the main focus, let's swap the charts and the table!

edavidsonsawyer:
@alexsoble I do like this better. 
buttonstop

alexsoble:
@edavidsonsawyer  👍

PR #1644, opened 2018-04-20
Work on absences dashboard info heirarchy

alexsoble:
# Who is this PR for?

Absences dashboard users.

# What does this PR do?

## GIF 

info-heirarchy

## Main changes

+ Moves the filter buttons above both the charts and the students table to make it visually clearer that clicking the filter buttons will change both the charts and the table
+ Adds a subtitle to the students table showing what time range the # of absences column represents 
+ Purely stylistically, removes some of the border lines to make the page feel less blocky

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashButton.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/DashButton.js
@@ -1,16 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-
 //Custom all purpose dashboard button
 class DashButton extends React.Component {
 
   render() {
+    const {isSelected} = this.props;
     return (
       <div className="DashButton">
         <a
           style={{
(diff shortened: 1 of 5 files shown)


edavidsonsawyer:
@alexsoble nice! I like the way the filter is set up here - it leaves room as well in case we want more options (or fewer, for that matter). The range on the chart is clear and unobtrusive. 

One smaller detail I noticed on IE is that the scrollbar looks like it overlaps the students table. Here's what I mean:

master: 
screen shot 2018-04-20 at 6 11 09 pm

This branch:
screen shot 2018-04-20 at 6 11 40 pm

Not a huge issue by any means! This is still a nice improvement!

edavidsonsawyer:
@alexsoble yeah, no reason to hold this PR. It looks like the table resizing might actually be a side effect of #1636. I might be able to take a look at that issue tonight.

alexsoble:
👍👍

PR #1645, opened 2018-04-21
Fixing table sizing

edavidsonsawyer:
# Who is this PR for?
Dashboard users

Dashboard tables don't resize correctly. cf #1644 
bug

# What does this PR do?
fix


alexsoble:
@edavidsonsawyer Just tested this out locally in IE after rebasing onto master — looks great! 👍👍

PR #1643, opened 2018-04-20
Tardies date range

edavidsonsawyer:
# Who is this PR for?
Dashboard users

# What problem does this PR fix?
The tardies dashboard had a fixed 3 month date range

# What does this PR do?
Gives it the same date range options as absences

# Screenshot (if adding a client-side feature)]
tardies

+ [x] Author checked latest in IE - Tardies Dashboard
+ [ ] Reviewer checked latest in IE - Tardies Dashboard

## Automated Testing

*Does the PR introduce a new feature or new section of code?*

+ [x] Author included specs for new code

PR #1646, opened 2018-04-22
Rebased/Tardies date range

alexsoble:
# Who is this PR for?
Dashboard users

The tardies dashboard had a fixed 3 month date range

# What does this PR do?
Gives it the same date range options as absences

# Screenshot
tardies-rebased

# QA

+ [x] Reviewer checked latest in IE - Tardies Dashboard


alexsoble:
@edavidsonsawyer What do you think of this branch? It rebases #1643 onto master and then adds two commits on top of it: one addresses the JSX structure item from https://github.com/studentinsights/studentinsights/pull/1643#issuecomment-383340897. The other one removes the `schoolYearFlag` prop API from StudentsTable in favor of the incidentSubtitle prop, incidentSubtitle can take any string as a subtitle and shows the string in the "incidents" column instead of in the header.

alexsoble:
@edavidsonsawyer 👍👍 — we've really been looking out for each other's PRs and doing great QA this whole past week! So the feeling is mutual!

PR #1649, opened 2018-04-24
Sorting homerooms by attendance and tardies

edavidsonsawyer:
# Who is this PR for?
Dashboard users

Homerooms aren't ordered in any particular way, but users are interested in seeing which ones have higher/lower attendance and tardies.

# What does this PR do?
Sorts homerooms by attendance and number of tardies.
Absences unsorted:
unsortedabsences

Absences sorted:
sortedabsences

Tardies unsorted:
unsortedtardies

Tardies sorted:
sortedtardies

# Checklists

## Javascript QA

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/tardies_dashboard/SchoolTardiesDashboard.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/tardies_dashboard/SchoolTardiesDashboard.js
@@ -132,7 +132,9 @@ class SchoolTardiesDashboard extends React.Component {
 
   renderHomeroomTardiesChart() {
     const homeroomTardyEvents = this.homeroomTardyEventsSince(this.state.startDate);
-    const homerooms = Object.keys(homeroomTardyEvents);
+    const homerooms = Object.keys(homeroomTardyEvents).sort((a,b) => {
+      return homeroomTardyEvents[b] - homeroomTardyEvents[a];
+    });
     const homeroomSeries = homerooms.map((homeroom) => {
       return homeroomTardyEvents[homeroom];
     });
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/school_administrator_dashboard/dashboard_components/tardies_dashboard/SchoolTardiesDashboard.js
```diff
@@ -132,7 +132,9 @@ class SchoolTardiesDashboard extends React.Component {
 
   renderHomeroomTardiesChart() {
     const homeroomTardyEvents = this.homeroomTardyEventsSince(this.state.startDate);
-    const homerooms = Object.keys(homeroomTardyEvents);
+    const homerooms = Object.keys(homeroomTardyEvents).sort((a,b) => {
+      return homeroomTardyEvents[b] - homeroomTardyEvents[a];
```
alexsoble: @edavidsonsawyer It looks like this is using the reverse of the function that the Absences Dashboard is using, and I see that show up on the graphs when I check this out locally. Do you think it would make sense for both dashboards to sort in the same direction? 
screen shot 2018-04-24 at 1 35 38 pm
screen shot 2018-04-24 at 1 35 12 pm

alexsoble: @edavidsonsawyer Oh, gotcha, that makes sense. I see how you're thinking about this and I think this PR is good to ship. One topic I'm interested in bringing up is if we should harmonize the tables — having the tables show attendance (high == good!) for absences and incidents (high == bad!) seems confusing to me. Clearly it tripped me up in reviewing this PR. I imagine it would be confusing for teachers and principals as well. I think we could gain a lot by using the same charts across the Absences page and the Tardies page. What do you think @edavidsonsawyer? Have you seen any evidence around this when you've been sitting in on meetings @kevinrobinson? We can also move this to an email thread so as not to have it get lost in this PR. 😀

kevinrobinson: Awesome, any sorting will make this better and definitely worth shipping! 🚢 

I agree it'd be good to make things consistent, but absences data is more important than tardies, so let's optimize for shipping improvements to absences first.

PR #1650, opened 2018-04-24
Export more attendance columns

alexsoble:
# Who is this PR for?

Educators looking at attendance events in Insights.

We're missing key data about whether absences/tardies are excused or not, related to medical reasons or not. 

# What does this PR do?

Adds additional columns to our SQL export so that we can start exploring how to add these additional fields into Insights. Changing the SQL export to add fields will send more fields to the SFTP site, but won't automatically send more fields to the Insights database — that will require modifying the importer code.

### DIFF
--- a/x2_export/attendance_export.sql
+++ b/x2_export/attendance_export.sql
@@ -4,6 +4,10 @@ SELECT
   'local_id', -- LASID
   'absence',
   'tardy',
+  'dismissed',
+  'reason',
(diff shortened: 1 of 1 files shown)


alexsoble:
John Breslin says he changed up the SQL, so I expect these new columns of data to be flowing to the SFTP site overnight and be ready to look at tomorrow. 👍

PR #1652, opened 2018-04-25
WIP – Show grade level on students table in Absences/Tardies dashboards

alexsoble:

Educators need to know the grade level of students when using the Absences/Tardies dashboards, because absences mean different things in PK/K than they do in higher level grades.

# What does this PR do?

Attempt to add grade level to the table. Adding another column made the layout cramped and some columns were squeezed up against each other. So I tried a `flex` solution, only to find that it didn't work well in Internet Explorer. I tried moving the styles from JS to CSS thinking that autoprefixer-rails might help (https://github.com/ai/autoprefixer-rails), but it didn't have any effect:

screen shot 2018-04-25 at 3 26 31 pm

Still WIP.

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/StudentsTable.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/StudentsTable.js
@@ -86,11 +86,16 @@ class StudentsTable extends React.Component {
             <DashResetButton clearSelection={this.props.resetFn} selectedCategory={this.props.selectedCategory}/>
           </div>
           <thead style={style.thead}>
-            <tr>
-              <th style={style.th}
+            <tr style={style.tr}>
+              <th style={{...style.th, ...style.name}}
                   onClick={this.onClickHeader.bind(null, 'last_name', 'string')}
                   className={this.headerClassName('last_name')}>
                 Name
+              </th>
+              <th style={style.th}
+                  onClick={this.onClickHeader.bind(null, 'grade', 'date')}
+                  className={this.headerClassName('grade')}>
+                Grade
               </th>
               <th style={style.th}
                   onClick={this.onClickHeader.bind(null, 'events', 'number')}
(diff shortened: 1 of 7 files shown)


alexsoble:
Good suggestion here from @edavidsonsawyer: this table is getting to be a headache, might be a good time to look at react-virtualized's table instead.

alexsoble:
Oooooh, initial results show react-virtualized working fairly well and being a fairly simple integration!

alexsoble:
This last commit adds a basic functional table to the page using react-virtualized. Still very rough around the edges. Need to re-implement sorting, making header and footer match the new table styles, remove the old table code and make the tests pass.

kevinrobinson:
I was just looking at this to see if I could help ship the low-budget version first for this morning, and thought I'd share a bit about the layout problems.  I think what's going on is that using `display: block` for table elements, which I think came up from trying to make the table body scroll, means that we tell the browser we don't want it to use the normal methods for doing layout on the table (eg, figure out how big column widths are are size them to fix contents and distribute extra space, etc).  So then that leads to adding in fixed pixel widths that are specific to the particular contents of the table, and then that means it's more involved to add in new columns.

So I'll leave this is and then we can ship the doubly-improved `react-virtualized` version when it's ready 👍

alexsoble:
@kevinrobinson Yep, that's exactly what happened. 

What time is the SST meeting? I have some non-Insights work to complete today, so I don't think I'll be able to ship this PR quickly. 😕 I also focused on #1653 yesterday instead of this PR, because working towards getting data on which absences are excused is higher priority. 

But if I'm not mistaken, this will be the first time we'll be seeing users use the re-designed Absence/Tardies dashboards in SST, with the new Filter button layout. Is that correct? So I wonder if there's still some valuable user insight to be gained at today's meeting. 

And yep, I agree with you that shipping the react-virtualized table should improve both how we do layout and performance for client-side rendering. 👍

kevinrobinson:
@alexsoble Sounds good!  SST was at 8:45am and I can share feedback from them, I pinged you on chat to see if you have input on the most helpful way to do that, but if you're busy that's cool and I'll just write up some notes  👍 

And this all sounds good - I think a bunch of this is input for me on how to better communicate and coordinate on sharing feedback, pointing out opportunistic time points, etc.  I was probably trying to be overly optimistic about that and should just let things keep making good progress, sorry if the comment this morning felt like over-the-shouldering at all.

kevinrobinson:
@alexsoble I took this up over in https://github.com/studentinsights/studentinsights/pull/1663 to get the last bit finished off.  I couldn't figure out why the layout for `react-virtualized` didn't work; but adding `{display: 'flex'}` into each row made it lay out like it should so I comment on it and went with  that for now.

PR #1653, opened 2018-04-26
Import more fields for attendance events 

alexsoble:
# Who is this PR for?

Educators who rely on Student Insights for analysis and visualizations about student attendance. 

Some attendance events are excused; Insights doesn't surface this right now, which makes educators need to turn to Aspen for more detailed data.

# What does this PR do?

Builds on #1650 by pulling in the new columns to our database. 

I tested what this would look like in production by running an import process into my local machine using this branch. Here's what I found (summary table above, calculations below):

### Summary table

field | Absences | Tardies
---|---|---
dismissed == true | 0.66% | 1.97%
excused == true | 6.27% | 2.78%
reason present? | 29.6% | 10.0%
comment present? | 18.3% | 4.4% 

### Exact queries + results

```
# Overall counts
2.3.0 :002 > Absence.count
 => 226182
2.3.0 :003 > Tardy.count
 => 188479

# Absences: dismissed 
2.3.0 :005 > Absence.where(dismissed: true).count
 => 1492
fraction of total: 0.006596457719889292 => 0.66%
2.3.0 :006 > Absence.where(dismissed: false).count
 => 224690
fraction of total: 0.9934035422801107 => 99.34%

# Absences: excused 
2.3.0 :007 > Absence.where(excused: true).count
 => 14185
fraction of total: 0.06271498174036838 => 6.27%
2.3.0 :008 > Absence.where(excused: false).count
 => 211997
fraction of total: 0.9372850182596316 => 93.73%

# Absences: reason
2.3.0 :049 > Absence.where(reason: ["", nil]).count
 => 159325
 fraction of total: 0.7044106073869716 => 70.4%
2.3.0 :050 > Absence.where.not(reason: ["", nil]).count
 => 66857
 fraction of total: 0.29558939261302847 => 29.6%

# Absences: comment
2.3.0 :052 > Absence.where(comment: ["", nil]).count
 => 184833
 fraction of total: 0.8171870440618617 => 81.7%
2.3.0 :053 > Absence.where.not(comment: ["", nil]).count
 => 41349
 fraction of total: 0.18281295593813832 => 18.3%

# Tardies: dismissed
2.3.0 :012 > Tardy.where(dismissed: true).count
 => 3713
fraction of total: 0.019699807405599564 => 1.97%
2.3.0 :013 > Tardy.where(dismissed: false).count
 => 184766
fraction of total: 0.9803001925944005 => 98.03%

# Tardies: excused
2.3.0 :015 > Tardy.where(excused: true).count
 => 5240
fraction of total: 0.027801505738039783 => 2.78%
2.3.0 :016 > Tardy.where(excused: false).count
 => 183239
fraction of total: 0.9721984942619603 => 97.2%

# Tardies: reason
2.3.0 :062 > Tardy.where(reason: ["", nil]).count
 => 169716
fraction of total: 0.9004504480605267 => 90.0%
2.3.0 :063 > Tardy.where.not(reason: ["", nil]).count
 => 18763
fraction of total: 0.09954955193947336 => 10.0%

# Tardies: comment
2.3.0 :064 > Tardy.where(comment: ["", nil]).count
 => 180251
fraction of total: 0.9563452692342383 => 95.6%
2.3.0 :065 > Tardy.where.not(comment: ["", nil]).count
 => 8228
fraction of total: 0.0436547307657617 => 4.4%
```

# Question

One thing I don't know: what does "dismissed" mean in this context? @kevinrobinson @edavidsonsawyer do you know this one? If not, maybe we can ask Uri tomorrow. Since it's the least frequent of all the new columns, I'm wondering if it's used by all Somerville schools. Or maybe there's another reason it has such a low frequency. 

# "Reason" categories 

There are a fixed number of categories used in the "reason" column for absences and tardies, in contrast with "comment" which is free-form.

```
=> puts Absence.pluck(:reason).uniq.compact.sort

Appointment
Bereavement
Holiday
Late
Legal
Medical
No Reason
Nurse
Parent
School
Sick
Suspended
Transportation
Truant
Vacation
Walked Out
Weather
Withdrawal
```


spec/fixtures/fake_attendance_export.txt
```diff
@@ -1,7 +0,0 @@
-"local_id","absence","tardy","event_date"
```
alexsoble: Removed this fixture because it isn't being used anywhere in the specs.

spec/importers/file_importers/attendance_importer_spec.rb
```diff
@@ -63,6 +73,33 @@
         end
       end
 
+      context 'row with absence (district does not have dismissed/reason/excused/comment fields)' do
```
alexsoble: I don't want to make adding these new data columns / features dependent on New Bedford. My guess is that New Bedford also has fields that map to dismissed/reason/excused/comment, but they might not, or their data structure might be different. 

Either way, we can't control how quickly or slowly we'd be able to get those columns exported for New Bedford. So I want to make sure that changing the database structure and import code won't break anything on the New Bedford side even if their incoming CSV rows have a different shape. The net effect of this change for New Bedford should be that their absences/tardies tables each have four additional empty columns.

db/migrate/20180426135126_add_new_fields_to_attendance_events.rb
```diff
@@ -0,0 +1,13 @@
+class AddNewFieldsToAttendanceEvents < ActiveRecord::Migration[5.1]
+  def change
+    add_column :absences, :dismissed, :boolean
```
alexsoble: I'm still thinking through whether this is the right shape or not — more information/data to come. 😄

app/importers/rows/attendance_row.rb
```diff
@@ -33,6 +35,13 @@ def build
       student_id: student.try(:id),
     )
 
+    attendance_event.assign_attributes(
```
kevinrobinson: One thing that might be good is to add a `PerDistrict` method to indicate whether we expect to be finding all these fields, since that seems different across districts.  That way it's clear in the code that "this is intentionally only for Somerville right now" rather than not being able to tell until looking at the data.

app/importers/rows/attendance_row.rb
```diff
@@ -16,6 +16,8 @@ def save; end
       def valid?; false end
 
       def errors; NullEventErrors.new end
+
+      def assign_attributes(attributes); end
```
kevinrobinson: This seems leftover, or am I missing what this is doing?

spec/importers/file_importers/attendance_importer_spec.rb
kevinrobinson: This might be a good thing to add as a method in`PerDistrict`, since we want to intentionally treat this differently across districts.

db/migrate/20180426135126_add_new_fields_to_attendance_events.rb
```diff
@@ -0,0 +1,13 @@
...
+    add_column :absences, :excused, :boolean
+    add_column :absences, :reason, :string
+    add_column :absences, :comment, :string
+
+    add_column :tardies, :dismissed, :boolean
+    add_column :tardies, :excused, :boolean
+    add_column :tardies, :reason, :string
+    add_column :tardies, :comment, :string
```
kevinrobinson: I'm assuming that by default these columns will be serialized and sent down to the UI when we ship this, is that your expectation too?  I don't think it matters a whole lot, but could increase the size a bit for the absences and tardies dashboard if they aren't already strict about limiting the fields they send down to the UI.

alexsoble: Yes, the columns will be serialized and sent to the UI for the Absences and Tardies dashboards, though both those queries are scoped to absences and tardies for the past year. Will also add more data to the UI for the Student Profile. Won't affect the School Overview page since that uses `most_recent_school_year_absences_count`. The Home Page for educators doesn't sent serialized absences to the UI. 

I think we should keep an eye out for how much of an effect this has on performance for the Absences/Tardies page and the student profile.

app/importers/rows/attendance_row.rb
alexsoble: This is because this PR uses `assign_attributes` in AttendanceRow. My mental model is that "reason" and "comment" are fields that might get updated after the fact. I don't want to treat absences for the same student on the same day with different reasons or comments as different events. NullRelation needs to have all of the same method names as AttendanceRow to follow the null object pattern. This is how we handle events that aren't Absences or Tardies through duck typing.

alexsoble:
@kevinrobinson Updated to use `PerDistrict` in the import code / specs

db/migrate/20180426135126_add_new_fields_to_attendance_events.rb
kevinrobinson: Cool, that makes sense!

Looking at the shape of the data here, one thing I notice is that this adds several string columns, and based on your awesome digging up above, these are present for ~30% of the absence events.  For the HS absences page, right now we're sending down ~10k absence events.  So if we take 30% of those and double or triple the size of the event object we send down, it seems like there's a good chance this would have a noticeable effect.  It'd be a bummer to ship this and then have the awesome performance improvement gains that you made last week regress.

So two suggestions might be: ship this and then measure the impact on the payload size for the absences page and page load time, or update `SchoolsController#absence_dashboard_data` first to whitelist only the fields we need in the UI.  Then this PR is less likely to have an impact on the absences view.

app/importers/rows/attendance_row.rb
kevinrobinson: Ah, thanks!  That's super helpful. 👍 I hear what you're saying describing how `NullRelation` works, and if I look back at [attendance_importer.rb](https://github.com/studentinsights/studentinsights/blob/68252fea01b64e8110f72ea73d9f5a648e0d83c8/app/importers/file_importers/attendance_importer.rb#L59) I see the other piece of what you're saying.

This is probably a discussion for a separate time, but I'll  think using the null object pattern in this situation is tricky, since AR models have a huge interface that you'd have to recreate.  Reading how the `NullRelation` gets created here, it looks like it's used to signal if the row isn't flagged as an absence or a tardy  I'm wondering if there are other ways to approach this, maybe having the `#build` method return exit early and return `nil` to indicate that case.  Or to have the `#build` method return either or AR object or an error object, and let the caller decide what to do rather than trying to hide that from the caller altogether.

And to be explicit, I don't think this is something we should tackle in this PR, just sharing as it comes up and maybe it's something we do later separately.

alexsoble: @kevinrobinson Yeah, I agree with that analysis, I don't think null object pattern makes the most sense here either.

db/migrate/20180426135126_add_new_fields_to_attendance_events.rb
alexsoble: I like the whitelisting strategy! I think that also gives us the opportunity to ship some exploratory features around this dark, where we can surface fields like `reason` and `excused` and get an understanding of how educators are interpreting what they see.

kevinrobinson: @alexsoble Sounds good!

alexsoble:
In a5bedf6332cd1faed27f4d63be513f39976cc303 I limited the fields we're sending down the Absences and Tardies dashboard as per @kevinrobinson's sharp suggestion here: https://github.com/studentinsights/studentinsights/pull/1653#discussion_r184662534

I did some measuring in my local environment for the queries loading the Absences dashboard with `ENV['MORE_DEMO_STUDENTS'] = true`. Here's the impact I found on server side performance. There are probably also positive impacts on client side performance here too but I didn't measure them.

**Before using `select` to limit fields**:

```
[0] Completed 200 OK in 1117ms (Views: 577.5ms | ActiveRecord: 54.1ms)
[0] Completed 200 OK in 1192ms (Views: 593.5ms | ActiveRecord: 67.0ms)
[0] Completed 200 OK in 1237ms (Views: 591.4ms | ActiveRecord: 47.6ms)
[0] Completed 200 OK in 1113ms (Views: 590.8ms | ActiveRecord: 46.7ms)
[0] Completed 200 OK in 1068ms (Views: 568.1ms | ActiveRecord: 42.3ms)

Average load time: 1145ms
```

**After using `select` to limit fields**:

```
[0] Completed 200 OK in 1384ms (Views: 471.7ms | ActiveRecord: 88.7ms)
[0] Completed 200 OK in 981ms (Views: 391.0ms | ActiveRecord: 43.3ms)
[0] Completed 200 OK in 941ms (Views: 397.3ms | ActiveRecord: 38.7ms)
[0] Completed 200 OK in 1288ms (Views: 520.5ms | ActiveRecord: 42.8ms)
[0] Completed 200 OK in 847ms (Views: 374.6ms | ActiveRecord: 41.3ms)

Average load time: 1088ms
```

This shows a ~5% improvement in the server side performance from limiting the fields we're sending down; it seems reasonable to expect a similar impact on the client side performance too.

alexsoble:
@edavidsonsawyer Take a look at the change I made to `dashboard_tardies` and `dashboard_absences` in a5bedf6332cd1faed27f4d63be513f39976cc303, these limit the columns we're sending down to just `student_id` and `occurred_at`. 

This means `created_at` and `updated_at` will be filtered out since we don't use them in the Dashboard view, because they're referring to when the records were created as opposed to when the attendance event occurred.

My assumption was that this would have no impact on the Absence/Tardies dashboard functionality besides speeding it up a bit, and that seemed to be the case when I tested it out locally. Wanted to make sure you got a chance to take a  👀 since you've written so much of the Absence/Tardy dashboard code.

Also @edavidsonsawyer, after this PR merges, you can run`rake db:reset` to re-seed the data and get fake but representative values for the new attendance event columns from 1fd830122627d2b30fb936a070df59434538c8db.

alexsoble:
@kevinrobinson Hmmmm, `select` isn't causing any bugs in this PR and is improving performance, so I think it should be fine to keep it here. We're using `select` in `StudentsQueryHelper` right now and it hasn't introduced any bugs.

kevinrobinson:
@alexsoble 👍 cool it's just my Rails brain that's buggy then :)

kevinrobinson:
@alexsoble Awesome!

Yeah so my understanding of `select` is that it applies at the level of the AR query and model.  It adds to the "SELECT" clause in the SQL that queries the database, and says "only get me the student_id and occurred_at field from the database."  So then those AR models that you get back literally only have those fields in them.  They don't have id fields or anything else like reason, etc. because `select` tells AR not to fetch them from the database.  I found that led to me making a few different mistakes, because most model methods won't work as you expect then (eg, other associations won't work, methods that read other fields on the model will raise).

It does end up influencing what happens when you later call `as_json`, since those fields aren't on the model they aren't serialized.  But if you want to limit the serialization then the more direct way is to pass `:only` when calling `as_json`.  So that's what I ended up doing, since reducing the bytes sent to the client reduced response time more than reducing the fields read from the db (since the db still has to read all the same rows anyway).  So one of my takeaways from that was `as_json(:only => ... )` was a simpler way to optimize the bytes sent to the client, and that might be a better place to start than bytes read from the db.

I was thinking of this more as "limit the data sent to the browser" than "limit the data read from the db."  But it might be different here, or you might have found other approaches that make more sense for this situation!

alexsoble:
@kevinrobinson, Alrighty subbed in `as_json` instead of `select` in 1ed017825cf253194ef1d62a3c16e6f6859eb0ae. I reviewed the absences and tardies pages manually to confirm that they are behaving just like before. @kevinrobinson, can you do the same QA and then flash a 👍 if this PR seems good to merge and deploy?

app/demo_data/fake_student.rb
```diff
@@ -212,6 +239,22 @@ def add_attendance_events
       attendance_event.occurred_at = occurred_at
       attendance_event.student = student
 
+      # Frequencies for attendance event reasons/comments/dismissed/excused
+      # modeled on Somerville data, for queries see:
+      # https://github.com/studentinsights/studentinsights/pull/1653#issue-184358100
```
kevinrobinson: awesomeness!

alexsoble: 👍

alexsoble:
@kevinrobinson 👍

PR #1663, opened 2018-05-01
Show grade level on students table in Absences/Tardies/Discipline dashboards

kevinrobinson:
Add-on to finish off @alexsoble's awesome work in https://github.com/studentinsights/studentinsights/pull/1652.

# Who is this PR for?
Educators need to know the grade level of students when using the Absences/Tardies dashboards, because absences mean different things in PK/K than they do in higher level grades.

Adds grade to absences, tardies and discipline incident table.  Swaps the table rendering with `react-virtualized`.  Cuts out sending fields we don't need to the client, reducing server response size.

# What does this PR do?
Locally, adding the serialization code here cuts down the response size by a lot.  The numbers here are gzipped sizes and ungipped sizes from Web Inspector, before and after this change:

#### /absences
before: 24.4 / 220kb
after: 16.5 / 128kb

#### /discipline
62.3 / 800kb
16.5 / 128kb

screen shot 2018-05-01 at 8 55 52 am
screen shot 2018-05-01 at 9 32 24 am
screen shot 2018-05-01 at 9 32 02 am

# Checklists
## Javascript QA
+ [x] Author checked latest in IE - /absences
+ [x] Author checked latest in IE - /tardies
+ [x] Author checked latest in IE - /discipline

## Automated Testing


app/controllers/schools_controller.rb
```diff
@@ -166,40 +166,34 @@ def authorized_students_for_overview(school)
   end
 
   # Methods for Dashboards #
+  def shared_student_fields(student)
```
kevinrobinson: @edavidsonsawyer For endpoints like this, where we're sending down a lot of events, using `as_json(only: [:foo, :bar])` to only serialize what the UI needs can save a lot of bytes!

kevinrobinson:
selfie cc @alexsoble @edavidsonsawyer

kevinrobinson:
This looks good.  In terms of performance, this cuts some of the response sizes down to as little as 30% of what they were previously (for HS especially).  This will help a lot for within-school internet speeds.  Response times are still as long as 20sec for the HS /discipline page on fast internet, which is from blocking on the query not transferring the data.

edavidsonsawyer:
@kevinrobinson I had no idea the number of incidents was going to be that high! Maybe we should build in a year range filter for the discipline incidents as we do for absences/tardies

edavidsonsawyer:
@kevinrobinson oh good, was just about to open a PR for this! Let me know what it looks like in production after this - I thought discipline numbers might be small enough that we don't need the hacky restrictions on the model for discipline, so if it's an issue I have a PR ready to go.

kevinrobinson:
@edavidsonsawyer Sure, https://github.com/studentinsights/studentinsights/pull/1668/files fixes the actual bug.

As for the data, are the hacky restrictions about adding `dashboard_discipline_events` to the model class itself?  As the code is now we can just add that `where` into the controller here: https://github.com/studentinsights/studentinsights/pull/1668/files#diff-7e285c6c367989eec0389d9fbe15e2ecR195.

edavidsonsawyer:
@kevinrobinson that sounds right. @alexsoble and I both tried that at different points and ran into some trouble. See, eg. https://github.com/studentinsights/studentinsights/pull/1629#discussion_r181898381

The problem seems to be that there isn't an obvious way to eager load only events within a specific time range.

kevinrobinson:
Gotcha.  So if I follow, there's two separate things: one is reducing the query for all the data, and the other is reducing the data sent down to the UI.

For reducing the data queried, the 'conditions' section in these docs might be helpful: https://apidock.com/rails/ActiveRecord/QueryMethods/includes, it seems like maybe we can put that in the controller code?

In https://github.com/studentinsights/studentinsights/pull/1663#issuecomment-385938969 I was thinking more about reducing the data we sent down to the client, but removing it from the query is even better!

PR #1664, opened 2018-05-01
Classrooms: Alpha version of Classroom list creator

kevinrobinson:
# Who is this PR for?
Teachers and principals grades KF-5 as they assign classrooms for the upcoming year.  Specifically, this is for the first pilot tests.

# What problem does this PR fix?
This feature is a tool to help teachers and principals create class lists and visualize some information about them in the process.  This adds a new `/balancing` URL for teachers to use this tool.

# What does this PR do?
The UI starts up in `ClassroomListCreatorPage` which tracks state changes and requests.  `ClassroomListCreatorWorkflow` renders the different steps of the process and `CreateYourClassroomsView` is the UI component for the main dragging-dropping and summarizing interaction.

Added dependencies:
- multi-select: `react-select` (requires `style-loader` and `css-loader` updates to Webpack config)
- dragging and dropping: `react-beautiful-dnd` (and `array#find polyfill` which it needs)
- percentile computations: `percentile`

Add components:
- `<Loading />` factored out from `<GenericLoader />` for other uses
- `<BoxAndWhisker />` for a small box-and-whisker plot
- `<Button />` factored out from ".btn" className
- `<Hover />` for rendering on hover
- In Ruby, `GradeLevels` does math with going forward or backward a year in grade levels

Updated user-facing routes:
- Rename `/equity/principal` to `/equity/explore` clarify
- Add `/balancing` and `/balancing/:id`

New endpoints:
- `#available_grade_levels_json` > what schools and grade levels can the user create?
- `#students_for_grade_level_next_year_json` > what's the list of students for that grade?
- `#update_classrooms_for_grade_json` > for recording changes while the user works

Changes to existing code:
- Change `<Bar />` API to take standard style instead of styles, update call sites.  Add story.
- Rename `<Breakdown />` to `<ExploreClassroomComparisons />`
- Add `overflow-x: hidden` on `index.html` to remove scrollbars in IE, add horizontal scroll on `/equity/explore` page so that's still supported.

# Screenshot (if adding a client-side feature)
### Choose your grade
screen shot 2018-05-01 at 9 54 54 am

### Make a plan
screen shot 2018-05-01 at 9 55 12 am

### Create your classrooms
screen shot 2018-05-01 at 9 55 48 am

#### Clicking on a student
(placeholder for now)
screen shot 2018-05-01 at 9 55 52 am

### Notes to principal
screen shot 2018-05-01 at 9 56 10 am

### Submit
screen shot 2018-05-01 at 9 57 49 am

#### react-select
[...]

PR #1665, opened 2018-05-01
Fix bug in mapping studentIdsByRoom to students

kevinrobinson:
Fixing bug in https://github.com/studentinsights/studentinsights/pull/1664

# Who is this PR for?
K1-6 teachers.

JS bug led to unexpected movement after letting go of the drag.

# Checklists


kevinrobinson:
selfie

PR #1669, opened 2018-05-03
Reducing scope of discipline incidents to one year

edavidsonsawyer:
# Who is this PR for?
Discipline dashboard users

The discipline dashboard has some slow load times.

# What does this PR do?
Reduces the discipline incidents handled by the dashboard to those within the last year, in keeping with the absences/tardies dashboard. 

### DIFF
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -49,7 +49,8 @@ class SchoolsController < ApplicationController
 
   def discipline_dashboard_data
     student_discipline_data = active_students(@school)
-      .includes([homeroom: :educator], :dashboard_discipline_incidents, :event_notes)
+      .includes([homeroom: :educator], :discipline_incidents, :event_notes)
+      .where('occurred_at >= ?', 1.year.ago).references(:discipline_incidents)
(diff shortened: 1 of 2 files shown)


edavidsonsawyer:
@kevinrobinson I just realized after our discussion in #1663 that you're right and there's no problem scoping the query in the controller here. It _is_ a problem with absences because that component is expecting all students whether or not they have absences within the timeframe, but that isn't the case here.

kevinrobinson:
@edavidsonsawyer Sounds good! 👍

kevinrobinson:
@edavidsonsawyer Doing just rough spot checks, this cuts down the SHS server response time from ~20sec to ~8sec, and puts K8 schools <5sec.  It probably meaningful helps the JS parse and render too, since I'd guess the change reduces the amount of data sent to probably 30-50% of what it was before.

Relatedly, in talking with @snoopyuri the other day, this long load time is generally okay for now; there's value for folks in being able to see data in this way at all, so they're generally super tolerant of waiting for this to load (versus say the profile page or home page).  As we iterate on this, we might work on this more and and measure it more rigorously, but for now it's better to focus on UX improvements and then some other product improvements than perf.

Issue #1660, opened 2018-04-27
Discipline: UX quality improvements


kevinrobinson:
From some feedback this morning, here's some small quality improvements to make:

- [x] Add ExperimentalBanner (https://github.com/studentinsights/studentinsights/pull/1670)
- [x] Page doesn't fit on a 1024x768 projector screen (https://github.com/studentinsights/studentinsights/pull/1670)
- [x] Filter dropdown doesn't show (maybe because page is zoomed to fit on screen).  Consider changing to react-select for dropdown (https://github.com/studentinsights/studentinsights/pull/1670)
- [x] Sort classrooms by count of incidents (https://github.com/studentinsights/studentinsights/pull/1670)
- [x] Update overall layout to match absences (eg, time filter up top, table on left) (https://github.com/studentinsights/studentinsights/pull/1670)
- [x] Speed up load time by only loading current school year's data (https://github.com/studentinsights/studentinsights/pull/1669)

EDIT:
- [x] Fix bug with clearing selection (https://github.com/studentinsights/studentinsights/issues/1692, fixed in https://github.com/studentinsights/studentinsights/pull/1707)
- [x] Sort locations by count of incidents (https://github.com/studentinsights/studentinsights/pull/1711)
- [x]  Student list doesn't match selection (eg, classroom) (https://github.com/studentinsights/studentinsights/issues/1708)

kevinrobinson:
@edavidsonsawyer awesome, 🚢ing https://github.com/studentinsights/studentinsights/issues/1708 was the last one 🎉 

If you want to chat more on some next things to work on, feel free to ping me or anyone else whenever you're back from your trip!

PR #1670, opened 2018-05-03
Discipline ux improvements

edavidsonsawyer:
# Who is this PR for?
Dashboard users cf #1660 

# What does this PR do?
Brings discipline dashboard in line with other dashboard layouts and adds some miscellaneous UX improvements:
improves dropdown,
updates layout,
sorts classrooms

uximprovements

# Checklists


edavidsonsawyer:
Hey @kevinrobinson sorry for the delay on this. One thing I didn't directly address here was the resolution problem. The screenshot above is at 1024x768 - as long as the window is maximized, everything seems to adjust appropriately. But because of that I also didn't add the experimental banner since it takes up real estate on the top which seems to be a problem.

edavidsonsawyer:
Er, to clarify, I'm not saying we shouldn't add the experimental banner, just that we should get a better idea why we're having problems fitting all the items onscreen first.

kevinrobinson:
@edavidsonsawyer I think we chatted about all this just now, but let me know if there's anything else I can help out with or whenever you want a review. 👍

edavidsonsawyer:
@kevinrobinson so I didn't do anything to explicitly fix the resolution issue, but after I merged in master, I can't reproduce the horizontal scroll problem in 1024x768. I think this one is ready for a look.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -90,27 +93,46 @@ class SchoolDisciplineDashboard extends React.Component {
 
   }
 
+  sortedClassrooms(chartKeys) {
+    if(this.state.selectedChart == 'classroom') { //should always be true when this method is called
```
kevinrobinson: If this is always supposed to happen, could this method always call `this.getChartData('classroom')` and then remove the if/else and comments?

edavidsonsawyer: @kevinrobinson I'm not wild about that because the chartkeys have to be the keys from the classroom and there's nothing to check that's the case. 

But looking at this again I'm not sure why I thought this check was necessary. This function isn't called unless the selectedChart is classroom in the switch statement above, so I'll just remove this.

edavidsonsawyer:
@kevinrobinson It's easy enough to make the ranged match, but I held off here only because @snoopyuri had mentioned someone playing around with the date ranges for discipline and I thought it might be the kind of data that could be interesting with arbitrary ranges. But he might also have been trying to find the specific dates that matter. 

What do you think of leaving this as-is and seeing what users say? Though if you've already got feedback that it's easier when they all share the same fixed ranges, that's good enough!

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
kevinrobinson: 👍 sounds good!

Yeah, I see what you're saying that you can't just pass any set of `chartKeys` to this method in any component state.  One thing to clarify might be naming the arguments for these methods more explicitly (and then `sortChartKeys` does the work of taking generic `chartKeys` and calling the right method).  Or i you want a stricter way to enforce this, you can move these sort methods to functions at the bottom of the file, outside the component.  Then you need to explicitly pass arguments, which would also show that `sortedClassrooms` needs all the chart data as well, in a way that the others don't.

kevinrobinson:
@edavidsonsawyer I think we should suggest time ranges, for the same kinds of reasons as in https://github.com/studentinsights/studentinsights/pull/1631.  In the observations I've seen, when people are fiddling with the date range it's because they want to see some notion of "recent" like a month or a quarter (eg 45 days).

We could separately add another button to the time range buttons to allow folks to pick an arbitrary date range (that say pops up the sliders).  I haven't heard that come up as a real need yet but it might as people are looking at this more!

edavidsonsawyer:
@kevinrobinson sounds good! Here's the range button version, and we'll table the slider unless someone really want it. 

Incidentally, if the slider can be removed from the app entirely, then we might be able to rewrite the logic in the absence dashboard to be a little more efficient. It may be that some of the slowness comes of needing to recalculate averages from an arbitrary date range. Something to consider if we get to that point in load times later.

kevinrobinson:
@edavidsonsawyer 👍 Want to add in the banner from https://github.com/studentinsights/studentinsights/issues/1660 too and then I can show folks this tomorrow morning?

edavidsonsawyer:
@kevinrobinson I missed that last comment. Should be all set!

kevinrobinson:
@edavidsonsawyer awesome! 🚢 👍

kevinrobinson:
@edavidsonsawyer This looks great!  I updated https://github.com/studentinsights/studentinsights/issues/1660 and added one small more improvement there, to do a similar sorting by count for "location" as you awesomely added for "classroom."  After that we can close out that issue and move on to what's next!

Issue #1659, opened 2018-04-27
Classrooms: Ship beta for pilot sessions


kevinrobinson:
Ship a beta version of the teacher classroom list creator for the pilot sessions.  See https://github.com/studentinsights/studentinsights/pull/1664 for alpha issue with more background.

Major pieces:
- [x] improvements to slicing: add DIBELS graph, change to "Limited" English, add discipline events > 3 (https://github.com/studentinsights/studentinsights/pull/1666)
- [x] saving work to the server
- [x] inline student profile (https://github.com/studentinsights/studentinsights/pull/1671)

Polish and small touches:
- [x] sort cards alphabetically initially (https://github.com/studentinsights/studentinsights/pull/1677)
- [x] simplify state management across steps, forward/back buttons depend on each page (https://github.com/studentinsights/studentinsights/pull/1678)
- [x] fix float moredots styling for long names (https://github.com/studentinsights/studentinsights/pull/1666)

(there are several more bits here, as part of responding to pilot sessions)

kevinrobinson:
Going to consider this closed.  We did three pilot sessions this week and will do one more next week.  The next step is https://github.com/studentinsights/studentinsights/issues/1695 and final polish items are tracked in https://github.com/studentinsights/studentinsights/issues/1696.

PR #1671, opened 2018-05-03
Classrooms: Beta improvements to server-side code, iterations to inline student profile

kevinrobinson:
# Who is this PR for?
Teachers and principals grades KF-5 as they assign classrooms for the upcoming year. Specifically, this is for the first pilot tests. https://github.com/studentinsights/studentinsights/issues/1659.

Updates the placeholder student profile in the classroom list creator page, including show the feed for that student.  Refactors and adds better test coverage to all server-side code.

# What does this PR do?
Factors out `InlineStudentProfile` and make some minimal UI iterations to improve how clearly and quickly it reads to folks.  We can do more iterations here, and use this as a design playground for what a redesigned profile might look (eg centering student voice), but this PR doesn't do much more yet.

Adds the feed to the `InlineStudentProfile` page, which involved factoring it out from the home page.  This involved some changes to the server-side code, pushing the "all cards" method out of the the home controller and into `Feed`, and adding a new `Feed` method specifying a particular set of students to meet the use case here.  On the UI side, this PR moves the feed components into a `feed` folder that's intended to be re-used in several places in the next few months (eg, grade-level page, classroom feed, house page, student profile).

This PR also does a refactoring pass on server code.  It adds foreign key constraints and validations to the model with a task to clean up earlier development data.  It also splits out the queries and authorization code from the controller in anticipation of re-using these when building principal access.  This also adds more test cases in particular around the authorization rules, all which got another pass for quality and accuracy in this PR.  In particular, this PR factors out the code in the home page allowing developers to view as different users, and moves it to `authorizer#authorized_when_viewing_as` along with another layer of authorization checks and set of tests.

Finally, in updated JS tests and stories, this PR updates the storybook config to mock out Jest functions, allowing us to import test setup functions from test files into story files.

#### Inline profile (early beta quality design)
screen shot 2018-05-03 at 3 57 58 pm

#### Feed in profile
screen shot 2018-05-03 at 3 41 51 pm

# Checklists
+ [x] Author checked latest in IE - Classroom list creator

## Automated Testing


kevinrobinson:
selfie, will catch up and do some post-facto reviews after vacation time (and still ahead of general release)

Issue #1656, opened 2018-04-27
Absences: Handle absences differently for PreK students


kevinrobinson:
Absences for kids below kindergarten are a little bit different.  School isn't mandatory at that age, and so the absence rates are a lot higher, but everyone is generally less concerned about absences.

- [x] Remove kids in grades lower than Kindergarten from the "high absences" box on the home page (https://github.com/studentinsights/studentinsights/pull/1672)
- [x] Show grade level in the dashboard table or let folks filter out PreK students (https://github.com/studentinsights/studentinsights/pull/1652 and https://github.com/studentinsights/studentinsights/pull/1663)

PR #1672, opened 2018-05-03
Update insights box for high absences to exclude kids younger than KF

kevinrobinson:
# Who is this PR for?
K8 educators, redirect counselors, APs and principals in particular.

See https://github.com/studentinsights/studentinsights/issues/1656.  Also updates to query only for active students.

# What does this PR do?
Filters out students that are considered for the "high absences" box.  It also moves creating absences out of `TestPals` and into controller specs.

#### before (Snow is PK)
screen shot 2018-05-03 at 5 10 29 pm

#### after
screen shot 2018-05-03 at 5 10 43 pm


kevinrobinson:
selfie

PR #1673, opened 2018-05-03
Update queries for Student.all to Student.active

kevinrobinson:
# Who is this PR for?
All educators

Some queries are for all students instead of just active.  This isn't causing any problems in practice, but I noticed and this fixes it up.


kevinrobinson:
selfie

kevinrobinson:
This impacted the high absences box quite a bit.

PR #1678, opened 2018-05-07
Classrooms: Load previous own workspaces, freer navigation, rename classes

kevinrobinson:
# Who is this PR for?
K1-6 educators, part of https://github.com/studentinsights/studentinsights/issues/1659.

This allows users to load their own previous sessions, removes some navigation restrictions with the stepper, and renames classes to be consistent now that the feature name has settled.

# Checklists

This could use another refactoring to simplify how  `triggerEffects` and fetching works, and to add better tests.  That's probably for later this week.


kevinrobinson:
selfie

PR #1682, opened 2018-05-09
Classrooms: Increase font size for table

kevinrobinson:
# Who is this PR for?
K1-6 educator teams, part of https://github.com/studentinsights/studentinsights/issues/1659.

Small fonts on large projectors with higher pixel ratios.

# What does this PR do?
Also, on load changes the loading so that it ignores last screen step and just loads the first screen.

# Screenshot (if adding a client-side feat
screen shot 2018-05-09 at 8 09 50 am


kevinrobinson:
selfie

PR #1683, opened 2018-05-10
Classrooms: View class lists

kevinrobinson:
# Who is this PR for?
K1-6 educator teams, especially school leaders.  Part of https://github.com/studentinsights/studentinsights/issues/1659.

# What problem does this PR fix?
Lets teacher teams view their own class lists.  Also lets principals and other admin view these lists, which is a step towards principal write-access.

# What does this PR do?
Adds a new page to `/classrooms` that lists class list workspaces and includes a button to create a new class list workspace.  Also factors out the introduction copy about the process to `<IntroCopy />` and renames the story grouping to `classlists`.

# Screenshot (if adding a client-side feature)
screen shot 2018-05-10 at 12 29 08 pm

# Checklists
+ [x] Author checked latest in IE - Class lists
+ [x] Author improved specs for code in need of better test coverage
+ [x] Author included specs for new code

PR #1684, opened 2018-05-10
Factor out Timestamp component from feed cards

kevinrobinson:
# Who is this PR for?
developers

In working on https://github.com/studentinsights/studentinsights/pull/1683, I initially wanted to use the same timestamp rendering (days ago plus the date) so factored this out.  I ended up removing that from https://github.com/studentinsights/studentinsights/pull/1683, but this seems useful to keep anyway since it's duplicated in two feed components now.

# What does this PR do?
Factors out `<Timestamp />` from two feed cards.

# Checklists


kevinrobinson:
selfie

PR #1686, opened 2018-05-11
Classrooms: support final balancing phase with column selections and highlights

kevinrobinson:
# Who is this PR for?
K1-6 educator teams, part of https://github.com/studentinsights/studentinsights/pull/1683.

In initial pilot runs, after folks create their classroom lists using their own process, they have a hard time figuring out how to revise class lists if they notice imbalances between classrooms (eg, more students with IEP in one classroom, higher STAR Math scores in one particular classroom).  This helps support them in that final part of the process.

This also avoids concerns with putting markers on student cards that would stay with them always - now we only "label" kids in these ways when teachers want to see this in the final phase of the process.  They still approach planning and do the initial placement process starting with student strengths or whatever else they choose (with all these various markers hidden and not influencing the process until the end).

# What does this PR do?
Adds new interactions to the tables at top, where clicking on a column selects that column and updates the student cards below to call out students in relation to that column.  For some attributes this highlights students (eg, if they have an IEP or 504 plan) while for others it colors the student's card (eg, for gender identity).

This also adds http://gka.github.io/chroma.js/ for color math.

### hovering
hover-first

### hovering enough for tooltip
hover

### click to select
clicked

### story showing different card stylings:
screen shot 2018-05-11 at 11 05 04 am

# Checklists
+ [x] Author checked latest in IE - Classlists


kevinrobinson:
selfie

PR #1689, opened 2018-05-11
Ignore order in school spec

kevinrobinson:
# Who is this PR for?
developers

Fixes flaky build: https://travis-ci.org/studentinsights/studentinsights/builds/377822776?utm_source=github_status&utm_medium=notification, not sure why this changed but we don't really care about order in this method.


kevinrobinson:
selfie

PR #1691, opened 2018-05-11
Remove code from Let's Encrypt SSL installation

alexsoble:
# Who is this PR for?

Developers; cleaning up old code.

Our SSL isn't managed by Let's Encrypt anymore, so let's clean out the code related to installing Let's Encrypt. The code added a special, unauthenticated endpoint that sent down a secret string so that Let's Encrypt could verify our ownership of the domain.

# What does this PR do?

Removes Let's Encrypt endpoint.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,9 +3,8 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
-  unless Rails.env.development?
-    force_ssl except: [:lets_encrypt_endpoint]
-  end
+
+  force_ssl unless Rails.env.development?
 
   before_action :redirect_domain!
   before_action :authenticate_educator!  # Devise method, applies to all controllers (in this app 'users' are 'educators')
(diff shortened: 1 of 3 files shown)


kevinrobinson:
@alexsoble This code seems like it would be good to remove if we're not using it anymore!    Checking my understanding, I thought we were using Let's Encrypt and this is what I see what looking at the SSL cert:
screen shot 2018-05-11 at 5 38 34 pm

So I'm wondering if there's something else we'll need to do to enable renewing that in July or if it will auto-renew.  I think that's all unrelated to removing this leftover cruft from the initial setup though, but am not entirely sure so figured I'd ask!

kevinrobinson:
Ah this came up in some other reading - we use Heroku's automated certificate management so I think it's doing all this for us, and it happens to also use Let's Encrypt: https://devcenter.heroku.com/articles/automated-certificate-management

alexsoble:
@kevinrobinson Yep, that's my understanding too. We're still using Let's Encrypt but it's being provided by Heroku, not by our application's custom installation.

kevinrobinson:
👍 🚢 😄

alexsoble:
@kevinrobinson Deployed, checked the cert in Chrome to triple check this, and tried visiting the site through http to verify that https force-redirect is working.

PR #1694, opened 2018-05-11
Classrooms: Handle changing number of rooms after students have been assigned

kevinrobinson:
# Who is this PR for?
Grade 1-6 educator teams.  This is part of https://github.com/studentinsights/studentinsights/issues/1659.

This came up in feedback from an AP who triggered https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/120/.

# What does this PR do?
Fixes the bug.  Previously the code assumed that the order of these operations was fixed; with relaxing some of the navigation this isn't fixed anymore, so this updates the code to handle when this event happens after students have been placed.

# Checklists


kevinrobinson:
selfie

PR #1697, opened 2018-05-14
Use Devise lockable strategy and notify on lock

alexsoble:
# Who is this PR for?

Security improvement.

Right now we don't have good visibility into repeated unsuccessful login attempts and we don't handle them in any particular way.

# What does this PR do?

Uses the Devise lockable strategy to lock out a user for 24 hours if they try logging in with the wrong password too many times in a row (maximum attempts currently configured to 4, but we could adjust this number).

Also adds an email hook to notify us when a user gets locked out to increase visibility. This could go to Insights developers, school-based IT staff, or a mix of both; configurable via an ENV variable. 

# GIF

lockout

# Checklists

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -5,37 +5,12 @@ Devise.setup do |config|
   # random tokens. Changing this key will render invalid all existing
   # confirmation, reset password and unlock tokens in the database.
   config.secret_key = ENV['DEVISE_SECRET_KEY']
-  # ==> Mailer Configuration
-  # Configure the e-mail address which will be shown in Devise::Mailer,
-  # note that it will be overwritten if you use your own mailer class
-  # with default "from" parameter.
-  config.mailer_sender = 'please-change-me-at-config-initializers-devise@example.com'
-
-  # Configure the class responsible to send e-mails.
-  # config.mailer = 'Devise::Mailer'
 
   # ==> ORM configuration
   # Load and configure the ORM. Supports :active_record (default) and
   # :mongoid (bson_ext recommended) by default. Other ORMs may be
   # available as additional gems.
   require 'devise/orm/active_record'
-
-  # ==> Configuration for any authentication mechanism
-  # Configure which keys are used when authenticating a user. The default is
-  # just :email. You can configure it to use [:username, :subdomain], so for
-  # authenticating a user, both parameters are required. Remember that those
-  # parameters are used only when authenticating and not when retrieving from
-  # session. If you need permissions, you should implement that in a before filter.
-  # You can also supply a hash where the value is a boolean determining whether
-  # or not authentication should be aborted when the value is not present.
-  # config.authentication_keys = [ :email ]
-
-  # Configure parameters from the request object used for authentication. Each entry
-  # given should be a request method and it will automatically be passed to the
-  # find_for_authentication method and considered in your model lookup. For instance,
-  # if you set :request_keys to [:subdomain], :subdomain will be used on authentication.
-  # The same considerations mentioned for authentication_keys also apply to request_keys.
-  # config.request_keys = []
 
   # Configure which authentication keys should be case-insensitive.
   # These keys will be downcased upon creating or modifying a user and when used
@@ -98,31 +73,6 @@ Devise.setup do |config|
   # Setup a pepper to generate the encrypted password.
   # config.pepper = 'd68602c07af77ffeb4817fad17efeda6ece23b89d3b2b42126b56b2a99a8cf5c653f47ad1298a1b39d8e866835b48148875e84883c308c3addf4d3acbc124e60'
 
-  # ==> Configuration for :confirmable
-  # A period that the user is allowed to access the website even without
-  # confirming their account. For instance, if set to 2.days, the user will be
-  # able to access the website for two days without confirming their account,
-  # access will be blocked just in the third day. Default is 0.days, meaning
-  # the user cannot access the website without confirming their account.
-  # config.allow_unconfirmed_access_for = 2.days
-
-  # A period that the user is allowed to confirm their account before their
-  # token becomes invalid. For example, if set to 3.days, the user can confirm
-  # their account within 3 days after the mail was sent, but on the fourth day
-  # their account can't be confirmed with the token any more.
-  # Default is nil, meaning there is no restriction on how long a user can take
-  # before confirming their account.
-  # config.confirm_within = 3.days
-
-  # If true, requires any email changes to be confirmed (exactly the same way as
-  # initial account confirmation) to be applied. Requires additional unconfirmed_email
-  # db field (see migrations). Until confirmed, new email is stored in
-  # unconfirmed_email column, and copied to email column on successful confirmation.
-  config.reconfirmable = true
-
-  # Defines which key will be used when confirming an account
-  # config.confirmation_keys = [ :email ]
-
   # ==> Configuration for :rememberable
   # The time the user will be remembered without asking for credentials again.
   # config.remember_for = 2.weeks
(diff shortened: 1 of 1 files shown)


app/models/educator.rb
```diff
@@ -37,6 +37,12 @@ def as_json(options = {})
     super(options.merge({ except: [:student_searchbar_json] }))
   end
 
+  def lock_access!(opts = { })
```
alexsoble: Override the `lock_access!` method provided by Devise in order to add an email notification to developers after user access is locked.

Original method defined here: 

https://github.com/plataformatec/devise/blob/7b3081760f32e7150601776276bd809601b8eaaa/lib/devise/models/lockable.rb#L42

app/views/lockout_mailer/lockout_report.erb
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body>
+    <h1>An educator has been locked out of Student Insights</h1>
+    <p>
+      The following educator has been locked out because they made too many login attempts:
```
alexsoble: Small improvement would be to read in the max login Devise config and include it here to make this more precise.

config/initializers/devise.rb
```diff
@@ -5,14 +5,6 @@
   # random tokens. Changing this key will render invalid all existing
   # confirmation, reset password and unlock tokens in the database.
   config.secret_key = ENV['DEVISE_SECRET_KEY']
-  # ==> Mailer Configuration
```
alexsoble: When I went to edit this initializer file, I saw a **lot** of commented-out cruft configuration that doesn't apply to the Student Insights use case for Devise. 

I decided to remove a bunch of the commented-out material, since it's not relevant to our app and it makes it harder to discover the sections that are relevant.

config/initializers/devise.rb
```diff
@@ -5,14 +5,6 @@
...
-  # Configure the e-mail address which will be shown in Devise::Mailer,
-  # note that it will be overwritten if you use your own mailer class
-  # with default "from" parameter.
-  config.mailer_sender = 'please-change-me-at-config-initializers-devise@example.com'
```
alexsoble: We don't send or receive any email from Devise, because passwords aren't recoverable over email.

config/initializers/devise.rb
```diff
@@ -98,31 +74,6 @@
   # Setup a pepper to generate the encrypted password.
   # config.pepper = 'd68602c07af77ffeb4817fad17efeda6ece23b89d3b2b42126b56b2a99a8cf5c653f47ad1298a1b39d8e866835b48148875e84883c308c3addf4d3acbc124e60'
 
-  # ==> Configuration for :confirmable
```
alexsoble: We don't use confirmable, so this seemed good to delete.

config/initializers/devise.rb
```diff
@@ -155,61 +106,27 @@
...
   # :time  = Re-enables login after a certain amount of time (see :unlock_in below)
   # :both  = Enables both strategies
   # :none  = No unlock strategy. You should handle unlocking by yourself.
-  # config.unlock_strategy = :both
+  config.unlock_strategy = :time
 
   # Number of authentication tries before locking an account if lock_strategy
   # is failed attempts.
-  # config.maximum_attempts = 20
+  config.maximum_attempts = 4
```
alexsoble: The default of 20 attempts seemed to high here, so I set it to a more conservative 4. We can always adjust.

app/mailers/lockout_mailer.rb
```diff
@@ -0,0 +1,17 @@
+class LockoutMailer < ActionMailer::Base
+
+  def lockout_report(educator)
+    @target_emails = ENV['LOCKOUT_MAILER_LIST'].split(',')
+    @subject = "LOCKOUT: Student Insights Lockout Mailer"
+    @educator = educator
+    @email = educator.email
+    @full_name = @educator.full_name
```
kevinrobinson: Looks like only `@email` and `@full_name` are used in the template?

app/models/educator.rb
kevinrobinson: 👍 want to put that right in as a comment?

kevinrobinson: How does this actually get called within the sequence of Devise steps?  I'm wondering because I don't think we have an educator model until our strategy runs here: https://github.com/studentinsights/studentinsights/blob/master/app/lib/ldap_authenticatable_tiny/strategy.rb#L18

So I'm wondering how `lockable` would be able to interrupt in the middle of that strategy's code.  That seems like the place we'd want the check to happen - once we verify the email is valid and we load that educator model, we check whether they've been locked.  Since that isn't happening in `strategy.rb`, I'm wondering how lockable can influence what happens.

spec/features/sign_in_feature_spec.rb
```diff
@@ -15,6 +15,25 @@ def expect_successful_sign_in_for(educator)
...
+
+      sign_in_attempt(rich_email, 'badpassword')
+      sign_in_attempt(rich_email, 'badpassword')
+      sign_in_attempt(rich_email, 'badpassword')
+      sign_in_attempt(rich_email, 'badpassword')
+      sign_in_attempt(rich_email, 'badpassword')
+
+      rich.reload
+
+      expect(rich.locked_at).not_to eq nil
```
kevinrobinson: maybe `expect(rich.failed_attempts).to eq 5` also?

kevinrobinson: Also, this won't be testing the production path since we're not using the LDAP configuration in dev or test.  Maybe a place where adding a mock LDAP server and swapping that in for this test would help.

config/initializers/devise.rb
kevinrobinson: Yeah I agree, I'd vote for 3 since that seems like what I'd expect but don't know if that's a normal expectation.

config/initializers/devise.rb
kevinrobinson: What do you think about `send_unlock_instructions` in https://github.com/plataformatec/devise/blob/7b3081760f32e7150601776276bd809601b8eaaa/lib/devise/models/lockable.rb#L42?

This would have the additional benefit of notifying a user of failed attempts on their account if it wasn't them.

kevinrobinson: (not that they could unlock over email)

alexsoble:
Rebasing this on top of #1701.

alexsoble:
OK, this spec failure is really good and important. The way I read it, Devise lockable worked out-of-the-box with database authenticable, but fails when implemented with our custom LDAP Tiny strategy. This is exactly the question you were bringing up @kevinrobinson — looks like you were 💯% right to be wondering how this would work with our `strategy.rb`. This is also a great example of how shipping #1701 helps us build & test better security features!

So the next step I'll look at is writing some custom code in our LDAP Tiny strategy that updates `locked_at` after a given number of failed attempts. Will start on that Monday. My hope is that it's relatively simple to glue lockable into our strategy.

alexsoble:
One thing that I see here, especially in 9bc9ef7fcfdac01e02b8958018ada1a127d3f604, is that Devise `:lockable` is not a drop-in security feature with our custom LDAP Tiny strategy. It only works as a drop-in with the `:devise_authenticable` strategy. This shifts the scope/prioritization for this work. Landing this PR will take more time and effort and testing than it would if it were a drop-in feature. I originally conceived of this as "low hanging fruit." Since it's not necessarily low hanging fruit anymore, it may make sense to land other, higher-value security features first, like incremental two-factor authentication.

alexsoble:
Broke the `devise.rb` config changes into their own pull request (#1824) so that this one is easier to review and has less unrelated diff, especially since at this point it looks like the work will involve writing our own custom `:lockable` instead of Devise `:lockable`.

alexsoble:
Hmmm, I tried to rebase this one on top of master using `git rebase -s ours master` so that it would favor the master branch's changes to `devise.rb` over this branch's changes to `devise.rb`... 

It seems the strategy I used was a bit too aggressive and wiped all the commits off of this branch. 😅  I'm sure I can get them back via git reflog, but honestly starting over wouldn't be the worst thing since I'm re-implementing `:lockable` here manually.

kevinrobinson:
@alexsoble 👍 fwiw, in that situation using `merge` might do what you're looking for:

```
$ git checkout set-up-lockable
$ git fetch
$ git merge origin/master
```

alexsoble:
@kevinrobinson Yeah, that would have done the trick I think! I like the way rebasing keeps a clean commit history, so I was curious if there are any strategies that I could pass to `git rebase` that would perform a regular rebase but always default to one branch or the other if there are conflicts.

PR #1698, opened 2018-05-16
Add Content-Security-Policy in report-only mode

kevinrobinson:
# Who is this PR for?
Everyone!

Adds Content-Security-Policy directives in report-only mode so we can see what we missed, then turn it on.  This PR adds updates Rollbar since it was previously served from an ambiguous cloudfront domain and now it's served from a more clearly scoped URL.  EDIT: (Rollbar changes removed)

Unfortunately CSP won't particularly help in IE11: https://stackoverflow.com/a/42940115 (which manual testing verifies).  But this seems worth adding regardless.

This reports to https://github.com/studentinsights/csp-logger and the new Heroku deployment of it, logging violations in the database.  I verified that this works as expected with the app locally in development mode and submissions look like this running the command [here](https://github.com/studentinsights/csp-logger#looking-in-the-database):

```
 id |   substr   |               documentURI               |                                                              blockedURI                                                              | sourceFile | lineNumber
----+------------+-----------------------------------------+--------------------------------------------------------------------------------------------------------------------------------------+------------+------------
 20 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery_ujs.self-784a997f6726036b1993eb2217c9cb558e1cbb801c6da88105588c56f13b466a.js?body=1              |            |
 19 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/effect-bounce.self-bfcb9288797ecd234ddeedf4fbc5b4c9de94a4e6d8d44e6faa5b5e9836fad9fd.js?body=1 |            |
 18 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/effect-clip.self-7368e4597c32360b162d264ff8bbd0cf9fb1535eb9b4564e6eb7c1db3c774bf3.js?body=1   |            |
 17 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/effect-blind.self-ed102209d7aa0b7374f3db8478cc7dcff5f9980d3146e8bf9fbfab959c6d5d8a.js?body=1  |            |
 16 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/effect.self-3acdb29e7da09c8f0195b994d2f04d73bfef50dc539c84fd2c835d964d33e5c1.js?body=1        |            |
 15 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/dialog.self-5a6c7980dfa3f6b74445b0f4b20eed799ee65363163500ed7f032e603d6b022d.js?body=1        |            |
 14 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/droppable.self-a7a2b4c2cec8fc176f51e4e4291d92bfd3eec40b6a580b47fc54c16334d52882.js?body=1     |            |
 13 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/resizable.self-60b870565f3cd62982c1907828119bb51f42ec8228d5e23868153edc3abc685c.js?body=1     |            |
 12 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/draggable.self-debcca3d0d5d9afe2fe91a6fd1aebdb41ca0a55f2f6017352e6c02b6d8a6e64c.js?body=1     |            |
 11 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/mouse.self-c513294e8da73f31f84ba3ef11e2a1180a47faea0eb2fea4a53fc26153dd21fd.js?body=1         |            |
 10 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/datepicker.self-6267be6d4d0f0e8665be3b4d6ac07f40ba23a32aff8eaa5d1bfc00f76a9eeae7.js?body=1    |            |
  9 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/button.self-4d4f21ba77bbf8cc08ae65e2b6329a7a1b952b32c3ba7bffca8313e7e8c93d0c.js?body=1        |            |
  8 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/autocomplete.self-3befc48aba87cead090cfd014562b2771a662ac6d2c8197b24c08b2d9f3d9f2d.js?body=1  |            |
  7 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/menu.self-7abc1c9401509a7e4fc609b1e461be6fd17e8116e8fda19cc59d05ba79aefb68.js?body=1          |            |
  6 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/position.self-e693ced4ecfa1a276f0b794f32c1d30d792764b08788bf68ecfa3b388c291333.js?body=1      |            |
  5 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/widget.self-c1602241ddc51216b58391768667068867b8e15b9efc722befcd25771eda6819.js?body=1        |            |
  4 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/core.self-93be4d22eda916787802a64d8f88b52db8a9027d4ccbbc0942625324a7d12d44.js?body=1          |            |
  3 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery-ui/accordion.self-c95f320275150a94241b004a91e21eef9c5ae9965a793cb3e2475f6078a1670d.js?body=1     |            |
  2 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/jquery.self-bd7ddd393353a8d2480a622e80342adf488fb6006d667e8b42e4c0073393abee.js?body=1                  |            |
  1 | script-src | http://localhost:3000/educators/sign_in | http://localhost:3000/assets/env.self-4f6714e1ca417f5bf8dd55419d21306ff71a53b5570b9498085969ee5786d6b9.js?body=1                     |            |
(20 rows)
```


kevinrobinson:
@alexsoble I second mention bot, want to take a look at this when you have a chance?

kevinrobinson:
huh also Rollbar has also since enabled [telemetry](https://github.com/rollbar/rollbar.js#telemetry) by default.  We don't want to increase our exposure on Rollbar so this opts out of DOM and log instrumentation, otherwise this would log form inputs to Rollbar.  Seems like a really bad default to opt people into.

I was planning on deploying this to demo first anyway, and then verifying, but this is another thing we'll need to manually verify when deploying.

kevinrobinson:
@alexsoble awesomely noticed that this blocks things in dev when it should be report only.  Need to look at this.

kevinrobinson:
@alexsoble also, can you review the Rollbar changes related to Telemetry here too?

config/initializers/secure_headers.rb
```diff
@@ -1,10 +1,40 @@
 SecureHeaders::Configuration.default do |config|
 
-  # Unblock PDF downloading for student report
+  # Unblock PDF downloading for student report and for IEP-at-a-glance
   config.x_download_options = nil
+  
+  if EnvironmentVariable.is_true('ENABLE_CSP')
+    config.csp = SecureHeaders::OPT_OUT # don't enforce yet
```
kevinrobinson: @alexsoble So this wasn't working locally because the default behavior of `secure_headers` is to add CSP unless you opt-out.  So this was adding both a CSP policy (the default) and then our own "report only" header too.

This updates it to opt-out of CSP defaults and only add a report-only header to start, and I verified this locally.  Good catch! 👍

app/views/shared/_rollbar.html.erb
```diff
@@ -29,19 +29,43 @@
...
...captureUncaughtExceptions:o,captureUnhandledRejections:t,wrapGlobals:a}},function(r,e){"use strict";function o(r,e){this.impl=r(e,this),this.options=e,n(o.prototype)}function n(r){for(var e=function(r){return function(){var e=Array.prototype.slice.call(arguments,0);if(this.impl[r])return this.impl[r].apply(this.impl,e)}},o="log,debug,info,warn,warning,error,critical,global,configure,handleUncaughtException,handleUnhandledRejection,_createItem,wrap,loadFull,shimId,captureEvent,captureDomContentLoaded,captureLoad".split(","),n=0;n<o.length;n++)r[o[n]]=e(o[n])}o.prototype._swapAndProcessMessages=function(r,e){this.impl=r(this.options);for(var o,n,t;o=e.shift();)n=o.method,t=o.args,this[n]&&"function"==typeof this[n]&&("captureDomContentLoaded"===n||"captureLoad"===n?this[n].apply(this,[t[0],o.ts]):this[n].apply(this,t));return this},r.exports=o},function(r,e){"use strict";r.exports=function(r){return function(e){if(!e&&!window._rollbarInitialized){r=r||{};for(var o,n,t=r.globalAlias||"Rollbar",a=window.rollbar,l=function(r){return new a(r)},i=0;o=window._rollbarShims[i++];)n||(n=o.handler),o.handler._swapAndProcessMessages(l,o.messages);window[t]=n,window._rollbarInitialized=!0}}}}]);
```
alexsoble: @kevinrobinson It looks like the Rollbar library snippet is changing here, what's driving that change?

app/views/shared/_rollbar.html.erb
```diff
@@ -29,19 +29,43 @@
     }
   }
 
+  // Only report if env exists.
+  function filterTelemetry(e) {
```
alexsoble: @kevinrobinson Why are we only reporting telemetry is env exists?

kevinrobinson:
hey @alexsoble, good question! You can see more about this in the earlier
comments about Rollbar. What initially drove it is that the version we were
using was from a generic cloudfront domain so in updating we can more
strictly scope the CSP scripts whitelist.

However, turns out that newer versions of Rollbar switch on a “telemetry”
feature that does some scary things like instruments all user interactions
and DOM events. So this is opting out of all that.

Alternately we could just add the current Rollbar cloudfront URL to the CSP
whitelist (no less secure than it is today) and ship just the CSP change
first.

app/views/shared/_rollbar.html.erb
kevinrobinson: Ah yeah good question.  I had thought this was similar to what we do for normal events but it shouldn't be and we can remove this.

For normal Rollbar events, the `transform` right above checks to see if env is set.  This happens because we load Rollbar independently from the rest of our app's code so that they're isolated from each other.  As a result, when Rollbar loads and then catches some kind of error, we don't know if the rest of the app's JS loaded or if something went wrong that prevented it.  If we can we get the env stuff and add it but if not we still report.  None of that really applies to telemetry (they will also be piped through the `transform` fn).

kevinrobinson:
@alexsoble okay, split this into two PRs.  This is just CSP with current Rollbar and has to allow `unsave-eval` since the current version of Rollbar uses it.  https://github.com/studentinsights/studentinsights/pull/1702 updates Rollbar and makes that stricter.  That one needs a review since Rollbar's new Telemetry feature is now turned on by default and we want to opt-out.

After that we can see if anything else comes up, and then turn this on from just report-only.

kevinrobinson:
🚢 and moved the Rollbar change out to https://github.com/studentinsights/studentinsights/pull/1702.

kevinrobinson:
I tested this in the demo site, and we'll need to resolve a few issues before turning this on.  They're noted in https://github.com/studentinsights/studentinsights/issues/1704.  I disabled the reporting for now since we got the info we need to do the next step of work.

alexsoble:
@kevinrobinson Great, thanks for shipping this incrementally Kevin. I think splitting the PR into two was a good move since it let you 🚢 faster.

app/views/shared/_rollbar.html.erb
alexsoble: 👍

PR #1701, opened 2018-05-17
Mock LDAP in dev/test/demo environments

alexsoble:
# Who is this PR for?

Developers.

Right now the auth code forks: in local development mode, test mode, and on the demo site we're using Devise's `database_authenticable` strategy. In production, we're using an LDAP strategy. That makes it difficult to exercise code as it's being used in production when developing. 

# What does this PR do?

Creates a mock LDAP service that we can call in development, in our test suite, and in our demo site. That way we can exercise our LDAP strategy code when creating and testing new features. 

Adds an initializer that checks for sane config and checks to make sure that the mock LDAP is never turned on for the real site by accident.

Removes all code related to the database authentication strategy. 

# Edge case

I was curious what would happen if I set `USE_MOCK_LDAP
` to `false` in development mode. 
This triggered an error from our LDAP code, but there was no corresponding user-facing text in `devise.en.yml`, so the user saw this:

screen shot 2018-05-18 at 3 47 19 pm

I added natural-language copy so that if our LDAP strategy sends down `:error`, the user will see this instead now:

screen shot 2018-05-18 at 3 47 50 pm

# Checklist 

# Deploy checklist 

+ [ ] Set `USE_MOCK_LDAP` and `MOCK_LDAP_PASSWORD` environment variables in demo site before deploying. 
+ [ ] Deploy
+ [ ] Verify in production


app/lib/ldap_authenticatable_tiny/strategy.rb
```diff
@@ -14,10 +14,13 @@ class LdapAuthenticatableTiny < Authenticatable
       # https://github.com/plataformatec/devise/blob/master/lib/devise/models/authenticatable.rb
       def authenticate!
         email = authentication_hash[:email]
+        should_use_mock_ldap = ::EnvironmentVariable.is_true('USE_TEST_PALS_LDAP')
```
kevinrobinson: Adding other guards here seems good, like disallowing this in production or only allowing it when the district_key is `demo`.

app/lib/ldap_authenticatable_tiny/strategy.rb
```diff
@@ -14,10 +14,13 @@ class LdapAuthenticatableTiny < Authenticatable
...
+        ldap_service = should_use_mock_ldap ? TestPalsMockLDAP : Net::LDAP
```
kevinrobinson: This is small but this is a Ruby class not an instance so maybe `ldap_class`?

app/lib/ldap_authenticatable_tiny/strategy.rb
```diff
@@ -29,14 +32,14 @@ def authenticate!
...
+      def is_authorized_by_ldap?(ldap_service, login, password)
         return false if password.nil? || password == ''
         options = ldap_options_for(login, password)
-        ldap = Net::LDAP.new(options)
+        ldap = ldap_service.new(options)
         is_authorized = ldap.bind
 
         if !is_authorized
-          ldap_error = ldap.get_operation_result
+          ldap_error = ldap_service.get_operation_result
```
kevinrobinson: Will this have to change?  I was thinking our dummy class would return an instance that implemented this method so this would stay `ldap.get_operation_result`.

app/lib/test_pals_mock_ldap.rb
```diff
@@ -0,0 +1,35 @@
...
+  def initialize(options)
+    @options = options
+    @email = options[:auth][:username]
+    @password = options[:auth][:password]
+  end
+
+  def bind
+    raise 'LDAP error' unless ::EnvironmentVariable.is_true('USE_TEST_PALS_LDAP')
+
+    return TEST_PAL_EMAILS.include?(@email) && @password == 'demo-password'
```
kevinrobinson: Could this method read the database?   It seems like the list of usernames and passwords duplicates what's already there.  Or are you trying to simulate that the lists are actually different to test those cases?  I could see benefits either way.

app/models/educator.rb
```diff
@@ -1,11 +1,5 @@
 class Educator < ActiveRecord::Base
-  devise :rememberable, :trackable, :timeoutable
-
-  if ::EnvironmentVariable.is_true('SHOULD_USE_LDAP')
-    devise :ldap_authenticatable_tiny
-  else
-    devise :database_authenticatable
-  end
+  devise :ldap_authenticatable_tiny, :rememberable, :trackable, :timeoutable
```
kevinrobinson: 👍

app/lib/test_pals_mock_ldap.rb
alexsoble: Oh interesting. I can see the benefit to reading the database for user emails. We won't be able to read passwords from the database anymore because we won't be storing passwords there.

app/lib/ldap_authenticatable_tiny/strategy.rb
alexsoble: Oh good catch here.

app/lib/ldap_authenticatable_tiny/strategy.rb
alexsoble: 👍

app/lib/ldap_authenticatable_tiny/strategy.rb
alexsoble: I have another guard in the TestPalsMockLDAP class itself. Unfortunately I don't think we can disallow this for `Rails.env.production == true` because the demo Heroku site is production from a Rails.env point of view.

kevinrobinson: Gotcha.  I think it's important to make it so that it takes more steps and efforts than flipping a switch to disable authorization.  So maybe we can find some other ways to do that!

Here's some brainstorming ideas of things we could try:
- Move this check to an initializer, so that we can check all these things at app start time (and crash the app) rather than not catching a bad config until someone tries to sign in.  In the initializer, only allow it to be run in production when it's the `demo` district and when the environment variable is set.  Add that initializer file to mention bot.
- Remove the demo passwords from code and put them in an env variable.  This would require two env variable changes rather than just one.
- Make the demo passwords env variable a map of `email => password` so that even it were copy-pasted into production by accident it wouldn't work

app/lib/test_pals_mock_ldap.rb
kevinrobinson: Yeah I was thinking that would still work because it's how the model worked before, but I agree removing them altogether would be better.  We should probably remove all those password fields from the model then too looks like there's quite a few.

And in https://github.com/studentinsights/studentinsights/pull/1701/files#r189039275 I was brainstorming other ways to store these passwords outside of code so that it would take a series of changes in order to disable normal production authentication (having multiple layers there seems important).

app/lib/test_pals_mock_ldap.rb
```diff
@@ -0,0 +1,17 @@
+class TestPalsMockLDAP
```
alexsoble: TODO (ARS): Add code comments here explaining what this class is/does.

app/lib/test_pals_mock_ldap.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def initialize(options)
+    @options = options
+    @email = options[:auth][:username]
+    @password = options[:auth][:password]
+  end
+
+  def bind
```
kevinrobinson: Should we simulate the worst-case scenario for how an LDAP server might be configured here?  That way tests can validate that we're defensive for this in our code.

alexsoble: Good idea 😄

alexsoble: Implemented this here: e871e596c6355128824da587cf9c91759bbd568b.

spec/lib/ldap_authenticatable_tiny_spec.rb
```diff
@@ -107,42 +107,58 @@ def test_options(login, password)
 
   describe '#is_authorized_by_ldap? calls Net::LDAP correctly' do
     before { set_env }
+    let!(:pals) { TestPals.create! }
```
alexsoble: In these tests we're now going to test against TestPals emails + default demo passwords instead of mocking everything.

app/lib/ldap_authenticatable_tiny/strategy.rb
alexsoble: I ended up implementing 2/3 of these suggestions:

* We have not one, not two, but three pieces of ENV required to disable real authentication and use the mock LDAP instead. `USE_TEST_PALS_LDAP` and `TEST_PALS_LDAP` password both need to be set and `DISTRICT_KEY` needs to be set to demo. 
* Also added an initializer to run checks to verify sane config and set good defaults on app start. 

I think this is a fairly decent improvement on the status quo, which all forks on a single ENV variable, `SHOULD_USE_LDAP`.

I didn't implement the map of `email => password` suggestion because if the Test Pals ever change down the line, we'd have to remember to replace the ENV variables everywhere and that seemed like something that could really easily fall out of sync.

alexsoble:
@kevinrobinson, when you're back on Monday, I think this is in a good place for another review! Overall I think this PR moves us forward security-wise. In addition to subbing in the mock LDAP interface, it adds a missing user-facing error message to the login flow and triples the number of ENV variable changes required for a developer to disable auth.

One question I'm wondering about is naming. When I started out, I named the mock LDAP class TestPalsMockLDAP. As I revised the PR, I realized the class is no longer hard-coded to expect TestPal usernames and passwords explicitly; it accepts whatever educator records are in the database as valid and passwords are set in ENV. I wonder if referencing TestPals would make this class more obscure to a new developer coming on board. So it might be worth it to rename.

app/lib/test_pals_mock_ldap.rb
```diff
@@ -0,0 +1,58 @@
...
+    raise 'LDAP error' unless ::EnvironmentVariable.is_true('USE_TEST_PALS_LDAP')
```
kevinrobinson: Let's make this error message explicit, saying that a method on the fake class was called when the env variable wasn't set.  Otherwise seeing "LDAP error" in the logs might make it seem like an LDAP server was down.

config/initializers/mock_ldap_config_check.rb
```diff
@@ -0,0 +1,19 @@
+should_use_mock_ldap = ::EnvironmentVariable.is_true('USE_TEST_PALS_LDAP')
```
kevinrobinson: What do you think about moving this to a class method on `MockLdap` and calling it here?  Then `MockLdap` could call that method each time as well, and you could remove the env check there.  That would also move all the things read the env related to this to the same file.

config/initializers/mock_ldap_config_check.rb
```diff
@@ -0,0 +1,19 @@
...
+
+env_is_production = Rails.env.production?
+
+district_is_demo = (ENV.fetch('DISTRICT_KEY') == 'demo')
```
kevinrobinson: Let's use`PerDistrict` for this, either put a method there or something like `PerDistrict.new.district_key == PerDistrict.DEMO`

config/initializers/mock_ldap_config_check.rb
```diff
@@ -0,0 +1,19 @@
...
+  raise 'Mocking LDAP not allowed in production except for demo site'
+end
+
+# Guard to make sure that if we're using the fake demo accounts for auth,
+# we also have demo passwords set in ENV:
+if should_use_mock_ldap && env_is_production && ENV['TEST_PALS_LDAP_PASSWORD'].nil?
+  raise 'Missing mock LDAP password'
+else
+  ENV['TEST_PALS_LDAP_PASSWORD'] = 'demo-password'
+end
```
kevinrobinson: Another guard might be to raise if any of the env variables like `DISTRICT_LDAP_HOST` are set.  Probably veering towards 🚲 here though :)

spec/features/database_authenticatable_sign_in_feature_spec.rb
```diff
@@ -6,7 +6,7 @@
 
   context 'teacher signs in' do
     def expect_successful_sign_in_for(educator)
-      educator_sign_in(educator)
+      sign_in_attempt(educator.email, 'demo-password')
```
kevinrobinson: I'm wondering how this works.  Won't this test fail since the password env variable isn't set?  Or wouldn't we want to test a sign in with a wrong password in this file?

spec/lib/ldap_authenticatable_tiny_spec.rb
```diff
@@ -107,42 +107,58 @@ def test_options(login, password)
...
-      expect(strategy.send(:is_authorized_by_ldap?, 'foo', 'bar')).to eq false
+
+      args = [
+        :is_authorized_by_ldap?,
+        TestPalsMockLDAP,
+        'uri@demo.studentinsights.org',
+        'bar'
+      ]
+
+      expect(strategy.send(*args)).to eq false
```
kevinrobinson: This seems tricksy to me versus saying it plain, is this to avoid long line length?

alexsoble: Line length but also making it easy for the reviewer to see which arguments are changing for each test case.

spec/features/database_authenticatable_sign_in_feature_spec.rb
alexsoble: The config check sets 'demo-password' automatically in test and dev environments.

config/initializers/mock_ldap_config_check.rb
alexsoble: Heh a little 🚲 sheddy but not a bad idea.

alexsoble:
@kevinrobinson This had merge conflicts with master, so I rebased and force-pushed. That's why some of your earlier comments say "We went looking everywhere, but couldn’t find those commits." if you try to click into the file names.

alexsoble:
@kevinrobinson I also renamed the class to MockLDAP and the ENV variables to USE_MOCK_LDAP, since they don't have anything to do with TestPals and are entirely generic. The only connection with TestPals is that in practice we'll use TestPals to populate the database in test/demo/dev environments.

spec/lib/ldap_authenticatable_tiny_spec.rb
kevinrobinson: I'd vote for finding another way to do this that doesn't involve `send(*args)` where the code hides what method is being called.  Another way to make it easier to see the arguments on each line might be to use the splat inline:

```
expect(strategy.is_authorized_by_ldap?(*[
  MockLDAP,
  'uri@demo.studentinsights.org',
  'bar'
])).to eq false

File: spec/features/database_authenticatable_sign_in_feature_spec.rb (hunk 13 above)
Comment (kevinrobinson) [reply spec/features/database_authenticatable_sign_in_feature_spec.rb]: Ah I see now, thanks!  Let's move this to `development.rb` and `test.rb`, then we can avoid any logic and code around setting env variables related to authorization automatically.  And in the production demo case, we have to set it manually in production for it to work.

`simple > easy`

File: config/initializers/mock_ldap_config_check.rb [hunk 15]
```diff
@@ -0,0 +1,19 @@
...
+if should_use_mock_ldap && env_is_production && !district_is_demo
+  raise 'Mocking LDAP not allowed in production except for demo site'
+end
+
+# Guard to make sure that if we're using the fake demo accounts for auth,
+# we also have demo passwords set in ENV:
+if should_use_mock_ldap && env_is_production && ENV['MOCK_LDAP_PASSWORD'].nil?
+  raise 'Missing mock LDAP password'
+else
+  ENV['MOCK_LDAP_PASSWORD'] = 'demo-password'
```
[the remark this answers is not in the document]
alexsoble: I agree that automatically setting env variables related to authorization is good to avoid. Where do you see `ENV['MOCK_LDAP_PASSWORD']` being set for the test environment? My understanding is that if we want it to be "demo-password" in test and development we have to set that ENV value somewhere in the code, but maybe I'm not thinking creatively enough here.

alexsoble:
@kevinrobinson Thanks, I appreciate the questions and suggestions! I'm down to keep talking because I think there are still some concrete improvements to make here.

config/initializers/mock_ldap_config_check.rb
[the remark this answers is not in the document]
alexsoble: Ah, I see. You're suggesting `ENV['MOCK_LDAP_PASSWORD'] = 'demo-password'` in `test.rb`, is that right?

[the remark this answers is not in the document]
kevinrobinson: Ah yeah good clarification, sorry!  re: setting env. variables I was suggesting adding `ENV['MOCK_LDAP_PASSWORD'] = 'demo-password'` in the dev.rb / test.rb files, yep!

And earlier when I said "if we move this bit of config into..." it sort of sounds like I meant that code snippet after.  Yeah I just mean moving the `ENV['MOCK_LDAP_PASSWORD'] = ...` bit into config, not the code.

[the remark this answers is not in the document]
alexsoble: OK, gotcha. In my mind moving `ENV['MOCK_LDAP_PASSWORD']` to `test.rb` means that we're still setting environment variables related to authorization in code. After all, `test.rb` is code. But I agree that it's an improvement because there's no logic involved if we set in `test.rb`.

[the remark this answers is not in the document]
kevinrobinson: Yeah, that's a better way to say what I'm trying to say :)

spec/lib/ldap_authenticatable_tiny_spec.rb
kevinrobinson: Got it!  Yeah `send` let's you call private methods in test.  The thing I'm trying to comment on is that using send with a variable together like `send(*args)` means you can't tell reading that code what method is being called.  I agree you should 🚢 what you think makes sense 👍

alexsoble:
Alright! @kevinrobinson, I removed the initializer check and replaced it with a `ShouldUseMockLDAP` class that implements the method you sketched out. Also added a unit test for all the different rules and cases we have for when Mock LDAP should/shouldn't be turned on. I think this is an improvement for sure — it centralizes the rules so we can call them in different places. Thanks for the suggestion @kevinrobinson!

I think next steps are for me to do a little bit of manual QA in development to make sure I didn't miss anything in this latest round of refactor, then set the mock LDAP ENV config in the demo Heroku site, then merge and 🚢.

kevinrobinson:
@alexsoble awesome, this whole part of the system will be simpler and sturdier! 👍

alexsoble:
✅ manually QA'd in local development mode
✅ got second look from @kevinrobinson on `ShouldUseMockLDAP` code
✅ set ENV vars from demo Heroku site in advance of merging/deploying

🚢

alexsoble:
Verified in production. 👍

PR #1702, opened 2018-05-18
Update Rollbar

kevinrobinson:
# Who is this PR for?
developers

The current version of Rollbar is old and it also uses `eval` in a way that we'd like to disallow with CSP.  This updates to the latest version, which lets us increase the strictness of the CSP policy and since Rollbar is now served from a different CDN it updates that as well.

Rollbar has also since enabled [telemetry](https://github.com/rollbar/rollbar.js#telemetry) by default.  We don't want to increase our exposure on Rollbar so this opts out of DOM and log instrumentation, otherwise this would log form inputs to Rollbar.  Seems like a really bad default for most developers.

This is another thing we'll need to manually verify when deploying.


kevinrobinson:
huh alsoRollbar has also since enabled [telemetry](https://github.com/rollbar/rollbar.js#telemetry) by default.  We don't want to increase our exposure on Rollbar so this opts out of DOM and log instrumentation, otherwise this would log form inputs to Rollbar.  Seems like a really bad default for most developers.

This is another thing we'll need to manually verify when deploying.

alexsoble:
@kevinrobinson What do you think is the best way to check the telemetry config? Maybe force-push this to the demo site, trigger some client-side errors, and look at the report that Rollbar sends back?

kevinrobinson:
@alexsoble Yeah that sounds great!  I was thinking to review the docs and this code, then deploy and verify.  But deploying to the demo site first sounds even better 👍

kevinrobinson:
deploying this to the demo site now to test!

kevinrobinson:
Confirmed that this looks good, testing manually on the demo site (forcing it to report errors and triggering some manually).  I tested this while searching for students, entering notes, entering services, and making requests to the server to save notes and services.


### Web inspector, looking at outbound requests to rollbar.com
This is what I see over the wire, showing no request/response bodies being captured in network telemetry, and no DOM interactions being captured either.

screen shot 2018-05-25 at 1 50 58 pm


### In the Rollbar product
And here's what I see in the Rollbar product:
screen shot 2018-05-25 at 1 52 39 pm

app/views/shared/_rollbar.html.erb
```diff
@@ -34,14 +34,30 @@
   var _rollbarConfig = {
     accessToken: "a79df03b6e734ef78f9149de2329d886",
     captureUncaught: true,
+    captureUnhandledRejections: true,
     transform: transform,
-    checkIgnore: checkIgnore
+    checkIgnore: checkIgnore,
+    autoInstrument: {
```
kevinrobinson: @alexsoble I've verified in the demo site that this looks right, but can you double-check and 👍 that this seems correct?

alexsoble:
Heya @kevinrobinson! I checked out the Rollbar UI for some of the test errors you triggered and I didn't see any sensitive data there.

I was curious about the telemetry options that we're enabling, so I went and read a little more about 
`network`, `navigation`, and `connectivity` here: https://docs.rollbar.com/docs/javascript.

Connectivity and navigation look useful and safe in terms of not leaking private data to Rollbar:

> Navigation events use the information from pushState on browsers that allow for this and gathers to and from information.

> Connectivity events try to capture changes in network connectivity status when this is exposed by the browser.

At first the `network` docs gave me pause:

> Network events are XHR and fetch requests. We store the status code, the url, and some timing events to determine how long requests take. For network events, we also support the following keys in the autoInstrument object: networkResponseHeaders, networkResponseBody, and networkRequestBody. These are only relevant if network is set to true. By default they all take the value false.

We definitely don't want to send `networkResponseBody` results over to Rollbar! But then I went ahead and read it a second time and realized that this will default to `false`, so this isn't enabled here. 

Overall, update looks good, glad that this will enable a better Content Security Policy, and good to learn more about RollbarJS details! Good job catching the telemetry settings in this update @kevinrobinson!

kevinrobinson:
@alexsoble awesome, thanks for double-checking! 👍 

And good catch re: network requests.  I also added those flags explicitly as `false` too so that it's clear we don't want this (regardless of whether it's the default):
```
      networkResponseHeaders: false,
      networkResponseBody: false,
      networkRequestBody: false,
```

🚢 ing!

kevinrobinson:
looks good 👍

PR #1706, opened 2018-05-18
Classrooms: Allow enabling class lists in demo site

kevinrobinson:
# Who is this PR for?
developers

Enables doing some simplistic load testing on the demo site, which isn't a perfect proxy for any production site but can help get some ballpark figures.

# What does this PR do?
Adds an ENV variable for controlling the class lists feature, and allows it to be set in the demo environment.

# Checklists


kevinrobinson:
added env first, selfie

Issue #1692, opened 2018-05-11
Discipline: Clear x throws error


kevinrobinson:
Related to https://somerville.studentinsights.org/schools/shs/discipline.

Clicking on the "x" on the select box throws an error:
<img width="505" alt="screen shot 2018-05-11 at 2 00 08 pm" src="https://user-images.githubusercontent.com/1056957/39939257-ab1af726-5523-11e8-8a5f-d45e6b665b0d.png">

kevinrobinson:
cc @edavidsonsawyer I spotted this locally but forgot about it earlier.

kevinrobinson:
Fixed in https://github.com/studentinsights/studentinsights/pull/1707.

PR #1707, opened 2018-05-21
Discipline Dashboard UX fixes

edavidsonsawyer:
# Who is this PR for?
dashboard users

# What does this PR do?
Completes the work in #1660 and makes the dropdown show all options

All charts except grade, time, and day of week are now sorted by number of incidents

uximprovements


edavidsonsawyer:
@alexsoble @kevinrobinson this should clear up #1660 including #1692 and also expands the dropdown menu (which may have been obscuring the "offense" option). 

There is a remaining issue I found that I haven't had the chance to address tonight. The chart sorting doesn't change when the date ranges are altered, so they may be out of order when not looking at the school year. I should be able to handle that early this week.

alexsoble:
@edavidsonsawyer Taking a look now!

app/assets/stylesheets/components/school_dashboard.scss
```diff
@@ -8,6 +8,12 @@
   display: flex;
   flex-direction: column;
   padding: 0 20px;
+  .Select-menu-outer {
```
alexsoble: This could be a good place to add a comment explaining that since the Discipline dropdown menu is large, without increasing the max-height some of the options get cut off. We could also link to the react-select default CSS here since it looks like what you're doing here is bumping up the value of `@select-menu-max-height` from 200px to 1000px. 

+ https://github.com/JedWatson/react-select/blob/master/less/menu.less
+ https://github.com/JedWatson/react-select/blob/3cf1b7d18844ec309725239f03a22fb98485b8a7/less/select.less

alexsoble: Also, taking a look on Internet Explorer VM, I'm still seeing scrolling behavior, which makes me think the CSS here isn't working the same way as in Chrome. There's also a scrollbar by default which I think would make it less likely that someone would miss the fact that they can scroll for more options. GIF from Internet Explorer 11 VM:

menu

alexsoble:
@edavidsonsawyer The clearable and sorting changes look 👌, left a few comments/questions about the Select-menu CSS changes.

app/assets/stylesheets/components/school_dashboard.scss
edavidsonsawyer: @alexsoble weird! It does show the full menu in my VM (gif above) but it doesn't look like a big deal b/c of the scroll bar. It was irritating on Chrome b/c the scrollbar isn't there until I try to scroll. 

Re linking the default css, I actually tried that, but rails objected to serving the css ([similar issue](https://stackoverflow.com/questions/7829480/no-route-matches-get-assets?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa)) I wasn't sure about changing our rails config just for that, but maybe I'm missing a better way.

edavidsonsawyer: @alexsoble ah, gotha. Yes, I'll update the PR with some comments here.

edavidsonsawyer:
@alexsoble I think this should be good to go if that comment works for you

alexsoble:
👍 🚢 👍 🚢

PR #1711, opened 2018-05-22
Sorting discipline events correctly in all date ranges

edavidsonsawyer:
# Who is this PR for?
Dashboard users

Charts were sorted by number of events, but only for the school year

# What does this PR do?
Sorts charts by number of events within the selected date range

Broken
sortingold

Fixed
sortingnew

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
@@ -21,17 +21,11 @@ class SchoolDisciplineDashboard extends React.Component {
       selectedChart: 'location',
       selectedCategory: null
     };
-    this.setDate = this.setDate.bind(this);
     this.setStudentList = this.setStudentList.bind(this);
     this.resetStudentList = this.resetStudentList.bind(this);
     this.selectChart = this.selectChart.bind(this);
   }
 
-  setDate(range) {
-    this.setState({
-      startDate: moment.unix(range[0]).format("YYYY-MM-DD")
-    });
-  }
   setStudentList(highchartsEvent) {
     this.setState({selectedCategory: highchartsEvent.point.category});
   }
@@ -42,8 +36,8 @@ class SchoolDisciplineDashboard extends React.Component {
     this.setState({selectedChart: selection.value, selectedCategory: null});
   }
 
-  filterIncidentDates(incidents) {
-    return incidents.filter((incident) => {
+  filterIncidentDates(incidentsArray) {
+    return incidentsArray.filter((incident) => {
       return moment.utc(incident.occurred_at).isSameOrAfter(moment.utc(this.state.startDate));
     });
   }
@@ -60,9 +54,10 @@ class SchoolDisciplineDashboard extends React.Component {
   }
 
   getChartData(selectedChart) {
+    const filteredEvents = this.filterIncidentDates(this.props.schoolDisciplineEvents);
     return {
       type: selectedChart,
-      disciplineIncidents: _.groupBy(this.props.schoolDisciplineEvents, selectedChart),
+      disciplineIncidents: _.groupBy(filteredEvents, selectedChart),
       title: "Incidents by " + selectedChart};
   }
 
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -60,9 +54,10 @@ class SchoolDisciplineDashboard extends React.Component {
   }
 
   getChartData(selectedChart) {
+    const filteredEvents = this.filterIncidentDates(this.props.schoolDisciplineEvents);
```
kevinrobinson: @edavidsonsawyer I'm wondering about how this is related to the sorting improvements in the awesome GIFs on the pull request.  This looks like it's adding in another step of filtering, which I thought worked correctly before.  Are these two things related?

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -21,17 +21,11 @@ class SchoolDisciplineDashboard extends React.Component {
       selectedChart: 'location',
       selectedCategory: null
     };
-    this.setDate = this.setDate.bind(this);
     this.setStudentList = this.setStudentList.bind(this);
     this.resetStudentList = this.resetStudentList.bind(this);
     this.selectChart = this.selectChart.bind(this);
   }
 
-  setDate(range) {
```
kevinrobinson: This is removed leftover code? 👍

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: @kevinrobinson oh good catch! The intention was to move this filter method here so that all the other components just get the filtered data without having to do it themselves. But apparently I forgot to remove the other instances of this filter. 

I updated the PR and commented here about why the filter is being applied and tested with the removed code. This should be clearer and have less repetition now.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: Yup. Just something leftover from the date slider, but doesn't have anything to do with this PR in particular.

app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -143,7 +139,7 @@ class SchoolDisciplineDashboard extends React.Component {
     const categories = this.sortChartKeys(Object.keys(selectedChart.disciplineIncidents));
     const seriesData = categories.map((type) => {
       if (!selectedChart.disciplineIncidents[type]) return [];
-      const incidents = this.filterIncidentDates(selectedChart.disciplineIncidents[type]);
+      const incidents = selectedChart.disciplineIncidents[type];
```
kevinrobinson: Ah so this is the import change, right?  I'm still a little confused because the PR descriptions are about improving sorting, but these changes all seem related to filtering.  Is the change that it was always sorting before, just not on the right subset of events based on the current filtering?

alexsoble:
Hey @edavidsonsawyer! Looks like @kevinrobinson reviewed this one and made some really good suggestions, which you implemented. I pulled down the changes and verified in IE that the classroom sort works correctly in different date ranges. 👍 👍 

I tried to follow the thread about how filtering impacts the sorting, and to be honest I'm still a bit unclear on that, so maybe at some point this could benefit from more code commenting or some refactoring. 

Since this is an improvement to a feature that's still dark to most users, I want to make sure we 🚢 this and keep forward momentum, so unless @edavidsonsawyer @kevinrobinson you see reasons not to, let's ship!

I also wonder if we could apply the same sorting function to the Location and Offense categories. These both seem like they could benefit from letting educators quickly see which locations/offenses are the most common and least common. Screenshots of how these charts look now, unsorted, below (with fake demo data). 

## Location

by_location

## Offense

by_offense

alexsoble:
🚢

Issue #1708, opened 2018-05-21
Discipline: Student list doesn't match selection (eg, classroom)


kevinrobinson:
Related to https://github.com/studentinsights/studentinsights/issues/1660.

When clicking on a particular item like "grade 1", the student list table updates.  But this seems like it's only sorting and not actually filtering - if you scroll the list you see what looks like probably all other students at all other grade levels too.  It may be that there's some bug in how the table is updated, since the "total incidents" count seems to update but not reflect the number of students visible when scrolling through the table.

### 1. Select "grade" in dropdown
<img width="1018" alt="screen shot 2018-05-21 at 11 37 06 am" src="https://user-images.githubusercontent.com/1056957/40316248-b108f686-5ceb-11e8-9376-ad3f5f9c5ab0.png">

### 2. Click on 1st grade
It looks like the list has updated, and the total count is smaller.
<img width="1014" alt="screen shot 2018-05-21 at 11 31 31 am" src="https://user-images.githubusercontent.com/1056957/40316026-fbb1e9a0-5cea-11e8-8728-acfa6caa714c.png">

### 3. Scroll down table
There are many students in other grades that shouldn't be on the list.
<img width="996" alt="screen shot 2018-05-21 at 11 31 40 am" src="https://user-images.githubusercontent.com/1056957/40315999-ec9f2ffe-5cea-11e8-8fec-fdd0059b60bf.png">

kevinrobinson:
@edavidsonsawyer FYI this came up looking at this with someone this morning.

edavidsonsawyer:
@kevinrobinson Ah! Nope, this never worked. Fixing!

kevinrobinson:
Closed by https://github.com/studentinsights/studentinsights/pull/1712.

PR #1712, opened 2018-05-22
restrict student table to selected category

edavidsonsawyer:
# Who is this PR for?
Dashboard users

students table doesn't restrict the students based on selection cf #1708 

# What does this PR do?
Properly restricts students in the table like so:

goodfilter

# Checklists

## Javascript QA

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
+++ b/app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
@@ -101,6 +101,15 @@ class SchoolDisciplineDashboard extends React.Component {
     });
   }
 
+  //For grades and classrooms, the students table should only show the relevant students
+  groupStudents() {
+    if (this.state.selectedChart === 'grade' && this.state.selectedCategory) {
+      return this.props.dashboardStudents.filter(student => student.grade === this.state.selectedCategory);
+    } else if (this.state.selectedChart === 'classroom' && this.state.selectedCategory) {
+      return this.props.dashboardStudents.filter(student => student.homeroom_label === this.state.selectedCategory);
+    } else return this.props.dashboardStudents;
+  }
+
   render() {
     const selectedChart = this.getChartData(this.state.selectedChart);
     const chartOptions = [
@@ -162,7 +171,7 @@ class SchoolDisciplineDashboard extends React.Component {
   }
 
   renderStudentDisciplineTable() {
-    const students = this.props.dashboardStudents;
+    const students = this.groupStudents();
     const studentDisciplineIncidentCounts = this.studentDisciplineIncidentCounts(this.state.selectedCategory);
     let rows =[];
     students.forEach((student) => {


app/assets/javascripts/school_administrator_dashboard/dashboard_components/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -101,6 +101,15 @@ class SchoolDisciplineDashboard extends React.Component {
     });
   }
 
+  //For grades and classrooms, the students table should only show the relevant students
```
kevinrobinson: I was expecting this would apply to all the selections the user might make.  Can you say more about what you're thinking here?

If we want this to only work for certain selections, then we might also want to think about communicating that somehow.  Right now it looks to me like if you click on the bar within any of the categories, the UI appears to sort by them all.

edavidsonsawyer: @kevinrobinson I've been thinking about clicking the bars as "show me the subcategory of students who fit in this group". Those subcategories are obvious with homeroom and grade, but less so for the other possible selections. If you click on, eg, Tuesday, should you see all students who might have incidents on that day, or just the ones who do? Showing all students seems more consistent to me, since in the case of grade/homeroom we show all eligible students whether or not they have incidents. 

Users might care less about the consistency I mentioned than just having a smaller list of students, in which case we could restrict those other categories just to students who have incidents. We could also do that in the case of the grade and homeroom categories, but I think that would get confusing since those are natural subgroups and users would probably expect to see all students within a grade or homeroom.

kevinrobinson: Ah got it, this is super helpful! 👍 

So when a user's looking at the chart showing a breakdown by a category, what might their intention be when they click on one of the categories (eg, a particular grade, a particular day of the week)?  I've been interpreting this intention as "tell me more about that."

So if Tuesday shows many more absences then Wednesday, when they click on that I'm interpreting that intention as "tell me more about what's going on on Tuesday."  So to map that to what you're describing, I'm thinking that means "update the table based on just incidents for that day I clicked on."

For homerooms, I've been interpreting clicking on a particular homeroom as "tell me more about what's going on with that homeroom."  This might suggest showing all students regardless of their absences, but since this page is all about helping folks find students who have been absent a bit, it seems reasonable to only include students with absences.

There might be other ways to think about this too, but my inclination would be towards "if a user clicks on a bar within a category, the selection is updated to filter by that particular subcategory (eg, 4th grade, absences on Tuesday)."  If you have other thoughts on this, or different ways we could resolve that's awesome too!

edavidsonsawyer: @kevinrobinson I see what you mean. I don't feel too strongly about this (say 4 on a scale of 1 to 10), but here are a few reasons why I wouldn't want to only show incidents in the grade/homeroom categories. 

1. If I'm trying to figure out what's happening in a homeroom and I select it for a deeper dive, students with no discipline incidents could be as interesting as the ones with a high number. Maybe nobody is using or will ever use the tool this way, but the composition and size of a classroom might tell you something about what's going on with that classroom and you lose that if you only show the students who've had some kind of problem. 

2. I might worry that the data is wrong if the list says it's showing a classroom, and I know a student is in this classroom, but I can't find them in the student chart. We would at least want to update the chart text to something like "Grade 6 (with incidents in timeframe)" or something like that. 

3. We show homerooms in the tardies and absences dashboards as well, and they also show all students in each homeroom when selected. So changing that here would mean we should change it there as well. It's not a lot of work, but it is more change for users. 

Outside of grade and homeroom, thinking about it more it does seem sensible to just show students with incidents for the other categories. The above arguments don't really apply to "things that happened in the lunchroom" or "at 3PM", so I'm good with only showing students who have incidents that match that criteria.

kevinrobinson: @edavidsonsawyer Gotcha.  Let's move this forward, do you feel comfortable deciding what you think is best given this good discussion, commenting or making any update, and then let's 🚢and make some new improvements?

kevinrobinson:
🚢 !

Issue #1696, opened 2018-05-11
Classrooms: Ship to all teaching teams


kevinrobinson:
This is the final step of https://github.com/studentinsights/studentinsights/issues/1659 and https://github.com/studentinsights/studentinsights/pull/1664.  After this will come https://github.com/studentinsights/studentinsights/issues/1695.

This includes any final polish, cleanup and updates before release.  The major pieces are:
- [x] code review on server code (https://github.com/studentinsights/studentinsights/pull/1720)
- [x] make submit steps explicit (https://github.com/studentinsights/studentinsights/pull/1713 and https://github.com/studentinsights/studentinsights/pull/1714).
- [x] updates to match video and copy (https://github.com/studentinsights/studentinsights/pull/1713 and https://github.com/studentinsights/studentinsights/pull/1717)
- [x] UX or performance polish (https://github.com/studentinsights/studentinsights/pull/1716, https://github.com/studentinsights/studentinsights/pull/1719, https://github.com/studentinsights/studentinsights/pull/1721, https://github.com/studentinsights/studentinsights/pull/1723, https://github.com/studentinsights/studentinsights/pull/1724, https://github.com/studentinsights/studentinsights/pull/1725, https://github.com/studentinsights/studentinsights/pull/1732)
- [x] data quality issues related to homerooms and grades (https://github.com/studentinsights/studentinsights/issues/1693, https://github.com/studentinsights/studentinsights/pull/1728)
- [x] DIBELS chart bug fixes (https://github.com/studentinsights/studentinsights/pull/1729, https://github.com/studentinsights/studentinsights/pull/1730)

kevinrobinson:
Going to consider this done for now, although the process is not done for all teams yet. 🚢   

https://github.com/studentinsights/studentinsights/issues/1695 and https://github.com/studentinsights/studentinsights/issues/1722 are next.

PR #1718, opened 2018-05-22
Classrooms: Tighten param log scrubbing a bit more

kevinrobinson:
Part of https://github.com/studentinsights/studentinsights/issues/1696, removing educator names from logging as well.

Example locally:
screen shot 2018-05-22 at 11 43 19 am


kevinrobinson:
selfie @alexsoble if you can look at some point later this week that'd be awesome!

PR #1721, opened 2018-05-23
Add ClassList index for (workspace_id, created_at)

kevinrobinson:
# Who is this PR for?

Speeds up queries doing scans and sorts.

# What does this PR do?
Adds an index, verified this works as expected locally:

```
> ClassList.all.where(workspace_id: workspace_id).order(created_at: :desc).limit(1).explain
 Limit  (cost=0.27..4.40 rows=1 width=350)
   ->  Index Scan using index_class_lists_on_workspace_id_and_created_at on class_lists  (cost=0.27..20.90 rows=5 width=350)
         Index Cond: ((workspace_id)::text = '8f6aa764-4ca9-441c-9e1f-99b0d7c1cfa6'::text)
```


kevinrobinson:
selfie

PR #1723, opened 2018-05-23
Classrooms: Performance optimization for class list validations

kevinrobinson:
# Who is this PR for?

The existing validation code is slow when there are many records (testing with ~10k total, ~5k in one `workspace_id`).

# What does this PR do?
This refactors the validation to optimize it, using a single query doing a SQL `group by` instead of doing this in ActiveRecord style.  No spec changes, specs pass as they are.


kevinrobinson:
selfie

PR #1727, opened 2018-05-23
Classrooms: Add restrictions on when users can change school and grade

kevinrobinson:
# Who is this PR for?

When this was first made, the sequence was linear and users could move back and forth freely.  Changing that made the state transitions more complicated, and a scenario that wasn't handled was if a user changed the school or grade after either the list of students or the existing class list had been loaded (or requested).  @alexsoble ran into this right away in helpful testing things out yesterday.

# What does this PR do?
This restricts when users can change those values.  An assumption is that folks will proceed linearly, but if not and they do want to go back to change this, there's a message telling them to just go create a new class list instead.

screen shot 2018-05-23 at 9 47 07 am

# Checklists


kevinrobinson:
selfie

PR #1731, opened 2018-05-23
Add High School Transition Note with prompt and styling, restrict to counselors 

alexsoble:
# Who is this PR for?

Counselor Educators at K-8 schools who are going to use Insights to write Transition Notes for 8th grade students the week of June 12th. The transition notes will be used in transition meetings with their Housemaster counterparts at the High School and will replace the old paper Transition packets.

# What does this PR do?

This adds a new Event Note Type for High School Transition, and adds new functionality and special changes for that type of note:

+ When an educator selects "High School Transition" note type, the note automatically populates with a set of prompts adapted from the Transition Packet with input from Les (Counselor Educator at West Somerville). 
+ Introduces a new `:is_counselor` field for educators so that only counselors will be able to write High School Transition Notes. This field defaults to false, which will let us ship the feature dark.
+ For educators who have permissions to read a student's profile (not just counselors), the High School Transition Note will appear "pinned" to the top of the note feed and emphasized with a bright border. 
  + The goal of these design choices is to make sure the transition notes don't get lost in a busy/large note feed, since they should contain important information about a student overall, as opposed to other notes which reflect day-to-day updates. 

# GIF (writing transition note)

Log in as Sofia Counselor, the Counselor role in Test Pals, to add a High School transition note: 

sofia

# GIF (reading transition note)

Log in as Uri, a non-counselor, to read the note Sofia left:

uri


db/schema.rb
```diff
@@ -124,6 +124,7 @@
     t.boolean "districtwide_access", default: false, null: false
     t.boolean "can_set_districtwide_access", default: false, null: false
     t.text "student_searchbar_json"
+    t.boolean "is_counselor", default: false
```
alexsoble: Since this will default to false for all educators, we can turn this on incrementally, first for user testing with Les and then opening it up to all Counselor Educators.

app/assets/javascripts/student_profile/take_notes.jsx
```diff
@@ -97,6 +98,45 @@ export default React.createClass({
...
+
+What is this student's involvement in the school community like?
+——————————
+
+
+How does this student relate to their peers?
+——————————
+
+
+Is this student receiving Social Services and if so, what is the name and contact info of their social worker?
```
alexsoble: Les's feedback on these initial prompts: 

> Questions 2 & 3 are NOT too similar, and I think they could be left as is.

> Questions 4 & 5 on the other hand could be merged into a combined question that I will think about wording.

Les also suggested merging the first and the last question into "Any additional strengths, comments, or good things to know about this student?" I wonder if there's some value into explicitly starting off the prompts with a positively-framed question.

app/assets/javascripts/helpers/FeedHelpers.js
```diff
@@ -23,7 +27,15 @@ export function mergedNotes(feed) {
   });
 
   const mergedNotes = eventNotes.concat.apply(eventNotes, [deprecatedInterventions]);
-  return _.sortBy(mergedNotes, 'sort_timestamp').reverse();
+
+  return mergedNotes.sort((note_a, note_b) => {
+    if (highSchoolTransition(note_a) && !highSchoolTransition(note_b)) return -1;
+    if (!highSchoolTransition(note_a) && highSchoolTransition(note_b)) return 1;
```
alexsoble: This is the "pin to top of Notes feed" code.

alexsoble:
Goal is to ship this PR by the end of the week to unblock initial user testing with Les in-app next week. Plenty of time to iterate both this week and next.

db/migrate/20180523162842_add_is_counselor_to_educator.rb
```diff
@@ -0,0 +1,5 @@
+class AddIsCounselorToEducator < ActiveRecord::Migration[5.1]
+  def change
+    add_column :educators, :is_counselor, :boolean, default: false
```
kevinrobinson: There's a `staff_type` field that has some "Counselor" values if I remember right, that might be worth checking out.

If not, this [EducatorLabel](https://github.com/studentinsights/studentinsights/blob/master/app/models/educator_label.rb#L1.) class might be useful for this purpose, it's what I used to label whether folks are on the NGE/10GE teams.

app/assets/javascripts/student_profile/take_notes.jsx
```diff
@@ -97,6 +98,45 @@ export default React.createClass({
...
   },
 
+  noteText() {
+    // In most cases, display the text the user has entered into the controlled
+    // text input, passed in as props. When the user is a counselor entering a
+    // High School Transition Note, show an initial prompt.
+
+    const {noteInProgressText, noteInProgressType} = this.props;
+
+    if (noteInProgressType !== 307) return noteInProgressText;
```
kevinrobinson: Adding branching like this in a few different places is a good signal that the new thing might be somewhat different than the existing similar things (here, the transition packet and notes).  This can be a great path for building quickly, since it reuses other related code!  The tradeoff is that it's adding complexity across a few separate components rather than isolating it to a separate component or place.  I don't have a suggestion here, this just seems a good thing to be mindful of as you're iterating!

alexsoble:
Closing this one in favor of #1742. This served as a great initial mockup to get feedback on though!

PR #1732, opened 2018-05-23
Classrooms: Improve sort order for class list workspaces

kevinrobinson:
# Who is this PR for?

The sort order is the `created_at` field, but as new `class_list` records are added, this keeps changed and the sort order isn't stable as people work and the "Created at" label is misleading.  It's also hard to scan to find a school or grade, and the revisions count isn't useful except for to developers.

# What does this PR do?
Reformats the table.

### before
screen shot 2018-05-23 at 5 45 09 pm

### after
screen shot 2018-05-23 at 5 44 25 pm

# Checklists


kevinrobinson:
selfie

Issue #1695, opened 2018-05-11
Classrooms: Principals review and edit classlists, then handoff to secretaries


kevinrobinson:
This is the next step from https://github.com/studentinsights/studentinsights/issues/1659.  Principals review and edit the class lists that teaching teams created, considering any other changes like updates to ACCESS scores, input from ELL or special education teachers, late parent requests, etc.

Principals review and edit, and then handoff to secretaries to enter into Aspen.

This may also involve assigning staff names to next year's classrooms or aligning to the homeroom numbers in Aspen, not sure yet.

The current target date for this is first week of June.

Major pieces:
- [x] review and edit (https://github.com/studentinsights/studentinsights/pull/1744, https://github.com/studentinsights/studentinsights/pull/1746)
- [x] handoff spreadsheet to secretaries, verify method works in IE11 (https://github.com/studentinsights/studentinsights/pull/1744)
- [x] polish, performance and bug fixes  (https://github.com/studentinsights/studentinsights/pull/1733, https://github.com/studentinsights/studentinsights/pull/1747)

After this, https://github.com/studentinsights/studentinsights/issues/1722 is the final cleanup and refactoring step for this round of work.

kevinrobinson:
Considering this shipped; there will be more analysis in the fall, and potentially other things that come up in the summer after the rollover, but nothing else to do here now.

PR #1733, opened 2018-05-24
Fix timezone in display of times in class lists"

kevinrobinson:
# Who is this PR for?
K-5 teaching teams, part of https://github.com/studentinsights/studentinsights/issues/1695.

Times are displayed in UTC time rather than local time.  This is an issue throughout the JS code and I tried to fix this more holistically in https://github.com/studentinsights/studentinsights/tree/patch/utc-time-for-class-lists but am tabling that for now to get this change out first.

# What does this PR do?
Expresses the time in local time.


kevinrobinson:
selfie

PR #1737, opened 2018-05-25
Classrooms: Fix coloring bug with medium STAR scores

kevinrobinson:
# Who is this PR for?

STAR colors aren't the same for the chart up top and student cards.  Cards don't color medium, and they should be yellow.

# What does this PR do?
Also removes some duplication and factors it out.  Also fixes a bug where one place was calculating "greater than" and the other was calculating "greater than or equal."

### before
screen shot 2018-05-25 at 5 47 46 pm

### after
screen shot 2018-05-25 at 5 47 58 pm

# Checklists


kevinrobinson:
selfie

PR #1739, opened 2018-05-29
Finish ES6 conversion and move tests next to code

kevinrobinson:
# Who is this PR for?
developers, part of https://github.com/studentinsights/studentinsights/issues/1180.

Making it so code is all in modules, simpler to work with, code and test and stories are in the same places, etc.

# What does this PR do?
- converts the rest of the student profile page components to use ES6 classes
- remove the use of `React.PropTypes` in favor of the standalone `prop-types` module
- moves test code from `spec` and next to the real code
- It also merges in https://github.com/studentinsights/studentinsights/pull/1757, see https://github.com/studentinsights/studentinsights/pull/1757#issuecomment-395983092 for rationale.

Tests and linter all pass, but this needs more manual testing before shipping and verification after.
+ [ ] Manual testing on demo site
+ [ ] Manual testing in production after deploy


kevinrobinson:
@alexsoble I thought this might be helpful as you're doing more work in https://github.com/studentinsights/studentinsights/pull/1731, making it a bit simpler to add new tests or stories.  If that feels helpful I can resolve merge conflicts and take care of those bits so you can stay focused on shipping features.

Or if this feels like it's a distraction from the feature work and not worth doing right now, it's totally cool to push this off for a few weeks and just let it sit for a while.  Let me know whatever you think is best I'm happy to help in whatever direction!

kevinrobinson:
@alexsoble Sounds good!  Let's put this on hold then.

kevinrobinson:
deploying to demo to shake the tires manually

kevinrobinson:
@alexsoble trying again on the demo site now

kevinrobinson:
@alexsoble We chatted about this earlier, but no need to review this.  The diff is huge since most of this is renaming and moving folders of files.  I'm going to test this manually on the demo site and then in production after.

kevinrobinson:
Okay, manual testing revealed two bugs, one with tooltips on profile charts (https://github.com/studentinsights/studentinsights/pull/1739/commits/0ad6ab25bd73218c7918eebd3d111f87922e59de) and another with reading checkboxes for PDF download (https://github.com/studentinsights/studentinsights/pull/1739/commits/999e7058654e2b7c684d5fac8fa2e9fba8897f31).  It also revealed a few other minor bugs, which I'll add issues for separately.

Going to 🚢 this now and manually verify in production.

alexsoble:
🎉 🎉 🎉  This is a major accomplishment @kevinrobinson!  🎉 🎉 🎉 

Thanks for pushing us past the final mile on ES6 conversion. 

I looked through the PR for important changes and noticed a new testing support class, `NowContainer.js` (not 100% sure if this was introduced in this PR but it looks like it). Next time we pair, would love to discuss this for a bit so I can learn how to use it.

Cheers and 🎉congrats🎉 again!

kevinrobinson:
@alexsoble 👍 There's no real new code here, just changes related to the plumbing.  So `NowContainer` isn't new but happy to chat whenever.  The short version is that it's for freezing time during tests, and you can see any example usage [here](https://github.com/studentinsights/studentinsights/blob/master/app/assets/javascripts/feed/BirthdayCard.test.js#L23), where the component needs to know the current time in order to know how to render something correctly.

alexsoble:
@kevinrobinson OK thanks, that's helpful! And thanks for pointing to a very isolated and easy-to-grasp example test — I saw examples all over the PR but looking at one that's as isolated as the Birthday Card spec was very helpful.

PR #1740, opened 2018-05-29
Analysis: Look at gaps in low grades and high absences across roles

kevinrobinson:
# Who is this PR for?
team members, dept. heads.

It's hard to see what the "insight boxes" show for all staff.  This is a rake task that prints that out as text file.  This isn't in the product since we want to limit this kind of evaluation to one-off conversations rather than having this become a kind of scorecard.

# What does this PR do?
Adds a rake task running the high absences / low grades suggestions across a group of teachers.


kevinrobinson:
selfie

PR #1741, opened 2018-05-30
Add code comments for iep import

alexsoble:
# Who is this PR for?

Developers. 

Lack of clear code comments around IEP import job.

# What does this PR do?

Clarifies how IEP ZIP files flow into the SFTP site from Somerville's IEP system; ensure that bulk IEP upload ENV variable is set before we call the `bulk_import!` method.


alexsoble:
selfie

PR #1742, opened 2018-05-30
High School Transition Note V2

alexsoble:
Iterate on design sketched in #1731 based on feedback from @kevinrobinson and @snoopyuri.

# Who is this PR for?

Counselor Educators at K-8 schools who are going to use Insights to write Transition Notes for 8th grade students the week of June 12th. The transition notes will be used in transition meetings with their Housemaster counterparts at the High School and will replace the old paper Transition packets.

# What does this PR do?

+ Adds a new table, `transition_notes` to store the note data. 
+ Adds new UI for K8 Counselors and House Masters to write and read Transition Notes within the Student Profile Page. This UI separates out regular transition note content from restricted content. 
+ Uses `EducatorLabel` to tag K8 Counselors and House Master; adds K8 Counselor to Test Pals. 

# Rollout 

In this PR, Transition Notes are visible to K-8 Counselor-Educators only. A future PR could merge in Transition Notes with Event Notes so that all educators with permissions to view a student can view their Transition Note. 

# GIF 

## Counselor-Educator writing transition note:

les

# Screenshots 

## Transition note data sent to UI (Counselor-Educator)
screen shot 2018-05-30 at 7 21 41 pm

## Transition note data sent to UI (non Counselor-Educator)
screen shot 2018-05-30 at 7 22 07 pm

### DIFF
--- a/spec/controllers/district_controller_spec.rb
+++ b/spec/controllers/district_controller_spec.rb
@@ -30,9 +30,19 @@ describe DistrictController, :type => :controller do
             "school_type"=>"HS",
             "name"=>"Somerville High",
             "local_id"=>"SHS",
-            "slug"=>"shs"
+            "slug"=>"shs"}
+        }, {
+          "enrollment"=>1,
+          "grade"=>"8",
+          "school"=>
+           {"id"=>pals.west.id,
+            "school_type"=>"ESMS",
+            "name"=>"West Somerville Neighborhood",
+            "local_id"=>"WSNS",
+            "slug"=>"wsns"
+          }
         }
-      }])
+      ])
     end
 
     it 'guards authorization' do


app/helpers/transition_notes_helper.rb
```diff
@@ -0,0 +1,2 @@
+module TransitionNotesHelper
+end
```
kevinrobinson: remove?

app/controllers/students_controller.rb
```diff
@@ -34,6 +34,7 @@ def show
       event_note_types_index: EventNoteSerializer.event_note_types_index,
       educators_index: Educator.to_index,
       access: student.latest_access_results,
+      transition_notes: student.transition_notes,
```
kevinrobinson: Will this send down restricted notes, regardless of whether the user has access to view restricted notes or to this feature?  Maybe add an authorization check function for this in `authorizer`.

app/controllers/transition_notes_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    if transition_note.save
+      render json: { result: 'ok', is_restricted: transition_note.is_restricted }
+    else
+      render json: { errors: transition_note.errors.full_messages }, status: 422
+    end
+  end
+
+  def update
+    transition_note = TransitionNote.find_by_id(params[:id])
```
kevinrobinson: Similar security comment here.  There's a few different student_ids referenced here: `params[:student_id]` the `params[:transition_note][:student_id]` and then the `student_id` that's on the `TransitionNote` record.

To be safe, you could raise if these aren't all the same.

app/controllers/transition_notes_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      render json: { result: 'ok', is_restricted: transition_note.is_restricted }
+    else
+      render json: { errors: transition_note.errors.full_messages }, status: 422
+    end
+  end
+
+  private
+
+    def transition_note_params
+      params.require(:transition_note).permit(
```
kevinrobinson: Is this a security hole?  I can't tell if it's guaranteed that the `student_id` param here is the same as the `Student.find(params[:student_id]` in the before filter.  It'll be safe as long as both places read the same value, or if you check they match and raise if not.

kevinrobinson: Actually I don't think `transition_note_params` can be the same for create / update.  We can't allow an `id` field in `create`.

app/models/transition_note.rb
```diff
@@ -0,0 +1,7 @@
+class TransitionNote < ApplicationRecord
+  belongs_to :educator
```
kevinrobinson: This is who recorded the note, right?

spec/controllers/students_controller_spec.rb
```diff
@@ -56,7 +56,8 @@ def make_request(options = { student_id: nil, format: :html })
 
         it 'assigns the student\'s serialized data correctly' do
           make_request({ student_id: student.id, format: :html })
-          expect(serialized_data[:current_educator]).to eq educator
+          expect(serialized_data[:current_educator]['id']).to eq educator['id']
+          expect(serialized_data[:current_educator]['email']).to eq educator['email']
```
kevinrobinson: Should this have labels here now too?

spec/controllers/transition_notes_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe TransitionNotesController, type: :controller do
+
+end
```
kevinrobinson: remove?

spec/factories/transition_notes.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
+  factory :transition_note do
+
+  end
+end
```
kevinrobinson: remove?

db/migrate/20180528181001_create_transition_notes.rb
```diff
@@ -0,0 +1,13 @@
+class CreateTransitionNotes < ActiveRecord::Migration[5.1]
+  def change
+    create_table :transition_notes do |t|
+
+      t.timestamps
+      t.belongs_to :educator, index: true, foreign_key: true
+      t.belongs_to :student, index: true, foreign_key: true
+      t.text "text"
+      t.datetime "recorded_at"
+      t.boolean "is_restricted", default: false
```
kevinrobinson: It seems like it might match the design more closely if this were one record for the restricted and non-restricted fields.  Then you could enforce a uniqueness on "one transition note per student" whereas here the database allows any number of these.

Or is the idea that multiple counselors would record multiple notes?

app/assets/javascripts/student_profile/student_profile_page.jsx
```diff
@@ -178,13 +180,34 @@ import {cumulativeByMonthFromEvents} from './QuadConverter';
...
         </div>
       );
     },
 
+    renderTransitionNote: function() {
+      const {currentEducator, actions, transitionNotes, requests} = this.props;
+      const labels = currentEducator.labels;
+
+      const isElemCounselor = _.includes(labels, 'k8_counselor');
+      const isHouseMaster = _.includes(labels, 'high_school_house_master');
```
kevinrobinson: Consider moving these to `PerDistrict`: https://github.com/studentinsights/studentinsights/blob/e68ab2fad2b10bf51bb1fc85415af4addff29939/app/assets/javascripts/helpers/PerDistrict.js#L46

app/assets/javascripts/student_profile/page_container.jsx
```diff
@@ -152,6 +153,33 @@ function fromPair(key, value) {
         .fail(this.onSaveNotesFail);
     },
 
+    onClickSaveTransitionNote: function(noteParams) {
```
kevinrobinson: This might feel too late, but something to consider here is cutting-pasting the changes in this component into `<TransitionNote />`.  The `TransitionNote` UI code is nicely wrapped up in its own component, and it could hold these callbacks and state itself, keeping it isolated and safer and cheaper to iterate over there.  Then you don't have to worry about any interactions with anything in this complicated bigger component and it's easier to test in isolation too.

Alternately you could cut-paste this code into a ` <TransitionNoteContainer />` that the main profile page UI calls, and that has these callbacks and state, and then it passes those all as props when rendering `<TransitionNote />` as it is.

app/assets/javascripts/student_profile/TransitionNotes.js
```diff
@@ -0,0 +1,169 @@
...
+    );
+  }
+
+  renderButton(onClickFn, buttonTextFn) {
+    const {readOnly} = this.props;
+
+    if (readOnly) return null;
+
+    return (
+      <button onClick={onClickFn} className='btn save'>
```
kevinrobinson: You can use `Button` in `../components/Button` if you like.  Then you don't need to remember the CSS detail.

app/assets/javascripts/student_profile/TransitionNotes.js
```diff
@@ -0,0 +1,169 @@
...
+            value={restrictedNoteText}
+            onChange={(e) => this.setState({restrictedNoteText: e.target.value})}
+            readOnly={readOnly} />
+          {this.renderButton(this.onClickSaveRestricted, this.buttonTextRestricted)}
+        </div>
+      </div>
+    );
+  }
+
+  renderButton(onClickFn, buttonTextFn) {
```
kevinrobinson: A small simplification might be to pass `buttonText` as a value.  Passing values instead of functions means less moving pieces.

kevinrobinson:
@alexsoble re: permissions for 9th grade housemasters - here's two options for 6/12 (with the caveat that I don't have much context for that actual meeting time):
- [ ] have K8 counselors use their computers during the meeting
- [ ] add an ENV switch that `is_authorized_for_student` respects, letting housemasters view 8th grade students who have transition notes when it's turned on

I'd also suggest bias towards something that's easy to turn off and throw away :)  If it's cheap to throw out, then it's easy to experiment and do something one-off since we know we don't have to solve the more general case and can tackle that when we have more time and knowledge this summer.  If the change is deeply embedded in other code, then it becomes harder to untangle after.

app/helpers/transition_notes_helper.rb
alexsoble: 👍

app/controllers/students_controller.rb
alexsoble: 👍

db/migrate/20180528181001_create_transition_notes.rb
alexsoble: Ooooh, I like that data structure you suggest. I think it's a little late to refactor in that direction though since I'm user-testing tomorrow.

app/controllers/transition_notes_controller.rb
alexsoble: Added a guard for this in `authorize!`.

app/lib/authorizer.rb
```diff
@@ -98,6 +98,7 @@ def authorized_when_viewing_as(view_as_educator, &block)
   def is_authorized_for_student?(student, options = {})
     begin
       return true if @educator.districtwide_access?
+      return true if @educator.labels.include?('high_school_house_master') && student.grade == '8' && ENV['HOUSEMASTERS_AUTHORIZED_FOR_GRADE_8']
```
alexsoble: There are currently no `"high_school_house_master"` labels in the database, so this auth feature can be manually QA'ed in production before we turn it on for housemasters.

alexsoble:
Specs are failing because of an issue unrelated to this PR. Our yarn.lock is expecting to find a package at https://registry.yarnpkg.com/@types/node/-/node-9.4.3.tgz but not finding anything. Checking here — https://registry.yarnpkg.com/@types/node/~ — looks like that version is out of date/unsupported anymore any we need to update.

alexsoble:
Rebased on top of master so that the specs can run after merging #1743.

alexsoble:
Deployed and added the `"k8_counselor"` label to the test user to do some manual verifying — looks good in prod! Excited for user testing tomorrow.

app/controllers/transition_notes_controller.rb
kevinrobinson: Yeah, for sure there are places where we should tighten this in other controllers too.

PR #1744, opened 2018-05-31
Classrooms: UI code for principals to make revisions and export as spreadsheet

kevinrobinson:
# Who is this PR for?
K-5 principals, part of https://github.com/studentinsights/studentinsights/issues/1695.

After https://github.com/studentinsights/studentinsights/issues/1659, teaching teams made their class lists.  Principals can review them, but can't make revisions and can't export the lists as spreadsheets (for those that send home class lists before the end of the year).

# What does this PR do?
It updates the UI code to allow a new phase where principals can make revisions to class lists and export them.  It determines the principal role with an `EducatorLabel` intended to be scoped just to this feature (rather than more generally describing who a principal is).

```
The overall flow is that:

1. A teacher makes a new class list workspace.
   For them, it's editable.
   For other teachers, or principals or district admin, it's readonly.
2. That teacher submits the class list.
   For them, it changes to readonly.
   For other teachers and district admin, it's still readonly too.
   For the principal of that school, it's "revisable."  The original teacher's
   work is still readonly, but the principal can make revisions stored in new fields,
   and can also export the class lists.
3. For the teaching team and district admin, the principal's revisions are visible
   but are readonly.
```

On the server, this also updates the controller to send down the `local_id` for export.  And it adds two methods for looking at drift in what students were in the grade at the time the lists were made, and what students are in the grade at a later point in time viewing the lists.  Using this server code, I looked at the few cases of this in production and updated the UI code to resolve this drift (right now it throws a JS error).

This PR *does not* update the server code to allow those users to actually save revisions.  So on deploy, if a user was labeled as `class_list_maker_finalizer_principal` the UI would show them the UI to make revisions and would send those revisions to the server, but the server would reject them as coming from a user who isn't authorized to make writes.  Since there's a lot of changes here, the intent here is to deploy incrementally and to verify that the UI changes don't regress and resolve the issues with drift before touching the authorization code for principal revisions next.

### when revising, navigation and principal badge in the upper right
screen shot 2018-05-31 at 6 08 04 pm

### principal revisions, highlighting students that were moved
screen shot 2018-05-31 at 6 09 49 pm

### adding teacher names and exporting
screen shot 2018-05-31 at 6 34 38 pm

### prompts to add names or notice unplaced students
screen shot 2018-05-31 at 6 34 56 pm

### exported spreadsheet
screen shot 2018-05-31 at 6 12 14 pm

screen shot 2018-05-31 at 6 20 30 pm

# Checklists


config/environments/development.rb
```diff
@@ -5,13 +5,12 @@
...
 
-  ENV['DISTRICT_KEY'] = 'somerville'
 
+  ENV['DISTRICT_KEY'] = 'somerville'
   ENV['DISTRICT_NAME'] = 'Localhost Public Schools'
-
   ENV['USE_MOCK_LDAP'] = 'true'
-
   ENV['MOCK_LDAP_PASSWORD'] = 'demo-password'
+  ENV['ENABLE_CLASS_LISTS'] = 'true'
```
kevinrobinson: @alexsoble This is what triggered mentionbot.  This adds in `ENABLE_CLASS_LISTS` as true in development.

kevinrobinson:
@alexsoble I was planning on deploying this tomorrow morning and then working on the server-side changes that will accept revisions from principals.

Given a choice, I think it'd be more helpful if you can review the next set of changes since they'll touch server-side authorization.  If you have time to review this PR that's awesome, but no worries if not, just trying to prioritize review requests :)

kevinrobinson:
The download button method doesn't work in IE.  Going to punt that from this PR and come back to that after.

app/assets/javascripts/class_lists/ClassListCreatorPage.js
```diff
@@ -341,6 +377,11 @@ export default class ClassListCreatorPage extends React.Component {
     });
   }
 
+  // Over time, there can be drift in the references within studentIdsByRoom 
+  // and the set of students that the server provides.  We filter this out or add this in
+  // at the view layer and don't proactively filter this.  So if the user looks at the
+  // lists, this will trigger as "dirty" and the document will be updated, but
+  // if they never look at it we preserve information about that inconsistency.
```
alexsoble: More really helpful and thoughtful documentation!

app/assets/javascripts/class_lists/ClassListCreatorPage.js
```diff
@@ -462,22 +509,30 @@ export default class ClassListCreatorPage extends React.Component {
...
+    labels: React.PropTypes.arrayOf(React.PropTypes.string).isRequired
+  }).isRequired,
   defaultWorkspaceId: React.PropTypes.string,
   disableHistory: React.PropTypes.bool,
   disableSizing: React.PropTypes.bool,
   autoSaveIntervalMs: React.PropTypes.number
 };
 ClassListCreatorPage.defaultProps = {
-  autoSaveIntervalMs: 5000
+  autoSaveIntervalMs: 1000
```
alexsoble: Just curious, why are we increasing the autosave interval to once per second here? That seems like a lot of saving and a lot of hitting the server.

app/assets/javascripts/class_lists/ClassListCreatorPage.story.js
```diff
@@ -4,13 +4,15 @@ import {storiesOf} from '@storybook/react';
 import ClassListCreatorPage from './ClassListCreatorPage';
 import mockWithFixtures from './fixtures/mockWithFixtures';
 import storybookFrame from './storybookFrame';
+import {testProps} from './ClassListCreatorPage.test';
 
 storiesOf('classlists/ClassListCreatorPage', module) // eslint-disable-line no-undef
```
alexsoble: Curious to learn more about your development flow here! When you develop this feature, are you mostly looking at it in the context of the locally-hosted Rails app, or the Storybook, or a mix of both?

alexsoble:
@kevinrobinson I wasn't able to do a deep review on this PR since it's getting late, but I read through the diffs and left comments on areas that seemed interesting or that I had questions about!

app/assets/javascripts/class_lists/ClassListCreatorPage.js
kevinrobinson: Good catch!  For teaching teams, one thing that came up is folks didn't understand if their work was saving.  We could add a message about that, but the natural space in the horizontal bar showing the steps is too crowded to fit it.  For principals there's also no final "submit" step in the same way to be like "I am done."  So I just reduced the interval here.

In terms of the number of requests, the variable name here is probably misleading - the value is used to throttle how often calls to the server are made, not to make a call to the server every 1000ms.  So on PD day when teaching teams were all doing this at once, it could increase the volume of requests to the server.  But on that day we were only at 1/3rd of the volume we had planned for, so there's plenty of headroom still.  The place the interval most matters is for text fields when folks are doing sustained typing.  There the throttling really would reduce the number of requests quite a bit (eg, if you type straight for 10 seconds this will make ~10 requests whereas before it'd make ~2).  But for principals there's no place they're doing sustained typing, and so the rate they're making changes will be lower anyway.

So I think this will be okay, but really the better thing would be to make the saving state more explicit.  We might have time to do that next week but probably not today.

app/assets/javascripts/class_lists/ClassListCreatorPage.story.js
kevinrobinson: Kind of a mix.  This might be a good thing to demo Friday or some other time!

I used a story as the primary way to develop the "create your lists" component, since I could set up the fixtures and then wanted to just iterate on the look and feel for a while.  I find stories super helpful for "view-only" kind of components.  For other components, I'll use the story to see them in different states at once.  Other times, when there's interactions with the server or it's complicated to do the fixture setup I would use local development like normal.

The other thing I tried with this work, which worked pretty well, was having the test files export their setup (eg, `testProps` here).  Then the story can re-use that.  And then it's also oneliner to add a snapshot test to ensure that changes don't regress how the Story is rendering.

app/assets/javascripts/class_lists/ExportList.js
```diff
@@ -0,0 +1,254 @@
...
+    } = this.props;
+    const studentIdsByRoom = principalStudentIdsByRoom || teacherStudentIdsByRoom; // TODO(kr)
+    const rows = this.mapToRows(studentIdsByRoom);
+
+    return (
+      <div className="SecretaryEnters">
+        <div style={headingStyle}>Have all students been placed?</div>
+        <div style={styles.spaceBelow}>{this.renderUnplaced(studentIdsByRoom)}</div>
+        
+        <div style={headingStyle}>Have you reviewed the teaching team's plan and notes before making changes?</div>
```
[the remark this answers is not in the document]
kevinrobinson: Yeah this is a pain.  I googled for minute yesterday and couldn't figure it out but we should definitely sort this out since it makes reviewing these diffs super tough...

kevinrobinson:
@alexsoble awesome, thanks!

kevinrobinson:
Pushed a few more improvements, many IE-related:
- changing the styling for inputting teacher names, and using a combo box instead of a textbox
- fixing the download link in IE
- fixing a sizing bug on the stepper bar up top on low resolutions in IE11 
- revised the layout on the export step so that it fits in smaller resolution IE11 screens with many classrooms
- fixing a bug with newlines within the CSV on Windows machines
- working around an IE11 bug with react-select autosizing

Updated layout:
screen shot 2018-06-01 at 9 39 49 am

kevinrobinson:
half selfie, half awesome review from @alexsoble :)

kevinrobinson:
Verified this in production for the various issues here.  Still need more verification after https://github.com/studentinsights/studentinsights/pull/1746 lands.

PR #1745, opened 2018-06-01
Autosave High School Transition notes as user types; fix bug with saving

alexsoble:
# Who is this PR for?

K-8 Counselors taking transition notes. 

During user feedback, Les mentioned that he'd have to remind some of the counselors to "hit" save, otherwise they'd lose all their work. He suggested that autosaving would make the process easier. 

# What does this PR do?

+ Implements autosaving for the Transition Note typing flow.
+ Fixes bug with saving

# GIF 

current-code

### DIFF
--- a/app/assets/javascripts/student_profile/TransitionNotes.js
+++ b/app/assets/javascripts/student_profile/TransitionNotes.js
@@ -58,11 +58,14 @@ class TransitionNotes extends React.Component {
       restrictedNoteId: (restrictedNote ? restrictedNote.id : null)
     };
 
-    this.onClickSave = this.onClickSave.bind(this);
-    this.onClickSaveRestricted = this.onClickSaveRestricted.bind(this);
+    this.autosaveRegularNote = _.throttle(this.autosaveRegularNote.bind(this), 5000);
+    this.autosaveRestrictedNote = _.throttle(this.autosaveRestrictedNote.bind(this), 5000);
+
+    this.onChangeRegularNote = this.onChangeRegularNote.bind(this);
+    this.onChangeRestrictedNote = this.onChangeRestrictedNote.bind(this);
   }
 
-  onClickSave() {
+  autosaveRegularNote() {
     const params = {
       is_restricted: false,
       text: this.state.noteText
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/student_profile/TransitionNotes.js
```diff
@@ -67,10 +67,22 @@ class TransitionNotes extends React.Component {
...
+  // readable when it's supposed to say "Autosaving..."
+
+  // This code adds an artificial delay so that the text that says "Autosaving..."
+  // can hang for around 2 seconds to let the user read it.
+
+  // It occurred to me that the server might respond more slowly on the Heroku app,
+  // which would make this code unnecessary. So I'm going to try commenting it out,
+  // deploying up to the Heroku demo site, and seeing what the user experience
+  // is like.
   componentWillReceiveProps(newProps) {
```
kevinrobinson: In general `componentDidUpdate` is the better hook for executing some kind of side-effect on a state/prop change.  You can read more here if you're curious: https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops, but should be able to just swap method and args and keep your same code.

The other thing that allows is to remove the manual autosave calls, and just save every time the state changes in ways you care about.  There's an example [here](https://github.com/studentinsights/studentinsights/blob/master/app/assets/javascripts/class_lists/ClassListCreatorPage.js#L116) if that's helpful - the other trick is to `_.throttle` or `_.debounce` calls to the [actual save method](https://github.com/studentinsights/studentinsights/blob/master/app/assets/javascripts/class_lists/ClassListCreatorPage.js#L84).

alexsoble:
After commenting-out the "delay autosaving message text" code and pushing this to the demo site, I saw the same effect I was seeing locally: the "Autosaving..." message appears and disappears too quickly for the user to see. So I'm going to restore that delaying code to make the autosaving-status messages readable.

app/assets/javascripts/student_profile/TransitionNotes.js
alexsoble: When I just swap in `componentDidUpdate` as the function name, it works as expected. When I try to update the args too and reflect the fact that one passes in `newProps` and the other, `prevProps`, it stops working as expected. 🤔 🤔 🤔

alexsoble:
selfie

alexsoble:
🚢

app/assets/javascripts/student_profile/TransitionNotes.js
kevinrobinson: @alexsoble sounds good!  I'm happy to look at this with you sometime next week, in person even :)

PR #1746, opened 2018-06-01
Classrooms: Update server and client code to allow principals to save revisions

kevinrobinson:
# Who is this PR for?
K5 principals, part of https://github.com/studentinsights/studentinsights/issues/1695.  Follow-on to https://github.com/studentinsights/studentinsights/pull/1744.

In https://github.com/studentinsights/studentinsights/issues/1695, I punted on updating the server code to handle principal revisions.  So with merging that, the server will reject revisions.

# What does this PR do?
This PR updates the UI code to send data to one endpoint for teacher updates (like before), and to send data to a new endpoint for principal revisions (new capability).  I'm not sure if this is the best design, but I split this so that the server code with authorization complexity would be simpler, and the client just branches mostly transparently between which endpoint it talks to.

This PR also adds tests for the new endpoint, and an `is_authorized_to_revise?` method, treating "revising" different than the initial "writing" since they have different authorization rules.

No visual differences, other than that principal revisions are also now loaded.

# Checklists


kevinrobinson:
@alexsoble If you could review the new endpoint code and authorization in particular, [https://github.com/studentinsights/studentinsights/pull/1746/files#diff-7d52aaa95c56c7d49ce1bc9fca49b169R163](class_list_controller#@principal_revised_class_list_json) and [class_list_queries#is_authorized_to_revise ](https://github.com/studentinsights/studentinsights/pull/1746/files#diff-2c1caf1907835bbdbe960a695cb62909R129), that would be awesome!

Ideally we'd ship https://github.com/studentinsights/studentinsights/issues/1695 this afternoon, and then ship this PR later today or Monday.

alexsoble:
@kevinrobinson Looking at those endpoints now!

app/lib/class_list_queries.rb
```diff
@@ -126,6 +126,16 @@ def is_authorized_for_writes?(workspace_id)
     true
   end
 
+  # Only principals of that school can revise.  The authorization is tightly scoped,
+  # as being marked a principal isn't done in a general way, so the intention here
+  # is that this lives just within this feature.
```
alexsoble: 👍

app/controllers/class_lists_controller.rb
```diff
@@ -161,6 +160,32 @@ def update_class_list_json
...
+
+    # They must be authorized to read it
+    grade_level_now = GradeLevels.new.previous(latest_class_list.grade_level_next_year)
+    raise Exceptions::EducatorNotAuthorized unless queries.is_authorized_for_grade_level_now?(latest_class_list.school_id, grade_level_now)
+
+    # They must be authorized to revise it
+    raise Exceptions::EducatorNotAuthorized unless queries.is_authorized_to_revise?(latest_class_list)
+
+    # Write a new record
+    class_list = latest_class_list.dup
```
alexsoble: Ooh, this is neat! I wasn't familiar with ActiveRecord `.dup`, dug in a little bit here: https://apidock.com/rails/ActiveRecord/Core/dup.

One thing that wasn't immediately obvious to me here: do we save the principal's educator ID anywhere on the `class_list` record here? How can we trace a set of class list revisions back to the principal who made them?

alexsoble:
@kevinrobinson Also, when I try creating a class list as Teacher, Sarah and revising it as Principal, Laura, I get the following JS stack trace:

screen shot 2018-06-01 at 3 31 12 pm

app/controllers/class_lists_controller.rb
kevinrobinson: @alexsoble thanks, this is a great question!  Right now, each change any user makes is stored as a separate `class_list` record.  They each have a `workspace_id` field, which describes a logical "4th grade class last that teacher Sarah started."

So if you do:

```
ClassList.where(workspace_id: 'foo')
```

You'll get a bunch of records that show the changes over time, including the teacher edits and then later the principal edits.  The `created_by_educator_id` would always refer to the original teacher.

This PR didn't save the principal's id anywhere, I updated that now.  I also renamed `created_by_educator_id` to `created_by_teacher_educator_id` to make it a little less foot-gun, since the created by field usually means who created the Active Record model.

Overall, this is all a consequence of putting things in one table to track all the changes over time. :)  A good improvement here would be to migrate everything from one table into three table: `workspaces`, `teacher_edits` and `principal_revisions`.  That would make the validations simpler and express more explicitly what's happening here.  For now I added more explanation around this here: https://github.com/studentinsights/studentinsights/pull/1746/files#diff-8130f19224e55dd36a1da7f6ae8e5ddeR1

kevinrobinson:
@alexsoble Thanks!  Those JS errors are real bugs :)  Updating this now, there was a test case that I thought was checking this it was hidden because there's a test bug too.

kevinrobinson:
Okay, this last commit fixes the bug and test bug.  The test wasn't running everything I had intended it to, and then when I updated it to, the setup wasn't entirely right either.  This fixes all the things, with some manual testing too.

This will need some manual testing in production anyway in order to enable this for any principals.

Thanks @alexsoble! 👍

kevinrobinson:
🚢

kevinrobinson:
@alexsoble FYI doing some manual testing with the developer user in production.

kevinrobinson:
done now. 👍

PR #1748, opened 2018-06-01
Performance optimizations for querying for educator authorizations

kevinrobinson:
# Who is this PR for?
developers

The "Educator permissions: overview" link has been broken since shipping the class list links.  This is because this page sorts based on the navbar links shown to each user, and the change to check whether to show the class list link made this significantly slower to do across all users.

# What does this PR do?
Speeds this up.  This is the data on the new `PerfTest` method here, running on 1% of educators:

```
# before
ALL_FOR_EDUCATOR, sample size: 98
  all_for_educator          median: 137ms   p95: 201ms

# after
ALL_FOR_EDUCATOR, sample size: 98
  all_for_educator          median: 6ms   p95: 36ms
```

### DIFF
--- a/app/controllers/admin/educators_controller.rb
+++ b/app/controllers/admin/educators_controller.rb
@@ -20,17 +20,19 @@ module Admin
     end
 
     def authorization
+      @all_educators = Educator.all
+      
       @districtwide_educators = []
       @can_set_educators = []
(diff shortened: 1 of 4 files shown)


kevinrobinson:
selfie

PR #1749, opened 2018-06-01
Allow masquerading as different users

kevinrobinson:
# Who is this PR for?
developers

As we create features that have more involved authorization, or show different things to different users, it's harder for us to verify in production that these are working correctly.  For the home page work, we modified the `home_controller` to allow this in a bespoke way but we don't want to be writing authorization code like this for each feature like https://github.com/studentinsights/studentinsights/pull/1746 or https://github.com/studentinsights/studentinsights/pull/1745.

This PR is an attempt at solving this more generally.  The intention is to scope this only to developers at first, and potentially after to project leads.

# What does this PR do?
Adds two endpoints to "become" another user, and to "clear" that.  This works by setting a session variable, and by overriding `current_educator` to return that other user if the session variable is set.

For tests, there are unit tests for each file, controller tests, and Capybara RackTest tests.

Here's the flow for a developer:

### 1. sign in as a user with developer access
screen shot 2018-06-01 at 7 24 23 pm

### 2. notice email and links in the upper right
This shows the current user email.
screen shot 2018-06-01 at 7 23 36 pm

### 3. click on the name to jump to the educators overview page
This page needs love, but does the trick for now.  You can now click on "become" next to any educator.  This will make a POST request to `/admin/masquerade/become?masquerading_educator_id=12345` which sets a flag on the session and then redirects to the home page.
screen shot 2018-06-01 at 7 23 16 pm

### 4. home page, masquerading as another user
The upper right shows the user name, and an `X` to indicate you're masquerading.  The data on the page is all different, because all of the various requests the page is making are all running into the bit set on the session.
screen shot 2018-06-01 at 7 19 48 pm

### 5. clearing the masquerade
Clicking the x drops the masquerade by making a POST to `/admin/masquerade/clear` which clears the session and then redirects to the home page.
screen shot 2018-06-01 at 7 23 09 pm

If we wanted to move forward the steps would be to talk it out with @alexsoble, do a thorough review, deploy to demo and verify, and then deploy and verify for real.


kevinrobinson:
@alexsoble This is just experimenting, but I'm wondering if you think this would be helpful or more trouble than it's worth..

It might be helpful for verifying changes that cut across permissions.  This gets dangerous when the operations involve writes, and we'd probably want to add more durable logging around this is we decide to do it, but an option to consider.

kevinrobinson:
@alexsoble This pushes a slew of tests, probably the tests are 10x as much as the code.  There's unit tests, integration tests and Capybara tests using RackTest.

I also updated the styling to be super super visible when this is happening to start:
screen shot 2018-06-02 at 4 55 31 pm

spec/features/masquerade_feature_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+    it 'healey_sarah_teacher' do expect_masquerading_to_fail(pals.healey_sarah_teacher) end
+    it 'west_marcus_teacher' do expect_masquerading_to_fail(pals.west_marcus_teacher) end
+    it 'shs_jodi' do expect_masquerading_to_fail(pals.shs_jodi) end
+    it 'shs_bill_nye' do expect_masquerading_to_fail(pals.shs_bill_nye) end
+    it 'shs_ninth_grade_counselor' do expect_masquerading_to_fail(pals.shs_ninth_grade_counselor) end
+    it 'shs_hugo_art_teacher' do expect_masquerading_to_fail(pals.shs_hugo_art_teacher) end
+    it 'shs_fatima_science_teacher' do expect_masquerading_to_fail(pals.shs_fatima_science_teacher) end
+  end
+
+  # it 'does not work for any other users to masquerade as anyone else' do
```
kevinrobinson: This seems most thorough, but takes ~30 seconds to run through.  A compromise is the lines above, which explicitly tests those users trying to masquerade as the highest-permission user, which is a bit faster and gives feedback on each one.

kevinrobinson:
selfie cc @alexsoble, it'd be great to review this post-facto

alexsoble:
@kevinrobinson Yep, reviewing this on #1793.

PR #1750, opened 2018-06-02
Add invalid grade to student grade validation

kevinrobinson:
# Who is this PR for?
developers

Make it easier to see why a grade validation fails.

# What does this PR do?
Adds in the invalid grade value to the validation error message.


kevinrobinson:
selfie

PR #1751, opened 2018-06-02
Add district key to subject line in ErrorMailer

kevinrobinson:
# Who is this PR for?
developers

The stack trace doesn't tell you which deployment it's coming from.


kevinrobinson:
selfie

PR #1752, opened 2018-06-02
Fix code and snapshots that rely on wall clock

kevinrobinson:
# Who is this PR for?
developers

Failed build: https://travis-ci.org/studentinsights/studentinsights/builds/387124655

The `ExportList` code used `moment` to grab time from the wall clock, and one of the tests used a snapshot based on this, making the test only pass on one day :)  This updates the code to grab the clock from context, and updates the tests and stories.  It also does the same for `ClassListCreatorPage`, unaffected by this bug but with the same use of the wall clock.


kevinrobinson:
selfie

PR #1754, opened 2018-06-04
Class lists: Update spreadsheet row order so student names are first

kevinrobinson:
# Who is this PR for?
secretaries, follow-on to https://github.com/studentinsights/studentinsights/pull/1746 and part of https://github.com/studentinsights/studentinsights/issues/1695.

The row order for spreadsheets was wrong; secretaries want this to be alphabetical by student name since that's how they have to enter it in Aspen.


kevinrobinson:
selfie

PR #1755, opened 2018-06-04
Add validation for educator full_name, add names to Student factory as well

kevinrobinson:
# Who is this PR for?
developers

This came up from https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/140/, a bug where an educator name was nil.

# What does this PR do?
Adds a presence validation to the `Educator` model, updates factory and specs that expected nil names.  Adds a presence validation to the `Student` model for first and last names, updates factories and specs.

In the process of doing this, I removed the Faker gem and added `FakeNames` to generate first and last names.  Both factory classes use this.

# Checklists


kevinrobinson:
hmm, this is stuck somehow.  Two test files fail in ways that make sense in https://travis-ci.org/studentinsights/studentinsights/builds/387824521?utm_source=github_status&utm_medium=notification, but locally when I removed the Faker gem I get errors like:

screen shot 2018-06-04 at 10 51 11 am


```
I18n::InvalidLocaleData:
       can not load translations from /Users/krobinson/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/activesupport-5.1.6/lib/active_support/locale/en.yml: expects it to return a hash, but does not
```

kevinrobinson:
So what was happening with the translation exceptions was:
1. I removed the faker gem
2. That resulted in a version change for i8n, which faker had depended on
3. ActiveSupport also uses the i8n gem, which happens to call `File.open` to read in translations
4. The `iep_storer_spec.rb` file was mocking `File.open` so that it always returned a dummy string
5. So depending on how the spec was run, this might raise the translation exception when ActiveSupport loaded translation files

https://github.com/studentinsights/studentinsights/pull/1755/commits/484978752f16a1728068bb34f2a099659782d25a fixes the mocking.

kevinrobinson:
selfie

kevinrobinson:
This triggered a validation error for NB because of a rename that happened in the upstream SIS.  I fixed that manually, but we need to solve it more generally.  It's on the backlog.

For Somerville there are several of these, also all from name changes.  For now backing out that bit in https://github.com/studentinsights/studentinsights/pull/1774 and then separately we'll need to improve how the import process handles drift like this more generally.

PR #1756, opened 2018-06-04
Improve Transition Note autosaving UX

alexsoble:
# Who is this PR for?

Les and the other K8 Counselor-Educators who will use the transition note feature. 

Current autosave feature has a bug. Bug report from Les: 

> "Noticed that the AutoSave (Caching/Refresh Rates) has a small glitch. When typing a more rapid rate keystrokes are being missed as the auto save is happening. The little message/indicator is how I noticed this."

Here's a GIF where I reproduce my interpretation of this bug report locally: 

the-bug

Here's my understanding of this bug: the autosaving user interface shows "Saved!" at the same time as requests are being sent off to the server to save the rest of a user's input. So the user feels free to exit, but some of their work remains unsaved.

Why does this happen? The current code sets both `{leading: true, trailing: true}` for the lodash `_.throttle`. This means  that API requests are being fired off as soon as the user types the first letter into the `<textarea>`. This causes something like this to happen:

```
- user presses key: "A"
- send request out to the network with body: "A"
- user presses keys: "bracadbra!"
- network responds to the original request, says "Saved!"
- user thinks that the network is saying "Saved! to "Abracadbra!", but network only processed "A"
```

This kind of mis-syncing between UI and server states is exacerbated by two other choices I made:

+ Delaying the text status to make it hang on "Autosaving..." for a longer amount of time, in order to make transitions appear smoother for the user
+ Setting a high throttle wait period of 3 seconds in order to make things easier on the serve

# What does this PR do?

+ Changes lodash options to `{leading: false, trailing: true}`
+ Reduces throttle wait timing to 1000ms (1 second)
+ Removes the custom "delay" code I wrote
+ Make the status text say `"Saving..."` instead of `"Autosaving..."` when in a pending state; this reduces the amount of visual "jumping" when the request is handled quickly, since most of the letters in the status text stay the same.

# GIF

the-fix

### DIFF
--- a/app/assets/javascripts/student_profile/TransitionNotes.js
+++ b/app/assets/javascripts/student_profile/TransitionNotes.js
@@ -58,7 +58,7 @@ class TransitionNotes extends React.Component {
       restrictedNoteAutosaveStatus: null,
     };
 
-    const throttleOptions = {leading: true, trailing: true};
+    const throttleOptions = {leading: false, trailing: true};
 
     this.autosaveRegularNote = _.throttle(this.autosaveRegularNote.bind(this), 3000, throttleOptions);
     this.autosaveRestrictedNote = _.throttle(this.autosaveRestrictedNote.bind(this), 3000, throttleOptions);
--- a/app/assets/javascripts/student_profile/TransitionNotes.js
+++ b/app/assets/javascripts/student_profile/TransitionNotes.js
@@ -60,8 +60,8 @@ class TransitionNotes extends React.Component {
 
     const throttleOptions = {leading: false, trailing: true};
 
-    this.autosaveRegularNote = _.throttle(this.autosaveRegularNote.bind(this), 3000, throttleOptions);
-    this.autosaveRestrictedNote = _.throttle(this.autosaveRestrictedNote.bind(this), 3000, throttleOptions);
+    this.autosaveRegularNote = _.throttle(this.autosaveRegularNote.bind(this), 1000, throttleOptions);
+    this.autosaveRestrictedNote = _.throttle(this.autosaveRestrictedNote.bind(this), 1000, throttleOptions);
 
     this.onChangeRegularNote = this.onChangeRegularNote.bind(this);
     this.onChangeRestrictedNote = this.onChangeRestrictedNote.bind(this);
--- a/app/assets/javascripts/student_profile/TransitionNotes.js
+++ b/app/assets/javascripts/student_profile/TransitionNotes.js
@@ -54,8 +54,6 @@ class TransitionNotes extends React.Component {
     this.state = {
       noteText: (regularNote ? regularNote.text : notePrompts),
       restrictedNoteText: (restrictedNote ? restrictedNote.text : restrictedNotePrompts),
-      regularNoteAutosaveStatus: null,
-      restrictedNoteAutosaveStatus: null,
     };
 
     const throttleOptions = {leading: false, trailing: true};
@@ -65,39 +63,6 @@ class TransitionNotes extends React.Component {
 
     this.onChangeRegularNote = this.onChangeRegularNote.bind(this);
     this.onChangeRestrictedNote = this.onChangeRestrictedNote.bind(this);
-  }
-
-  // Note on componentWillReceiveProps:
-  // This method will be deprecated in React 16.3. But we're still using
-  // React 15.4 in Insights, and have some legacy-cleanup work to finish
-  // til we can migrate to 16. Noting that this func will have to be refactored.
-
-  // When the user types into the Transition Note textarea, the server
-  // handles autosaving very quickly. In fact, so fast that in the "pending" state,
-  // the status text is barely readable when it's supposed to say "Autosaving..."
-
-  // This code adds a delay so that the text that says "Autosaving..."
-  // can hang for  2 seconds to let the user read it.
-  componentWillReceiveProps(newProps) {
-    if (newProps.requestState !== this.props.requestState) {
-      if (this.props.requestState === 'pending') {
-        setTimeout(() => {
-          this.setState({regularNoteAutosaveStatus: newProps.requestState});
-        }, 2000);
-      } else {
-        this.setState({regularNoteAutosaveStatus: newProps.requestState});
-      }
-    }
-
-    if (newProps.requestStateRestricted !== this.props.requestStateRestricted) {
-      if (this.props.requestStateRestricted == 'pending') {
-        setTimeout(() => {
-          this.setState({restrictedNoteAutosaveStatus: newProps.requestStateRestricted});
-        }, 2000);
-      } else {
-        this.setState({restrictedNoteAutosaveStatus: newProps.requestStateRestricted});
-      }
-    }
   }
 
   autosaveRegularNote() {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/student_profile/TransitionNotes.js
```diff
@@ -54,52 +54,17 @@ class TransitionNotes extends React.Component {
     this.state = {
       noteText: (regularNote ? regularNote.text : notePrompts),
       restrictedNoteText: (restrictedNote ? restrictedNote.text : restrictedNotePrompts),
-      regularNoteAutosaveStatus: null,
-      restrictedNoteAutosaveStatus: null,
     };
 
-    const throttleOptions = {leading: true, trailing: true};
+    const throttleOptions = {leading: false, trailing: true};
```
kevinrobinson: The default behavior here is `leading: true, trailing: false`, I can't quite follow why we want to change this, can you explain?  Here's what I'm thinking would happen with the default behavior:

```
1. state changes, calls throttled fn
2. throttle calls underlying fn right away
3. other state changes get ignored for 1000ms
4. after 1000ms, then throttle calls underlying fn again
```

app/assets/javascripts/student_profile/TransitionNotes.js
```diff
@@ -133,7 +98,7 @@ class TransitionNotes extends React.Component {
   autosaveStatusText(status) {
     if (status === 'saved') return 'Saved.';
```
kevinrobinson: @alexsoble Here's two thoughts on simplifying this.  I think what you have here is a great improvement, so I think it completely makes sense to 🚢 this, and it's up to you if you think these other ideas are worth doing now or after or not at all.

First, this has two separate "threads" running that can modify the state on the server, one for the normal note and the other for the restricted note.  And the throttling is happening on a fn that's processing the state changes for the different text boxes - a few layers away from the actual server call.  This invites race conditions, since there are two separate ways that can make server calls, and they're throttled independently, rather than adding the throttling around the server call directly.  If the user modifies one text box it fires off a request, and then if they modify the second text box it fires off another request.  It'd be simpler to understand the communication with the server if you could throttle a single function that is communicating with the server.  Here, that might mean adding a `doSave` method, updating the other methods so that they call that instead of calling `this.props.onSave(params)` directly, and then only throttle `doSave`.

Second, as you described in your awesome PR description, the problem is related to when there are multiple requests.  Tracking only the state of the latest request isn't sufficient for understanding what's happening there, just like you describe.  This works in other places in the UI like when you're taking notes, because the UI essentially blocks everything to make sure there can only be one server request in-flight at once.

So instead of storing the state of the request to the server, another approach is to store "what does the client think the state of the server is?"  And then if that's different than the client state, you know there's unsaved changes.  Here that could mean instead of tracking the last change to any request state, track what the `lastServerState` is.  On load it's the same as whatever data is read from the server.  Then that's the value to check to see if the client thinks everything is saved.  If there are multiple requests to the server, and they go out and come back in out of order, this value gets updated whenever a response from the server comes back.  This still isn't perfect (since requests could finish in order on the server, but come back to the client out of order), but is probably good enough for most normal purposes.

app/assets/javascripts/student_profile/TransitionNotes.js
```diff
@@ -54,52 +54,17 @@ class TransitionNotes extends React.Component {
...
-    this.autosaveRegularNote = _.throttle(this.autosaveRegularNote.bind(this), 3000, throttleOptions);
-    this.autosaveRestrictedNote = _.throttle(this.autosaveRestrictedNote.bind(this), 3000, throttleOptions);
+    this.autosaveRegularNote = _.throttle(this.autosaveRegularNote.bind(this), 1000, throttleOptions);
+    this.autosaveRestrictedNote = _.throttle(this.autosaveRestrictedNote.bind(this), 1000, throttleOptions);
 
     this.onChangeRegularNote = this.onChangeRegularNote.bind(this);
     this.onChangeRestrictedNote = this.onChangeRestrictedNote.bind(this);
   }
 
-  // Note on componentWillReceiveProps:
```
kevinrobinson: 👍 simplifying!

app/assets/javascripts/student_profile/TransitionNotes.js
```diff
@@ -54,52 +54,17 @@ class TransitionNotes extends React.Component {
...
-  // React 15.4 in Insights, and have some legacy-cleanup work to finish
-  // til we can migrate to 16. Noting that this func will have to be refactored.
-
-  // When the user types into the Transition Note textarea, the server
-  // handles autosaving very quickly. In fact, so fast that in the "pending" state,
-  // the status text is barely readable when it's supposed to say "Autosaving..."
-
-  // This code adds a delay so that the text that says "Autosaving..."
-  // can hang for  2 seconds to let the user read it.
-  componentWillReceiveProps(newProps) {
```
kevinrobinson: Another thing that might be good to add is a warning when the user tries to close the page or navigate away, but the UI thinks there are unsaved changes.  There's a browser hook for this called `beforeunload` so this is easy to add, here's an example you could check out: https://github.com/studentinsights/studentinsights/blob/584775c509f94d433cb06c1ef769e472a02e0c40/app/assets/javascripts/class_lists/ClassListCreatorPage.js#L112

app/assets/javascripts/student_profile/TransitionNotes.js
alexsoble: My thinking was that in step (1), the state has only changed by one character since the user has just started typing.

If we enable `leading`, we're firing off a server call with a single character's worth of data. If we disable `leading`, we let the user type a meaningful word or two and then send that to the server. 

I tried restoring the default `_.throttle` settings locally and (subjectively) found that the overall experience didn't change too much, but did notice an immediate "Saving..." notice as soon as I started typing.

app/assets/javascripts/student_profile/TransitionNotes.js
alexsoble: 👍 — yup, I tried to get way too clever hear and it definitely made the user experience worse!

app/assets/javascripts/student_profile/TransitionNotes.js
kevinrobinson: Ah gotcha!  Yeah from a "saving all the data" perspective, it does make sense to batch a little bit.  In that case, `debounce` is the function for that I think.  The lodash docs link to a good article with examples about the difference.

Another perspective might be "we want to change everything as soon as possible, but just don't call to the server too much."  That has the nice effect of communicating to the user right away as soon as they change that saving has started.

Thanks for talking this out, either seems 👍 to me!

app/assets/javascripts/student_profile/TransitionNotes.js
alexsoble: Woah, this a fantastic way to change how I think about tracking state! This helps me see the advantages of some of the changes you suggested earlier, like consolidating down to one `doSave` call. I won't implement this today since it involves big API changes, and I want to get an improved UX version of this feature to Les by today. But I definitely want to implement this pattern either here or in the next component I write that involves sharing multiple streams of information between client/server!

alexsoble:
Thanks for the review comments @kevinrobinson! I'm going to ship this one incrementally to get a new-and-improved version to Les — that said I learned a ton from your comments and I can't wait to try out some of the ideas you shared!

PR #1762, opened 2018-06-05
Fix cursor jumping bug by wrapping Transition Note textarea

alexsoble:
# Who is this PR for?

Les and the other K8 Counselor-Educators who will use the transition note feature.

Cursor jumps to the end of the Transition Note textbox while user typing. 

My theory is that this happens when the server responds to a POST request from the client, updating its props. This forces re-render, which loses cursor position of inputs. 

# What does this PR do?

Wraps the textarea in question in its own component. My theory here is that the textarea inputs won't lose their cursor position if they're wrapped in a component this way. 

Why? The TransitionNoteTextbox.js component has three props. Two of them should never change (`readOnly` and `onChange`). The `value` prop is governed by the state of the parent component, which in turn is governed by user input. 

None of these props change when the server responds to the client's POST request: the state and props values that change, like `requestState`, and `requestStateRestricted`, aren't passed down to this lower-level component.

### DIFF
--- a/app/assets/javascripts/student_profile/TransitionNoteTextbox.js
+++ b/app/assets/javascripts/student_profile/TransitionNoteTextbox.js
@@ -1,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const styles = {
+  textarea: {
+    marginTop: 20,
+    fontSize: 14,
+    border: '4px solid rgba(153,117,185, 0.4)',
+    width: '100%'
+  }
+};
+
+class TransitionNoteTextbox extends React.Component {
+
+  render() {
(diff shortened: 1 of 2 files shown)


alexsoble:
Deploying this to demo site to test it there.

alexsoble:
Doing some manual testing on Chrome and IE confirmed that this bug is only happening on IE. That's why I never caught it. Need to respect the Github PR Template!! 

Here's a GIF of the bug in IE on the demo site:

bug-ie-res

alexsoble:
This latest commit didn't fix the bug. I've got another idea to try!

alexsoble:
Closing this because we decided to back out the autosave feature instead.

PR #1763, opened 2018-06-05
Back out autosaving feature from Transition Notes

alexsoble:
# Who is this PR for?

K8-Counselors. 

Autosaving was not good/usable in Internet Explorer; see #1762. Reverting back to a healthy state.

# What does this PR do?

Removes the autosaving.

saving


kevinrobinson:
with @alexsoble

PR #1766, opened 2018-06-06
Remove line breaks from transition note prompts

alexsoble:
# Who is this PR for?

K8 Counselor Educators.

The manual line breaks I added between prompts are confusing and ugly for users. 

# What does this PR do?

Removes the line breaks so that we see only the prompts. 

screen shot 2018-06-06 at 10 36 51 am


alexsoble:
selfie

PR #1767, opened 2018-06-06
Run migrations and seed database at the end of cleanup script

alexsoble:
# Who is this PR for?

Developers. 

Running cleanup script leaves database un-setup. 

# What does this PR do?

Runs migrations and seeds fake demo data at the end of cleanup script.


kevinrobinson:
👍 do it!

Issue #1722, opened 2018-05-23
Classrooms: Cleanup and refactoring


kevinrobinson:
After https://github.com/studentinsights/studentinsights/issues/1696.

- [x] refactoring and code cleanup (https://github.com/studentinsights/studentinsights/pull/1718, https://github.com/studentinsights/studentinsights/pull/1769, https://github.com/studentinsights/studentinsights/pull/1770, https://github.com/studentinsights/studentinsights/pull/1772, https://github.com/studentinsights/studentinsights/pull/1773)
- [x] edge cases around multi-grade classrooms (https://github.com/studentinsights/studentinsights/issues/1693, done enough for now)
- [x] fix layout bug for "expand" on "create your lists" page for small numbers of students (https://github.com/studentinsights/studentinsights/pull/1830)
- [x] fix bug report in https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/146/ (https://github.com/studentinsights/studentinsights/pull/1860

kevinrobinson:
Considering this closed out and 🚢 for now.  There's more we could do, but nothing pressing and we've run out of the time box for this.

PR #1769, opened 2018-06-08
Set ENABLE_CLASS_LIST for test, randomize rspec tests

kevinrobinson:
# Who is this PR for?
developers, cleanup as part of https://github.com/studentinsights/studentinsights/issues/1722.

sets env variable for test mode, remove conditions from tests.  Also randomizing rspec runs to surface implicit dependencies.


kevinrobinson:
I was playing with code climate, removed it now.  @alexsoble FYI this sets the rspec tests to be run in random order so implicit dependencies can creep in.  Going to selfie but if you want to chat more next week just @ !

PR #1774, opened 2018-06-09
Remove educator full_name validation

kevinrobinson:
# Who is this PR for?
developers, follow-on to https://github.com/studentinsights/studentinsights/pull/1755

There are several duplicates for Somerville, all related to name changes and revealing that we don't handle this during import.  Backing out the validation for now, since it's not as trivial to patch manually and it's not helpful to alert on this every night until we improve this.

# What does this PR do?
Removes the validation


kevinrobinson:
selfie

Issue #1775, opened 2018-06-09
Improve StudentsImporter to handle students with no grade set


kevinrobinson:
from this:
```
Validation failed: Grade invalid grade:

      /app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.6/lib/active_record/validations.rb:78:in `raise_validation_error'
	/app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.6/lib/active_record/validations.rb:50:in `save!'
	/app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.6/lib/active_record/attribute_methods/dirty.rb:43:in `save!'
	/app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.6/lib/active_record/transactions.rb:313:in `block in save!'
	/app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.6/lib/active_record/transactions.rb:384:in `block in with_transaction_returning_status'
	/app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.6/lib/active_record/connection_adapters/abstract/database_statements.rb:235:in `block in transaction'
	/app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.6/lib/active_record/connection_adapters/abstract/transaction.rb:194:in `block in within_new_transaction'
	/app/vendor/ruby-2.3.0/lib/ruby/2.3.0/monitor.rb:214:in `mon_synchronize'
	/app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.6/lib/active_record/connection_adapters/abstract/transaction.rb:191:in `within_new_transaction'
	/app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.6/lib/active_record/connection_adapters/abstract/database_statements.rb:235:in `transaction'
	/app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.6/lib/active_record/transactions.rb:210:in `transaction'
	/app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.6/lib/active_record/transactions.rb:381:in `with_transaction_returning_status'
	/app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.6/lib/active_record/transactions.rb:313:in `save!'
	/app/vendor/bundle/ruby/2.3.0/gems/activerecord-5.1.6/lib/active_record/suppressor.rb:46:in `save!'
[...]

PR #1776, opened 2018-06-09
Log and skip instead of raising on trying to import invalid student record

kevinrobinson:
# Who is this PR for?
addresses https://github.com/studentinsights/studentinsights/issues/1775


kevinrobinson:
selfie

Issue #1777, opened 2018-06-09
Missing or stray quote in import file


kevinrobinson:
Missing or stray quote in line 236980
```
      /app/vendor/ruby-2.3.0/lib/ruby/2.3.0/csv.rb:1852:in `block (2 levels) in shift'
	/app/vendor/ruby-2.3.0/lib/ruby/2.3.0/csv.rb:1845:in `each'
	/app/vendor/ruby-2.3.0/lib/ruby/2.3.0/csv.rb:1845:in `block in shift'
	/app/vendor/ruby-2.3.0/lib/ruby/2.3.0/csv.rb:1805:in `loop'
	/app/vendor/ruby-2.3.0/lib/ruby/2.3.0/csv.rb:1805:in `shift'
	/app/vendor/ruby-2.3.0/lib/ruby/2.3.0/csv.rb:1747:in `each'
	/app/app/importers/file_importers/streaming_csv_transformer.rb:48:in `with_index'
	/app/app/importers/file_importers/streaming_csv_transformer.rb:48:in `each_with_index'
	/app/app/importers/file_importers/attendance_importer.rb:19:in `import'
	/app/app/importers/import_task.rb:136:in `block in import_all_the_data'
	/app/app/importers/import_task.rb:129:in `each'
	/app/app/importers/import_task.rb:129:in `import_all_the_data'
	/app/app/importers/import_task.rb:29:in `connect_transform_import'
	/app/app/jobs/import_job.rb:9:in `perform'
```

kevinrobinson:
I can't see this looking at the file manually, although it might be that the file has since changed, I didn't look further.

kevinrobinson:
Fixed in #1855

PR #1778, opened 2018-06-10
Add additional steps for character encoding and escaping in CSV files

kevinrobinson:
# Who is this PR for?
Addressing https://github.com/studentinsights/studentinsights/issues/1777.

There are a few issues with character encoding and escaping in the CSV files (for Somerville in particular).  This addresses them and adds them into `StreamingCsvTransformer` and `CsvTransformer`, along with more verbose logging to see what these files are doing.

# What does this PR do?
Adds a `ParseableCsvString` class.  This moves the code handling encoding and non-standard quote escaping into one place, and adds another pass to strip newlines within fields that the Ruby CSV parser can't handle.

# Checklists

### DIFF
--- a/app/importers/data_transformers/csv_transformer.rb
+++ b/app/importers/data_transformers/csv_transformer.rb
@@ -11,7 +11,7 @@ class CsvTransformer
 
   # Performs whole-file transformations first
   def transform(csv_string)
-    cleaned_contents = ParseableCsvString.new(@log).parseable_string_from_string(csv_string)
+    cleaned_contents = ParseableCsvString.new(@log).from_string(csv_string)
 
     csv = CSV.parse(cleaned_contents, headers: @headers,
                           header_converters: :symbol,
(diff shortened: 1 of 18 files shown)


kevinrobinson:
Example usage for Somerville:
```
irb(main):106:0> ParseableCsvString.new(STDOUT).from_filename('attendance_export.txt');nil
#read...
#encode_as_utf8...
  converted 25 known invalid character encodings
#convert_quotes...
  stripped 59 quotes.
#strip_inline_newlines...
  stripped 1983 newlines.
```

kevinrobinson:
I don't entirely understand the history here.  I can reproduce this on production data, and verify this code works as expected.  In looking into that, all the "stray quote" issues were coming from data in the `comment` field, introduced as part of getting more information about reasons and medical excuses for absences [here](https://github.com/studentinsights/studentinsights/pull/1650/files#diff-9d754a821f6747b2c2f3cd2f0d892f23R10).  It's strange that we didn't get error notifications about this earlier, and that when we did it didn't happen every day.  I don't see how this could have been working correctly since that change, assuming the export data is reliable.  That may have been a problem with error reporting, but just guessing.

kevinrobinson:
selfie

kevinrobinson:
verified this looks good in production

PR #1780, opened 2018-06-10
Fix date range for PDF download

kevinrobinson:
# Who is this PR for?
educators

This came up in manual testing for https://github.com/studentinsights/studentinsights/pull/1739, the "end date" defaults to today's date.  Using that in a `where` query to look for notes means that it excludes data from the current day.

# What does this PR do?
Makes the querying more inclusive on the end date.


kevinrobinson:
selfie

PR #1781, opened 2018-06-11
Fix PDF download, change back to date, add spec, remove methods from application_helper

kevinrobinson:
# Who is this PR for?
educators, fixes bug introduced in https://github.com/studentinsights/studentinsights/pull/1780.

https://github.com/studentinsights/studentinsights/pull/1780 changed the query from a date to a time, but these objects were used in the view template and this broke the contract.  This wasn't tested, so it didn't come up.

# What does this PR do?
Changes these back to dates, fixing the underlying issue by querying for one day ahead of what they ask for.  It also adds a smoke test for the controller calling through to the view without raising.  And separately removes some methods for this particular feature that were in `application_helper` and moves them back into just this controller.

# Checklists


kevinrobinson:
selfie

PR #1782, opened 2018-06-11
Flush import logging after each write

kevinrobinson:
# Who is this PR for?
developers

Importer jobs don't seem to have any log output while they're running, and then have a lot all at the same instant if the process dies.  Seems like they are buffering output too much, and makes it hard to debug what's going on.

# What does this PR do?
flush after each log


kevinrobinson:
selfie

PR #1783, opened 2018-06-11
Avoid collisions on randomly generated educator names being mapped to emails

kevinrobinson:
# Who is this PR for?
developers

failures like https://travis-ci.org/studentinsights/studentinsights/builds/390783980?utm_source=email&utm_medium=notification, which I think are from individual test cases that generate a bunch of educator records with random names, and sometimes then generate collisions on email addresses (with had been written as `(first initial, last name)`).  This was introduced with https://github.com/studentinsights/studentinsights/pull/1755.

# What does this PR do?
Changes emails to be `first_name.last_name` for educators made with factories.

### DIFF
--- a/spec/factories/educators.rb
+++ b/spec/factories/educators.rb
@@ -11,7 +11,7 @@ FactoryBot.define do
     end
     email do
       last_name, first_name = full_name.split(', ')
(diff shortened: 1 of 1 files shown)


kevinrobinson:
selfie

Issue #1657, opened 2018-04-27
Absences: Include medically excused absences in Insights


kevinrobinson:
There's more data about absences in Aspen that isn't imported into Insights.  This comes up when folks look at absences dashboards and see students with high absences; it's one of the first things they ask about or look Aspen to see.

Overall, one suggestion is that throughout the product when we have only one number for absences shown, let's keep showing the total absences number.  In specific places though, we'll show more targeted information based on what people expect and what they're going to do with the information.

#### Import:
- [x] Update Aspen export to send this data (https://github.com/studentinsights/studentinsights/pull/1650)
- [x] Update import task to read it in (https://github.com/studentinsights/studentinsights/pull/1653)
- [x] Build tools to look at data for quality (first step is in https://github.com/studentinsights/studentinsights/pull/1890)

#### Data quality:
- [x] CSV rows not imported (fixed within 90 day window)
- [x] more Insights records than CSV rows (https://github.com/studentinsights/studentinsights/pull/1890)
- [x] need to run backfill past `only_recent_attendance`
- [x] Migrate Absence and Tardy models for stricter validations and field types ([#1908](https://github.com/studentinsights/studentinsights/pull/1908))

#### Product uses:
- [x] Send down all absences to the dashboard, but add a filter for excused/total and set it to exclude excused by default (https://github.com/studentinsights/studentinsights/pull/1811)
- [x] In the student profile, update the "details" charts to show total absences, but use a stacked bar chart to indicate whether it's excused or not excused (https://github.com/studentinsights/studentinsights/pull/1895)
- [x] In the student profile, show the full data about absences, including the reason and comment (no design suggestions included here, but ping if you want some!) (#1895)
- [x] Update dashboard filters for only when `excused` is supported (https://github.com/studentinsights/studentinsights/pull/1916)
- [x] Update the absences insights box to only be about unexcused absences.  If a student has medically excused absences, it's a different kind of action to take (https://github.com/studentinsights/studentinsights/pull/1921)

kevinrobinson:
re: data quality, right now in Somerville there are 15% more records in the database between `Absence` and `Tardies` than there are lines in the CSV file.  The CSV file also includes rows with no flags set (https://github.com/studentinsights/studentinsights/issues/1779), and includes rows that are flagged as dismissal instead of absences or tardies.
 
Also related: https://github.com/studentinsights/studentinsights/pull/1778#issuecomment-396062524

kevinrobinson:
[...]

Issue #1626, opened 2018-04-13
Importer issues with failure modes


kevinrobinson:
related issues:
- https://github.com/studentinsights/studentinsights/pull/1751
- https://github.com/studentinsights/studentinsights/pull/1750
- https://github.com/studentinsights/studentinsights/issues/1738

Fixed or improved!
--------------------------------
- @alexsoble your DNS and MailGun config changes look like they worked great!  I staretd getting SIGTERM ErrorMailer emails and ImportJob#failure emails (and I think you did too).
- I cleared out the Delayed::Job queue for New Bedford.  I have the raw data on the errors if we need it, but there have been a bunch of changes in how and what we log, so don't think this matters.
- Way too much logging for attendance and behavior importer (logging a line for every row read).  In the process I realized that the `index` being passed in for each row was nil, because of a bug in how we thought the CSV parsing worked.  That's fixed and it fixed the "undefined" showing up in the attendance importer logs too.
- The import tasks were logging a bunch of raw data for invalid rows, and this included things like teacher and student names (and maybe even could include things like addresses or phone numbers), so I removed logging for any invalid rows.  We're not looking at it now anyway, we can build a better way to look at this later if we need it.

Failure modes
--------------------------------
- Looking at the stuck Delayed::Job records, they don't show the number of attempts or failures accurately (although I think this is improved after some of these changes (eg, see #72).
- On memory errors, it looks like Heroku sends SIGKILL directly, so that just kills the process and no Delayed::Job cleanup runs.  That's not good and so doesn't help improve how we handle out of memory errors.  But, it helps improve how we handle config changes or dyno restarts or deploys.  Those will send at SIGTERM now.  But...
- It doesn't look like "destroyed_failed_jobs = false" worked.  Logs show this line: `Job ImportJob (id=73) REMOVED permanently because of 1 consecutive failures`.  That's the opposite of what I thought that config line would do :\  So I think we are still in a state where if there's a deploy, the job just gets interrupted and dropped.
- Basically, https://github.com/studentinsights/studentinsights/pull/1613 didn't really improve much.
- And in trying to debug this now, I can't seem to get the workers to pick up jobs in the queue.  It may be there's some "backoff on failure" config that I can't understand.  The README in their repo doesn't describe this though so I got stuck and couldn't debug further because I couldn't get any jobs to run.  Before stopping, I cleared the queue again so it was empty when I stopped.

Other big issues to look into
--------------------------------
[...]

PR #1785, opened 2018-06-11
Allow running single importer jobs, remove Thor and add initial import job specs

kevinrobinson:
⚠️ Merging this would require updating the Heroku Scheduler.

# Who is this PR for?
developers
For https://github.com/studentinsights/studentinsights/issues/1657, I want to be able to run a single importer.  I want to be able to do this for https://github.com/studentinsights/studentinsights/issues/1626 as well, and to have some more spec coverage if we make any changes around logging or handling failures.

Right now we can't run a single importer.  Making changes to the thor defaults requires using the DSL (eg, factoring a constant out to a method didn't just work) and without being able to write tests changes there are more dangerous.  There are multiple files describing the importer classes, and it's hard to see how they relate.  Running the import task seeds the database with schools in a way that is unexpected for an import task.

# What does this PR do?
- allow running a single importer on the command line
- replace Thor with a Rake task using builtin `OptionParser`, and add specs for command line parsing
- move code describing the importers into `FileImporterOptions` and make order explicit
- removing seeding schools from the import job
- add some tests for these changes
- adds changes to the set of importers or their priority to mentionbot

This is the rake task:
```
$ bundle exec rake import:run -- --help
Usage: bundle exec rake import:run -- [options]
        --background                 Add job into DelayedJob queue.
        --only-recent-attendance     Only import attendance rows from the past 90 days for faster attendance import.
        --school-local-ids x,y,z     Scope import to only these school local_ids (not Insights school.id).
        --importer-keys x,y,z        Import data from one of EducatorsImporter, CoursesSectionsImporter, EducatorSectionAssignmentsImporter, StudentsImporter, StudentSectionAssignmentsImporter, BehaviorImporter, AttendanceImporter, StudentSectionGradesImporter, X2AssessmentImporter, StarMathImporter, StarReadingImporter.
```

# Checklists


app/importers/import_task.rb
```diff
@@ -1,38 +1,45 @@
...
 
-    # options["school"] is an optional filter; imports all schools if nil
-    @school = @options.fetch("school", nil)
+    # which schools to include
+    @school_local_ids = validate_school_local_ids!(@options[:school_local_ids])
 
-    # options["source"] describes which external data sources to import from
-    @source = @options.fetch("source", ["x2", "star"])
+    # which importers to run
+    @importer_descriptions = validate_importer_keys!(@options[:importer_keys])
```
kevinrobinson: The substantive change here is to require callers to pass in all `importer_keys` they want to run instead of only the "x2" and "star" groupings.

app/importers/import_task.rb
```diff
@@ -1,38 +1,45 @@
...
+      log('Setting up logging...')
       @record = create_import_record
+      log('Done logging.')
+
+      log('Running inital report...')
       @report = create_report
-      log('Starting validation...')
-      validate_district_option
-      seed_schools_if_needed
-      validate_school_options
```
kevinrobinson: Validation is moved up to the constructor so invalid args fail before any work can be done, and seeding is removed here.

alexsoble:
Hey @kevinrobinson! I'm open to this change and happy to dig into the details with a review, but I'm not sure I understand the motivation behind it yet. The top level "Who is this PR for?" section says, "I want to be able to run a single importer."

Is the `thor import:start --source attendance` option not working for you? If not, how would you want it to work differently?

kevinrobinson:
@alexsoble Thanks!  Yeah a better description here would help, and sounds like talking it out first might have been good too.  This doesn't seem super important to me, so if we have other things going on and it slips a bit that seems fine to me too :)

I wasn't sure if I could run a single importer, so I looked at the code.  And then I saw there were four places were the set of importers were defined in five places `FileImporterOptions.keys`, `FilterImporterOptions.options`, `X2Importers.list`, `StarImporters.list`, `UnwrapFileImporterOptions::PRIORITY`.  This seemed like an opportunity to simplify, so I started doing that.

Having moved these all into `FileImporterOptions`, I tried to rename `scope` to `importer_keys` and change the default value to all of them in `import.thor`.  This didn't just work, and I could figure out how to make method calls from there.  Then I Google "ruby command line options" and saw that there's a parser built-in to Ruby, and starting updating that.  Another benefit of that change is that we can test this, although we probably could do that with Thor tasks also, but I didn't look up how.

So another path we could take it:
1. move forward with simplifying the import options into one file
2. back out the changes related to the Thor task and spec
3. change the default value for importers to be `nil`, and then inside the `#import` method in the Thor file, make the method call ourselves to set the default value if the param is nil
4. either ship or back out removing seeding from the import task, or talk about that separately

What do you think?

kevinrobinson:
Closing; we're not going to merge this as-is.

PR #1768, opened 2018-06-06
WIP – Add student photos to S3

alexsoble:
# Who is this PR for?

Initially, K8 counselors; eventually, all educators. 

# What problem does this PR fix?

Brings us closer to fixing the problem of educators who receive a large influx of new students (say, high school house masters at the start of a new school year). Educators may not recognize every student by name when looking at their Student Profile. 

# What does this PR do?

Adds rake task that imports student photos from the Somerville SFTP site to a secure Amazon S3 bucket. Next step is to add authenticated server code that checks educator authorization before sending down the image; then we can add student photos to student profiles, starting with 8th graders in time for next week's K8 Counselor/High School House Master meetings.

PR #1787, opened 2018-06-11
Add student photos to S3

alexsoble:
➡️ This PR is a cleaner rebased/cherry-picked version of #1768

# Who is this PR for?

Initially, K8 counselors; eventually, all educators. 

Brings us closer to fixing the problem of educators who receive a large influx of new students (say, high school house masters at the start of a new school year). Educators may not recognize every student by name when looking at their Student Profile. 

# What does this PR do?

Adds rake task that imports student photos from the Somerville SFTP site to a secure Amazon S3 bucket. Next step is to add authenticated server code that checks educator authorization before sending down the image; then we can add student photos to student profiles, starting with 8th graders in time for next week's K8 Counselor/High School House Master meetings.

### DIFF
--- a/app/models/student.rb
+++ b/app/models/student.rb
@@ -20,6 +20,7 @@ class Student < ActiveRecord::Base
   has_one :iep_document, dependent: :destroy
   has_many :student_section_assignments
   has_many :sections, through: :student_section_assignments
+  belongs_to :student_photo
 
   has_many :dashboard_tardies, -> {
     where('occurred_at >= ?', 1.year.ago)
--- a/app/models/student.rb
+++ b/app/models/student.rb
@@ -8,7 +8,7 @@ class Student < ActiveRecord::Base
 
   belongs_to :homeroom, optional: true, counter_cache: true
   belongs_to :school
-  belongs_to :student_photo
+  has_one :student_photo
   has_many :student_assessments, dependent: :destroy
   has_many :assessments, through: :student_assessments
   has_many :interventions, dependent: :destroy
--- a/app/models/student.rb
+++ b/app/models/student.rb
@@ -8,7 +8,7 @@ class Student < ActiveRecord::Base
 
   belongs_to :homeroom, optional: true, counter_cache: true
   belongs_to :school
-  has_one :student_photo
+  has_many :student_photos
   has_many :student_assessments, dependent: :destroy
   has_many :assessments, through: :student_assessments
   has_many :interventions, dependent: :destroy
--- a/app/models/student.rb
+++ b/app/models/student.rb
@@ -8,6 +8,7 @@ class Student < ActiveRecord::Base
 
   belongs_to :homeroom, optional: true, counter_cache: true
   belongs_to :school
+  belongs_to :student_photo
   has_many :student_assessments, dependent: :destroy
   has_many :assessments, through: :student_assessments
   has_many :interventions, dependent: :destroy
@@ -20,7 +21,6 @@ class Student < ActiveRecord::Base
   has_one :iep_document, dependent: :destroy
   has_many :student_section_assignments
   has_many :sections, through: :student_section_assignments
-  belongs_to :student_photo
 
   has_many :dashboard_tardies, -> {
     where('occurred_at >= ?', 1.year.ago)
--- a/db/migrate/20180606173823_add_student_photo_to_student.rb
+++ b/db/migrate/20180606173823_add_student_photo_to_student.rb
@@ -1,0 +1,5 @@
+class AddStudentPhotoToStudent < ActiveRecord::Migration[5.1]
+  def change
+    add_column :students, :student_photo_id, :integer
+  end
+end
--- a/db/migrate/20180606173823_add_student_photo_to_student.rb
+++ b/db/migrate/20180606173823_add_student_photo_to_student.rb
@@ -1,5 +1,0 @@
-class AddStudentPhotoToStudent < ActiveRecord::Migration[5.1]
-  def change
-    add_column :students, :student_photo_id, :integer
-  end
-end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180606164642) do
+ActiveRecord::Schema.define(version: 20180606173823) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -367,6 +367,7 @@ ActiveRecord::Schema.define(version: 20180606164642) do
     t.string "primary_email"
     t.text "house"
     t.text "counselor"
+    t.integer "student_photo_id"
     t.index ["homeroom_id"], name: "index_students_on_homeroom_id"
     t.index ["local_id"], name: "index_students_on_local_id"
     t.index ["school_id"], name: "index_students_on_school_id"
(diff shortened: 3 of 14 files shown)


alexsoble:
Hmmm, specs are failing but the failing spec looks like it has nothing to do with this PR: 

```
rspec ./spec/controllers/homerooms_controller_spec.rb:18 # HomeroomsController#show happy path returns the right shape of data
```

Suspect this is an unrelated nondeterministic test failure that can be fixed with `match_array` rspec helper. I'm going to re-run the test suite.

app/models/student.rb
```diff
@@ -8,6 +8,7 @@ class Student < ActiveRecord::Base
 
   belongs_to :homeroom, optional: true, counter_cache: true
   belongs_to :school
+  belongs_to :student_photo
```
alexsoble: I'd prefer to model this as "student has one student photo." That feels more logical and natural-language to me: you'd naturally say that a student has a photo of him/her, but you wouldn't say that a photo has a student. Here's the Rails guide on this one: http://guides.rubyonrails.org/association_basics.html#choosing-between-belongs-to-and-has-one.

db/schema.rb
```diff
@@ -299,6 +299,15 @@
     t.index ["student_id"], name: "index_student_assessments_on_student_id"
   end
 
+  create_table "student_photos", force: :cascade do |t|
+    t.string "student_id"
```
alexsoble: Should be an integer, not a string. Also this needs an index since it's a foreign key.

db/migrate/20180606173823_add_student_photo_to_student.rb
```diff
@@ -0,0 +1,5 @@
+class AddStudentPhotoToStudent < ActiveRecord::Migration[5.1]
```
alexsoble: This migration isn't clear to me and I'd want to remove it. It seems to add a second foreign key pointing in the other direction. As far as I understand we never want to have two foreign keys, because there should only be a single source of truth about what belongs to what.

alexsoble:
Hey @kevinrobinson @snoopyuri: progress update on the student photo import!

Today I:

+ Updated the data model. 
+ Added a bunch of tests.
+ Fixed a few bugs small that the tests found.
+ Set up a developer error email if any of the student photos fail to save.

I'm feeling good about test coverage: `PhotosStorer` is well covered. There's no spec for the Importer class yet, but that's mostly downloading, unzipping, and file-system handling and I've already seen it work locally, so I don't think it needs an rspec test. 

My plan for tomorrow is to do a little light QA in the morning, then merge, deploy, and run a photo import job. @kevinrobinson, if you spot anything you want to comment on or suggest tomorrow morning, that's cool. No formal review requested though because I don't have any outstanding questions and we verified the critical part (S3 bucket settings) together in person in Somerville. 😎

app/models/student.rb
```diff
@@ -8,6 +8,7 @@ class Student < ActiveRecord::Base
 
   belongs_to :homeroom, optional: true, counter_cache: true
   belongs_to :school
+  has_one :student_photo
```
kevinrobinson: I saw some earlier comments about the keys here, so not sure if this is entirely relevant, but this looks to me like the storer code might create a `StudentPhoto` record for each photo as it is imported.  If that's still the case, then I'm not sure how `has_one` works if there are multiple records on the other side of the association (I think `has_one` just adds `LIMIT 1` to the query and nothing smarter).

This also doesn't matter at all for the initial import, just seems like a place a sneaky bug might come up.

app/models/student.rb
kevinrobinson: Yep!  The idea previously was to store all new photos we get.  They each have a reference to the student, so we'd store that (a foreign key).  Separately, we also want to know what the most recent photo for the student is, so that's what this was (another foreign key).  So that data has meaning separate from how the Rails language is on top.  So `current_student_photo` might have been a more accurate name previously.

And cool to do whatever you think is best to 🚢 this!

db/migrate/20180606173823_add_student_photo_to_student.rb
kevinrobinson: 👍 The idea before was to hold two different truths :)

app/models/student.rb
alexsoble: Cool, thanks for the explanation!

app/models/student.rb
alexsoble: Ooooh good catch here. Do you feel strongly about `has_one` versus `has_many`? We could either store the most recent photo for each student in AWS and our database, or store all the photos and select the most recent one in application code.

You're right to point out that the import code looks like it's currently configured for `has_many`. The `"student / upload date / image"` file structure implies multiple images per student and we don't overwrite anywhere. 

I think the `has_one` approach is better because it means we store fewer photos. That means cheaper AWS bills and less clutter lying around.

My instinct is to change this so that the import code keeps us strictly `has_one` (and maybe write some stronger validation on the model), but if you want to offer any reasons why `has_many` is the better choice, I'd really like to hear because I can't think of any myself off the top of my head.

kevinrobinson: @alexsoble I'm happy to go with whatever you decide!  I think it's good to just :ship this since we're not setting up a recurring job yet anyway (on our end or on John's).

Initially, I was thinking that writing them all would be simpler.  We don't have to track changes over time - we just get photos, and we write them to S3, there's no other moving parts.  Nothing can change over time, except for which one a given student points to.  And we'll have the information in the database and logs to understand or debug that if it seems weird.

For costs, I'm assuming that any S3 storage costs here will be almost nothing, and if it does cost anything it'll cost less than an hour of our time each month, so I wasn't thinking it was super meaningful to optimize.

So what I was initially thinking was sort of a hybrid of the two options you described: we store everything, and we make it easy to query for the current photo.  Maybe another iteration on that might be making a `latest_student_photo` method on the `student` model?  That would remove the second key that you were wondering about, although it would require adding an index in its place to make it fast to query for the latest photo.

alexsoble:
🚢ing now. This PR's only goal is to get student photos into our secure Amazon S3 bucket; it doesn't include any code to optimize finding student photo records in the database or display them on the student profile page.

alexsoble:
Deployed and ran the task. Most records saved successfully and the task completed. But we got a fair amount of errors like this (don't have an exact number yet):

```
Error! 74440076668 is out of range for ActiveModel::Type::Integer
```

This drew my attention to the schema. I see a few clear mistakes in the `student_photos` table:

```
  create_table "student_photos", force: :cascade do |t|
    t.bigint "student_id"
    t.integer "file_digest"   # should be a string
    t.integer "file_size"     # should maybe be bigint? not sure yet
    t.integer "s3_filename"   # should be a string
    ...
  end
```

I'm not sure how well Rails and Postgres will handle migrating an integer column to a string column. We also have a bunch of orphan records up in Amazon S3 where the file was successfully uploaded but the database record failed to create.

So my preferred path forward would be:

1. Delete all objects in the S3 bucket.
2. Use `destroy_all` in the console to empty all student photos in the database. 
3. Write a migration updating the schema for student_photos to be correct.
4. Deploy and migrate.
5. Re-run the entire job.

Seem reasonable @kevinrobinson?

alexsoble:
There's an AWS "Empty Bucket" button, so that's super easy. https://docs.aws.amazon.com/AmazonS3/latest/user-guide/empty-bucket.html

alexsoble:
OK. The largest file in `photos.zip` is 154,927 bytes. Postgres integer can handle numbers up to ~2 billion. So the `file_size` field can stay a regular integer. The overflows are being caused by`file_digest`/`s3_filename`, which are being converted into giant integers, rendering all the student photo records we have in the database useless. This confirms that we need to wipe the database and start from scratch.

alexsoble:
Emptied out the s3 bucket.

PR #1764, opened 2018-06-05
Unify Transition Note endpoints

alexsoble:
# Who is this PR for?

K-8 Counselors. 

# What problem does this PR fix?

Closes a bug in a note-saving edge case, when multiple requests are fired off to the server quickly.

# What does this PR do?

+ Changes the controller code for Transition Notes to only implement one action, `#update`. This removes the possibility of bugs around create/update confusion.
+ Fixes the validation logic on Transition Note to say that there can only be one regular and one restricted note. 
+ Adds unit tests for the model, controller tests for the controller, and smoke test for the JS component. 
+ Disables firing off more than one request per "thread" (restricted/non-restricted note) at a time.

# GIF

unify

PR #1789, opened 2018-06-11
Patch transition note bug with looking at server response

kevinrobinson:
This is for merging into https://github.com/studentinsights/studentinsights/pull/1764

# Who is this PR for?
8th grade counselors

"Save" button appears disabled on restricted note after first save.

# What does this PR do?
Updates `PageContainer` to not look at the server response for deciding which note was updated (this data isn't in the response) and instead binds the request params to tell which kind of note this is.


alexsoble:
Good catch, verifying this now.

Issue #1760, opened 2018-06-04
Simplify: Routing improvements


kevinrobinson:
Follow-on from https://github.com/studentinsights/studentinsights/issues/1180.
- [ ] remove code in index.js
- [ ] remove legacyRouter, migrate to App
- [ ] remove HashRouter in admin dashboard code, migrate to App
- [ ] use react-router routing instead of history / location directly

PR #1791, opened 2018-06-11
Ship "Is Service Working for Student?" viz (no links in product)

alexsoble:
# Who is this PR for?

School leaders and district leaders.  

No visibility into whether services are working for students. We're collecting a lot of data on student service assignment, why not create a UI that pairs that data with relevant student outcomes? 

# What does this PR do?

Ship a V1 experimental spike dark so that Uri can get feedback on it. 

# Screenshot (Attendance Officer)

daisy view

# Screenshot (Attendance Contract)

micky view


alexsoble:
Since this is an experimental feature meant to allow Uri to gather feedback, I don't think it needs test coverage. I manually checked that the charts rendered by the "Is Service Working?" view match the charts rendered in the Student Profile view for demo students; quick manual check to make sure this experimental feature is surfacing the right data. 

Compare the Student Profile charts below with the "Is Service Working?" charts above:

# Mickey 
micky profile

# Daisy
daisy profile

alexsoble:
Hey @kevinrobinson! If you could give this a very quick review in two places, that would be awesome! 

1. Security, in particular authorization controller code 
2. Triple-confirm that data shown in the charts match to the correct student data points

app/controllers/service_types_controller.rb
```diff
@@ -3,8 +3,40 @@ class ServiceTypesController < ApplicationController
 
   # Authentication by default inherited from ApplicationController.
 
+  before_action :authorize_for_districtwide_access_admin, except: [:index]
```
kevinrobinson: @alexsoble This line was surprising to me at first, until I realized these endpoints are unrelated bits of code.  What do you think about cutting-and-pasting this feature into its own controller so it's safely isolated, and can slap its authorization check on everything?

app/controllers/service_types_controller.rb
```diff
@@ -3,8 +3,40 @@ class ServiceTypesController < ApplicationController
...
+
   def index
     render json: ServiceType.pluck(:name).sort
   end
 
+  def is_service_working_json
+    service_type_id = params[:service_type_id]
+
+    service_type = ServiceType.find(service_type_id)
+    student_ids = service_type.services.where('date_started > ?', Time.current - 1.year).map(&:student_id)
```
kevinrobinson: I think this is a good place to use `authorized` as an extra layer.

```
students = authorized do
  service_type.services.where('date_started > ?', Time.current - 1.year).map(&:students)
end
```

That will bounce each student off of `authorizer.is_authorized_for_student?` (the block is just stylistic, it functions the same as if you passed an argument).

The benefit of moving the authorization check close to the query is that if the code is refactored and the controller guard changes, the query itself has its own auth check already.

app/controllers/service_types_controller.rb
```diff
@@ -3,8 +3,40 @@ class ServiceTypesController < ApplicationController
...
+
+    chart_data = Student.active.where(id: student_ids).map do |student|
```
kevinrobinson: This is another "unscoped" query, so good to wrap it with `authorized`.  But this seems like it could collapse with the one up above into a single query:

```
students = authorized do
  service_type.services.where('date_started > ?', Time.current - 1.year).map(&:students).select(&:active?)
end
```

config/routes.rb
```diff
@@ -78,7 +81,11 @@
     end
   end
   resources :services, only: [:destroy]
-  resources :service_types, only: [:index]
+  resources :service_types, only: [:index] do
+    collection do
+      get :is_service_working
```
kevinrobinson: If you like, you can change this to `get :is_service_working => 'ui#ui'` and then remove the controller method for this that isn't doing anything meaningful.

app/controllers/service_types_controller.rb
```diff
@@ -3,8 +3,40 @@ class ServiceTypesController < ApplicationController
 
   # Authentication by default inherited from ApplicationController.
 
+  before_action :authorize_for_districtwide_access_admin, except: [:index]
+
+  def authorize_for_districtwide_access_admin
+    unless current_educator.admin? && current_educator.districtwide_access?
+      error = 'You don\'t have the correct authorization.'
```
kevinrobinson: You can use `raise Exceptions::EducatorNotAuthorized` here and let `application_controller` decide how to translate that to an HTTP response.

app/controllers/service_types_controller.rb
```diff
@@ -3,8 +3,40 @@ class ServiceTypesController < ApplicationController
 
   # Authentication by default inherited from ApplicationController.
 
+  before_action :authorize_for_districtwide_access_admin, except: [:index]
+
+  def authorize_for_districtwide_access_admin
+    unless current_educator.admin? && current_educator.districtwide_access?
```
kevinrobinson: Should this be `can_set_districtwide_access?`  This will give access to a more people (eg, `test_pals.rich` as an example).

ui/legacyRouteHandler.js
```diff
@@ -59,5 +59,10 @@ export default function legacyRouteHandler(el) {
     return true;
   }
 
+  if ($('body').hasClass('service_types') && $('body').hasClass('is_service_working')) {
```
kevinrobinson: For new things, it's better to put them into `App.js`, but no need to change this, we can just migrate this with the rest of them later as part of https://github.com/studentinsights/studentinsights/issues/1760.

app/assets/javascripts/service_types/IsServiceWorking.js
```diff
@@ -0,0 +1,186 @@
...
+  render() {
+    return (
+      <div className="IsServiceWorking">
+        <ExperimentalBanner />
+        <SectionHeading>
+          "Is this Service Working for this Student?"
+        </SectionHeading>
+        <p style={{padding: '10px 0 0 10px'}}>
+          Every student is different.
+          One student might respond well to a service, while another might respond poorly.
```
kevinrobinson: I like this intro text, short and trying to set the tone for curiosity. 👍 

Maybe consider rephrasing "might respond poorly" - it sort of suggests this happens because of an essential quality in the student, rather than something that's in educators' control (eg, might not have been the right fit for that student, not well-matched what they need).

app/assets/javascripts/service_types/IsServiceWorking.js
```diff
@@ -0,0 +1,186 @@
...
+      name: 'Attendance Contract',
+    };
+
+    return (
+      <div style={style.buttonSectionStyle}>
+        <Button style={style.buttonStyle}
+          onClick={() => {
+            this.setState({serviceTypeSelected: null}, () => {
+              this.setState({serviceTypeSelected: attendanceOfficer});
+            });
```
kevinrobinson: What's with the nesting, why set it to one value and then another?  Is this here to show a loading state?

app/assets/javascripts/service_types/IsServiceWorking.js
```diff
@@ -0,0 +1,186 @@
...
+        {this.renderStudents(json)}
+      </div>
+    );
+  }
+
+  renderPhasebands(services) {
+    return services.map(service => {
+      return {
+        momentUTCstart: moment.utc(service.date_started),
+        momentUTCend: moment.utc(service.discontinued_at),
```
kevinrobinson: not sure if null handling matters here, it's probably fine to ship without caring about that, but will some of these be null?

app/assets/javascripts/service_types/IsServiceWorking.js
```diff
@@ -0,0 +1,186 @@
...
+        text: this.state.serviceTypeSelected.name
+      };
+    }, this);
+  }
+
+  renderStudents(json) {
+    const chartData = json.chart_data;
+    const sortedData = _.sortBy(chartData, datum => {
+      return (datum.services.length > 0)
+        ? _.map(datum.services, 'date_started').sort()[0]
```
kevinrobinson: Will this sort correctly?  I'm assuming `date_started` is a string and that this is sorting by string order.

If it is string order, `datum.services.map(service => toMomentFromTime(service.date_started).getDate()).sort()[0]` might do the trick to convert the Rails timestamp to a JS `Date`.

app/assets/javascripts/service_types/main.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import IsServiceWorking from '../service_types/IsServiceWorking.js';
+
+export default function renderIsServiceWorking() {
+  const serializedData = $('#serialized-data').data();
+
+  ReactDOM.render(
+    <IsServiceWorking
+      serializedData={serializedData}
```
kevinrobinson: I think you can remove `serializedData` from this file?

app/assets/javascripts/service_types/IsServiceWorking.js
```diff
@@ -0,0 +1,186 @@
...
+        </div>
+      );
+    }, this);
+
+  }
+
+}
+
+IsServiceWorking.propTypes = {
+  serializedData: React.PropTypes.object.isRequired,
```
kevinrobinson: I don't think this is used, looks like `GenericLoader` is loading the data for this component.

app/assets/javascripts/student_profile/ProfileBarChart.js
```diff
@@ -70,24 +70,50 @@ export default class ProfileBarChart extends React.Component {
     });
   }
 
+  makePlotbands(monthKeys) {
```
kevinrobinson: So this is reaching into a component called `ProfileBarChart` inside the `student_profile` folder for an experimental feature.  Ideally we'd keep these things separated, so that experimental code doesn't introduce more risk to core production features, and so that we don't have to maintain backwards compatibility with experimental code when refactoring or improving core production features.

To 🚢 this and keep moving, what do you think about copying this component file into the `service_types` folder with this patch, and keeping the existing one as-is.  That way it's fast to ship this feature, and we avoid coupling together these two separate parts of the app.

app/assets/javascripts/service_types/IsServiceWorking.js
alexsoble: 👍

app/assets/javascripts/service_types/main.js
alexsoble: 👍

app/assets/javascripts/service_types/IsServiceWorking.js
alexsoble: Yes, to show loading state.

config/routes.rb
alexsoble: 👍

app/assets/javascripts/service_types/IsServiceWorking.js
kevinrobinson: What happens if you just call `this.setState({serviceTypeSelected: attendanceOfficer})`?  Perhaps this is something that isn't working the way it should in `GenericLoader`.

As a workaround, you could also set `key={serviceTypeSelected}` on `GenericLoader`, and that would make a new one each time that value changed, which I think would enable removing the nested `setState` calls.

alexsoble: When I just call `this.setState` without the callback, it switches from one set of charts to the other without showing a "Loading..." state.

alexsoble:
Hey there @kevinrobinson, thanks for the great comments and feedback on this!

I made a bunch of changes that you suggested or pointed to:

+ Isolated the chart into its own component to leave `ProfileBarChart` untouched
+ Isolated the controller code into its own controller 
+ Used the newer `App.js` path for client-side routing 

I didn't add `authorized` blocks around the student queries. I think that would be a great improvement later on, because it would mean we could open this feature up to more users, like homeroom teachers. But for now, since this is meant to be used by Uri to spark conversation, I think the controller `before_action` authorization block is sufficient. 

Going to 🚢, QA in production, and then ping Uri!

app/assets/javascripts/service_types/IsServiceWorking.js
kevinrobinson: Isn't that good? :)

kevinrobinson:
@alexsoble 👍 FWIW, my vote would be that even for experimental code, we wrap queries in `authorized`, regardless of whether there are other layers of guards (like the `before_action` here).  The idea there is that we we break the habit of writing "unscoped queries" that rely on code somewhere else to enforce authorization security.  And that hopefully we can do this with little development overhead by defaulting to wrapping all queries, since experimental endpoints are just as exposed in production as non-experimental endpoints :)

PR #1792, opened 2018-06-12
Revert "Allow masquerading as different users"

kevinrobinson:
Used this to temporarily verify authorization, which checked out.  Reverting this now so we can review and test more thoroughly before it's deployed for real.

Reverts studentinsights/studentinsights#1749


kevinrobinson:
selfie

PR #1793, opened 2018-06-12
Double revert: Allow masquerading as different users

kevinrobinson:
This is for discussing and merging https://github.com/studentinsights/studentinsights/pull/1749, which was merged temporarily and then reverted in studentinsights/studentinsights#1792.

See https://github.com/studentinsights/studentinsights/pull/1749 for background, and let's do code review on this PR.


app/controllers/admin/masquerade_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Admin
+  class MasqueradeController < ::ApplicationController # does not inherit from administrate
```
alexsoble: Does it not inherit from Administrate so that it's impossible to masquerade into the Administrate dashboard and change user permissions?

spec/features/masquerade_feature_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+    it 'healey_sarah_teacher' do expect_masquerading_to_fail(pals.healey_sarah_teacher) end
+    it 'west_marcus_teacher' do expect_masquerading_to_fail(pals.west_marcus_teacher) end
+    it 'shs_jodi' do expect_masquerading_to_fail(pals.shs_jodi) end
+    it 'shs_bill_nye' do expect_masquerading_to_fail(pals.shs_bill_nye) end
+    it 'shs_ninth_grade_counselor' do expect_masquerading_to_fail(pals.shs_ninth_grade_counselor) end
+    it 'shs_hugo_art_teacher' do expect_masquerading_to_fail(pals.shs_hugo_art_teacher) end
+    it 'shs_fatima_science_teacher' do expect_masquerading_to_fail(pals.shs_fatima_science_teacher) end
+  end
+
+  # it 'does not work for any other users to masquerade as anyone else' do
```
alexsoble: This test should be un-commented.

alexsoble:
After doing a quick scan through the code, I did a quick scan through the UI and found a few things. I think ultimately a real review should encompass a thorough scan through the code, a thorough scan through the UI, and a dive into Devise sessions.

app/controllers/admin/masquerade_controller.rb
kevinrobinson: Ah good question!  I haven't looked at how this would interact with the administrate routes.  I did this here because I wanted to know exactly what code was running here, and with the administrate `Admin::ApplicationController` there are more moving parts that aren't visible in our codebase (eg, I think it defines `requested_resource` and reflects on the controller class name to know what model to look at, and I wasn't sure how that applied here).

alexsoble:
There were a couple of things that confused me when using the authorization page.

+ The authorization page always has a white background, and never turns red to indicate Masquerading mode. Wouldn't we want to keep the red Masquerade mode consistent through the authorization page?
+ We render "become" links for the currently-logged-in user, which isn't supported because it would be confusing. One improvement would be not render "become" links for educators who you can't "become." 

screen shot 2018-06-12 at 9 34 45 am

spec/features/masquerade_feature_spec.rb
kevinrobinson: I'm not sure if we should do this so left it here so we could discuss, but the double revert lost the comment :)

https://github.com/studentinsights/studentinsights/pull/1749#discussion_r192570372
> This seems most thorough, but takes ~30 seconds to run through. A compromise is the lines above, which explicitly tests those users trying to masquerade as the highest-permission user, which is a bit faster and gives feedback on each one.

What do you think is the right tradeoff here?

alexsoble:
The text in the upper right corner in Masquerade mode confused me a few times. I thought it said, "Sign Out silva X", but it really is two separate links: "Sign out", and "silva X" (which goes to the authorization page). I think putting these on separate lines, adding space, or using different colors would help here. This is a developer-facing feature so we don't have to worry as much about UX as we would on an educator-facing feature, but trying it out as a developer I stumbled on this a few times.

screen shot 2018-06-12 at 9 33 27 am

kevinrobinson:
@alexsoble Awesome, thanks!  👍 

I'll incorporate your great suggestions in https://github.com/studentinsights/studentinsights/pull/1793#issuecomment-396615450 and fix up adminstrate as well.  I'll also revise the design for the links that you called out in https://github.com/studentinsights/studentinsights/pull/1793#issuecomment-396616626 as well.

re: https://github.com/studentinsights/studentinsights/pull/1793#issuecomment-396613249, I'll ping back when these other things are addressed and ask you to review.

alexsoble:
One final note from my UI journey: I can still access the authorization view when Masquerading as Teacher Silva. Is that necessarily what we want? Teacher Silva can't access the authorization view her/himself. There are no actions to take there — you can't become another educator if you're already Masquerading as one.

alexsoble:
@kevinrobinson Sounds like a plan! I spent 1 hour reviewing, happy to come back and review in more depth again. For a feature this consequential I think something around ~4 hours of review on my end would make sense.

spec/features/masquerade_feature_spec.rb
alexsoble: Ah, gotcha! Yeah, I see why this would be really slow with the nested looping across `Educator.all`. 

Two thoughts. First, I think it would be good to bring back the outer Educator loop instead of naming and testing each educator individually. That's because we sometimes add TestPal educators. Looping means that each new TestPal will be covered automatically and there's no need to remember to update the spec.

In terms of the inner loop, I agree that testing for masquerade-as-Uri is a good idea. One compromise would be to add two tests to the inner loop, one against masquerade-as-Uri and one against masquerade-as-another-user. That way if something changes about the Uri TestPal that might screw with this test (unlikely but just trying to think about future-proofing), at least it's not the only test role we're testing against.

kevinrobinson: @alexsoble sure, sounds good!

app/views/admin/educators/authorization.html.erb
```diff
@@ -1,100 +1,108 @@
...
+            <tr style="vertical-align: top;">
+              <td style="padding: 5px;">
+                <span style="display: inline-block; width: 2em; text-align: right; color: white; padding: 5px; background: <%= color %>;">
+                  <%= days_since_last_sign_in %>
+                </span>
+              </td>
+              <td style="padding: 5px;"><%= educator.sign_in_count %></td>
+              <td style="padding: 5px;"><%= educator.email %></td>
+              <td style="padding: 5px;"><%= link_to educator.full_name, "/educators/view/#{educator.id}" %></td>
+              <td>
```
kevinrobinson: This is the meaningful change; the rest is whitespace noise from adding a wrapping div that does some styling.

app/controllers/admin/application_controller.rb
```diff
@@ -9,9 +9,14 @@ class ApplicationController < Administrate::ApplicationController
     before_action :authenticate_educator!, :authenticate_admin
 
     def authenticate_admin
-      redirect_to(new_educator_session_path) unless current_educator && current_educator.can_set_districtwide_access?
+      redirect_to(not_authorized_path) unless current_educator && current_educator.can_set_districtwide_access?
```
kevinrobinson: This was not-quite-right previously, and it came up in testing this PR so I fixed it up.

app/views/admin/educators/_navigation.html.erb
```diff
@@ -10,6 +10,7 @@ as defined by the routes in the `admin/` namespace
 <% if action_name == "show" || action_name == "edit" %>
   <nav class="navigation" role="navigation">
     <% Administrate::Namespace.new(namespace).resources.each do |resource| %>
+      <% next unless resource.name == :educators %>
```
kevinrobinson: Administrate reflects on the routes in the admin namespace, and assumes it can map all those 1:1 to controllers and models.  Since it can't, this whitelists it to only do this for the educator model.

kevinrobinson:
Showing the overlay in effect on the `admin/authorization` page (in the unanticipated case where you masquerade as another user with access to masquerade):
screen shot 2018-06-18 at 6 58 08 pm

Showing links are removed for masquerading as the real underlying user and the current user you are masquerading as:
screen shot 2018-06-18 at 6 58 49 pm

kevinrobinson:
@alexsoble This is ready for another go, and includes updates related to your great suggestions and questions, as well as a few others.  Here's an overview of the various pieces that might be helpful in reviewing.

### Server code:
1. `MasqueradeController`, only be accessible to elevated users, allows making calls into the `Masquerade` class to change the state of masquerading.
2. `MasqueradeHelper` patches `current_educator`.  It's mixed into `ApplicationController` and `Administrate::ApplicationController`, and this is what leads to the server treating the user like they're someone else.
3. The `Masquerade` class is what actually changes the state of the user session, so that masquerading is in effect across multiple server requests.
4. The `MasqueradingLog` model class stores a log of these operations for auditing.


### UI code:
1. `admin/educators/authorization.html.erb` renders links for elevated users to masquerade as another user.
2. `navbar_signed_in` puts links for elevated users to see who they're masquerading as and to clear that.  If they're not masquerading, it provides a link to jump to the list of users so they can masquerade.
3. The application layout calls into a `shared/masquerade` partial that renders a red overlay on every page if the user is masquerading.


### Plumbing:
1. administrate doesn't inherit from `ApplicationController` and doesn't use the main application layout.  So this PR mixes in `MasqueradeHelper` to the base adminsitrate controller.
2. It also runs the administate generator to let us check the administrate layout into source control.  It then updates that layout to render the masquerade warning partial.
3. Finally, the administrate layout didn't previously include other things in our main layout (like the CSRF tag, necessary for making POST requests on the admin/authorization page here).  This factors out the head of the template from our main layout, so we can call it from the administrate layout as well.

kevinrobinson:
And to be explicit, there's no pressing need to finish this immediately anymore.

kevinrobinson:
🚢and verifying; this is needed now to verify https://github.com/studentinsights/studentinsights/pull/1931.

PR #1796, opened 2018-06-12
Class lists: Rake task to take "snapshots" of student data, since it will change over time

kevinrobinson:
# Who is this PR for?
project team, K5 principals and teaching teams. part of https://github.com/studentinsights/studentinsights/issues/1722

The work teaching teams and principals do on class lists are stored in the database as a set of student ids.  This doesn't contain all the information about each student (eg, latest STAR or DIBELS scores, notes) and these values can change over time.  This is fine in general, but means in the future we won't be able to see what these values were _at the time teams and principals made the lists_.  Since we're interested in seeing how these plans change from the end of the school year to the beginning of the next school year, and making this visible to teaching teams and principals, we'll need to store that information.

# What does this PR do?
Adds a rake task to "snapshot" across all class list workspaces, and write them to a new `ClassListSnapshot` table if the student data value has changed.  The intention is that we could run this every night as a task either for a critical window (eg, now until the end of the school year) or every night to keep it simple and with less maintenance.

There's no UI to view this data in this PR, this would just start tracking and storing it.

# Checklists


kevinrobinson:
The immigrant task doing its thing! :) https://travis-ci.org/studentinsights/studentinsights/builds/391383160?utm_source=github_status&utm_medium=notification

kevinrobinson:
selfie

kevinrobinson:
With https://github.com/studentinsights/studentinsights/pull/1822, verified that this looks good, and that the more conservative equality check works to detect changes.  I added this as a daily job in the scheduler, and will check in on it in a few days.

Issue #1675, opened 2018-05-04
HS Counselors: Show only students on their caseload in feed


kevinrobinson:
The feed is too much information for HS counselors, who typically have access to all students in the school because of the nature of their role and support teams, but only focus on supporting a particular set of students.

One implementation path might be to update the `Feed` object so it can query with a different filter. 
 And for implementing that filter, we have data on a student's counselor's last name and we could manually map these, try to infer the link, or build a structure for HS folks to enter and update this themselves.

kevinrobinson:
Going to extend this to the "high absences" box on the home page and the analysis dashboards as well, so that this is simpler and more coherent across the product for these folks.  This doesn't impact the actual authorization rules (which might be something to tighten in a next phase).  This is in https://github.com/studentinsights/studentinsights/pull/1806.

kevinrobinson:
🚢 ed

PR #1798, opened 2018-06-12
Show HS counselors only their students in the home page feed

kevinrobinson:
# Who is this PR for?
HS counselors, addresses https://github.com/studentinsights/studentinsights/issues/1675.

HS counselors see the feed for all HS students per authorization rules, but updates for students on their caseload are most meaningful.

# What does this PR do?
Updates the home page feed to filter further when a feature switch is enabled, and when the educator has the `use_counselor_based_feed` `EducatorLabel` set.  When both are set, it using the new `CounselorNameMapping` model to filter down to only students who have a "counselor" field that references them.

Shipping this wouldn't change anything, since both switches need to be set to enable this for an educator.

This would benefit from shipping https://github.com/studentinsights/studentinsights/pull/1793 to verify impact, but we can also use the `/educators/view/:id` endpoint to do this in the meantime.

# Checklists


kevinrobinson:
I initially did this by updating the authorization code, which I think might be what we really want here.  But this approach is simpler and less risky, and keeps the scope smaller.

spec/controllers/educators_controller_spec.rb
```diff
@@ -67,19 +67,24 @@ def included_student_ids(response)
     it 'works for Harry as an example, showing he is authorized for 8th graders' do
       response = get_my_students(pals.shs_harry_housemaster)
       expect(response).to be_success
+      puts response.body
```
kevinrobinson: @kevinrobinson oops, go fix this :)

kevinrobinson:
selfie

kevinrobinson:
Migrated, updated manually, verified this is good in production and ready to switch on.

PR #1799, opened 2018-06-12
Minimal change to add counselor to profile page

kevinrobinson:
# Who is this PR for?
HS classroom teachers

HS teachers can't easily see the counselor.

# What does this PR do?
Adds it minimally to the profile page; this design needs some more love before next fall.

screen shot 2018-06-12 at 6 30 28 pm


kevinrobinson:
selfie

PR #1800, opened 2018-06-12
Show transition notes only for 8th graders

kevinrobinson:
# Who is this PR for?
8th grade counselors and 9th grade house masters.  Noticed this working on https://github.com/studentinsights/studentinsights/pull/1799.

Transition note shows for all students, not just those in 8th grade.

# What does this PR do?
Doesn't show the note for students unless they're in 8th grade.

# Checklists

### DIFF
--- a/app/assets/javascripts/student_profile/StudentProfilePage.js
+++ b/app/assets/javascripts/student_profile/StudentProfilePage.js
@@ -57,12 +57,13 @@ export default class StudentProfilePage extends React.Component {
   }
 
   renderTransitionNote() {
-    const {currentEducator, actions, transitionNotes, requests} = this.props;
+    const {student, currentEducator, actions, transitionNotes, requests} = this.props;
     const labels = currentEducator.labels;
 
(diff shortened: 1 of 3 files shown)


kevinrobinson:
@alexsoble Does this seem like a good idea?  Just noticed it verifying https://github.com/studentinsights/studentinsights/pull/1799

alexsoble:
@kevinrobinson Hey hey. I noticed this too, but I don't think it's a big deal because counselors don't have any reason to add Transition Notes for anyone but 8th graders. I don't think this PR hurts but we should be careful not to introduce any regressions since this isn't a feature request, so I think the extra specs you added are great. 👍 

One change though: transition notes should be shown for 9th graders as well as 8th graders. We want House Masters to be able to see the notes when these students start High School in the fall.

alexsoble:
@kevinrobinson Ah, ok, those are great use cases I hadn't thought of. I'm cool with you shipping this for only 8th graders. 👍

Issue #1179, opened 2017-10-20
High School Risk Levels


alexsoble:
Status: waiting for input from HS stakeholders on how to calculate risk levels.

snoopyuri:
Had a productive meeting with high school team on Wednesday.  Here's what they are hoping to see in risk level calculations.  It is more complex than the elementary, but I think it's thoughtful and maybe a good way to start thinking when we look again at elementary.

+ 5 levels (0 to 4)
+ Rename to call them "Tiers" not "Risk Levels"
+ Looking at discipline actions from Aspen, not just the conduct description (not sure if we're pulling actions in as well)
+ Level 4: At least 4 F's OR less than 80% attendance over last 45 school days OR 7 or more discipline actions over the last 45 school days
+ Level 3: 3 F's OR less than 85% attendance over last 45 OR 5-6 discipline actions over the last 45 school days
+ Level 2: 2 F's OR less than 90% attendance over last 45 (no discipline involved in calculation)
+ Level 1: 1 F and 2 Ds OR less than 95% attendance over last 45 days (no discipline involved)
+ Level 0: Not any of the other levels

Happy to go back to them with any more clarification needed.  This looks big, but I think they're looking to reorganize how they tier and look at kids based on this which is very cool.

jhilde:
Thanks @snoopyuri! I'll churn through it later today to do some thinking on:

a) if we have all the data (discipline actions)
b) how we might think about this in a broader way [i.e. other schools / districts]

jhilde:
Some initial thoughts.

First, the level of effort is fairly significant here. Might be similar to the first high school implementation.

Looks like there are a few different data points they want to use in the calculation:

Current grade in classes: We will likely need Student Insights to know what term it currently is and how terms map to schedules.

Attendance over the last n days: This gets into the problem we identified in the attendance work. We could assume calendar days but that would reduce the accuracy. (Seems reasonable to me though for this purpose).

Discipline Actions: Pretty sure we’re not pulling actions. From looking at Aspen I have a general sense. Seems like educators can input a behavior incident which then enters into a workflow. From the data I looked at, seems the workflow is mostly just a step of what action, if any to take. Seems like there’s 0 or 1 action taken per behavior incident. We could likely pull that action, if any. (Same issue with number of days as attendance)

jhilde:
This is likely the most Somerville-specific piece. The current elementary school risk level also falls into this category.

[...]

PR #1801, opened 2018-06-13
Tiering: SHS prototype for "levels formula" and levels of support

kevinrobinson:
# Who is this PR for?
HS dept. heads, housemasters, principal, part of https://github.com/studentinsights/studentinsights/issues/1802.

Ships prototype-quality feature for end-of-year conversations with HS staff, and for further conversations and research over the summer.  This is more intended as a conversation piece and aid for user research than as a product feature.  In other words, the design here isn't super intentional towards solving any real problem.  The most intention choices are to minimize the "negative" aspects of tiering or the "if-then" aspect, and call out the "what supports are being provided" more directly.

The intention is to have the "Experimental" banner on this indefinitely, until after the school year starts next year.  Depending on other planning, we may revisit this again in the fall but there's no commitment yet.

This will be open to all educators who have the link, with the authorization standard and filtered further by school.  The expectation is that this might show some useful things for MS or K8 students, so we can explore that, but the only intentional focus here is the HS prototype.

# What does this PR do?
Adds a new `/levels/shs` page, intended for HS admin team educators.  This PR makes that only visible to members of the project team first, and doesn't add a link anywhere.  We may white list for release to particular educators later this week.

The UI is essentially a table, with the full list of active students in the school.  There are filters for house/grade and a textbox to search by name (since the table is virtualized, Ctrl+F in the browser won't work).

This also implements an iteration on the "tiering" formula from back in https://github.com/studentinsights/studentinsights/issues/1179, from some conversations with JR.  This is called "levels" in this iteration to distinguish that this is a different concept than "tiers" in MTSS language and that they mean different things.

Finally, there's a rough heuristic to map student courses to "subjects" in an effort to both simplify the presentation of grades and to expand it beyond just ELA/Math as in the Aspen report.

Page:
screen shot 2018-06-12 at 6 21 19 pm

Clicking the "summary" link in the upper right:
screen shot 2018-06-12 at 6 06 26 pm

# Checklists
+ [ ] Author checked latest in IE - HS Levels

This doesn't have amazing test coverage relative to what it is; this is just experimental and the prototype banner will stay for on here for the next few months at least.


kevinrobinson:
selfie

PR #1805, opened 2018-06-14
Add student photo endpoint

alexsoble:
# Who is this PR for?

All educators who use the Student Profile.

No student photo in the profile. This PR moves us a step closer to implementing student photos by adding an authenticated endpoint for us to use to add student photos to the profile. 

# What does this PR do?

Adds controller, route, and controller spec for `StudentsController#photo`.

# GIF (Internet Explorer)

photo-download-via-url-ie

# Next steps 

+ [ ] Merge and ship 
+ [ ] Set `USE_PLACEHOLDER_STUDENT_PHOTO` in the demo Heroku site
+ [ ] Verify student photo download via URL in production
+ [ ] Add student photo to the Student Profile front-end


alexsoble:
@kevinrobinson This is part two of the student photo work! Adds a controller, route, and spec for authenticated student photo endpoint. Also adds a dummy "MissingAvatar" photo in development for manual testing. (This is liked to a separate public S3 bucket.) 

Like the last PR, I don't think this one needs a formal code review, but if you have the time/energy to look it over and have any comments before it ships, that's welcome on my end. Planning to 🚢 and verify tomorrow afternoon.

config/environments/development.rb
```diff
@@ -5,13 +5,13 @@
...
   end if File.exists?(env_file)
 
-
   ENV['DISTRICT_KEY'] = 'somerville'
   ENV['DISTRICT_NAME'] = 'Localhost Public Schools'
   ENV['USE_MOCK_LDAP'] = 'true'
   ENV['MOCK_LDAP_PASSWORD'] = 'demo-password'
   ENV['ENABLE_CLASS_LISTS'] = 'true'
   ENV['HOUSEMASTERS_AUTHORIZED_FOR_GRADE_8'] = 'true'
+  ENV['AWS_S3_PHOTOS_BUCKET'] = 'fake-demo-student-insights-photos'
```
kevinrobinson: So this is real, public and doesn't require credentials to read or write?  That's probably worth a comment here.  Another improvement we could make some other time is to make a fake S3 client object (like with LDAP), and have the `put_object` and `get_object` and methods just write into `tmp/photos` or something on the filesystem so there's no external things involved.

config/routes.rb
```diff
@@ -62,6 +62,7 @@
 
   resources :students, only: [:show] do
     resources :event_notes, only: [:create, :update]
+    resources :student_photos, only: [:show]
```
kevinrobinson: This looks like it requires the UI to know the `student_photo_id`.  Using something like `/students/:id/photo` might be easier for us to work with when making UI and product developers, but that's probably also a question for another PR :)

db/seeds.rb
```diff
@@ -102,6 +102,14 @@ def sample_numeric_grade_from_letter(grade_letter)
   end
 end
 
+Student.all.each do |student|
+  StudentPhoto.create!(
+    student_id: student.id,
+    file_digest: Digest::SHA256.hexdigest(student.local_id),
+    file_size: 37010,
+    s3_filename: 'MissingAvatar.png'
```
kevinrobinson: This code seems like it's simulating "there is a photo for this student!" which is awesome.  But using a "missing avatar" image makes it seem like that's a feature - that the endpoint will show this if the server can't find an image for them.  Maybe consider changing the image to a cartoon image of a student to clarify?

config/routes.rb
alexsoble: Ooooh no you're totally right. Thought about this as soon as I woke up this morning! No need to pull the student's latest photo's ID out of the database, we already know which photo we want.

db/seeds.rb
alexsoble: 👍 — that makes sense... I'll change the filename at the very least.

config/environments/development.rb
alexsoble: Hey @kevinrobinson! Good thoughts. 

+ The bucket is real. Credentials are required to: list objects, write objects, read bucket permissions and write bucket permissions. 
+ The image is readable by everyone. Credentials are required to write/update the image. 
+ A code comment here is a good idea! Thanks for suggesting.
+ What are some of the pros/cons of mocking S3 instead of calling a real S3 bucket in development? 
  + For LDAP, setting up a real LDAP service is non-trivial and would involve learning a new technology, so mocking LDAP was much quicker to implement. 
  + One pro to using the real service is that it's better for development — if there's a mistake in the way the code calls S3, using a real bucket for development is more likely to surface that mistake than a mock. We depend on ourselves to build an accurate mock object that handles edge cases correctly. 
  + One con to using a real service is that it introduces the possibility of confusion. i named the bucket `"fake-demo-student-insights-photos"` to make its function as explicit as possible.
    + If somehow we screw up big-time and the demo bucket is used in production, there will be no effect because none of the student photo records in the database will match S3.
    + If somehow we screw up big-time and the production bucket is used on the demo Heroku site, we have additional layers of protection because the S3 creds need to be set in Heroku to access those images.
    + If somehow we screw up big-time and send real photos to the demo bucket, they'll be private, non-readable/writeable to others by default and we can quickly delete them from S3.
+ So my overall thought is that using this S3 bucket in development/demo Heroku site is very safe. But I may be missing something and I'm curious to hear your thoughts on this!

kevinrobinson: @alexsoble This is awesome and super thorough!  I see the same things you're seeing with regards to security, it seems like there are a lot of steps here that would require us to make a series of mistakes to confuse this with a production bucket, which is great!

I was thinking about this more in terms of the developer experience, and isolation across different environments.  I see the service is mocked out in test so that test runs locally and in Travis won't use this bucket if I follow.  So next I'm wondering about if the same bucket would be used in development locally by everyone on the team, or maybe on the demo site too, and whether that involves coordinating and sharing credentials.

Your pros and cons for mocking/using a real service are great too!  I do think that the way we are using the S3 API is simple enough (eg, get/put) where it could be a good candidate for mocking the service in development.

alexsoble:
I'm using the planet image as a placeholder "photo" in development environments for now. Some kind of more-realistic-but-still-obviously-fake image might be better in the future, but for now I wanted to focus on adding a working, well-tested endpoint for student photos. We can always swap in a better demo image in the future. 

I think this PR should be ready to 🚢! Made a list of Next Steps in the PR description for what comes after it ships.

alexsoble:
@kevinrobinson That said, if you spot anything that needs fixing or raises questions, let me know!

kevinrobinson:
@alexsoble all sounds great! 👍

alexsoble:
⚙️  Set ENV variable on Heroku demo site . . .

alexsoble:
⚙️  Rebuilding demo site using script . . .

alexsoble:
⚙️  Verified student photo download via URL in production . . .

PR #1806, opened 2018-06-14
HS counselors: scope high absences and dashboards to their caseload

kevinrobinson:
# Who is this PR for?
HS counselors

Follow on to https://github.com/studentinsights/studentinsights/pull/1798, related to https://github.com/studentinsights/studentinsights/issues/1675.

# What does this PR do?
Updates the "high absences" insights box on the home page to only show HS counselors' caseload students.  This matches the set of students shown in their feed.  Similarly, update the absences / tardies / discipline dashboards to do the same thing.  For HS counselors, this means that while they currently still have access to all students (and can see search within the search bar, see homerooms and sections, and the HS roster), they will only see their caseload in most links they click that aren't explicitly scoped more widely (eg, a section page).

This also factors out the dashboard queries and serialization from the `schools_controller` into a `DashboardQueries` file.  @edavidsonsawyer had already awesomely organized this within the file and commented which were associated, so this just enforces that nice organization within its own class.

# Checklists


kevinrobinson:
selfie, reworked this to not touch authorization and just be a filter on top

PR #1804, opened 2018-06-14
Import SPED liaison for student, show it on the student profile page

kevinrobinson:
# Who is this PR for?
all educators, driven by HS classroom teachers

# What problem does this PR fix?
HS classroom teachers asked for this, for more easily coordinating services across gen ed/sped.

# What does this PR do?
Exports a new field from Somerville Aspen that is read by the `students_importer` into the `Student` model.  Sends this down to the UI and renders it in the student profile page.

Since this crunched the information on the "Interventions" column, I updated the styling slightly, as well as the layout and what information is included.  To help with this, I reworked the tests and added stories to make this visible for a few students, and added snapshot tests for those stories.

I'm assuming we'll do some redesign work on the profile here, so this is mostly aimed at meeting the immediate needs of something folks asked for during the last week and a half of school.

# Screenshot (if adding a client-side feature)
screen shot 2018-06-14 at 11 40 43 am
screen shot 2018-06-14 at 11 40 39 am

# Checklists
+ [x] Author improved specs for code in need of better test coverage
+ [x] Author included specs for new code

PR #1807, opened 2018-06-15
Fix bug with reading 504 plan values in student profile page

kevinrobinson:
# Who is this PR for?
everyone

https://github.com/studentinsights/studentinsights/pull/1804 has a JS bug that impacts students with string values in the `plan_504` field and crashes rendering of the profile page. I triggered this manually testing, and it didn't come up in tests because we don't test exhaustively for all permutations of student profile values, and the three students we test didn't have this case.  This PR adds another student who has a 504 plan to add a bit more coverage.

I was verifying this manually, and tested on New Bedford (with doesn't have Rollbar enabled, which is why there are no alerts even though the profile page is effectively down).

# What does this PR do?
fixes it, adds model validations to document production values, fixes JS bug, adds better JS test coverage.

screen shot 2018-06-14 at 9 34 52 pm

# Checklists


kevinrobinson:
selfie

PR #1810, opened 2018-06-15
HS levels prototype: Performance optimizations

kevinrobinson:
# Who is this PR for?
HS pilot users

https://github.com/studentinsights/studentinsights/pull/1801 takes a while to load, often over 30 seconds, which hits the Heroku timeout.

# What does this PR do?
Makes performance optimizations to the querying, primarily batching up the query for latest notes into an optimized single query that then gets merged back into the response.  It also adds a new `PerfTest`.  Here's production data, although note that there's a fair bit of variability here:

```
# before
[["all_for_educator", 105746],
[["all_for_educator", 71790],

# after
[["all_for_educator", 10728],
[["all_for_educator", 15650]]
```

So still not fast :) but under the 30 second timeout threshold.

### DIFF
--- a/app/assets/javascripts/tiering/StudentLevelsTable.js
+++ b/app/assets/javascripts/tiering/StudentLevelsTable.js
@@ -121,7 +121,7 @@ export default class StudentLevelsTable extends React.Component {
 
   renderStudent({rowData}) {
     const student = rowData;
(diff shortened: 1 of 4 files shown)


kevinrobinson:
selfie

PR #1811, opened 2018-06-15
Filter excused absences

edavidsonsawyer:
# Who is this PR for?
Dashboard users

We have data on whether absences have been excused, but the dashboard doesn't currently show that information. 

# What does this PR do?
Shows users only unexcused absences by default and adds a button to toggle to all absences. 

excusedfilter

# Checklists

## Javascript QA

## Automated Testing


edavidsonsawyer:
@kevinrobinson @alexsoble this should be ready for review now. I tried out a few checkbox-y things for the excused filter, but I think it ended up being confusing to have those work differently than the day filters, so I went with the button you see in the screenshot. 

Right now this is filtering all excused/dismissed absences, regardless of the reason. If it's better to ship this with only medically excused absences removed, that's a relatively small change to make.

edavidsonsawyer:
I'm not totally sure why the student profile spec is failing here.

kevinrobinson:
@edavidsonsawyer https://github.com/studentinsights/studentinsights/pull/1853 should help, sorry about that.  Restarting the build.

kevinrobinson:
Okay, JS tests look good now.  The build is still red for this PR but it's just a small Rubocop fix: https://travis-ci.org/studentinsights/studentinsights/builds/396914392#L1961

app/assets/javascripts/school_administrator_dashboard/absences_dashboard/SchoolwideAttendance.js
```diff
@@ -42,30 +67,36 @@ class SchoolwideAttendance extends React.Component {
...
     return (
         <SchoolAbsenceDashboard
           schoolAverageDailyAttendance = {this.schoolAverageDailyAttendance()}
-          homeroomAverageDailyAttendance = {this.homeroomAverageDailyAttendance()}
-          schoolAbsenceEvents = {this.props.schoolAbsenceEvents}
+          schoolAverageDailyAttendanceUnexcused = {this.schoolAverageDailyAttendanceUnexcused(schoolUnexcusedAbsenceEventsByDay)}
+          homeroomAverageDailyAttendance = {this.homeroomAverageDailyAttendance(this.homeroomAbsenceEventsByDay(studentsByHomeroom))}
+          homeroomAverageDailyAttendanceUnexcused = {this.homeroomAverageDailyAttendance(this.HomeroomAbsenceEventsByDayunexcused(studentsByHomeroom))}
+          schoolAbsenceEventsByDay = {this.props.schoolAbsenceEventsByDay}
+          schoolUnexcusedAbsenceEventsByDay = {schoolUnexcusedAbsenceEventsByDay}
```
kevinrobinson: This looks to me like it is calculating everything for both conditions all the time.  I'm wondering if there are ways to simplify this do calculate only what is needed based on the user's selection.  Are there ways we could push this filtering upward, so that the component doesn't change - we apply a filter to the absence events, and the rest of the code stays unchanged?

Looking more closely at the component, I'm wondering if maybe having the separation between placing the computation in `SchoolwideAttendance` and the state about the filters in `SchoolAbsenceDashboard` is what's driving this.  Concretely, would it help simplify if `SchoolAbsenceDashboard` were the main component, that owned the bits of state, and then it just called into the various functions it needed to compute things (that now live in `SchoolwideAttendance`?  It seems like that might remove the need to compute both things.

Or if the idea that computing both is intended as a performance optimization related to the comment [here](https://github.com/studentinsights/studentinsights/pull/1811/files#diff-6f3e04125c9062d8f9711609485b94a8R8) about responsiveness?

app/assets/javascripts/school_administrator_dashboard/absences_dashboard/SchoolwideAbsences.js
```diff
@@ -7,9 +7,11 @@ import SchoolwideAttendance from './SchoolwideAttendance';
 class SchoolwideAbsences extends React.Component {
 
   render() {
+    const allAbsenceEvents = DashboardHelpers.absenceEvents(this.props.dashboardStudents);
     return (
       <SchoolwideAttendance
-        schoolAbsenceEvents = {DashboardHelpers.absenceEventsByDay(this.props.dashboardStudents)}
+        schoolAbsenceEvents = {allAbsenceEvents}
+        schoolAbsenceEventsByDay = {DashboardHelpers.eventsGroupedByDay(allAbsenceEvents)}
```
kevinrobinson: I don't follow why this API is changing or why the same data in two different shapes are passed in, is this to try to re-use the grouping computation?

app/assets/javascripts/school_administrator_dashboard/absences_dashboard/SchoolwideAttendance.js
edavidsonsawyer: @kevinrobinson yeah, you have this right. The trouble here is that we have to go from students -> absences -> daily averages -> monthly averages. In early versions of the dashboard, those were computed with each state change and it was irritatingly slow. Not unusable, but not ideal. If the dashboard gets the daily averages as a prop, this is much faster. However, if we start to do more with filtering based on other absence attributes, this model gets unwieldily. 

I agree this is kind of clumsy. One thought might be to allow filtering in the class list (which seems like the most useful part of the component) and not in the charts, which is the piece that cares about these averages. I hadn't come up with a good way to communicate that visually, though, so I haven't tried it. 

One important point is that this problem is unique to the absence dashboard and tardies/discipline have a more streamlined functional approach.

edavidsonsawyer:
@kevinrobinson I updated with two buttons. I'd gone back and forth on this design myself so I already had a commit ready. I left a response on the architecture and where to calculate. I'm absolutely open to alternative ways to handle it.

app/assets/javascripts/school_administrator_dashboard/absences_dashboard/SchoolwideAbsences.js
edavidsonsawyer: @kevinrobinson it is, but frankly this may be more than is necessary. V1 of this feature was slow enough to render than everything got separated out wherever possible, but for readability's sake it might be better to bite the bullet and call this function a bunch of times in the next component. If you think that's better, I'll consolidate them.

app/assets/javascripts/school_administrator_dashboard/absences_dashboard/SchoolwideAttendance.js
kevinrobinson: @edavidsonsawyer Thanks, this is super helpful!  and now I'm more on the same page with how this works and what you're thinking 👍

We could make these improvements to simplify, but let's keep moving forward.  I'll look into the data quality on this tomorrow morning and see what comes up there.  Assuming that's mostly in good shape, then we can 🚢 this tomorrow, and we can see what other needs come up down the line (eg, next time we add something, if the factoring makes that change more involved, maybe do some refactoring then).

app/assets/javascripts/school_administrator_dashboard/absences_dashboard/SchoolwideAbsences.js
kevinrobinson: Got it.  Did you do this because of real performance issues you encountered making this change, or defensively to avoid those kinds of problems?  I think let's 🚢 this regardless, I'm  wondering on the current state.

kevinrobinson:
@edavidsonsawyer Got it, thanks for talking through this!  Let's keep moving and I'll verify the data tomorrow and then we can :ship this.

Here's some ideas on optimizing compute without changing component APIs that I've found helpful.  Sometimes you have to rearrange the ways components are factored, but this can be helpful for optimizing without having to do that kind of surgery :)

If something is computed multiple times in `render` within a component, one optimization is to move up that computation into the `render` method itself, and then pass that computed value into other render methods within the component.  This lets methods within the component re-use computation while still staying stateless and not having to move the computation outside of the component.

If something is computed multiple times between components, or over and over again in one component because something unrelated changed, one optimization is to use `_.memoize` on a helper function, which will store the value of past computations in memory outside of the component tree.  That can let the computation stay lazy and on-demand, but get some of the same kinds of benefits as precomputing different combinations up front.

Anyway, this might be a place to do some refactoring next time around, but I think let's 🚢  and get into the "heatmap" component, since this will unlock looking at the distinctions between this data, and we can always refactor more later if we find it necessary :)

kevinrobinson:
@edavidsonsawyer As an update, there's data issues we need to resolve (https://github.com/studentinsights/studentinsights/issues/1657) before we can ship this PR.  The most important one is that we have only imported these fields for the last ~100 days, so any counts for "unexcused absences" over a longer period of time will be wrong.

I'll keep working on this tomorrow.

kevinrobinson:
🚢 and verifying this!

PR #1812, opened 2018-06-15
Levels: Allow querying by school slug

kevinrobinson:
# Who is this PR for?
HS educators

Makes the URL more stable and sensible, since this isn't visible by a link right now.


kevinrobinson:
selfie

Issue #1809, opened 2018-06-15
Discipline: Total incidents and student incident counts don't update when changing time range


kevinrobinson:
This was definitely working before (see GIF in https://github.com/studentinsights/studentinsights/pull/1670), but seems to have regressed.

To reproduce:
- Load the page for any school and click on the time filter.
- I was expecting the "total incidents" at the bottom left to update, and the "incidents" for each student to update.
- They seem to be unchanged regardless of the time filter.

edavidsonsawyer:
@kevinrobinson I ran into this while working on #1811. I should have a fix within a day or so.

kevinrobinson:
@edavidsonsawyer awesome, sounds great! 👍

edavidsonsawyer:
Should be all set with #1815

PR #1815, opened 2018-06-17
Fix discipline incident counts

edavidsonsawyer:
#1809 which was I broke in #1712 causing the discipline incident counts to remain the same across all time ranges

# What does this PR do?
Correctly updates the discipline counts

goodfilter

# Checklists

## Javascript QA

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
+++ b/app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
@@ -46,7 +46,9 @@ class SchoolDisciplineDashboard extends React.Component {
   studentDisciplineIncidentCounts(incidentCategory) {
     let studentDisciplineIncidentCounts = {};
     const selectedChartData = this.getChartData(this.state.selectedChart);
-    const incidents = incidentCategory ? selectedChartData.disciplineIncidents[incidentCategory] : this.props.schoolDisciplineEvents;
+    const incidents = incidentCategory ?
+                      selectedChartData.disciplineIncidents[incidentCategory] :
+                      this.filterIncidentDates(this.props.schoolDisciplineEvents);
     incidents.forEach((incident) => {
       studentDisciplineIncidentCounts[incident.student_id] = studentDisciplineIncidentCounts[incident.student_id] || 0;
       studentDisciplineIncidentCounts[incident.student_id]++;


edavidsonsawyer:
@kevinrobinson @alexsoble this should be all set when there's time to look at it

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -46,7 +46,9 @@ class SchoolDisciplineDashboard extends React.Component {
   studentDisciplineIncidentCounts(incidentCategory) {
     let studentDisciplineIncidentCounts = {};
     const selectedChartData = this.getChartData(this.state.selectedChart);
-    const incidents = incidentCategory ? selectedChartData.disciplineIncidents[incidentCategory] : this.props.schoolDisciplineEvents;
+    const incidents = incidentCategory ?
+                      selectedChartData.disciplineIncidents[incidentCategory] :
+                      this.filterIncidentDates(this.props.schoolDisciplineEvents);
```
kevinrobinson: @edavidsonsawyer This seems like it could work, but then it also seems like it doesn't fix the root problem, which is keeping the data in the table in sync with the data in the charts.

I'm wondering if moving the filtering into the `render` method, and then have it pass the filtered results down to the subsequent `renderStudentDisciplineTable` and `renderDisciplineChart` methods, if this would remove this class of bugs altogether.

So to sketch that:
```
render() {
  const students = this.filterStudentsBasedOnSelections(selectedChart, selectedCategory);
  const incidents = this.filterDisciplineIncidentsBasedOnSelections(selectedChart, selectedCategory);
  return (
     ...
     {this.renderStudentsTable(students, incidents)}
     {this.renderCharts(students, incidents)}
     ...
  );
```

I'm not sure if this is worthwhile, so I think that's your call to make, and since this fixes the immediate bug let's 🚢 it now :)

kevinrobinson: @edavidsonsawyer In testing this, this PR has a bug so I reverted shipping it in this revert PR: https://github.com/studentinsights/studentinsights/pull/1816.

To reproduce, select a category (eg, "Time"), click a time of day, then change the time filter.  This is what I see locally:

screen shot 2018-06-18 at 12 39 01 pm

So we can't ship this PR as-is.  I'd recommend a refactoring similar to what I suggested [here](https://github.com/studentinsights/studentinsights/pull/1815#discussion_r196145564) as a path forward, since this class of bug seems to be making it tricky to make changes, and changing that factoring might help.  If you want to chat about adding another test or two to help catch these kinds of things, I'm happy to chat about that too.

PR #1817, opened 2018-06-18
Add index to students table on enrollment_status

kevinrobinson:
# Who is this PR for?
K8 educators, especially HS classroom teachers

In looking at query speed for the `/educators/my_students` page, one thing that came up is the common `Student.active` query that is passed to the authorizer (and in several other places too) is filtering down to only active students, but there's no index on that field.  This would add one, which in looking at SPS production data would cut this to 68% of all student row.

I'm not certain the query planner will use this - it doesn't in development but suspect it will in production.

# What does this PR do?
Adds an index on `enrollment_status` on the `students` table.

### DIFF
--- a/app/models/student.rb
+++ b/app/models/student.rb
@@ -38,7 +38,7 @@ class Student < ActiveRecord::Base
   validates :first_name, presence: true
   validates :last_name, presence: true
   validate :validate_grade
-  validate :validate_registration_date_cannot_be_in_future
+  validate :registration_date_cannot_be_in_future
   validate :validate_free_reduced_lunch
 
   VALID_GRADES = [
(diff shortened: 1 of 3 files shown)


alexsoble:
@kevinrobinson Ooooh, interesting, great optimization find here Kevin!

kevinrobinson:
selfie

kevinrobinson:
@alexsoble Yeah, not sure if this will really help yet, but it's easy to find out! :)

kevinrobinson:
huh, the Postgres planner doesn't seem to use this index on `Student.active` but would use it for other queries related to enrollment status if we made those.  I didn't look into this more to understand why it doesn't think the index is helpful, but this change has no impact on the speed of these queries.

PR #1818, opened 2018-06-18
My students: Add more filters and fields

kevinrobinson:
# Who is this PR for?
educators, primarily HS classroom teachers

Browser search doesn't work within react-virtualized, so this adds it.  It also makes more visible fields that would be useful for coordinating services.

# What does this PR do?
Adds searching and shows columns for program, SPED liaison.

screen shot 2018-06-18 at 2 44 20 pm

# Checklists
+ [x] Author checked latest in IE - My students


kevinrobinson:
selfie

PR #1820, opened 2018-06-18
Remove unused student enrollment status index

kevinrobinson:
# Who is this PR for?
developers, clean up

https://github.com/studentinsights/studentinsights/pull/1817 added an index to see if it would be used in a common query; it's not.

# What does this PR do?
Removes the index.


kevinrobinson:
selfie

PR #1821, opened 2018-06-18
Tune MyStudents column widths

kevinrobinson:
# Who is this PR for?
HS educators in particular.

follow-on to https://github.com/studentinsights/studentinsights/pull/1818, the column width tuning doesn't look great with production data, this updates it to compress school name over other values.


kevinrobinson:
selfie

PR #1822, opened 2018-06-18
Add logging for class list snapshot

kevinrobinson:
# Who is this PR for?
developers, follow-on to https://github.com/studentinsights/studentinsights/pull/1796

Running this revealed that the diff check doesn't work.

# What does this PR do?
Adds more verbose logging for seeing progress and debugging; adds a more conservative equality check for differences in date/time classes.


kevinrobinson:
selfie

PR #1823, opened 2018-06-18
Add student photo to Student Profile Header

alexsoble:
# Who is this PR for?

All the educators who use the student profile.

In large meetings at larger schools, some educators won't recognize a particular child by name alone. For example, at the start of the new school year in the High School, House Masters have hundreds of new students in their caseload and it takes time for them to learn every student's face. 

# What does this PR do?

Add student photo to the Student Profile page. The goal here is to add a small-sized photo in order to create as little design change in the StudentProfileHeader component as possible, since we'll be working on a more robust & thorough redesign for the Student Profile Page over the summer.

## Screenshot (demo data, small-sized original "photo")
student-small
## Screenshot (demo data, larged-sized original "photo")
student-large
## Screenshot (demo data, student has no "photo")
student-no-photos
## Screenshot (demo data, New Bedford: student photo feature not enabled)
Wasn't able to test this one manually — for some reason, no matter what I did, `window.shared.Env` always returned `somerville` in development, even if the value was set correctly on the server side. 

I spent some time trying to figure that one out and got stuck. I decided I shouldn't let this manual testing issue block this PR for a few reasons: 

+ The automated JS tests show that the UI responds correctly when `districtKey` is set to something other that "somerville."
+ Even if this code fails and the student photo `img` tag is rendered, it will show a single white pixel because the associated student will have no student photo records.
+ In the New Bedford prod site, `window.shared.Env` returns the expected results. 

# Handling edge cases 

## Students without a student photo

I wondered how many students are missing a student photo.

```
> Student.active.count
=> 4768
StudentPhoto.count
=> 4777
```

We've stored more student photos than active students. But some of those stored photos may be for non-active students.

```
active_no_photo = 0

Student.active.each do |s|
  if s.student_photos.size == 0
   active_no_photo += 1
 end
end

active_no_photo
=> 269
```

So about 5.6% of active students don't have any imported photos from Aspen/X2.

## Different student photo dimensions 

Running `ls -S -lat > sorted.txt` showed a few things about the photos we have from Somerville:

+ The smallest photo we have is 4,787 bytes. Its dimensions are 172x207.
+ The largest photo we have is 154,927 bytes. Its dimensions are 356x433.
+ Excluding three outliers on the large end, most photo dimensions are between 308x364 and 172x207.


alexsoble:
Very initial mocks, figuring out how to position the student image in the profile. Very ugly and leaves a ton of white space on the table. I wonder what it would look like to have a fixed sidebar with student name, key details, and photo? That way the photo is visible even when the SST teams have scrolled down to take notes.

screen shot 2018-06-19 at 12 48 27 pm

kevinrobinson:
@alexsoble Nice! 👍 

My vote would be to punt incremental design changes here and tackle this more holistically (eg as part of the "best list" work).  I think an ideal way to do that design work would be to batch up a few known issues (eg, profile data not useful when taking notes, doesn't show student voice info yet, doesn't yet show transition note strengths to all educators).  Then we could spend ~1-2 weeks on design iteration to address those all at once, since those kinds of design and layout changes for different improvements will all be interrelated and in tension with each other.  It might be harder to tackle iteratively and still have the page feel cohesive.  If that's something you wanted to start up now, I'm up for chatting about it! 

Or we could 🚢 the smallest incremental change and get photos live this week.  But spending design energy on other layouts seems like it won't have immediate impact (school gets out Thursday) and it would probably be overwritten by other design work later this summer.

alexsoble:
@kevinrobinson Interesting! Going to give you a quick phone call to talk about some of the tradeoffs here.

kevinrobinson:
@alexsoble For `ClassListsViewPage`, you can try out https://github.com/studentinsights/studentinsights/pull/1833, which merges into this PR.  I can verify that the patch affects the behavior of tests locally for me if I change the timezone on my computer.

After merging that change into your PR, you'll also need to re-run the tests and update the snapshot for `ClassListsViewPage`.  If I read the diff here and the test failures right, that spec doesn't properly fix the timezone in test.  So you probably ran the tests on your computer and saw this snapshot test fail locally.  You then updated the snapshot, which isn't great, since your change shouldn't have impacted the `ClassListsViewPage`.  But now the snapshot in this PR essentially says "only pass when run in central time zone."  The test had passed in Travis before, but when you checked in the updated snapshot run in CT, that's why it started failing in Travis (which is in ET).

Let me know if that helps issue no. 2 above, or you want to pair on any remaining bits for it.  That one's a bug with the class list test code.

alexsoble:
@kevinrobinson Alrighty, we can try that. 😄 

Overall I agree with this comment from October 2017: https://github.com/studentinsights/studentinsights/pull/1216#issuecomment-340110958. I feel the fatigue from snapshots failing on every change, and I'm not super good at catching bugs in snapshot diffs. I'm happy to run with it for a while and then discuss it more but right now I'm just not seeing the upside.

kevinrobinson:
@alexsoble Yep, re: snapshots that's my bad if there's first-time learning friction, and I'm happy to pair with you anytime to share what workflow's worked well for me here.  It sounds like the mixture of figuring out snapshots along with a good ol' pesky timezone bug is making this a pain.  It also might be helpful to check out the FAQs https://facebook.github.io/jest/docs/en/snapshot-testing.html.  And it'd be awesome if you can help make find ways to make testing smoother and simpler.

Snapshots are a little different than normal tests, and if the snapshot files themselves are committed with mistakes, it can be confusing to untangle.  Looking back at commits, I think that's what happened early on here: https://github.com/studentinsights/studentinsights/pull/1823/commits/ad0564a5407f839a138dc26a501bc1cc83858d17 - that checks in snapshots with the timezone change and the different pixel values for the SVG bits (which I think is coming from the same underlying timezone problem).  And from that point on it comes harder to understand what's happening because there's drift between local test runs and Travis and it's unrelated to the changes you're actually making :\

Here's a plan to fix the timezone problem going forward:
- Let's improve the way we set the timezone in tests, so that when you and I and Travis run the tests we're running the same time :)  Merging https://github.com/studentinsights/studentinsights/pull/1835 to master will do that.
- That will let us remove the short-term patch into this PR yesterday. 
 Reverting https://github.com/studentinsights/studentinsights/pull/1833 will do that.
- Then we can update this PR with the code to set the timezone.  Merging master into this branch will do that (since master will now have https://github.com/studentinsights/studentinsights/pull/1835).

At that point, you can check how things work in local development or with the student profile stories using `yarn storybook` and verify things look good.  If they do:
- check out the two snapshot files on this PR back to what's on master to start fresh
- run the tests
- only the student profile snapshot should be changed
- and since you just verified it looks good, you can update it
- and then 🚢 !

alexsoble:
Thanks for the help with fixing these flaky tests @kevinrobinson! I think we successfully got my machine to run them as if I were in Eastern Time! ✨

alexsoble:
@kevinrobinson Yay! Tests are passing! Thanks again for your help. 😄🌈 

Next up today, I'm going to add screenshots to the PR description, do a review of the code, and 🚢.

kevinrobinson:
@alexsoble Awesome!  No problem, thanks for spotting this so we can fix it up, and I'm happy to help out so you can keep moving with shipping some awesome 👩‍🎓 📷 📸s!

alexsoble:
Ah! Remembered that we want to only run this front-end code for Somerville, so we need to add this to `PerDistrict.js` on the front end to match the per-district function on the Ruby side.

alexsoble:
Updated the pull request description with screenshots and comments about manual testing . . . time to 🚢.

PR #1824, opened 2018-06-18
Trim cruft from devise config file

alexsoble:
Note: This PR takes one change from #1697, a larger authentication-related pull request, and isolates it into its own pull request so it can be shipped independently.

# Who is this PR for?

Developers working with authentication.

Our `devise.rb` config file is extremely verbose and contains large paragraph-style comments on Devise features that we aren't using and don't intend to use. 

# What does this PR do?

Trims out the "cruft" sections from `devise.rb`.

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -5,37 +5,12 @@ Devise.setup do |config|
   # random tokens. Changing this key will render invalid all existing
   # confirmation, reset password and unlock tokens in the database.
   config.secret_key = ENV['DEVISE_SECRET_KEY']
-  # ==> Mailer Configuration
-  # Configure the e-mail address which will be shown in Devise::Mailer,
-  # note that it will be overwritten if you use your own mailer class
-  # with default "from" parameter.
-  config.mailer_sender = 'please-change-me-at-config-initializers-devise@example.com'
-
-  # Configure the class responsible to send e-mails.
-  # config.mailer = 'Devise::Mailer'
 
   # ==> ORM configuration
   # Load and configure the ORM. Supports :active_record (default) and
   # :mongoid (bson_ext recommended) by default. Other ORMs may be
   # available as additional gems.
   require 'devise/orm/active_record'
-
-  # ==> Configuration for any authentication mechanism
-  # Configure which keys are used when authenticating a user. The default is
-  # just :email. You can configure it to use [:username, :subdomain], so for
-  # authenticating a user, both parameters are required. Remember that those
-  # parameters are used only when authenticating and not when retrieving from
-  # session. If you need permissions, you should implement that in a before filter.
-  # You can also supply a hash where the value is a boolean determining whether
-  # or not authentication should be aborted when the value is not present.
-  # config.authentication_keys = [ :email ]
-
-  # Configure parameters from the request object used for authentication. Each entry
-  # given should be a request method and it will automatically be passed to the
-  # find_for_authentication method and considered in your model lookup. For instance,
-  # if you set :request_keys to [:subdomain], :subdomain will be used on authentication.
-  # The same considerations mentioned for authentication_keys also apply to request_keys.
-  # config.request_keys = []
 
   # Configure which authentication keys should be case-insensitive.
   # These keys will be downcased upon creating or modifying a user and when used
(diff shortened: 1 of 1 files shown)


config/initializers/devise.rb
```diff
@@ -5,38 +5,13 @@
   # random tokens. Changing this key will render invalid all existing
   # confirmation, reset password and unlock tokens in the database.
   config.secret_key = ENV['DEVISE_SECRET_KEY']
-  # ==> Mailer Configuration
```
alexsoble: We never mail our educator users. They can't sign up and can't recover or change their passwords through Devise, because we are using our district partners' LDAP system for authentication.

config/initializers/devise.rb
```diff
@@ -5,38 +5,13 @@
...
-  # Configure the class responsible to send e-mails.
-  # config.mailer = 'Devise::Mailer'
 
   # ==> ORM configuration
   # Load and configure the ORM. Supports :active_record (default) and
   # :mongoid (bson_ext recommended) by default. Other ORMs may be
   # available as additional gems.
   require 'devise/orm/active_record'
 
-  # ==> Configuration for any authentication mechanism
```
alexsoble: We are using one instance per district, so we don't need to check for things like subdomain in the auth mechanism.

config/initializers/devise.rb
```diff
@@ -98,31 +73,6 @@
   # Setup a pepper to generate the encrypted password.
   # config.pepper = 'd68602c07af77ffeb4817fad17efeda6ece23b89d3b2b42126b56b2a99a8cf5c653f47ad1298a1b39d8e866835b48148875e84883c308c3addf4d3acbc124e60'
 
-  # ==> Configuration for :confirmable
```
alexsoble: We aren't using confirmable because responsibility for authentication lies in our district partners' LDAP systems.

config/initializers/devise.rb
```diff
@@ -151,66 +101,6 @@
...
-  # is failed attempts.
-  # config.maximum_attempts = 20
-
-  # Time interval to unlock the account if :time is enabled as unlock_strategy.
-  # config.unlock_in = 1.hour
-
-  # Warn on the last attempt before the account is locked.
-  # config.last_attempt_warning = true
-
-  # ==> Configuration for :recoverable
```
alexsoble: We don't use recoverable.

config/initializers/devise.rb
```diff
@@ -151,66 +101,6 @@
   # time the user will be asked for credentials again. Default is 30 minutes.
   config.timeout_in = 6.hours
 
-  # ==> Configuration for :lockable
```
alexsoble: I explored implementing lockable in #1697. Turns out lockable doesn't work as a drop-in replacement for us because we're using our own ldap_authenticable_tiny strategy instead of database_authenticable. So my plan is to not use Devise's out-of-the box lockable at all and roll our own implementation.

config/initializers/devise.rb
```diff
@@ -151,66 +101,6 @@
...
-  #
-  # Defines which key will be used when recovering the password for an account
-  # config.reset_password_keys = [ :email ]
-
-  # Time interval you can reset your password with a reset password key.
-  # Don't put a too small interval or your users won't have the time to
-  # change their passwords.
-  config.reset_password_within = 6.hours
-
-  # ==> Configuration for :encryptable
```
alexsoble: We're not using our own encryption algorithm because we're running with the default bcrypt.

config/initializers/devise.rb
```diff
@@ -151,66 +101,6 @@
...
-  # Allow you to use another encryption algorithm besides bcrypt (default). You can use
-  # :sha1, :sha512 or encryptors from others authentication tools as :clearance_sha1,
-  # :authlogic_sha512 (then you should set stretches above to 20 for default behavior)
-  # and :restful_authentication_sha1 (then you should set stretches to 10, and copy
-  # REST_AUTH_SITE_KEY to pepper).
-  #
-  # Require the `devise-encryptable` gem when using anything other than bcrypt
-  # config.encryptor = :sha512
-
-  # ==> Scopes configuration
```
alexsoble: We only have one user scope.

config/initializers/devise.rb
```diff
@@ -224,32 +114,4 @@
 
   # The default HTTP method used to sign out a resource. Default is :delete.
   config.sign_out_via = :delete
-
-  # ==> OmniAuth
```
alexsoble: We're not using OAuth.

alexsoble:
@kevinrobinson Force-pushing this up to the demo site. I want to triple-check that removing un-commented `config...` lines that pertain to modules we aren't using won't break the site.

kevinrobinson:
@alexsoble and remember that with your awesome work in https://github.com/studentinsights/studentinsights/pull/1701 the full Devise code path is being exercised in the RSpecs too! 👍

alexsoble:
@kevinrobinson Heyyyy great point! Demo site looks good so I'm going to 🚢.

alexsoble:
Deploying to all.

PR #1826, opened 2018-06-19
Clear sessionStorage on sign out

kevinrobinson:
# Who is this PR for?
educators, students

For signed-in users, student names for the search bar are cached in `sessionStorage` to keep the searchbar navigation speedy across pages.  The browser clears this data when the tab or window is closed.  In the event that a signed-in user signs out and keeps the tab or window open afterward, these names aren't cleared and would be accessible to someone on the JS console, or by signing into the product as a different user with different permissions.

# What does this PR do?
Clears `sessionStorage` on sign out, and on loading a signed-out page as an additional guard.

### DIFF
--- a/app/assets/javascripts/studentSearchbar.js
+++ b/app/assets/javascripts/studentSearchbar.js
@@ -26,7 +26,7 @@ function downloadStudentNames () {
   });
 }
 
-export default function studentSearchbar() {
+export function initSearchBar() {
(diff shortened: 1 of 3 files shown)


kevinrobinson:
selfie

kevinrobinson:
verified this works as expected

alexsoble:
👍 👍

PR #1827, opened 2018-06-19
Update ClassListSnapshot spec to capture log output

kevinrobinson:
# Who is this PR for?
developers

# What problem does this PR fix?
The spec for ClassListSnapshot logs in test, making test output noisy.

# What does this PR do?
Logs to `FakeLog` in test.

PR #1828, opened 2018-06-19
Add diff function for looking at ClassListSnapshot changes

kevinrobinson:
# Who is this PR for?
developers

The class list snapshotting strategy is conservative, capturing all the information needed to reconstruct changes over time in what the user would have seen, but this makes it hard to directly see what has actually substantively changed for the user over a longer period of time.

# What does this PR do?
This adds a `diff_across_workspaces` method that makes this easier, for validating https://github.com/studentinsights/studentinsights/pull/1796.  I think this probably will be the last step before actually building tools or visualizations to see these changes.

# Checklists

### DIFF
--- a/app/models/class_list_snapshot.rb
+++ b/app/models/class_list_snapshot.rb
@@ -7,7 +7,7 @@ class ClassListSnapshot < ActiveRecord::Base
   # that has changed.  This is to track changes as student data changes over time,
   # while the student_ids in the class list record stay the same.
   def self.snapshot_all_workspaces(options = {})
-    log = options[:log] || STDOU
+    log = options[:log] || STDOUT
     snapshots_taken = []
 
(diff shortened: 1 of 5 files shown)


app/models/class_list_snapshot.rb
```diff
@@ -7,7 +7,7 @@ class ClassListSnapshot < ActiveRecord::Base
   # that has changed.  This is to track changes as student data changes over time,
   # while the student_ids in the class list record stay the same.
   def self.snapshot_all_workspaces(options = {})
-    log = options[:log] || STDOU
+    log = options[:log] || STDOUT
```
kevinrobinson: bug in https://github.com/studentinsights/studentinsights/pull/1827

kevinrobinson:
selfie

PR #1829, opened 2018-06-19
Remove some logging from class list snapshots, filter out nil diffs

kevinrobinson:
# Who is this PR for?
developers, follow on to https://github.com/studentinsights/studentinsights/pull/1828

The rake task logs more than it should, no need for that getting into logentries.  The diff function also includes nil values for workspaces that haven't changed at all, no need for that either.

# What does this PR do?
Removes the logging, filters out the nils.


kevinrobinson:
selfie

Issue #1704, opened 2018-05-18
Improvement: Resolve warnings and enable CSP


kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/1698.

Enabling CSP in report-only mode on the demo site shows a few things we need to resolve:
- [x] Update Rollbar (https://github.com/studentinsights/studentinsights/pull/1702)
- [x] switch to `CSP_REPORT_ONLY_WITHOUT_ENFORCEMENT` (https://github.com/studentinsights/studentinsights/pull/1837)
- [x] fix `unsafe-eval` on profile page chart tooltips (https://github.com/studentinsights/studentinsights/pull/1836)
- [x] Make script rule stricter by noncing or hashing the inline Rollbar and Mixpanel JS stubs, which we want to keep in the initial HTML payload (added in Rails 5.2 http://guides.rubyonrails.org/5_2_release_notes.html#content-security-policy) - instead factor these out (https://github.com/studentinsights/studentinsights/pull/1841 and https://github.com/studentinsights/studentinsights/pull/1842)
- [x] `style-loader` uses inline styles (open issue: https://github.com/webpack-contrib/style-loader/issues/306, punted in punted in https://github.com/studentinsights/studentinsights/pull/1842)
- [x] resolve Highcharts style issues (punted in https://github.com/studentinsights/studentinsights/pull/1842).
- [x] enforce on all sites
- [x] add some notification mechanism for new violations sent to csp-logger (send to rollbar)

kevinrobinson:
@alexsoble FYI I'm going to be fiddling with CSP settings on the demo site.  If this messes you up in any way just let me know and I'll hold off and do it another time.

kevinrobinson:
Okay, with https://github.com/studentinsights/studentinsights/pull/1831 shipped, the more relaxed CSP policy is now enabled on all sites in report-only mode.  Going to let it sit for a day or two to see what is reported, then we can turn it on for realz.

kevinrobinson:
@alexsoble The `csp-logger` app will now send Rollbar warnings on a CSP violation (although right now they don't say more than "CSP violation").  We now have monitoring for these, and we can see if it's useful or too noisy and adjust it.

With that, going to consider this set of improvements done.

PR #1831, opened 2018-06-19
Relax CSP style rules

kevinrobinson:
# Who is this PR for?
developers

I tried a few approaches to resolving the issues in https://github.com/studentinsights/studentinsights/issues/1704, but neither worked.  One source comes from `react-select`, which injects some styles for IE workarounds and also an autosize components.  Disabling these didn't help, I think because we're still using `style-loader` to inject the stylesheet for the library itself.  We could work around this by adding hashes, but `react-beautiful-dnd` also injects styles.  I tried a few times to see if those were deterministic, but ran into a few that came from what looked like autosizing code and gave different hashes.

So rather than continuing to block on not enabling any CSP protection, let's relax this a bit and get it out, then we can revisit and tighten further.

# What does this PR do?
Relaxes CSP for styles to allow inline styles.  With the env flag on, this is still in report-only mode.

I verified this looks good in the demo site, and will watch the CSP logger after flipping this on in production.

### DIFF
--- a/config/initializers/secure_headers.rb
+++ b/config/initializers/secure_headers.rb
@@ -13,7 +13,12 @@ SecureHeaders::Configuration.default do |config|
       connect_src: %w('self' https:),
       form_action: %w('self' https:),
       script_src: %w('unsafe-inline' https: api.mixpanel.com cdn.mxpnl.com https://cdnjs.cloudflare.com/ajax/libs/rollbar.js/),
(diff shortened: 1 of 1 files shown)


kevinrobinson:
selfie

kevinrobinson:
@alexsoble Featuring switching this on will still only add CSP in reporting-only mode.

PR #1832, opened 2018-06-19
Upgrade to Rails 5.2

kevinrobinson:
# Who is this PR for?
developers

I initially started this to add in CSP nonces as part of https://github.com/studentinsights/studentinsights/issues/1704, since there's some helper functions that do that new in Rails 5.2 (http://guides.rubyonrails.org/5_2_release_notes.html#content-security-policy).  I ended up shifting directions with https://github.com/studentinsights/studentinsights/pull/1831 to try to incrementally ship CSP, but had already read up on the Rails upgrade and started it, so this finishes it off.

# What does this PR do?
Upgrades Rails using the helper task [here](http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-1-to-rails-5-2).  The only major migration issue is that Rails 5.2 changes the default validations for `belongs_to` associations (see https://blog.bigbinary.com/2016/02/15/rails-5-makes-belong-to-association-required-by-default.html).  Instead of addressing that here, I opted out of that, so that we can make that improvement separate from the actual version upgrade in this PR.


config/application.rb
```diff
@@ -8,6 +8,12 @@
 
 module SomervilleTeacherTool
   class Application < Rails::Application
+    # Initialize configuration defaults for originally generated Rails version.
+    config.load_defaults 5.1
+
+    # see https://blog.bigbinary.com/2016/02/15/rails-5-makes-belong-to-association-required-by-default.html
+    Rails.application.config.active_record.belongs_to_required_by_default = false
```
kevinrobinson: I added this line.

config/cable.yml
```diff
@@ -1,9 +1,9 @@
...
-  adapter: async
+# test:
+#   adapter: async
 
-production:
-  adapter: redis
-  url: redis://localhost:6379/1
+# production:
+#   adapter: redis
+#   url: redis://localhost:6379/1
```
kevinrobinson: The upgrade task suggested we add some more config in here; instead I commented it all out since it looked like we were using this but we're not.

config/environments/production.rb
```diff
@@ -44,7 +49,7 @@
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
-  config.log_level = :info
+  config.log_level = :debug
```
kevinrobinson: The update task changed this, seems fine to me.

config/initializers/assets.rb
```diff
@@ -3,9 +3,13 @@
 # Version of your assets, change this if you want to expire all your assets.
 Rails.application.config.assets.version = '1.0'
 
-# Add additional assets to the asset load path
+# Add additional assets to the asset load path.
 # Rails.application.config.assets.paths << Emoji.images_path
 
+# Add Yarn node_modules folder to the asset load path.
+# Rails.application.config.assets.paths << Rails.root.join('node_modules')
```
kevinrobinson: I commented this out, we don't want Rails getting more tangled with any JS.

config/puma.rb
```diff
@@ -30,18 +30,27 @@
 # you need to make sure to reconnect any threads in the `on_worker_boot`
 # block.
 #
-preload_app!
+# preload_app!
```
kevinrobinson: The upgrade task suggested removing this, and the comments here suggest we need to do some other config for this to work correctly (which we had commented out).  Rails 5.2 separately adds [bootsnap](http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#bootsnap), so I think better to look at this separately.

kevinrobinson:
selfie cc @alexsoble @edavidsonsawyer This'll require a `bundle install` locally.

config/environments/production.rb
kevinrobinson: It's much more verbose now, which has been somewhat useful in some debugging, but also much noisier.  I think probably best to turn this back down to :info after the holiday.

Issue #1759, opened 2018-06-04
Remove window.shared for Env


kevinrobinson:
Follow-on from https://github.com/studentinsights/studentinsights/issues/1180

- [x] Env (env / mixpanel / rollbar) (https://github.com/studentinsights/studentinsights/pull/1834)
- [x] MixPanel (https://github.com/studentinsights/studentinsights/pull/1834)
- [x] remove all use of window.shared (eslint rule on window?) (https://github.com/studentinsights/studentinsights/pull/1834)

PR #1834, opened 2018-06-19
Remove window.shared and factor out Rails env variables sent to JS

kevinrobinson:
# Who is this PR for?
developers

The immediate use for this is to enable moving Rollbar and Mixpanel tokens into env variables that we could set differently for different instances.  It also removes the last direct use of `window.shared` and adds `readEnv` so that JS code has a nice plain JS interface into this data without using globals.  That cleanup is part of https://github.com/studentinsights/studentinsights/issues/1759.

# What does this PR do?
Refactors `window.env.shared` from an ERB file in the asset pipeline into a plain Ruby class and a JS stub that writes that as a `ENV_FOR_JS` global variable on the page.  Refactors the stubs for Rollbar and Mixpanel from `.erb` templates into plain HTML that reads from `ENV_FOR_JS`.  Adds `readEnv` which other JS code can use to read this.

In a future step, we could move `readEnv` into `App.js` as context.


app/views/shared/_rollbar.html
```diff
@@ -0,0 +1,54 @@
...
+function transform(payload) {
+  payload.environment = window.ENV_FOR_JS.railsEnvironment;
+  payload.deploymentKey = window.ENV_FOR_JS.deploymentKey;
+  payload.districtKey = window.ENV_FOR_JS.districtKey;
+}
+
+// Sets global configuration for the Rollbar error reporting library,
+// based on the environment, and then loads the Rollbar code or not.
+var _rollbarConfig = { // eslint-disable-line no-var
+  accessToken: window.ENV_FOR_JS.rollbarJsAccessToken,
```
kevinrobinson: The only substantive change here is to read the token from `window.ENV_FOR_JS`, and to read from that in other places instead of `window.shared.Env`.  This code doesn't use `readEnv()` since this isn't normal JS code - it's not compiled by Webpack and is inlined into the HTML since it's on the critical path for loading and error reporting.

app/views/shared/_env_for_js.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= json_div(id: 'env-for-js', 'data-env-for-js-json': EnvForJs.new.to_json) %>
+
+<script type="text/javascript">
+const el = document.getElementById('env-for-js');
+const jsonString = el.getAttribute('data-env-for-js-json');
+window.ENV_FOR_JS = JSON.parse(jsonString);
```
kevinrobinson: This works differently than the `serialized_data` template because it needs to load before any other JS, and jQuery isn't loaded at this point.

config/environments/development.rb
```diff
@@ -4,14 +4,8 @@
...
-
-  ENV['DISTRICT_KEY'] = 'somerville'
-  ENV['DISTRICT_NAME'] = 'Localhost Public Schools'
-  ENV['USE_MOCK_LDAP'] = 'true'
-  ENV['MOCK_LDAP_PASSWORD'] = 'demo-password'
-  ENV['ENABLE_CLASS_LISTS'] = 'true'
-  ENV['ENABLE_COUNSELOR_BASED_FEED'] = 'true'
-  ENV['HOUSEMASTERS_AUTHORIZED_FOR_GRADE_8'] = 'true'
+  
+  Env.set_for_development_and_test!
```
kevinrobinson: @alexsoble FYI, this PR factors the shared env setup for dev/test into a method.

kevinrobinson:
verified this in demo, going to verify in production on deploy here too.

PR #1835, opened 2018-06-19
Fix timezone to NYC in Jest tests

kevinrobinson:
# Who is this PR for?
developers

The JS tests run in local timezone, but we use test and snapshots that expect to be running in a particular timezone.

# What does this PR do?
Sets TZ env variable when running tests.  We tried doing this with moment-timezone and I also tried just setting this env value in `setupTests` but seems that is more complicated (eg https://github.com/nodejs/node/issues/3449#issuecomment-271295444).

One wrinkle is that for developers who are working in a different timezone, there will be differences between what they see in a story/development mode and what they see in a test for the same fixture.  I'm not sure how to work around that but that's more icing while this current issue blocks development.

This is a better and more global solution than https://github.com/studentinsights/studentinsights/pull/1833, which was a short-term patch.

### DIFF
--- a/app/assets/javascripts/helpers/toMoment.test.js
+++ b/app/assets/javascripts/helpers/toMoment.test.js
@@ -10,25 +10,10 @@ it('#toMoment and then back to text', () => {
   expect(toMoment('01-5-18').format('YYYY-MM-DD')).toEqual('2018-01-05');
 });
 
-
-describe('#toMomentFromTime works as expected across timezones', () => {
-  /* eslint-disable no-undef */
-  describe('in NYC', () => {
+describe('#toMomentFromTime', () => {
+  it('works as expected because local timezone is set as expected', () => {
     const string = '2018-05-09T12:03:26.664Z';
-    const tz = process.env.TZ; 
-    process.env.TZ = 'America/New_York';
-    expect(toMomentFromTime(string).local().format('dddd M/D, h:mma')).toEqual('Wednesday 5/9, 7:03am');
+    expect(toMomentFromTime(string).local().format('dddd M/D, h:mma')).toEqual('Wednesday 5/9, 8:03am');
     expect(toMomentFromTime(string).toDate().toLocaleTimeString()).toEqual('08:03:26');
-    process.env.TZ = tz;
   });
-
-  describe('in Chicago', () => {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/helpers/toMoment.test.js
```diff
@@ -1,9 +1,19 @@
...
   expect(toMoment('3/5/2018').format('YYYY-MM-DD')).toEqual('2018-03-05');
   expect(toMoment('1/15/18').format('YYYY-MM-DD')).toEqual('2018-01-15');
   expect(toMoment('01/5/18').format('YYYY-MM-DD')).toEqual('2018-01-05');
   expect(toMoment('01-5-18').format('YYYY-MM-DD')).toEqual('2018-01-05');
-});
\ No newline at end of file
+});
+
+describe('#toMomentFromTime', () => {
+  it('works as expected because local timezone is set as expected', () => {
```
alexsoble: 👍 Gotcha — so this will fail for me locally unless I'm running the tests on EST.

package.json
```diff
@@ -14,9 +14,10 @@
     "lint-cli": "eslint --ext jsx --ext js -c .eslintrc app/assets/javascripts spec/javascripts ui",
     "dev": "NODE_ENV=development webpack --config ui/config/webpack.dev.js --watch",
     "build": "NODE_ENV=production webpack --config ui/config/webpack.prod.js",
-    "start": "concurrently --kill-others \"bundle exec rails s\" \"yarn run dev\"",
-    "test": "jest --config jest.json --watch",
-    "test-cli": "yarn run lint-cli && jest --config jest.json",
+    "start": "concurrently --kill-others \"bundle exec rails s\" \"yarn dev\"",
+    "jest": "TZ='America/New_York' jest --config jest.json",
```
alexsoble: 👍

alexsoble:
🚢

app/assets/javascripts/helpers/toMoment.test.js
kevinrobinson: Yep, that's the idea, so that if for some reason timezones are wonky and fail a bunch of other tests, this test will fail to try to help pinpoint that as the issue.

PR #1836, opened 2018-06-20
Remove unsafe use of _.template in older lodash

kevinrobinson:
# Who is this PR for?
developers, security

CSP violations in using `_.template`, which uses `eval` in the current older version of lodash.  This in code working with Highcharts, where we have to pass it HTML strings.

# What does this PR do?
Removes the use of `_.template` and remove special-casing for clicking on the tooltip date to navigate on the discipline chart (which didn't seem to work anyway).

screen shot 2018-06-20 at 7 29 13 am


kevinrobinson:
selfie

PR #1837, opened 2018-06-20
Update CSP config to be on by default, with env switch to report-only

kevinrobinson:
⚠️ This will require setting ENV variables before deploying

# Who is this PR for?
developers, security, part of https://github.com/studentinsights/studentinsights/issues/1704

CSP is not yet enforced by default.

# What does this PR do?
Changes CSP to be enforced by default, with env flag to switch to report-only mode.


kevinrobinson:
selfie

kevinrobinson:
This isn't yet enabled by default - each env has the override flag set so they're still report-only.

PR #1841, opened 2018-06-21
Refactor JS stubs out from Rails

kevinrobinson:
# Who is this PR for?
developers, part of https://github.com/studentinsights/studentinsights/issues/1704.

We have to be careful because this JS isn't compiled, and tightening CSP rules would require hashing these.

# What does this PR do?
Instead, let's just factor them out into plain JS.


kevinrobinson:
selfie

PR #1842, opened 2018-06-21
Move env-for-js parsing stub from inline into file for CSP 

kevinrobinson:
# Who is this PR for?
developers, part of https://github.com/studentinsights/studentinsights/issues/1704.

This JS stub is inline.

# What does this PR do?
Moves it to within two existing bundles.


kevinrobinson:
verified on the demo site

PR #1843, opened 2018-06-21
Disable extra CSP reporting for styles

kevinrobinson:
# Who is this PR for?
developers, security, part of https://github.com/studentinsights/studentinsights/issues/1704.

I looked at hashing each inline style in https://github.com/studentinsights/studentinsights/tree/patch/try-hashing-styles, but this was more involved than I expected since there are a few libraries that do this, and setting these hashes manually would be tough to maintain without tooling support.

# What does this PR do?
Remove additional CSP reporting for styles.


kevinrobinson:
selfie

PR #1844, opened 2018-06-21
Remove CDN for Mixpanel and Rollbar code

kevinrobinson:
# Who is this PR for?
security

Rollbar uses cloudflare as a CDN, and this appears to set all sorts of cookies 😠.  We don't want any of this tracking whatsoever.

# What does this PR do?
Uses yarn to package Rollbar and Mixpanel libraries.

### before
screen shot 2018-06-21 at 5 18 14 pm

### after
screen shot 2018-06-21 at 5 35 45 pm


kevinrobinson:
selfie

kevinrobinson:
verified this is 👍

PR #1845, opened 2018-06-21
Fix bug with secure_headers setup

kevinrobinson:
# Who is this PR for?
part of https://github.com/studentinsights/studentinsights/issues/1704

`secure_headers` needs a full CSP policy for report only; a fragment will raise and prevent the app from booting.

# What does this PR do?
merges report-only additions in with enforced policy


kevinrobinson:
selfie, will enable incrementally with env switch

PR #1846, opened 2018-06-21
Increase strictness for CSP for scripts

kevinrobinson:
# Who is this PR for?
security, part of https://github.com/studentinsights/studentinsights/issues/1704.

Raises the strictness of scripts.

# What does this PR do?
Also removes a report-only directive that is harmless but noisy in Chrome devtools.


kevinrobinson:
selfie

PR #1848, opened 2018-06-22
Upgrade to Ruby 2.5.1

kevinrobinson:
# Who is this PR for?
developers, security

Bug and security fixes (https://www.ruby-lang.org/en/news/2018/03/28/ruby-2-5-1-released/)

# What does this PR do?
Updates Ruby.  Devise 4.3.0 has syntax errors in Ruby 2.5.0 which have been patched in 4.4.0 (https://github.com/plataformatec/devise/issues/4739#issuecomment-354220177), so this updates Devise as well.  Devise includes some changes related to enabling autocomplete on sign in form fields that we might not want ([changelog](https://github.com/plataformatec/devise/blob/master/CHANGELOG.md))., but these don't apply as we're using our own view template to render this.


kevinrobinson:
Steps to update with Homebrew on Macs locally:
```
$ brew update && brew upgrade ruby-build
$ rbenv install 2.5.1
$ gem install bundler
$ bundle install
```

PR #1849, opened 2018-06-22
Add hardcoded paths for new Somerville SFTP box

alexsoble:
# Who is this PR for?

Developers/security.

Cutting over to a new box.


alexsoble:
🚢

PR #1851, opened 2018-06-25
Add extra check: wrap "Is Service Working?" query in authorized block

alexsoble:
# Who is this PR for? / What problem does this PR fix?

Adds extra security layer and removes "unscoped" query; see https://github.com/studentinsights/studentinsights/pull/1791#issuecomment-400004769.

### DIFF
--- a/app/controllers/is_service_working_controller.rb
+++ b/app/controllers/is_service_working_controller.rb
@@ -9,11 +9,16 @@ class IsServiceWorkingController < ApplicationController
 
   def is_service_working_json
     service_type_id = params[:service_type_id]
-    service_type = ServiceType.find(service_type_id)
 
-    student_ids = service_type.services.where('date_started > ?', Time.current - 1.year).map(&:student_id)
+    students = authorized do
+      Student.active.select do |student|
+        student.services.where(service_type_id: service_type_id)
+                        .where('date_started > ?', Time.current - 1.year)
+                        .present?
+      end
+    end
 
-    chart_data = Student.active.where(id: student_ids).map do |student|
+    chart_data = students.map do |student|
       {
         student: student,
         school: student.school.local_id,
--- a/spec/controllers/is_service_working_controller_spec.rb
+++ b/spec/controllers/is_service_working_controller_spec.rb
@@ -47,7 +47,13 @@ describe IsServiceWorkingController, :type => :controller do
         sign_in(pals.uri)
         make_request(502)
         expect(response).to be_successful
-        expect(JSON.parse(response.body)).to eq({'chart_data' => []})
+
+        json_response = JSON.parse(response.body)
+        expect(json_response.keys).to eq(['chart_data'])
+
+        chart_data = json_response['chart_data']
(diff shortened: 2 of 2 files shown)


alexsoble:
Only educators with districtwide access are authorized for this endpoint. That means we can't test if the `authorized` block is winnowing down the array of student JSON being sent down the endpoint; the only users who can get a response are authorized to see all students. 

To make the specs for this endpoint stronger and more useful, I added tests to verify other logic in the controller: are the right students being sent down as JSON? 

We could potentially remove the districtwide educator auth check for this controller and make the page accessible to any educator. We'd rely on the `authorized` block to send down the right students. I don't have strong feelings about whether that's a good move to make right now or not, since the goal of this feature at this stage is to allow Uri to dig into the data and spark conversations with other districtwide admin educators.

kevinrobinson:
@alexsoble Sounds good!  👍 Yeah I don't think we should remove the districtwide auth check - my suggestion in https://github.com/studentinsights/studentinsights/pull/1791#issuecomment-400004769 is just that we remove unscoped queries.  Having another layer guarding access like that is still great!

alexsoble:
@kevinrobinson I've been trying to write tests for this but finding it difficult without removing the auth check. I tinkered with moving the `authorized` query to a private method and using `send` but ran into issues with how the Authorizer works. Do you think this needs to be explicitly tested or looks good to ship as-is?

kevinrobinson:
@alexsoble It's hard to say more without knowing what the actual problem has been - if it's an error from `filter_relation` related to missing fields on the `Student` model, you can work around by adding `.to_a` at the end of the expression within the `authorized` block.  That was a premature optimization that we should remove from the authorizer code.

If that's not it, you could try using `TestPals` students here (and then in the test setup code, adding attendance contracts like you've done here).  In general, I think `FactoryBot.create(:student)` is a testing anti-pattern, since it creates a student that doesn't have homerooms, sections, etc. set in a way that all authorization checks require.  So using that essentially makes it hard to write test code that tests realistic code paths.

Alternately, if this seems painful, splitting this into methods like you suggest might be helpful, especially if you factor out `authorized_students` as a method that does `authorized { Student.active }` and then the `where` query is applied after.

kevinrobinson:
also, I think testing "Uri can see students" and "no one else can" is definitely good enough - I was suggesting adding `authorized` since I thought it would add another layer of guards and we could add it in without any other costs.

kevinrobinson:
@alexsoble https://github.com/studentinsights/studentinsights/pull/1853 should help with the JS, sorry about that.  The root problem is that I added those snapshot tests that have really wide coverage, and that code contained a few places that used wall clock instead of the clock that was passed down.  That should be fixed now, but I also it's a good example of adding snapshot tests over large pieces of the codebase without really shaking it down first, and I probably shouldn't have done that.

Restarting the build for ya now.

alexsoble:
OK — this adds and additional auth guard and more specs to verify that the right data is being sent down. Going to 🚢.

PR #1853, opened 2018-06-26
Fix some JS tests that weren't passing along the time, remove creeping React.PropTypes

kevinrobinson:
# Who is this PR for?
developers

build failures on unrelated features like https://github.com/studentinsights/studentinsights/pull/1851 and https://github.com/studentinsights/studentinsights/pull/1811#issuecomment-400331739

# What does this PR do?
There's some related cleanup, but the substantive change is in `ServicesList` - this PR updates it to not use `moment` as the wall clock but to grab a `nowFn` from context.  Because this is used in pages that are outside of `App` and still in `legacyRouteHandler`, there's a plumbing step in `PageContainer` so that the component can use the same API before we migrate the routing to it.

This also updates one of the test fixtures for the student profile, since it had dates that didn't make sense with the fixed time for JS tests.

This also removes use of `React.PropTypes`, since I copy-pasted it accidentally when first making this change.


app/assets/javascripts/student_profile/ServicesList.js
```diff
@@ -146,26 +138,27 @@ export default class ServicesList extends React.Component {
...
       );
-    } else {
-      const startedToday = moment().utc().subtract(1, 'day') < momentStarted;
+    }
 
-      // Don't show how long service has been going if it was added today
-      if (startedToday) return null;
+    // Don't show how long service has been going if it was added today
+    const {nowFn} = this.context;
+    const nowMoment = nowFn();
```
kevinrobinson: This is the substantive change - changing from `moment.utc()` to using the time provided by context.

kevinrobinson:
selfie

PR #1855, opened 2018-06-28
Fix bug with overmatching on slash quote conversions

kevinrobinson:
# Who is this PR for?
developers, educators

Addresses import errors in https://github.com/studentinsights/studentinsights/issues/1777, step towards resolving problem with attendance import job aborting and not completing.  This also blocked work on https://github.com/studentinsights/studentinsights/issues/1657.

# What does this PR do?
Fixes a bug introduced in https://github.com/studentinsights/studentinsights/pull/1778 which was trying to address this problem originally.  The regex cleanup would overmatch on particular escape sequences and cause the cleaner to produce unparseable strings.

It adds test cases and also has each test case verifying that the output of `ParseableCsvString` is in fact parseable.

# Checklists


kevinrobinson:
selfie

PR #1857, opened 2018-06-29
Add Rollbar notification for LDAP connection errors

kevinrobinson:
# Who is this PR for?
developers

If there are connection issues (eg, during firewall upgrades), we log this but don't notify.

# What does this PR do?
Notifies on connection errors via Rollbar.

# Checklists


kevinrobinson:
selfie

PR #1858, opened 2018-06-29
Simplify CSV cleaning regex, add more tests for different edge cases

kevinrobinson:
# Who is this PR for?
NB educators

https://github.com/studentinsights/studentinsights/pull/1855 and https://github.com/studentinsights/studentinsights/pull/1778 have been trying to address https://github.com/studentinsights/studentinsights/issues/1777, but there are interactions between the different kinds of cleaning.  Fixing one issue for Somerville created a separate one for New Bedford.

# What does this PR do?
Improves the regex for translating `\"` to `""`, to satisfy what the strict Ruby CSV parser expects.  In particular, this prevents overmatching on the `\\"` sequence, and makes sure it matches the `\"\"` sequence, both of which were problematic.

# Checklists

### DIFF
--- a/app/importers/data_transformers/parseable_csv_string.rb
+++ b/app/importers/data_transformers/parseable_csv_string.rb
@@ -71,9 +71,9 @@ class ParseableCsvString
     @log.puts '#convert_slash_quote_to_double_quote...'
 
     quotes_converted_count = 0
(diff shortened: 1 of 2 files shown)


kevinrobinson:
Verified this on production dynos as well.

PR #1859, opened 2018-06-29
Really upgrade to Rails 5.2

kevinrobinson:
# Who is this PR for?
developers, initially from looking at CSP helpers for security

Finishing off https://github.com/studentinsights/studentinsights/pull/1832, which ran the Rails upgrade task that it turns out doesn't actually update the gems :)

# What does this PR do?
Update Rails to 5.2, with explicit bumps to administrate ([changelog](https://github.com/thoughtbot/administrate/blob/master/CHANGELOG.md)) and delayed_job_activerecord too ([commits](https://github.com/collectiveidea/delayed_job_active_record/commits/master)).


kevinrobinson:
selfie, verified on demo site.

PR #1860, opened 2018-06-29
Fix bug with resolving drift when trying to export class lists

kevinrobinson:
# Who is this PR for?
K5 principals

Fixes https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/169/; when students are withdrawn or move schools, there's drift between the set of students in the class lists and the current set of students.  The class list UI code resolves this in a few places, but not in the export list.  So if a student is withdrawn, this page raises an error.

# What does this PR do?
Calls the resolve function for the `ExportList` component too.


kevinrobinson:
selfie

PR #1861, opened 2018-06-29
Migrate assessment and STAR importers to StreamingCsvTransformer, optimize Attendance importer

kevinrobinson:
# Who is this PR for?
NB educators, developers

Memory errors on the NB import, slow import cycles.

# What does this PR do?
It swaps `StreamingCSVTransformer` into `X2AttendanceImporter`, `StarReadingImporter`, and `StarMathImporter`.

For the STAR importers, this PR makes a few related changes:
- Removes the StarCsvTransformer, StarReadingCsvTransformer and StarMathCsvTransformer classes.  These provided two functions: telling the CSV parser not to symbolize the column keys, and then manually converting them to symbols.  This layer of indirection isn't doing anything functionally, so we can remove these classes rather than maintaining and swapping them to streaming.
- Modifies `StreamingCSVTransformer` to allow setting parse options for the CSV.  The STAR importers then use this option to prevent any transformations of the column keys.
- Updates the importer code to look for the column keys that are actually in the files (eg, 'PercentileRank' rather than :percentile_rank).
- Updates spec setup code.

It also updates the `AssessmentReport` to remove some slow queries that determine the count of assessments for active students.  I don't think we use this, and it slows down iteration now, so removing rather than optimizing further now.

Finally, it adds an option to the import task to `skip_index_updates`, since these index updates can be expensive when trying to iterate on a single importer.

### Performance
I didn't benchmark performance on this change, but based on https://github.com/studentinsights/studentinsights/pull/1427 it's reasonable to expect this will help in a similar way.

#### STAR
Looking at past server logs, it took about ~20 minutes to run the two STAR importers on the last run.  Testing this PR in production, it takes about ~60 seconds to run both STAR importers and the updated `AssessmentReport`.

#### X2Assessment
Based on past server logs, this took 20 minutes before the task OOMed.  In testing now, I couldn't get it to even finish the CSV parsing process without OOMing, when running only this importer alone on a fresh one-off dyno.  Testing early iterations of this PR in production now, it ran for ~23 minutes without an OOM before I interrupted the job.  With changes to only import the last 365 days, and other changes to reduce queries, it finishes in ~6 minutes for NB and ~2 minutes for Somerville.


app/importers/data_transformers/json_transformer.rb
```diff
@@ -1,16 +0,0 @@
-class JsonTransformer
```
kevinrobinson: unrelated change, this just isn't used anywhere

app/reports/assessments_report.rb
```diff
@@ -34,9 +38,7 @@ def subject_column(assessment)
   end
 
   def active_students_count(assessment)
-    Student.active.inject(0) do |sum, student|
-      sum + student.student_assessments.where(assessment: assessment).count
-    end
+    assessment.student_assessments.count
```
kevinrobinson: This was doing a lot of extra work, so this cuts the time down to a few seconds.

kevinrobinson:
Working on some other things in https://github.com/studentinsights/studentinsights/pull/1861/commits/c714e6b06f6339fbab580a4189a2f81535987236 to speed up the `assessment_importer` here too, unrelated to memory usage, which is stable now.

kevinrobinson:
There are several more commits here, with work related to the attendance importer.  These aim to optimize the speed of the task, improve logging, and fix some validation and data quality bugs.  This also surfaced a bunch of other issues that aren't addressed here, but here's what these commits do:

1. Improve handling of assessment types.  The importer class had a whitelist, branching and mutation of the record, this is now a `switch` and there's a validation on the model that works across instances.  The importer whitelist didn't include `WIDA` assessments previously, so these records weren't being imported, including a set from early June.  Fixing this imported 903 new records.
2. `McasRow`, `DibelsRow` and `AccessRow` each were doing lookups for students and for the assessment model, before they did the real work to find a matching `student_assessment` record.  This PR refactors those to push these queries to be done in a batch once fro the importer rather than each row.  This reduces the query volume significantly and speeds up the job.  Incidentally, this PR removes the `NullRelation` so the single public method returns a `StudentAssessment` record or `nil`.
3. The row classes were written to create new assessment subjects as it found them, but the `StudentAssessment` model code is written to validate that subjects fall within a whitelist.  This meant the importer would have tried to do things that the model would prevent, so this PR removes that from the importer.  Any CSV rows with references to assessment records that aren't in the database are logged as part of `invalid_rows_count`.  Additionally, the importer task logs the count of the test names that it encounters with `encountered_test_names_count_map` so that we can review the logs and see what tests are in the export that we might want to add to the import process.
4. Adds a `skip_old_records` flag that the attendance importer respects, and causes it to only import records for the last calendar year (of wall clock time).
5. Updates the importer to only call `.save!` on records that change.  Because of the validation on `StudentAssessment` that check its associations, this leads to loading the associated student record and assessment record.  None of this is meaningful work, as we can tell beforehand that the CSV data is the same as the model data that's already loaded.  So this reduces the writes to the database to only meaningful ones.  It also updates logging to track updated/created counts separately.

kevinrobinson:
selfie.  @alexsoble if you want to check this out when you're back, the `#import` and `#import_row` methods of the importer are the main changes: https://github.com/studentinsights/studentinsights/pull/1861/files#diff-44dc7155975da0487bccaac27758812dR12

PR #1862, opened 2018-06-29
Tune how ImportJob handles failure

kevinrobinson:
# Who is this PR for?
NB educators, developers

The NB importer jobs haven't been super healthy, and have been retrying in a few different ways. 
 See https://github.com/studentinsights/studentinsights/issues/1626#issue-314269511, and also https://github.com/studentinsights/studentinsights/pull/1861 which addresses the memory usage issues.

Right now, there are two separate mechanisms for replaying that don't know about each other - the `attempt` option passed to the `ImportJob`, and the `attempts` value owned by `Delayed::Job`.  And we're seeing jobs that are repeatedly retried at increasing frequency for NB, suggesting these two mechanisms may be interacting to amplify the total number of retries.

# What does this PR do?
Changes how ImportJob handles failure.

1. Moves `Delayed::Job` configuration from `ImportJob` into `delayed_job_config` so it's altogether.  This also addresses https://github.com/collectiveidea/delayed_job_active_record/issues/159.
2. Tunes down `max_run_time` to two hours.  This param tunes how long a job can run, but also tunes how long it is until `Delayed::Job` considers the worker to have died and allows reserving the job again.  So having it so high makes it so that retries get delayed 24 hours now, which isn't what we want.  I looked back at the logs to see how long jobs take, and they are consistently well under two hours.
3. Removes special behavior on `SignalException` - catch it and log, but re-raise.  The intention here is that re-raising will "causing the running job to abort and be unlocked, which makes the job available to other workers." (https://github.com/collectiveidea/delayed_job#gory-details).  I confirmed this works as expected locally, and sending worker tasks SIGTERM leads to the task being retried right away (from the teardown process setting `attempts` and `last_error` and `run_at` on the record).

There are two other related threads to this work.  One, with https://github.com/studentinsights/studentinsights/pull/1861#issuecomment-401558466, we haven't had any issues over the last week, and all jobs look healthy.  So this isn't as much of a concern anymore anyway.  Two, right now we're paying for 24 hours of worker dynos that are just polling the database, and we may be able to switch to using one-off dynos for all tasks to reduce those costs.

### DIFF
--- a/app/importers/file_importers/csv_downloader.rb
+++ b/app/importers/file_importers/csv_downloader.rb
@@ -10,10 +10,13 @@ class CsvDownloader
   end
 
   def get_data
+    log('Downloading...')
     file_contents = download_file
 
+    log('Transforming...')
     data = @transformer.transform(file_contents)
 
+    log('CleanupReport...')
     CleanupReport.new(
       @log, @remote_file_name, @transformer.pre_cleanup_csv_size, data.size
     ).print
@@ -22,7 +25,7 @@ class CsvDownloader
   end
 
   def download_file
-    log("\nDownloading #{@remote_file_name}...")
+    log("Downloading remote_filename: #{@remote_file_name}")
 
     downloaded_file = @client.download_file(@remote_file_name)
     File.read(downloaded_file)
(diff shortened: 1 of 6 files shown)


kevinrobinson:
selfie cc @alexsoble when you're back

kevinrobinson:
I also manually verified jobs work as expected when run as one-off scheduled dynos.  That's the current setup, we can discuss separately whether the cost-saving there is worth the tradeoff of retrying that comes with `Delayed::Job`.

PR #1863, opened 2018-06-30
Migrate ErrorMailer to use Rollbar instead

kevinrobinson:
# Who is this PR for?
developers

We have two channels for monitoring server-side errors now, no need for that.  Not sure why we added `ErrorMailer`, I think it was because we hadn't used or setup Rollbar on the server yet back then.

# What does this PR do?
Migrates `ErrorMailer` to use Rollbar instead, and removes `ErrorMailer`.


kevinrobinson:
selfie cc @alexsoble when you're back

PR #1864, opened 2018-06-30
Add more logging to ImportTask index updates

kevinrobinson:
# Who is this PR for?
developers

These index updates at the end are slow and do a lot of work, let's see which piece is slowest.

# What does this PR do?
Adds logging around each step.


kevinrobinson:
selfie

PR #1865, opened 2018-06-30
Add logging to attendance importer

kevinrobinson:
# Who is this PR for?
developers trying to confirm that log flushing isn't working right


kevinrobinson:
selfie

PR #1866, opened 2018-06-30
Optimize StudentsImporter to use StreamingCsvTransformer, and to only save changed records

kevinrobinson:
# Who is this PR for?
developers

`StudentsImporters` takes ~8 minutes and is one of the slowest importers for Somerville.

# What does this PR do?
Uses `StreamingCsvTransformer` and only calls `.save!` if anything has changed, taking approach from https://github.com/studentinsights/studentinsights/pull/1861#issuecomment-401558466.

### DIFF
--- a/app/importers/file_importers/students_importer.rb
+++ b/app/importers/file_importers/students_importer.rb
@@ -12,7 +12,7 @@ class StudentsImporter
       log: @log, remote_file_name: remote_file_name, client: client, transformer: data_transformer
     ).get_data
 
-    @data.each.each_with_index do |row, index|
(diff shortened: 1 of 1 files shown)


kevinrobinson:
selfie

kevinrobinson:
Actually we can't do this; the risk level computation is based on `limited_english_proficiency` and if this changes during this import, we'd miss it.

For now, going to leave this merged knowing that it impacts the risk level computation - in the case of the normal nightly import job, the risk level will be computed afterward anyway.

And, it's possible we'll remove the current risk level indicators altogether shortly, so optimistically leaving this merged :)

PR #1867, opened 2018-07-09
Tune down production log level

kevinrobinson:
# Who is this PR for?
developers

Upgrading Rails in https://github.com/studentinsights/studentinsights/pull/1832 changed the default log level to DEBUG, so all queries are logged now.  This is somewhat helpful for debugging performance issues (eg, in import jobs), but also adds a ton of log noise.  It's also hitting some logentries limits:
screen shot 2018-07-09 at 10 47 31 am

We can always enable this in more fine-grained way for profiling particular code if we want to.

# What does this PR do?
Sets log level back to :info.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -49,7 +49,7 @@ Rails.application.configure do
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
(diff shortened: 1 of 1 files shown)


kevinrobinson:
selfie

PR #1868, opened 2018-07-09
Try forcing STDOUT.sync true for logging timestamps

kevinrobinson:
# Who is this PR for?
developers

Import task timestamps still aren't always flushed synchronously, making it hard to understand how long parts of the task are taking by looking at logs.

# What does this PR do?
Based on https://github.com/heroku/rails_stdout_logging#migrating-to-rails-5 and related PRs, it seems there is a config value controlling this, but that Heroku requires one other change in additional to the default Rails 5 changes.  The env variable seems magical too, as it doesn't show up in `heroku config` but does if you run a Rails console and check it.  I added those explicitly now.


kevinrobinson:
selfie

kevinrobinson:
this looks good now 👍

PR #1869, opened 2018-07-09
Factor out StudentIdMap, optimize AttendanceImporter student lookups, use skip_old_records

kevinrobinson:
# Who is this PR for?
developers

AttendanceImporter takes 11 minutes, reducing this time a bit.

# What does this PR do?
Optimizes `AttendanceImporter` time to complete by reducing queries for students by using approach from `AssessmentImporter`, now factored out into `StudentIdMap`.  It also checks for which records actually change and only save those ones and log about updated/created/unchanged counts.  Migrates `only_recent_attendance` to use `skip_old_records` so both attendance and assessment work similarly.


app/importers/data_transformers/json_object_cleaner.rb
```diff
@@ -1,81 +0,0 @@
-class JsonObjectCleaner < Struct.new :json_object
```
kevinrobinson: unrelated; this isn't used anywhere.

kevinrobinson:
Trying this out in Somerville, it cuts the time from 11 min to 6 min and adds the more detailed logging.

kevinrobinson:
selfie

PR #1870, opened 2018-07-09
Migrate CsvTransformer to StreamingCsvTransformer

kevinrobinson:
# Who is this PR for?
developers

# What problem does this PR fix?
Deprecated code.

# What does this PR do?
Removes deprecated `CsvTransformer` altogether.

PR #1872, opened 2018-07-09
Remove unused cleanup_report

kevinrobinson:
# Who is this PR for?
developers

With https://github.com/studentinsights/studentinsights/pull/1870, the data this logs is useless for `StreamingCsvTransformer` and so this class doesn't do useful work anymore.  Also, we have no visibility into rows that are dropped because of `CsvRowCleaner` which is what the class was trying to get at originally.

# What does this PR do?
Removes the class, simplifies the API for `StreamingCsvTransformer` and updates the two longest importer tasks to log stats on discarded rows.


kevinrobinson:
selfie

PR #1876, opened 2018-07-10
Remove log message about expected files being missing in IEP import

kevinrobinson:
# Who is this PR for?
developers

remove log noise:
```
File #<File:0x0000000005bc8d38> has zero size. Did you mean to pass the create flag?
```
we didn't mean to pass the create flag, but we don't need to be making this call at all since we know there's no file there (as expected and noted in other log messages).


kevinrobinson:
selfie

PR #1878, opened 2018-07-10
Update IEP PDF importer with paths for new instance

kevinrobinson:
# Who is this PR for?
Somerville educators

I noticed this in verifying that https://github.com/studentinsights/studentinsights/pull/1876 was correct.  It turns out that when we cut over to a new instance, we didn't update this importer and it has been broken.  This wasn't obvious, since it's normal that IEPs aren't updated every day, and would be even more normal if this happened during the summertime.  But these have been updated (at least in this past week), and the import task was not picking them up since it wasn't looking at the correct path for the new instance.

# What does this PR do?
Updates the paths and moves them into `PerDistrict` and yaml config, and adds a bit more logging.

before, not finding any files:
```
I, [2018-07-10T19:25:04.248109 #13]  INFO -- : open /home/jbreslin/data/student-documents-6.zip: no such file (2)
I, [2018-07-10T19:25:04.248189 #13]  INFO -- : No file found but no worries, just means no educators added IEPs into the EasyIEP system that particular day
...
```

after, finding files:
```
I, [2018-07-10T18:57:05.220782 #13]  INFO -- : parsing unzipped pdfs from <path>...!
```

before, not info to debug why student not found:
```
student not in db!
```

after, with info about student not found and total count of PDFs dropped:
```
student local_id: 123456 not found, dropping the IEP PDF file...
...
IepPdfImportJob: stored 8 IEP PDFs and dropped 1 PDF files...
```

Verified this manually.

### DIFF
--- a/app/config_objects/per_district.rb
+++ b/app/config_objects/per_district.rb
@@ -116,6 +116,14 @@ class PerDistrict
     @district_key == SOMERVILLE
   end
 
+  def filenames_for_iep_pdf_zips
+    if @district_key == SOMERVILLE
+      LoadDistrictConfig.new.remote_filenames.fetch('FILENAMES_FOR_IEP_PDF_ZIPS', [])
+    else
+      []
+    end
+  end
+
   private
   def raise_not_handled!
     raise Exceptions::DistrictKeyNotHandledError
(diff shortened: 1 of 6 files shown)


kevinrobinson:
selfie

PR #1881, opened 2018-07-10
Add new instance doc

kevinrobinson:
# Who is this PR for?
developers, consolidating work from https://github.com/studentinsights/studentinsights/issues/1371 for NB ahead of other districts.

Capturing things we remember so we can re-use them, automate them, etc.

# What does this PR do?
Moves the "create a new Insights instance docs" from README into `new_instance.md` and adds the other items noted in https://github.com/studentinsights/studentinsights/issues/1371 with it.


kevinrobinson:
@alexsoble Can you review this?  This is aiming close out "New Bedford is setup" from https://github.com/studentinsights/studentinsights/issues/1371 and to get a good-enough draft for "setting up a new instance" since we'll probably actually pull doing this work shortly.

docs/technical/new_instance.md
```diff
@@ -0,0 +1,188 @@
+# Creating a new Insights instance
+The overall idea is strong isolation across districts, which means some initial setup overhead.  Here's how you do it!  More detailed notes for various steps are below.
+
+### 1: Stand up an instance
+Goal: Visit the home page signed-out
```
alexsoble: I like the Goals for each section and the checklist structure a lot! 

The checklist structure makes me wonder if it would be nice to make these markdown checkboxes — `[ ]` — so we could copy these lists for each district and and tick off each step as we go. 

The Goal for each one makes me think that each section could be accompanied by a more detailed guide down the line if that's needed.

docs/technical/new_instance.md
```diff
@@ -0,0 +1,188 @@
...
+- Add domain
+- SSL certificate
+- New Heroku app with Node/Ruby buildpacks, grant access to team
+- Review and set environment variables (minimally)
+- Review and set PerDistrict values (minimally)
```
alexsoble: Might be good to mention here that PerDistrict values are set in code, as opposed to ENV. Big difference from all the other steps in this section which are set through various web UIs.

docs/technical/new_instance.md
```diff
@@ -0,0 +1,188 @@
+# Creating a new Insights instance
+The overall idea is strong isolation across districts, which means some initial setup overhead.  Here's how you do it!  More detailed notes for various steps are below.
+
+### 1: Stand up an instance
+Goal: Visit the home page signed-out
+- Add domain
+- SSL certificate
```
alexsoble: Heroku auto-supplies SSL, so I think we can cut this one?

docs/technical/new_instance.md
```diff
@@ -0,0 +1,188 @@
...
+- Set up Mixpanel
+- Set up CSP and CSP logger
+
+### 2: Sign in
+Goal: Sign in as developer, see empty home page, roster page, districtwide pages
+- Create developer user
+
+### 3: Import first data
+Goal: Sign in as districtwide educator, use the home page, roster page, student profile page, take notes for students in support meetings
+- Create an instance for receiving data (maybe with static IPs, grant access to team)
```
alexsoble: You mean "create an SFPT box instance" here, right?

alexsoble: This one line is a whole process onto itself: setting up new box, communicating with district parter to connect with SIS, setting up security procedures and updates for the box.

docs/technical/new_instance.md
```diff
@@ -0,0 +1,188 @@
...
+
+### 5: Final review
+- Review ENV variables
+- Review `PerDistrict`
+- Review scheduled jobs
+- Review Puma tuning (eg, WEB_CONCURRENCY)
+- Load test or scale accordingly
+
+
+# Details
```
alexsoble: When I read through I don't see much logical flow between the checklists above and the "Details" section below. What do you think about mapping out from each checklist (or each step, if it's a major step) to a relevant "Details" doc or section?

alexsoble:
@kevinrobinson Reviewed!

docs/technical/new_instance.md
kevinrobinson: Yeah I think of this more as a "remember what we did for New Bedford?" aid for you and I than something that we could copy-paste and follow super literally.  We don't do this process often enough that we really have to optimize this, so this is more about "capture the list of things we just did" than "make a how-to doc that a new person could just follow."

And next time we do this, it seems better to incrementally automate parts of this (like your "make a new Heroku app" script).

docs/technical/new_instance.md
kevinrobinson: ah cool, thanks!

docs/technical/new_instance.md
kevinrobinson: Ah yeah I think this file is "new (Insights) instance" and here it's "new (EC2) instance" thanks!

Yep, lots of these things have a bunch of discrete steps, and if there's more details that are helpful to capture we can add those at the bottom of this doc too.  If you have notes on this from when we did, you could push them here or add them in another PR, that'd be awesome.

docs/technical/new_instance.md
kevinrobinson: Yeah there's no connection, I'll write that explicitly :)  I think we need more "a list" than we do a user-friendly tutorial.  How about next time you and I sit down to actually do this work, we can just scan through the bottom of this doc and find what we need, and update this based on our experience from the next district?

PR #1883, opened 2018-07-12
Remove Risk Level bubbles from front-end

alexsoble:
# Who is this PR for?

Educators. 

Showing a colorful number could cause educators to have a simplified view of a student and hinder seeing the student in their best light. 

# What does this PR do?

Removes the RiskBubble component and story from the front end. To be followed by a PR removing risk level calculations, model, and table from the back end. 

screen shot 2018-07-12 at 3 17 53 pm


kevinrobinson:
@alexsoble I think this snapshot test is failing correctly - the snapshot is like "hey, isn't there supposed to be a risk bubble on the page" and now it's missing :)  So you should be able to use `u` to update it and then commit the updated snapshot but if you run into test headaches feel free to ping and I'm happy to help out.

alexsoble:
@kevinrobinson Snapshots updated! All looks good on the profile page locally. Selfie 🚢ing!

PR #1877, opened 2018-07-10
Fix layout bug with note and note attachment overflow

kevinrobinson:
# Who is this PR for?
educators

# What problem does this PR fix?
When there is long text that's unbroken (like a URL), the note text overflows.  This impacts the event note attachments as well.

# What does this PR do?
Fixes this layout bug.

# Screenshot (if adding a client-side feature)
### before, home
screen shot 2018-07-10 at 1 48 55 pm

### before, profile, IE
screen shot 2018-07-10 at 1 49 17 pm

### before, profile, Chrome
screen shot 2018-07-10 at 1 49 32 pm

### after, home
screen shot 2018-07-10 at 1 48 14 pm

### after, profile
screen shot 2018-07-10 at 1 47 55 pm

# Checklists
+ [x] Author checked latest in IE - Student Profile
+ [x] Author checked latest in IE - Home page

PR #1884, opened 2018-07-12
Update wrapping for long text like URLs in NoteText

kevinrobinson:
# Who is this PR for?
educators

In https://github.com/studentinsights/studentinsights/pull/1877 I tried to fix layout of long URLs in notes, but this led to notes breaking mid-word.

# What does this PR do?
Changes back to `break-word` and adds `overflow-x`: hidden for IE - the behavior here is different in Chrome (which forcibly breaks the long text) and IE (which does not).  So in IE this is less than ideal, it truncates the text without visual indication and allow horizontal scrolling in the area with the keyboard if it's editable.  This only impacts a small number of pages, so shipping the improvement here of "not breaking the whole page layout" seems okay.

### before, showing forcibly breaks within words
screen shot 2018-07-12 at 6 37 47 pm

### after, showing long links are forcibly broken but not wrapped in IE11
screen shot 2018-07-12 at 6 30 58 pm
screen shot 2018-07-12 at 6 29 50 pm

### DIFF
--- a/app/assets/javascripts/components/NoteText.js
+++ b/app/assets/javascripts/components/NoteText.js
@@ -2,13 +2,20 @@ import PropTypes from 'prop-types';
 import React from 'react';
 
 
+// The sizing for long text with no space (eg, URLs) is different
+// in IE11 vs. Chrome.  Chrome forces `break-all` essentially when the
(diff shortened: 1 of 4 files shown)


kevinrobinson:
selfie

PR #1885, opened 2018-07-12
Revise copy for URL on roster page

kevinrobinson:
# Who is this PR for?
educators, from feedback from LL

Make more explicit that particular filter combinations can be loaded by URL.

# What does this PR do?
Changes the copy.

### before
screen shot 2018-07-12 at 6 39 51 pm

### after
screen shot 2018-07-12 at 6 40 15 pm


kevinrobinson:
selfie

PR #1886, opened 2018-07-13
Fix sort order on ProfileBarChart date tooltips

kevinrobinson:
# Who is this PR for?
educators, came up working on https://github.com/studentinsights/studentinsights/issues/1657

On hover, dates for absences are in reverse order, recent to oldest.  This sort of makes sense in a "what happened most recently sense" but not in what you'd expect just reading a chart.

# What does this PR do?
Sorts oldest - recent.

### before
screen shot 2018-07-13 at 9 36 16 am

### after
screen shot 2018-07-13 at 9 36 09 am


kevinrobinson:
selfie

PR #1890, opened 2018-07-13
Add RecordSyncer for AttendanceImporter, to track changes to records and delete records removed upstream

kevinrobinson:
# Who is this PR for?
administrators, redirect, counselors

Part of https://github.com/studentinsights/studentinsights/issues/1657.

Absence counts are slightly inflated (1-2 records) for ~9% of students in Somerville.  `AttendanceImporter` doesn't handle if the CSV snapshot no longer has records that it previously did, so the hypothesis is that these records were deleted upstream.  I looked at a few of these manually, and found one that had a clear paper trail showing it had been entered before the event for a vacation, and that the last date was removed, and that stopped showing up in the CSV snapshot after that day.  That might not be representative, but fits the hypothesis.  Other spot checks also confirmed that they had last been imported the day after the absence, suggesting it was maybe removed the next day (and no longer updated in the import process).

# What does this PR do?
Adds a `RecordSyncer` class for the `AttendanceImporter`.  The idea is that the importer class has to do the work to 1) figure out what CSV rows are in scope and what to skip and then 2) figure out how to translate a CSV row into an Insights record (in memory).  It can then ask the `RecordSyncer` to sync that to Insights (update an existing record, create a new one, or do nothing if nothing has changed).  `RecordSyncer` keeps tracks of the records it sees and what it does.

Then when the importer is done all CSV rows, it can ask `RecordSync` to delete the Insights records that it would have expected to be updated, that were never encountered in the CSV.  So for a particular time range or a school, the import might expect that after processing all CSV rows, that all of the Insights Absence records within that time range for students in that school would have been matched up.  If any didn't, `RecordSync` can delete them (ie, because they have been deleted upstream).

This adds memory usage with tracking the records touched, but this shouldn't impact the job to where we see any memory usage warning.

It's possible that we could re-use the `RecordSync` class in other places but this PR isn't trying to design for that.  `EducatorSectionAssignmentsImporter` and `StudentSectionAssignmentsImporter` both use the same strategy already so they could get metrics by migrating.  We haven't yet looked at how much drift there is in other models but `DisciplineIncidents` would be another one where inflation like this would have the most direct product impact.

# Checklists


kevinrobinson:
@alexsoble are you up for reviewing this?  It's got a bunch of tests, but I didn't verify any of this manually yet.

alexsoble:
@kevinrobinson Reviewing the code and specs now! What are your thoughts on verifying manually? Since this introduces new deletion powers, maybe worth testing manually locally with a fixture CSV file?

app/importers/helpers/record_syncer.rb
```diff
@@ -0,0 +1,108 @@
...
+    @marked_ids = []
+  end
+
+  # Given a new or persisted record, with attributes updated in memory
+  # based on the CSV, update it in the Insights database.
+  # Also track that the Insights record still exists in the current CSV snapshot.
+  def validate_mark_and_sync!(insights_record)
+    # Passed nil, something failed upstream
+    if insights_record.nil?
+      @passed_nil_record_count = @passed_nil_record_count + 1
```
alexsoble: Small style thing, but maybe we want to use the `@passed_nil_record_count += 1` pattern here. That would reduce the surface for variable name mistakes.

alexsoble: (And in the blocks below.)

app/importers/file_importers/attendance_importer.rb
```diff
@@ -43,30 +40,39 @@ def import
...
+      .where(:schools => {:local_id => @school_local_ids})
+      .where('occurred_at >= ?', @old_threshold)
   end
 
   def school_filter
-    SchoolFilter.new(@school_scope)
+    SchoolFilter.new(@school_local_ids)
   end
 
   def is_old?(row)
```
alexsoble: I wrote a "note to self" here as I reviewed saying that if the event date falls exactly on the `@old_threshold`, the record is considered not old. That made it easy for me to look at the `records_within_scope` query and verify that records falling on the threshold are correctly handled as within scope.

app/importers/file_importers/attendance_importer.rb
```diff
@@ -43,30 +40,39 @@ def import
...
-    row[:event_date] < @time_now - 90.days
+    row[:event_date] < @old_threshold
   end
 
   def import_row(row)
```
alexsoble: I'm looking at the conditionals here:

```
if !school_filter.include?(row)
  @skipped_from_school_filter = @skipped_from_school_filter + 1
  return
end

if @skip_old_records && is_old?(row)
  @skipped_old_rows_count = @skipped_old_rows_count + 1
  return
end

...
```

It looks like these filters are narrowing down to the exact same set of records as `records_within_scope`:

```
record_class
  .joins(:student => :school)
  .where(:schools => {:local_id => @school_local_ids})
  .where('occurred_at >= ?', @old_threshold)
```

Could having this same logic in two places potentially introduce bugs? 

Since we have to query for `records_within_scope` anyways, what if we query for it at the beginning, use the results to check if each record is within scope in `import_row`, and then pass the same result as an argument to `syncer.delete_unmarked_records!` at the end?

kevinrobinson:
@alexsoble Yeah, that would be a good idea!  I've been running it with more logging in a one-off dyno limited to the last two days, and looking at each of the rows and records manually.  There was a bug in the `RecordSyncer` - the "marking" was happening for new records before they got an id, so each run would create a new record and then destroy it since it wasn't marked.  The latest commit fixes that, so it might be ready to try a larger window, and then expand that out.

app/importers/helpers/record_syncer.rb
kevinrobinson: cool, thanks!  for some reason I thought you couldn't do that in Ruby (I think that's `++`).

app/importers/file_importers/attendance_importer.rb
kevinrobinson: Yeah, good question!  These are filtering on different sides.  The "skip" things are filtering the source side (CSV rows), and the `records_within_scope` is filtering on the destination side (Insights records).

So we can't have them literally be the same, and another wrinkle is that the filter on the destination side (Insights records) is applied to two different sets of Insights models.  We could describe the filters as a data structure that is shared by those different filters.  That's kind of what's happening here anyway though by storing `@school_local_ids` and `@old_threshold` but if we factored it out, it would enforce that one of the filters doesn't forget about the other.  Looking at this, I think there might be a bug here where the query doesn't work right without `skip_old_records`, so I'll try this out 👍

alexsoble: Ah ok! Thanks for the explanation!

kevinrobinson:
Okay, I've run this manually and verified that it looks correct in a bunch of different ways across both sites.  I'm going to ship this and consider this done but there are two related issues that came up that are worth calling out:

1. The way I was checking that the CSV data and Insights data matched up wasn't accurate, because using `grep` with multiple filters when text fields have newlines in them won't work.  So the figure that ~9% of students had inaccurate counts isn't right (https://github.com/studentinsights/studentinsights/issues/1657#issuecomment-396062667), and without building other tools to look at that we can't get a better number.  I did manually verify this across several records on both sites, with only single filters, so I think we can consider this resolved when this PR ships.

2. New Bedford has a significant number of records that are skipped because it can't match the student `local_id`.  This appears to be happening because of differences in how the Somerville and New Bedford export processes export the `school_local_id` field for attendance.  Somerville does a join on the student, then the school.  The NB Java export code appears to grab the school at the time that the attendance event occurred (rather than the student's school at export time).  That means that the "school filter" for Insights importer code don't work the same way they do for Somerville, and would skip records for a student who is in a school we want to import now (ie, it would skip the records from when they were at their previous school).  This is split out into https://github.com/studentinsights/studentinsights/issues/1894, and we can deal with it separately.

kevinrobinson:
Merging this and will run one more manual verification once it's deployed.  Also, in verifying manually I mistakenly delete all of New Bedford's absence and tardy records (so historical `created_at` timestamps were blown away).

PR #1896, opened 2018-07-16
Optimize what records to delete in RecordSyncer

kevinrobinson:
# Who is this PR for?
developers

The operation to find what records in scope haven't been marked takes ~2 min for all Somerville records, this should be able to finish in a small number of seconds at most.  This won't matter in the regular case, just in the case of backfills.

# What does this PR do?
Optimizes this computation, also adds logging for longer-running delete processes (which shouldn't happen anymore after the initial runs for https://github.com/studentinsights/studentinsights/pull/1890).


kevinrobinson:
selfie

PR #1892, opened 2018-07-13
WIP – Remove student risk levels from back-end

alexsoble:
# Who is this PR for?

# What problem does this PR fix?

# What does this PR do?

# Screenshot (if adding a client-side feature)

# Checklists

## Javascript QA

*If the PR touches Javascript, follow this checklist. If the PR doesn't touch Javascript, you can delete this section.*

*Which features or pages does this PR touch? Make a list here so that you can double-check each affected page in a Internet Explorer before merging.*

*Make one list for the author and another for the reviewer.*

*For more info on how to download and use a Virtual Machine to test in IE, see the README.*

*Example:*

+ [ ] Author checked latest in IE - Student Profile
+ [ ] Author checked latest in IE - Student Report PDF
+ [ ] Author checked latest in IE - School Overview
+ [ ] Reviewer checked latest in IE - Student Profile
+ [ ] Reviewer checked latest in IE - Student Report PDF
+ [ ] Reviewer checked latest in IE - School Overview

## Automated Testing

*Does the PR touch an area of the codebase that needs more specs? More examples within the specs? More scenarios covered by specs?*

+ [ ] Author improved specs for code in need of better test coverage

*Does the PR introduce a new feature or new section of code?*

+ [ ] Author included specs for new code

PR #1897, opened 2018-07-16
Store user login attempts with authtrail 

alexsoble:
# Who is this PR for?

Security. 

We have decent visibility into successful educator authentication right now, but no visibility into failed authentication attempts (i.e. attack vectors). In addition to adding this visibility, this is a first step towards being able to lock out user accounts that attempt to log in too many times in a short period. 

# What does this PR do?

Adds https://github.com/ankane/authtrail into our Devise stack.

### DIFF
--- a/config/initializers/authtrail.rb
+++ b/config/initializers/authtrail.rb
@@ -1,0 +1,1 @@
+AuthTrail.geocode = false
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -344,6 +344,18 @@ ActiveRecord::Schema.define(version: 2018_07_16_183457) do
     t.index ["student_id"], name: "index_student_photos_on_student_id"
   end
 
+  create_table "student_risk_levels", id: :serial, force: :cascade do |t|
+    t.integer "student_id"
+    t.integer "level"
+    t.datetime "created_at"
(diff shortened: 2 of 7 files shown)


db/schema.rb
```diff
@@ -323,18 +344,6 @@
     t.index ["student_id"], name: "index_student_photos_on_student_id"
   end
 
-  create_table "student_risk_levels", id: :serial, force: :cascade do |t|
```
alexsoble: This schema cruft is from working on #1892 at the same time.

config/initializers/authtrail.rb
```diff
@@ -0,0 +1 @@
+AuthTrail.geocode = false
```
alexsoble: We can implement geocoding incrementally; not a must-have to start recording login activity. My assumption is that we can also geocode old login attempts with a backfill job since we'll have all the IPs.

alexsoble:
I think this is in good shape to 🚢! I added comments to the `LoginActivity` model describing a few key things devs should know when looking at the data. Turned off geocoding; we can consider adding that functionality later. Key value-add is to store successful/unsuccessful user login attempts.

Issue #1655, opened 2018-04-27
Absences: Change suggestions to students without any notes rather than SST only


kevinrobinson:
At some schools, folks have started changing the way they enter data in response to the insights box and absences dashboard, by entering notes as being part of an SST meeting even if they aren't.  They're doing this to make the "scoreboard" or "todolist" that Insights shows reflect the reality of the work they're doing to support students.

To improve this, we'll talk with folks in person and encourage them to accurately capture what's happening as a general rule.  And then place responsibility on us to make the "scoreboard" or "todolist" more accurate.  In this case, that means:

- [x] Update absences insight box on the home page to use "any note" rather than SST (https://github.com/studentinsights/studentinsights/pull/1674)
- [x] Update the absences dashboard to use "any note" rather than SST (https://github.com/studentinsights/studentinsights/pull/1898)

kevinrobinson:
Closed with https://github.com/studentinsights/studentinsights/pull/1898.

PR #1898, opened 2018-07-16
Update Absences, Tardies and Discipline dashboards to use "any note" rather than SST

kevinrobinson:
# Who is this PR for?
K12 SST staff, administrators, counselors, housemasters.  part of https://github.com/studentinsights/studentinsights/issues/1655.

These dashboards show data on the last SST note, but often there are other notes going on, and we want to stay at "discuss what's going on with this student" rather than the more narrow technical SST focus per se (and avoid folks adding notes as SST to impact this view).

# What does this PR do?
Changes server query to look for just latest note, and then updates UI to show it along with the type inline.

### before, absences (others are similar)
screen shot 2018-07-16 at 3 56 25 pm

### after, absences (others are similar)
screen shot 2018-07-16 at 3 39 44 pm

# Checklists

### DIFF
--- a/app/assets/javascripts/components/eventNoteType.js
+++ b/app/assets/javascripts/components/eventNoteType.js
@@ -10,6 +10,18 @@ export function eventNoteTypeText(eventNoteTypeId) {
   return textMap[eventNoteTypeId] || 'Other';
 }
 
+const miniTextMap = {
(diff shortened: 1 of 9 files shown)


kevinrobinson:
selfie

PR #1899, opened 2018-07-16
Reorder dashboard time filter buttons

kevinrobinson:
# Who is this PR for?
K12 SST, administrators, counselors, housemasters

Make the left-most value selected by default, default to shorter time range.

# What does this PR do?
Re-orders the buttons.


kevinrobinson:
selfie

kevinrobinson:
This doesn't work; this is just UI display state and so this leads to the button being selected but the tables showing other date ranges.

PR #1900, opened 2018-07-16
Revert "Reorder dashboard time filter buttons"

kevinrobinson:
Reverts studentinsights/studentinsights#1899


kevinrobinson:
selfie

PR #1901, opened 2018-07-17
Add hacky fix for risk level bug

alexsoble:

Bugfix; https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/178/.

# What does this PR do?

Explicitly remove risk_level from students query, even though risk level was removed from the students table in #1892. (Seems there's some kind of bug with the column_names ActiveRecord cache.)

# Note

```
2.5.1 :001 > ['a'] - ['c']
 => ["a"]
```

So we're safe removing an item (column name) that shouldn't be there long-term.


alexsoble:
selfie

PR #1903, opened 2018-07-17
Add alternative `all.sh` deploy script

alexsoble:
# Who is this PR for?

Devs.

Slightly different dev environments mean that our `all.sh` deploy script doesn't work consistently across computers.

# What does this PR do?

Adds an alt script that works on my machine. Kicking this particular problem down the road. 👋


alexsoble:
selfie

PR #1904, opened 2018-07-17
Revert "Store user login attempts with authtrail "

alexsoble:
Reverts studentinsights/studentinsights#1897. 

Deploying this brought down the site with the following error message:

```
app/jobs/auth_trail/geocode_job.rb:2:in `<module:AuthTrail>': uninitialized constant AuthTrail::ApplicationJob (NameError)
```


alexsoble:
Backing out the parts of the revert that touch the `/db` directory, since messing with those migrations could get us into an ugly place. Based on the error message above I'm very confident the database migrations are not the source of the bug.

alexsoble:
selfie

PR #1905, opened 2018-07-17
Add empty ApplicationJob as part of Rails 5 migration

alexsoble:

The Rails docs mention that to upgrade to Rails 5, we need to add an `ApplicationJob` class that inherits from `ActiveJob::Base`:

http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#activejob-now-inherits-from-applicationjob-by-default

This wasn't a big issue in our app. We only used Delayed::Job in one instance (and we're not using it there anymore). And we hadn't added any libraries that called `ApplicationJob`... until recently! 

#1897 caused the site to go down because authtrail expects `ApplicationJob` to be present. My understanding is that once this PR is merged, we could un-revert authtrail safely. 

# Question to consider!

Rails recommends that after you add `ApplicationJob`, "make sure that all your job classes inherit from it." 

Is this something we should go ahead and do?

Looking at the jobs in our app/jobs folder: two are for testing purposes, three are for import jobs that run nightly. None of them are "background jobs" in the sense that they contain large volumes of work coming in continually to be processed in the background by a background processor like Sidekiq or Delayed::Job. This makes me wonder if we should move them out of app/jobs and into another top-level folder (/tasks?). This way we aren't at odds with Rails expectations about what a "job" is and what classes it should inherit from.

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -1,0 +1,2 @@
+class ApplicationJob < ActiveJob::Base
+end


kevinrobinson:
@alexsoble nice writeup, and quick recovery from the problem earlier too! 👍 

This seems like it would fail locally too, why do you think it only comes up in production?  It looks like https://github.com/ankane/authtrail/blob/67ff2eec6b5ae6af394db0d04e40ee95af4637f2/app/jobs/auth_trail/geocode_job.rb#L2 is the immediate problem, so adding the job line you described seems like a good way to unblock.

re: jobs separately, I didn't know [ActiveJob](http://guides.rubyonrails.org/active_job_basics.html) was a thing, but I don't think any of those things in `/app/jobs` are used by that.  We could rename them which seems fine, that seems separate to getting authtrail working so also could just punt.

alexsoble:
@kevinrobinson I think this error didn't surface in development because we have `eager_load` set to `false` in development and `true` in production. Here's what `production.rb` has to say:

```
# Eager load code on boot. This eager loads most of Rails and
# your application in memory, allowing both threaded web servers
# and those relying on copy on write to perform better.
# Rake tasks automatically ignore this option for performance.
config.eager_load = true
```

kevinrobinson:
⚡️ 👍

re: `eager_load`, I think the main benefit for development is that you can change code, reload the page, and it runs the new code (versus having to restart the server on each code change).  We could probably catch this by adding some kind of thing that does this in Travis, but doesn't seem super worth it, considering it's really rare to make these kinds of changes, get these kinds of bugs, and when we do we get alerted and can fix it quickly like you awesomely did here. 👍

PR #1909, opened 2018-07-17
Un-revert: Store user login attempts with authtrail

alexsoble:
Reverts studentinsights/studentinsights#1904.

Re-attempting to deploy authtrail. #1905 means that the app shouldn't blow up on first boot. 

I'm going to rebase this on top of #1907 so that Travis can tell us whether it will boot successfully before we merge and deploy.


alexsoble:
No rebasing required: `eager_load` is set to `true` on this branch so I'll watch Travis.

alexsoble:
Tests pass! 🚢

alexsoble:
Manually verified by attempting to log in a few times and checking the results in the `LoginActivities` table. Looks good! 👍🌈✨

kevinrobinson:
@alexsoble FYI the masquerading PR (https://github.com/studentinsights/studentinsights/pull/1793) doesn't show up in the `LoginActivity` table, which I think is good and what we'd want.

PR #1910, opened 2018-07-17
Update FakeStudent to make Absence and Tardy events on dates

kevinrobinson:
# Who is this PR for?
developers, prospective districts

The demo site can't run the migration from https://github.com/studentinsights/studentinsights/pull/1908#issuecomment-405705919, because the generates create data in the wrong shape - it creates times for `Absence` and `Tardy` records when production systems use dates.  So the migration making this stricter fails.

# What does this PR do?
Updates the seed task to generate dates, so we can rebuild the demo site and migrate.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -6,16 +6,11 @@ start_time = Time.now
 
 def sample_numeric_grade_from_letter(grade_letter)
   case grade_letter
-    when "A"
-      grade_numeric = rand(90..100)
-    when "B"
-      grade_numeric = rand(80..89)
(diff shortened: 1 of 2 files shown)


db/seeds.rb
```diff
@@ -6,16 +6,11 @@
 
 def sample_numeric_grade_from_letter(grade_letter)
   case grade_letter
-    when "A"
```
kevinrobinson: unrelated linter fix

kevinrobinson:
selfie

PR #1911, opened 2018-07-17
Remove model validations on Absence and Tardy, since they regress Importer performance

kevinrobinson:
# Who is this PR for?
developers, follow on to https://github.com/studentinsights/studentinsights/pull/1908

https://github.com/studentinsights/studentinsights/pull/1908 migrated the database but also added two Rails validations.  These validations mean that each call to `valid?` or `save!` also leads to database queries on associated models to enforce the validations.  This slows down the Somerville `AttendanceImporter` significantly since it has to do this for every row it processes - after 74 minutes on a 1x box it's only about halfway done.  We don't really care about how long this task takes in general, but it really slows down iterating on the last steps for https://github.com/studentinsights/studentinsights/issues/1657.

# What does this PR do?
Removes the model validations.  These are enforced by the database already:
```
ActiveRecord::InvalidForeignKey (PG::ForeignKeyViolation: ERROR:  insert or update on table "absences" violates foreign key constraint "fk_rails_dc2c1be879")
DETAIL:  Key (student_id)=(99999999) is not present in table "students".
: INSERT INTO "absences" ("occurred_at", "created_at", "updated_at", "student_id") VALUES ($1, $2, $3, $4) RETURNING "id"
...
ActiveRecord::RecordNotUnique (PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_absences_on_student_id_and_occurred_at")
DETAIL:  Key (student_id, occurred_at)=(7068, 2018-07-16) already exists.
: INSERT INTO "absences" ("occurred_at", "created_at", "updated_at", "student_id") VALUES ($1, $2, $3, $4) RETURNING "id"
```

### DIFF
--- a/app/demo_data/demo_data_util.rb
+++ b/app/demo_data/demo_data_util.rb
@@ -1,6 +1,15 @@
 class DemoDataUtil
   FIVE_YEARS_OF_SECONDS = 157766400
 
+  def self.generate_unique(n, &block)
+    values = []
+    while values.size < n
+      values << block.call
+      values = values.uniq
+    end
+    values
+  end
+
   def self.random_time(options = {})
     time_now = options[:time_now] || DateTime.now
     seconds_back = options[:seconds_back] || FIVE_YEARS_OF_SECONDS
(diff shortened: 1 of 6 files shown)


kevinrobinson:
An additional commit updates `FakeStudent`, since it has been generating multiple events for the same student on the same day.

kevinrobinson:
selfie

PR #1912, opened 2018-07-18
Stop Rails from auto-generating helper files

alexsoble:
# Who is this PR for?

Developers.

Cut generators that add extra cruft around helper files when using `rails g controller`, we're not using helpers extensively in this app.

# What does this PR do?

Fix small mistake: `helpers` shouldn't be plural, which is why Rails was autogenerating lots of helper files despite our attempt to turn this feature off.


alexsoble:
selfie

PR #1913, opened 2018-07-18
Visualize login activity patterns

alexsoble:
# Who is this PR for?

District leaders, district IT, and developers.

No good insight right now into failed login attempts. Many failed login attempts could indicate a very frustrated user, or a possible attacker trying to get into the app).

# What does this PR do?

Adds a new component, `LoginActivityPage`, which visualizes the past 30 days of successful and unsuccessful login attempts. 

# Screenshot (Internet Explorer, test pals data)

login-attempts-chart-ie-screenshot

# Checklists

## Javascript QA

+ [x] Author checked latest in IE - Login Activity Page 

## Automated Testing

### DIFF
--- a/app/assets/javascripts/login_activity/LoginActivityPage.js
+++ b/app/assets/javascripts/login_activity/LoginActivityPage.js
@@ -132,20 +132,32 @@ export default class LoginActivityPage extends React.Component {
         </div>
         {pastThirtyDaysArray.map((day) => {
           return (loginData[day])
-            ? this.renderCell(loginData[day])
+            ? this.renderCell(email, loginData[day], day)
             : this.renderEmptyCell()
         })}
       </div>
     );
   }
 
-  renderCell(data) {
+  renderCell(email, data, day) {
     return (
-      <div style={{...style.cell, ...style.squareCell}}>
+      <div style={{...style.cell, ...style.squareCell}} className='tooltip'>
+        {this.renderTooltipText(email, data, day)}
         {this.renderSuccessLoginSegment(data.success || 0)}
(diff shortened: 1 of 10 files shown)


alexsoble:
# Status update

I'm thinking this PR is in good shape for some Friday afternoon shipping!  🚢🎉 

I'm going to ship this one dark, without a link in the product itself, so we can take a look at production data viz before making it a live link. 

Using this note to track areas that look solid and those we might attend to in future iterations once we're past next week's sprint. 

## What's solid 

+ Key security checks: controller has authorization checks and those authorization checks are double-checked by Ruby controller test code. 
+ Chart renders correctly in Internet Explorer, needed an `array.prototype.fill` polyfill to work.

## What might be awkward in production 

+ Sort order of email addresses: didn't address that yet. Logically we might want to put the addresses with the most activity up at the top. Could also look weird if a spam bot type email attempting to log in as up at the top.

## What's a little messy but not worth fixing right now 

+ The `LoginActivityPage` doesn't use the same awesome pattern of sending down `nowFn()` or `Time.now` that's followed in the rest of the app. This leaves us open to a class of bugs around different parts of the component not sharing the exact same current time. I didn't observe any of those bugs when developing the feature so I'm going to leave this as a future refactor opportunity. 
+ I'm not 100% happy with how we handle dates in the component. I ended up using `moment.utc()` to get date comparisons to work, but that may not be the best way to do this. No concrete suggestion for improving this yet, just noting it's an area to potentially look at later. 

cc @kevinrobinson — I don't think this PR needs line-by-line review, but here's a status update!

PR #1914, opened 2018-07-19
Update seed task to raise on errors and respect validations for duplicates

kevinrobinson:
# Who is this PR for?
developers

https://github.com/studentinsights/studentinsights/pull/1908 added stricter validations for absences and tardies, which the seed task can violate, depending on how the randomization comes out.  Separately, `FakeStudent` used `.save` in several places instead of `.save!` which meant that it silently ignored Rails validation errors (although `.save` does respect database-enforced constraint violations like in https://github.com/studentinsights/studentinsights/pull/1911).  Changing this to `.save!` revealed that the code was trying to create two separate sets of Access assessments and was violating the `StudentAssessment` Rails validations.

# What does this PR do?
Updates `FakeStudent` to use `save!` and refactors the call to be `FakeStudent.create!` reflecting this instead of `FakeStudent.new`.  Renames the methods for assessments to be about individual assessments types so it's clear what's in their scope.

### DIFF
--- a/app/demo_data/fake_student.rb
+++ b/app/demo_data/fake_student.rb
@@ -25,24 +25,34 @@ class FakeStudent
     'Mom called, did not leave a reason.'
   ]
 
+  def self.create!(school, homeroom)
+    FakeStudent.new(school, homeroom).create!
+  end
+
   def initialize(school, homeroom)
     @school = school
     @homeroom = homeroom
-    @student = Student.create!(data)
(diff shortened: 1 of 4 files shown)


kevinrobinson:
selfie, cc @alexsoble this is related to https://github.com/studentinsights/studentinsights/pull/1908, and so if you have pulled that, you might need to pull this in to be able to regenerate local development data.  let me know if you run into any snags!

Issue #1676, opened 2018-05-04
Absences: Update "Last SST/note" column to be "days ago"


kevinrobinson:
Right now, this shows a date and we can sort.  But this leaves the user to do date math and we could do that for them, supporting this more directly and making visible what really matters - which kids need follow-up.

So an improvement proposed here is that we could update the date column to be "days ago" or "days since."  This number is always relative and so it's easier for folks to see right away how long it's been since the last follow-up.  We could also dually encode this information with the color of the text (eg, highlight students who haven't had a note in the last 45 days, reinforcing the same idea that's in the insight box on the home page).  Other design ideas would be welcome that tackle the underlying problem too!

kevinrobinson:
Closed by https://github.com/studentinsights/studentinsights/pull/1915 and https://github.com/studentinsights/studentinsights/pull/1917, and would be improved further by https://github.com/studentinsights/studentinsights/issues/1918.

PR #1915, opened 2018-07-20
Revise layout for dashboard pages, add "days ago" for notes

kevinrobinson:
# Who is this PR for?
K12 admin, counselors, housemasters

https://github.com/studentinsights/studentinsights/issues/1676, to change from showing the dates of notes to showing how many days ago they occurred, to norm this for users so they don't have to do the date math.  When this feature first launched, this didn't matter since folks could easily see which students had notes and which didn't, but as this list fills up (good!) it becomes hard to see what's old and what's recent at a glance.

In making this change, I wanted to make the table ~50px wider to accommodate this, but found that changes to the layout sizing didn't work as expected.  In looking to change this, I found that some of the CSS was shared across other pages, so changes here impacted other pages, which could be good.  But the use of those styles was different on different pages, so sharing these styles made making changes to fix layout on one page regress it on another page.

# What does this PR do?
Adds in the column.  Also, revises the layout for the absences, discipline and tardies pages.  These now use less shared styles, and these are only for the structure of the layout.  This is that the table on the left takes the full height of the page and a fixed width, the charts take the rest of the width, and then size their height to divide the screen in half.  This means these pages take the full screen and do not scroll.

Relatedly, this removes a wrapper element from `HighChartsWrapper` and revises it to allow styling the container.

Separately, it revises the controller response to pass down the school, so that we can add a title to the page that says what the data is and what school it's for.

screen shot 2018-07-20 at 9 52 34 am
screen shot 2018-07-20 at 9 52 49 am
screen shot 2018-07-20 at 9 55 34 am

# Checklists
+ [x] Author checked latest in IE - Absences


kevinrobinson:
The dates in the screenshots are unrealistic; I looked at updating the demo data to address this but cut that from this work.

kevinrobinson:
selfie

PR #1916, opened 2018-07-20
Don't show unexcused absences on Absences dashboard if feature is unsupported

kevinrobinson:
# Who is this PR for?
K12 New Bedford educators

Part of https://github.com/studentinsights/studentinsights/issues/1657#issuecomment-396062667, related to https://github.com/studentinsights/studentinsights/pull/1811.  We don't have this data for New Bedford, but the UI currently makes it seem like we do and that these buttons will filter by them.

# What does this PR do?
Removes these options for New Bedford.  There are a few related bits necessary to enable this - providing the `districtKey` as context within `<App />` and updating the Rails config to allow passing env values on the command line (which previously, prevented us from starting the app in a different instance with `DISTRICT_KEY=new_bedford rails s`).

### for new bedford
screen shot 2018-07-20 at 11 10 20 am

### DIFF
--- a/app/assets/javascripts/components/PerDistrictContainer.js
+++ b/app/assets/javascripts/components/PerDistrictContainer.js
@@ -1,0 +1,25 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+
+// A container that provides `districtKey`, assumed to be static across
+// the lifetime of all JS execution.  For React components, use this instead of
+// `readEnv`, which is intended for bootstrap code.
+export default class PerDistrictContainer extends React.Component {
+  getChildContext() {
+    const {districtKey} = this.props;
+    return {districtKey};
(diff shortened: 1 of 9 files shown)


kevinrobinson:
Note that this doesn't fully branch or address the discussion in https://github.com/studentinsights/studentinsights/pull/1811#discussion_r198544142; there's a bunch of computation that's unnecessarily when this feature isn't supported, which was done originally out of performance concerns as users interact with the page.  This PR punts that for now since 1) that time is probably better spent adding these features for NB and 2) we don't have a good way to measure those performance issues now so we should add that first before spending any other time optimizing.  Separately, switching the default selection to 45 days will change the relative benefit for different optimizations as well.

kevinrobinson:
selfie

PR #1919, opened 2018-07-20
Add FilterBar and Select components, refactor to use in My Students and HS Tiering

kevinrobinson:
# Who is this PR for?
developers

There are four implementations of filters (my students, tiering, absences/tardy/discipline dashboard filters, discipline select).  This starts splitting out small pieces so we can have a single UX for these.

# What does this PR do?
Factors out `FilterBar` as a container, and Select components for House, Grade and Counselor.  We could factor out other ones like the text search bar as well.  The absence/tardy/discipline dashboards would be a next target for these, making a single UX for filtering.  After that it's possible we could add this to the home page to address the HS needs around "Simplify feed for department heads and housemasters."

This also includes some unrelated refactoring that came from trying to adding this into the Absence dashboard as well.  That work is cut out here since that got too involved, but I kept the isolated changes to other code where it seemed like it made sense.  These include:
- adding `EscapeListener` that has a hook for the escape key
- factoring out functions in the student profile `QuadConverter` to `helpers/schoolYear` so that the Absence dashboard and others can use those instead of duplicating

### no regressions:
screen shot 2018-07-20 at 3 21 20 pm
screen shot 2018-07-20 at 1 58 45 pm


kevinrobinson:
selfie

Issue #1918, opened 2018-07-20
Set default time range on Absences, Discipline and Tardy dashboards to be 45 days


kevinrobinson:
This is line with other places in the product; this is our opinion on the time frame to be looking at, aimed at taking specific actions (eg, add to SST, talk with student of family and note it) rather than evaluating or analyzing per se.

This involves some refactoring, and it maybe be beneficial to pull in work related to `<FilterBar />` in the process (https://github.com/studentinsights/studentinsights/pull/1899 was a smaller reverted attempt, in the absence page there are three bits of state controlling this, and separately the precomputing code on top).

kevinrobinson:
Closed by https://github.com/studentinsights/studentinsights/pull/1920

PR #1920, opened 2018-07-21
Absence/Tardy/Discipline: Select components for filter and time range, default to 45 days

kevinrobinson:
# Who is this PR for?
K12 APs, redirect, housemasters

These default to full school year now which isn't aligned with taking action now, or with the insights box nudges on the home pages (https://github.com/studentinsights/studentinsights/issues/1918).  There isn't space in the UI to allow other filters that folks at the HS need (eg, house, counselor) or at K8 (filter out Pre-K).

# What does this PR do?
Migrate from the dashboard button UX to select components, factored out from the My students and Tiering pages, adding new ones for excused absences and time range.  Moves the time range to the right, and filtering actions on the left.  It also uses the factored out `schoolYear` functions from https://github.com/studentinsights/studentinsights/pull/1919.

This also revealed that the absence dashboard tests set up dates in a way that didn't make sense (eg, `displayDates` were for 2001 but `schoolAbsenceEventsByDay` were for the current date, counts for total absences were lower than counts for when broken down by homeroom).  Some of the good unit tests for this math are included and factored out into functions, but most of this is removed in this PR for a single fixture file generated from development data.

Separately, this PR removes unrelated stale code for routing and loading these dashboard pages from their own route with data inline; these haven't been used for a few months.

### before, absences
screen shot 2018-07-21 at 8 24 35 am

### after, each
screen shot 2018-07-21 at 8 23 35 am

screen shot 2018-07-21 at 8 22 52 am

screen shot 2018-07-21 at 8 23 06 am

# Checklists
+ [x] Author checked latest in IE - Absence


kevinrobinson:
selfie

PR #1921, opened 2018-07-21
Update insight box for absences to exclude excused absences by default

kevinrobinson:
# Who is this PR for?
K12 APs, redirect, counselors, housemasters, classroom teachers

Part of https://github.com/studentinsights/studentinsights/issues/1657.  These suggestions include students who already have been documented separately as being absence for excused reasons, so this results in them being false positives here (ie, an educator is already aware and has note that).

# What does this PR do?
This PR ignores those excused absences by default, if that data is present.  For districts without that data, nothing changes.

# Checklists


kevinrobinson:
selfie

PR #1923, opened 2018-07-21
Fix bug in InsightsHighAbsences spec

kevinrobinson:
Fixes https://travis-ci.org/studentinsights/studentinsights/builds/406589658, spec bug in https://github.com/studentinsights/studentinsights/pull/1921.  The random number of absences could fall below the threshold.


kevinrobinson:
selfie

PR #1924, opened 2018-07-21
Remove Filter by text for absence dashboard if excused absences not enabled

kevinrobinson:
This was branching on the select; should be branching on the whole bar:

screen shot 2018-07-21 at 11 20 16 am


kevinrobinson:
selfie

PR #1927, opened 2018-07-21
Add placeholder element for flex layout to work when no absences filter on dashboard

kevinrobinson:
# Who is this PR for?
NB educators

https://github.com/studentinsights/studentinsights/pull/1924 removed this filter; this means `justify-content: space-around` doesn't push this to the right.

# What does this PR do?
Adds a placeholder element so the layout works for both configurations.

### before
screen shot 2018-07-21 at 11 50 55 am

### after
screen shot 2018-07-21 at 11 50 21 am


kevinrobinson:
selfie

PR #1928, opened 2018-07-23
Tune down Devise timeoutable

kevinrobinson:
# Who is this PR for?
K12 educators, security

I mistakenly increased this in https://github.com/studentinsights/studentinsights/pull/32, thinking it was an absolute measure from sign-in, and not the total time since the last request.

# What does this PR do?
Cuts it down to 30 minutes, and comments about the heuristic in the UI warning about this.

### DIFF
--- a/app/assets/javascripts/sessionTimeoutWarning.js
+++ b/app/assets/javascripts/sessionTimeoutWarning.js
@@ -6,15 +6,20 @@ function show() {
   $('#renew-session').slideDown();
 }
 
+// Show a warning that the user's session is likely to timeout shortly.
+// This will be reset by Ajax calls or single-page navigation, so isn't entirely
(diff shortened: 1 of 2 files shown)


kevinrobinson:
@alexsoble I think I messed this up earlier, does this seem right to you?

kevinrobinson:
@alexsoble ping again on reviewing this when you get a chance.

kevinrobinson:
@alexsoble sounds good, thanks for merging!  👍 

Yeah the tricky thing here is that this approach won't work the same with long-lived pages (like the class list pages).  There each XHR resets the server-side timeout, but the JS code here doesn't know about that so it will alert too soon as a false positive.  So yeah not something to solve today :)

PR #1932, opened 2018-07-24
Tune down log level in test, since it is taking up GBs :)

kevinrobinson:
# Who is this PR for?
developers

Seeding the database with more students and running the tests a few times can lead to log sizes of a few GB.

# What does this PR do?
Tune down log level in test; I don't think we ever use this anyway.


kevinrobinson:
selfie, just local dev

PR #1935, opened 2018-07-24
Allow school_type to be null in HomeroomTable

kevinrobinson:
# Who is this PR for?
developers

propType warning that can happen, since `school_type: nil` is valid, and not set for any schools in NB.

# What does this PR do?
Relaxes propType, other code treats this as a whitelist so this is fine.


kevinrobinson:
selfie

PR #1946, opened 2018-07-25
Add more New Bedford Schools to admin page (+ update thor import task)

alexsoble:
# Who is this PR for?

Group of New Bedford principals who are going to be trained on Student Insights the week of 8/20. Depending on how the import process goes, we may be able to show them data from their own schools (beyond the two pilot schools) at that meeting.

# What does this PR do?

Adds additional schools to the admin page, using the schools whose leaders/staff will be attending the demo meeting this week. 

This way the admin page will include all the schools whose data we've imported from New Bedford, since we're going to try to import data for those pilot schools before the meeting. 

# Heroku deploy notes

+ [ ] Update task to: `thor import:start --background=false --skip-old-records --school '115' '123' '078' '063' '050' '040' '045' '405' '415' '410'`

### DIFF
--- a/app/config_objects/per_district.rb
+++ b/app/config_objects/per_district.rb
@@ -52,7 +52,22 @@ class PerDistrict
   # with pilot schools in New Bedford shown first.
   def ordered_schools_for_admin_page
     if @district_key == NEW_BEDFORD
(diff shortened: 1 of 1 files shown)


alexsoble:
Kicked off manual import job with `--no-skip-old-records`.

PR #1947, opened 2018-07-25
Fix denominator in STAR Reading integrity check

alexsoble:
# Who is this PR for?

Developers.

Wrong denominator.


alexsoble:
selfie

PR #1948, opened 2018-07-25
Validate STAR assessment data; add more data quality code comments

alexsoble:

+ Unclear and unenforced validations around STAR assessment data attributes.

# What does this PR do?

+ Enforces validations for STAR records, records based on analysis of what the raw data coming in from STAR looks like. (Data quality is high from STAR with no missing fields found in their CSV exports).
+ Add notes about STAR `:grade_equivalent` data in Somerville's production instance. 
+ Delete validation code that has been commented out.

### DIFF
--- a/app/importers/file_importers/star_math_importer.rb
+++ b/app/importers/file_importers/star_math_importer.rb
@@ -63,16 +63,16 @@ class StarMathImporter
       return
     end
 
-    star_assessment = StudentAssessment.where({
+    star_assessment = StudentAssessment.find_or_initialize_by(
       student_id: student.id,
       date_taken: date_taken,
       assessment: star_mathematics_assessment
-    }).first_or_initialize
+    )
 
-    star_assessment.update_attributes({
+    star_assessment.assign_attributes(
       percentile_rank: row['PercentileRank'],
       grade_equivalent: row['GradeEquivalent']
-    })
+    )
 
     star_assessment.save!
   end
--- a/app/importers/file_importers/star_math_importer.rb
+++ b/app/importers/file_importers/star_math_importer.rb
@@ -67,12 +67,14 @@ class StarMathImporter
       student_id: student.id,
       date_taken: date_taken,
       assessment: star_mathematics_assessment
-    }).first_or_create!
+    }).first_or_initialize
 
-    star_assessment.update!({
+    star_assessment.update_attributes({
       percentile_rank: row['PercentileRank'],
       grade_equivalent: row['GradeEquivalent']
     })
+
+    star_assessment.save!
   end
 
   def log(msg)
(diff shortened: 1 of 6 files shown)


app/importers/file_importers/star_math_importer.rb
```diff
@@ -63,16 +63,18 @@ def import_row(row)
       return
     end
 
-    star_assessment = StudentAssessment.where({
+    star_assessment = StudentAssessment.find_or_initialize_by(
```
alexsoble: This method matches the `find_or_initialize_by` used in `mcas_row.rb`.

app/importers/file_importers/star_math_importer.rb
```diff
@@ -63,16 +63,18 @@ def import_row(row)
...
       student_id: student.id,
       date_taken: date_taken,
       assessment: star_mathematics_assessment
-    }).first_or_create!
+    )
 
-    star_assessment.update!({
+    star_assessment.assign_attributes(
```
alexsoble: This method matches the `assign_attributes` used in `mcas_row.rb`.

app/models/student_assessment.rb
```diff
@@ -9,34 +9,31 @@ class StudentAssessment < ActiveRecord::Base
...
     when 'STAR'
-      # errors.add(:scale_score, "invalid attributes") unless valid_star_attributes
-    when 'DIBELS'
-      # errors.add(:scale_score, "invalid attributes") unless valid_dibels_attributes
+      errors.add(:percentile_rank, "invalid") unless valid_star_attributes?
+
+      # TODO: Add validation for STAR grade_equivalent field.
+      # This should be present for all STAR records, but because this wasn't
+      # backfilled to older records, ~40% of STAR records in production
+      # have grade_equivalent of nil.
```
alexsoble: Not clear to me if the best practice is to add a validation even if it would mark ~40% of existing data as invalid, or just leave this note here and add in the validation once we've cleaned up the data. 

My instinct is to hold off on adding the validation for now. That makes all the other validations we're adding useful, because all things equal we expect our data to be in a valid state and any validation errors are from something unexpected.

alexsoble:
Trying to validate that STAR import works correctly locally before I merge, but having some trouble doing that. When I run the import task against this branch locally, no STAR rows are created. More mysteriously, when I add `log('...')` and `@log('...')` statements to `star_math_importer.rb` to learn more, they aren't getting printed to the console.

alexsoble:
Ah, I got it. Looks like there's an issue with `SchoolFilter` here.

app/importers/file_importers/star_math_importer.rb
kevinrobinson: FWIW, what motivates the factoring in `McasRow` is that it only does one read.  And then the importer uses `RecordSyncer` to do the write, and to track which Insights records are updated, created or left unchanged.

alexsoble: Gotcha! That would be a good pattern to follow for the STAR importers.

alexsoble:
Verified that the STAR data import process works on this branch! 👍

Added a bunch of code comments on student assessment validations in addition to the STAR validation work. 

And thanks for the review/code comments @kevinrobinson!

🚢ing this one now.

PR #1950, opened 2018-07-26
Fix up STAR import by passing only School Local ID to school filter

alexsoble:
# Who is this PR for?

Data import process. 

Bug in STAR data import; the bug blocks new STAR data from importing.

The SchoolFilter class assumes each row has the same shape and checks for `row[:school_local_id]`; STAR rows have a different shape and store this data as `row['SchoolLocalID']`. 

# What does this PR do?

The implementation details of the key name shouldn't matter to the SchoolFilter class, and it doesn't need to know anything about a `row` other than it's school local ID. This changes the code so that only school local ID gets passed in.

### DIFF
--- a/app/importers/file_importers/attendance_importer.rb
+++ b/app/importers/file_importers/attendance_importer.rb
@@ -111,7 +111,7 @@ class AttendanceImporter
 
   def import_row(row)
     # Skip based on school filter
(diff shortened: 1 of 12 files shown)


alexsoble:
Verified locally; 🚢ing.

PR #1951, opened 2018-07-27
Profile v3: Prototype for design iterations (not completely functional)

kevinrobinson:
# Who is this PR for?
project team, designers 

This is a prototype of a design iteration related to the student profile page.  I'll skip describing the design intentions and features here, since we've talked about those separately over email and in person, and they're likely to change.

To enable iterating on the design, it'll be helpful to look at some real student data and see what comes out.  This PR ships that capability, aiming for minimal changes to existing product code, since this design iteration is likely to change and isn't even fully functional.  The intention is to use this for further design iterations and feedback, and a bunch of other work would be necessary to ship this for real.

I didn't include the `<ExperimentalBanner />` here like we normally would because we want feedback on the feel of the page and the layout changes, which the banner would interfere with.  But this code is all experimental and should be written so that we can delete it all with no impact to the existing profile page or other pages.

# What does this PR do?
Makes a new profile page accessible at `/students/:id/v3`.  It isn't fully functional or complete.

The overall approach for iterating on this is to create a parallel `<LightProfilePage />` that has the same API as the existing `<StudentProfilePage >/` so changes are completely isolated.  This new page uses many existing components, but in places where it would require significant changes, this PR copies those files (eg, `<LightProfileHeader />`) so these changes are isolated from the existing profile page.  This also means that all files here have this "Light" prefix (which doesn't mean anything in particular, I think maybe because it was more visually "lightweight" initially :)).

In a few places, existing components are changed minimally, such as allowing styles to be passed in, or adding an additional boolean prop.  The changes are listed below:

New work:
- `LightProfilePage`
- `LightProfileHeader`
- `LightNotesDetails`
- `LightServicesDetails`

UI elements:
- `LightProfileTab`
- `LightHelpBubble`
- `LightCarousel`

Changes to existing code:
- `<AttendanceDetails />`, `<ElaDetails />` and `<MathDetails />` take prop to hide navbar links
- Updates `<TakeNotes />`, `<HelpBubble />`, `<SectionHeading />` and `<StudentPhoto />` to take standard `style` prop for optional styles
- add `helpers/language` copied out from class lists

leaving these out, this is still in design phase and we'll iterate on this over email or in-person instead

# Checklists

I added a few smoke tests, but didn't really do this yet.  That's probably for the next iteration, when we get to a decision point to go forward in this direction or not.

### DIFF
--- a/app/assets/javascripts/components/HelpBubble.js
+++ b/app/assets/javascripts/components/HelpBubble.js
@@ -31,9 +31,10 @@ class HelpBubble extends React.Component {
   }
 
   render(){
+    const {style, linkStyle} = this.props;
     return (
-      <div style={{display: 'inline', marginLeft: 10}}>
-        <a href="#" onClick={this.openModal} style={{fontSize: 12, outline: 'none'}}>
+      <div style={{display: 'inline', marginLeft: 10, ...style}}>
+        <a href="#" onClick={this.openModal} style={{fontSize: 12, outline: 'none', ...linkStyle}}>
           {this.props.teaser}
         </a>
         {// The modal is not logically here, but even while not displayed it needs a location in the DOM.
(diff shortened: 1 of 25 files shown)


kevinrobinson:
selfie

PR #1952, opened 2018-07-27
Validate SchoolFilter inputs and make "all schools" option more explicit 

alexsoble:
# Who is this PR for?

Developers working with the import process.

Classes that interact with `SchoolFilter` are unclear about what inputs get sent down; makes it confusing to work with as a developer. 

The code in `SchoolFilter` says that if's passed a school local IDs list of `nil`, then it won't filter at all and will include all schools. But the code in `ImportTask` is set up differently and never sends in `nil`, it sends down an array of all school local IDs if none are supplied by the user. 

# What does this PR do?

+ Use special `:all` keyword for `SchoolFilter` to accept rows from all schools.
+ Validate argument inputs to `SchoolFilter`.

### DIFF
--- a/app/importers/helpers/school_filter.rb
+++ b/app/importers/helpers/school_filter.rb
@@ -1,5 +1,18 @@
-class SchoolFilter < Struct.new(:school_local_ids)
+class SchoolFilter
+  def initialize(school_local_ids)
+    raise "Invalid school_local_ids: #{school_local_ids}" unless valid?(school_local_ids)
+    @school_local_ids = school_local_ids
+  end
+
+  def valid?(school_local_ids)
+    return true if school_local_ids == :all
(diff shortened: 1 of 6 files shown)


alexsoble:
Did some testing. Turns out this branch is breaking the import process. Still not quite sure why. 😢

When I run `thor import:start --source students` against master, it successfully imports the students file. When I run it against this branch, I crash against:

```
💾  ImportTask: Error! no implicit conversion of nil into Hash

💾  ImportTask: ["/studentinsights/app/importers/rows/student_row.rb:37:in `merge'" ...]
```

Still not sure why the changes in this branch are causing the error, but this branch is more cleanup and less critical bugfix, so might not solve and merge this right away.

alexsoble:
Closing; this isn't on the critical path right now.

PR #1953, opened 2018-07-27
Profile v3: Layout and quote iterations

kevinrobinson:
# Who is this PR for?
educators

Follow on to https://github.com/studentinsights/studentinsights/pull/1951.

# What does this PR do?
- Removes `?transition` flag in query string, read real data by default now
- Fixes layout for larger images
- Fix links to IEP document
- Adds in an overflow for long quotes for now, so they don't bust the layout
- Adds in a sketch of what adding a new insight inline might look like

This PR doesn't touch other production code.  Also, this is busted in IE11 as is https://github.com/studentinsights/studentinsights/pull/1951.


kevinrobinson:
selfie

PR #1954, opened 2018-07-27
Fix layout on share an Insight

kevinrobinson:
# Who is this PR for?
educators

Layout bug in https://github.com/studentinsights/studentinsights/pull/1953.

# What does this PR do?
Fixes it.

After:
screen shot 2018-07-27 at 6 20 22 pm

# Checklists
This feature is isolated from production code.


kevinrobinson:
selfie

PR #1955, opened 2018-07-27
Grades can be null, so skip them in profile

kevinrobinson:
# Who is this PR for?
HS educators

`grade_numeric` can be null and often is, so profile v3 mistakenly considers that a course with a lot grade when it should skip it.

# What does this PR do?
Fixes the bug.

# Checklists
This code doesn't touch production paths.


kevinrobinson:
selfie

PR #1956, opened 2018-07-27
Null is not isNaN

kevinrobinson:
# Who is this PR for?
me learning JS :)

null is not isNaN in https://github.com/studentinsights/studentinsights/pull/1955

# What does this PR do?
just check for null :)


kevinrobinson:
selfie

PR #1957, opened 2018-07-30
Create new table and model for STAR results; add database constraints and total_time field

alexsoble:
# Who is this PR for?

Educators who work with STAR data.

As part of the monolithic `student_assessments` table, there's no way to apply database validations specific to STAR test results, for example:

```
:grade_equivalent, null: false
```

While pulling STAR into its own table and improving data quality guards, I took the opportunity to make a few other forward-looking changes as well:

# Key changes in new model/table 
* Enforce `null: false` validations at the database level for all fields
* Import `:total_time` field
* Import time as well as date for `:date_taken` 
* Validate `:percentile_rank` values at the Rails code level

# Migration strategy 

This is step one of a three-step migration strategy to create separate tables for STAR Math and Reading results. The tables need to be separate because Reading has an additional field for instructional reading level.

We'll migrate over in stages:
* First, dump data from the import process into the new model's table.
* Next, point endpoints in the app over to read from the new table. (The app may be showing slightly stale data until this step is shipped.)
* Finally, remove STAR results from the old monolithic `student_assessment` model/table.

# State of the database after deploying

After this PR is deployed and new data is imported, the database will be in a temporary state with STAR data stored in two different tables. I'll use this opportunity to check for any data quality differences or issues between the two tables. This state will be resolved at the end of the migration process, when I'll delete STAR data from the old monolithic `student_assessment` table.

There is now another STAR ZIP file to read from, with Historical data. I may need to import Historical and current data into the new table separately with two different import jobs.

### DIFF
--- a/app/importers/file_importers/star_math_importer.rb
+++ b/app/importers/file_importers/star_math_importer.rb
@@ -58,8 +58,8 @@ class StarMathImporter
       return
     end
 
-    date_in_cst = row.fetch('AssessmentDate') + ' -6'  # All times in the CSV are CST
-    date_taken = DateTime.strptime(date_in_cst, "%m/%d/%Y %H:%M:%S %z")
+    date_in_cst = row.fetch('AssessmentDate') + ' CST'  # All times in the CSV are CST
+    date_taken = DateTime.strptime(date_in_cst, "%m/%d/%Y %H:%M:%S %Z")
 
     test_result = StarMathResult.find_or_initialize_by(
       student_id: student.id,
--- a/app/importers/file_importers/star_math_importer.rb
+++ b/app/importers/file_importers/star_math_importer.rb
@@ -60,8 +60,7 @@ class StarMathImporter
 
     datetime_string = row.fetch('AssessmentDate')
     day = DateTime.strptime(datetime_string, "%m/%d/%Y")
-    time = Time.strptime(datetime_string, "%m/%d/%Y %H:%M:%S")
-               .in_time_zone('America/Chicago')  # Times from STAR are in CST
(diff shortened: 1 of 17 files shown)


alexsoble:
selfie 🚢

alexsoble:
Deployed and imported for Somerville & New Bedford.

app/importers/file_importers/star_math_importer.rb
```diff
@@ -51,30 +51,39 @@ def filter
...
+    student = Student.find_by_local_id(row.fetch('StudentLocalID'))
     if student.nil?
       log("skipping, StudentLocalID not found: #{row['StudentLocalID']}")
       return
     end
 
-    star_assessment = StudentAssessment.find_or_initialize_by(
+    datetime_string = row.fetch('AssessmentDate')
+    day = DateTime.strptime(datetime_string, "%m/%d/%Y")
+    time = Time.strptime("#{datetime_string} CDT", "%m/%d/%Y %H:%M:%S %Z")
```
kevinrobinson: Can you say more about why this is parsing a time string in CDT here?

db/schema.rb
```diff
@@ -328,6 +328,29 @@
...
 
+  create_table "star_math_results", force: :cascade do |t|
+    t.integer "percentile_rank", null: false
+    t.string "grade_equivalent", null: false
+    t.integer "total_time", null: false
+    t.bigint "student_id", null: false
+    t.datetime "date_taken", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["student_id"], name: "index_star_math_results_on_student_id"
```
kevinrobinson: this looks awesome! :)

PR #1959, opened 2018-07-31
Enforce student_id on discipline incidents at database level

alexsoble:
# Who is this PR for?

Data quality. 

Rails validations aren't sufficient to ensure data quality; database-level guards needed. We need to know the `:student_id` of a discipline incident, otherwise the record isn't useful.

# What does this PR do?

Adds a migration to enforce `null: false`. 

# Existing data quality / Deployment

## Somerville 

Digging into the Somerville database showed that there are 1,047 discipline records missing a student ID, compared with 17,650 valid records that have a student ID. This means that 5.9%. are missing a student ID.

How was this introduced? The current Rails validation on the discipline incident model (`validates_presence_of :student`) was introduced Jun 2, 2017, when we moved from associated discipline incidents with student school years to associating them with students: #929. All of the records without a student ID were introduced on or before May 25 2017. So these are older records that were added to the database before the validation was added. 

Some looking at exampled showed that these are duplicates of other discipline incident records. Because they have no student IDs, they would be excluded from scoped queries like `authorized_students_for_dashboard` and should not have ever been surfaced in the product. Of course there's no way for them to appear on any student profile since they have no student ID.

Before this PR is merged and the database migration run, I'll need to clear out the invalid data with something like:

```
DisciplineIncident.where(student_id: nil).destroy_all
```

Otherwise the migration won't run, because it would create data that violates the database constraint. 

## New Bedford

```
irb(main):005:0> DisciplineIncident.where(student_id: nil).count
=> 0
```

### DIFF
--- a/app/importers/file_importers/behavior_importer.rb
+++ b/app/importers/file_importers/behavior_importer.rb
@@ -59,7 +59,7 @@ class BehaviorImporter
       return
     end
 
-    behavior_event = BehaviorRow.build(row, student.id)
+    behavior_event = BehaviorRow.new(row, student.id).build
     if !behavior_event.valid?
       @invalid_rows_count += 1
       return
--- a/app/importers/file_importers/behavior_importer.rb
+++ b/app/importers/file_importers/behavior_importer.rb
@@ -14,6 +14,7 @@ class BehaviorImporter
 
     log('Starting loop...')
     @skipped_from_school_filter = 0
+    @skipped_from_invalid_student_id = 0
     @touched_rows_count = 0
     @invalid_rows_count = 0
     streaming_csv.each_with_index do |row, index|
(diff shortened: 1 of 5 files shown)


alexsoble:
Tested out importing Somerville discipline incidents on this branch locally and it worked smoothly: 

```
BehaviorImporter: Starting loop...
BehaviorImporter: processed 0 rows.
BehaviorImporter: processed 10000 rows.
StreamingCsvTransformer#stats: {:skipped_dirty_rows_count=>0, :processed_rows_count=>14140}
BehaviorImporter: Done loop.
BehaviorImporter: @skipped_from_school_filter: 1708
BehaviorImporter: @skipped_from_invalid_student_id: 0
BehaviorImporter: @touched_rows_count: 12432
BehaviorImporter: @invalid_rows_count: 0
```

Note that there were no rows skipped because of a missing or invalid school local ID.

Going to remove dirty data from the Somerville instance now and then 🚢 this one.

PR #1961, opened 2018-08-01
Sync and delete Discipline records missing from Aspen/X2 export CSV 

### LINKED ISSUE #1709: Discipline: Investigate duplicate events
Have to search for test cases.  This might be happening because a principal is revising something in Aspen, and that's creating a new record that then is importing in Insights in a way that we can't match and do an update.

alexsoble:
# Notes 

Digging into the discipline data set revealed records that don't exist anymore in the upstream CSVs from New Bedford and Somerville. Queries suggested that about 2.7% of Somerville student records and 1.7% of New Bedford student records contained one of these records. 

Often the records would look highly similar to another discipline incident in the student's event feed. This suggested a problem of educators editing or deleting discipline incident records in Aspen/X2. These changes failed to sync to the Insights database because the discipline import code has been create-only, with no delete path. Uri confirmed that this might be the case, adding that he thought editing/deleting is rare. 

This PR uses the `RecordSyncer` class that @kevinrobinson introduced in order to remove those events that are missing from the upstream CSVs. It also changes the `BehaviorRow` model to be extremely "picky": if any data about a behavior event changes upstream, the old record is destroyed and entirely new one is created. This create/destroy should be a more reliable method than trying to detect which events are the same and updating. 

# Issue

+ Fix #1709.

# Testing and QA

+ [x] Added specs for `BehaviorImporter#import` to test deleting behavior. 
+ [x] Manually imported a small amount of behavior data locally; used the Rails console to manually edit one of the events' `description` attribute; and re-imported to test create/delete behavior: 

```
Sync stats: {
:passed_nil_record_count=>0, 
:invalid_rows_count=>0, 
:unchanged_rows_count=>1225,
:updated_rows_count=>0, 
:created_rows_count=>1,
:marked_ids_count=>1226, 
:destroyed_records_count=>1
}
```

# Checklist 

+ [x] Merge and deploy
+ [ ] Run `thor import:start --source behavior` for New Bedford and Somerville; note how many discipline incident records are deleted

### DIFF
--- a/app/importers/file_importers/behavior_importer.rb
+++ b/app/importers/file_importers/behavior_importer.rb
@@ -34,6 +34,8 @@ class BehaviorImporter
     log("@invalid_rows_count: #{@invalid_rows_count}")
     log("@touched_rows_count: #{@touched_rows_count}")
 
+    log('Calling #delete_unmarked_records...')
+    @record_syncer.delete_unmarked_records!(records_within_scope)
     log("Sync stats: #{@record_syncer.stats}")
   end
 
@@ -94,7 +96,6 @@ class BehaviorImporter
     maybe_matching_record = BehaviorRow.new(row, student.id).build
 
     @record_syncer.validate_mark_and_sync!(maybe_matching_record)
-    @record_syncer.delete_unmarked_records!(records_within_scope)
   end
 
   def within_date_range?(row)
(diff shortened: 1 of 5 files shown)


alexsoble:
*TODO:*

+ [ ] Add integration spec for Behavior Importer + Record Syncer
+ [ ] Test manually

alexsoble:
selfie 🚢

PR #1962, opened 2018-08-01
Import recent STAR events to new database tables

alexsoble:
# What does this PR do?

Follow-on to #1957: now let's import more recent STAR data from `SM.csv` and `SR.csv` to the new STAR database tables with extra rows and database-level validation.


alexsoble:
selfie

PR #1964, opened 2018-08-03
Update lodash to 4.17.10, most recent

alexsoble:

Additional friction on front-end pull requests that comes with having to remember to use and older version when looking up lodash methods.

# What does this PR do?

Upgrades to 4.0.0 using lodash codemod: https://www.npmjs.com/package/lodash-codemods.

Still a WIP because tests pass, but there are warnings and I haven't done an actual product QA yet.

### DIFF
--- a/app/assets/javascripts/class_lists/ClassListCreatorPage.test.js
+++ b/app/assets/javascripts/class_lists/ClassListCreatorPage.test.js
@@ -27,7 +27,7 @@ function mountWithContext(props) {
 }
 
 function anyServerCallsIncludePath(string) {
-  return _.any(fetchMock.calls(), call => {
+  return _.some(fetchMock.calls(), call => {
     const path = call[0];
     return path.indexOf(string) !== -1;
   });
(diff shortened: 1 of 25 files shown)


alexsoble:
Sample of warnings we're getting from the codemod:

WARNING: In file `app/assets/javascripts/school_overview/SchoolRoster.js:37:20`, please check that the iteratee passed to `_.groupBy` only uses its first argument `value`.
WARNING: In file `app/assets/javascripts/student_profile/QuadConverter.js:94:9`, please check that the iteratee passed to `_.sortBy` only uses its first argument `value`.
WARNING: In file `app/assets/javascripts/tiering/TieringView.js:239:38`, please check that the iteratee passed to `_.orderBy` only uses its first argument `value`.
WARNING: In file `app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js:127:29`, please check that the iteratee passed to `_.groupBy` only uses its first argument `value`.

PR #1967, opened 2018-08-03
Fix Capuano school roster bug

alexsoble:

Fix broken roster page for Capuano school (CAP), Parent Information Center (PIC), SPED.

# What does this PR do?

These schools don't have a school type, i.e. they're not High Schools or Elementary/Middle School. Capuano is an Early Childhood Center.

 This case wasn't handled in `PerDistrict#studentTableEventNoteTypeIds` and led to a bug with no School roster rendering. 

# What's the fix?

Handle "unsupported district key" at the top, then handle different Somerville school type cases below and and provide default for no school type.


kevinrobinson:
Nice fix!

PR #1968, opened 2018-08-07
Add total time to STAR chart tooltips

alexsoble:
# Notes

One critical piece of context for understanding STAR data results is the time the student spent taking the test. If the student spent less than a minute, that means they quit early and the data point is essentially junk data. This is a first pass at visualizing total time along with STAR results, by adding it to the additional information displayed in the tooltip.

# Screenshot (Internet Explorer, fake data)

virtualbox_ie11 - win7_08_08_2018_10_42_48

# Checklist 

+ [x] Checked latest commit in Internet Explorer
+ [x] Added specs for additional JS

### DIFF
--- a/app/assets/javascripts/student_profile/ProfileChartSettings.js
+++ b/app/assets/javascripts/student_profile/ProfileChartSettings.js
@@ -3,6 +3,15 @@ import moment from 'moment';
 const ProfileChartSettings = {};
 export default ProfileChartSettings;
 
+function humanizeStarTotalTime (seconds) {
+  if (seconds < 60) return `${seconds} seconds`;
+
+  const minutes = Math.floor(seconds / 60);
+  const remainingSeconds = seconds % 60;
+
+  return `${minutes} minutes and ${remainingSeconds} seconds`;
+}
+
 ProfileChartSettings.base_options = {
   chart: {
     renderTo: 'chart',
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/student_profile/QuadConverter.js
```diff
@@ -56,7 +56,8 @@ export function toStarObject(starObject) {
   return {
     x: toMomentFromRailsDateTime(date_taken).valueOf(),
     y: starObject.percentile_rank,
-    gradeLevelEquivalent: starObject.grade_level_equivalent
+    gradeLevelEquivalent: starObject.grade_equivalent,
```
alexsoble: This is an active bug.

app/assets/javascripts/student_profile/ProfileChartSettings.js
```diff
@@ -87,9 +87,13 @@ ProfileChartSettings.star_chart_base_options = {
       const gradeLevelEquivalent = this.points[0].point.gradeLevelEquivalent;
       const formattedGradeEquivalent = `<br>Grade Level Equivalent: <b>${gradeLevelEquivalent}</b>`;
 
+      const totalTime = this.points[0].point.totalTime;
+      const formattedTotalTime = `<br>Total Time: <b>${totalTime}</b>`;
+
       return `${formattedDate}
               ${formattedPercentileRank}
-              ${gradeLevelEquivalent ? formattedGradeEquivalent : ''}`;
```
alexsoble: We now have database-side constraints guaranteeing that `gradeLevelEquivalent` will always be there, so removing this.

alexsoble:
Realizing that `moment.humanize()` doesn't supply a super accurate or precise result. It's rounds 23 minutes and 40 seconds up to say "24 minutes", which is less precise than we want. Looking at https://github.com/moment/moment/issues/348.

alexsoble:
selfie 🚢

PR #1969, opened 2018-08-09
Add new DIBELS table forked from monolith student assessments table

alexsoble:
# Notes 

Similar to the pattern we followed with STAR, this creates a new table, model, and data import path for DIBELS test results. Using a separate table allows us to:

+ Have more accurate column names on each individual assessment table
+ Use database-enforced presence validations with `null: false`
+ More easily write Rails validations

# Deployment path

As with STAR, this change will be deployed in steps:

1. Add new table; change import path to point to new table; import data to new table (<= this PR)
2. Change controller and client side code to read from new table
3. Remove code that references old table and destroy records in old table 

There will be a period after (1) ships and before (2) ships where the Insights app can show stale data. Since DIBELS isn't likely to be administered over the summer, this isn't a major concern. 

# District notes

New Bedford DIBELS data comes down the import path in a significantly different shape from Somerville DIBELS data. This PR imports `DibelsResult` records for Somerville only. Once we have more detailed conversations with New Bedford about DIBELS we can devise a plan and write a code path for New Bedford DIBELS.


alexsoble:
Tested against New Bedford and Somerville in production... 🚢.

Issue #1758, opened 2018-06-04
Simplify: Remove global JS dependencies vendored from Rails gems


kevinrobinson:
Follow-on from https://github.com/studentinsights/studentinsights/issues/1180

+ [ ] all the other files imported in legacy.js (https://github.com/studentinsights/studentinsights/pull/1739)
+ [ ] all references to `window.shared` or other `window.` variables our code put there
+ [ ] ProvidedByEducatorDropdown uses $.autocomplete and touches body
+ [ ] datepicker reliance on jQuery plugin, globals, server interaction, `window.datepicker_options`
+ [ ] remove vendored jquery UI gem
+ [ ] include jquery_ujs in JS build, remove gem
+ [ ] remove application.js / remove sprockets-shims / remove js gems / remove global in eslintrc 
+ [ ] move to ui folder

PR #1971, opened 2018-08-13
Migrate global datepicker_options into Datepicker component, limit jQuery UI imports

kevinrobinson:
# Who is this PR for?
developers

It's a step towards https://github.com/studentinsights/studentinsights/issues/1758.  This came up as I was looking to see what it would take to upgrade to jQuery 3.x, since Firefox raises CSP violations on load for jQuery 1.12.x.

# What does this PR do?
Removes two global bits that `<Datepicker />` relied on - `window.datepicker_options` and an asset path sent down in `application.html.erb`.  Also scopes down the sprockets imports for jQuery UI.


kevinrobinson:
selfie

PR #1973, opened 2018-08-13
Cut over to new DIBELS table in controller and JS code; remove deprecated DIBELS records 

alexsoble:
# Notes 

+ This moves #1969 forward by reading from the new DIBELS table with improved column names and database-enforced validation. 
+ The next step after this PR will be to purge DIBELS data from the old monolithic `student_assessments` table. 
+ This introduces a small regression in the amount of DIBELS data we show on the Student Profile Page for Somerville. Instead of showing benchmark plus subtest results ("CORE 100 NWF-CS"), we now show just the benchmark score ("CORE"). 
  + The tradeoff here is that we won't show any invalid data, like, because the `DibelsResults` table has stricter validation. I think this tradeoff is worth it: it's not clear how valuable educators find subtest data, but junk values showing up on the Student Profiles is a big no.
  + We can display subtest results in the Student Profile with a separate PR, and probably with a better UI (hover for subtest results, or click a "subtest results" link). 

# QA Checklist 

## Student Profile
+ [x] Checked Student Profile Page (Header) in Internet Explorer VM
+ [x] Checked Student Profile Page (Profile Details) in Internet Explorer VM
+ [x] Checked Student Profile PDF 

### Screenshots (Student Profile, Internet Explorer, fake data)

student-profile-header-dibels

profile-detail-dibels

dibels-pdf

## Class Lists
+ [x] Checked InlineStudentProfile in Internet Explorer VM
+ [x] Checked StudentCard in Internet Explorer VM
+ [x] Checked ClassroomStats in Internet Explorer VM

### Student cards highlighted based on DIBELS 

class-lists-dibels-highlighted

### Inline profile with DIBELS data

inline-profile

# Deploy checklist

+ [ ] Merge and deploy
+ [ ] Run `heroku run rake data_migration:remove_deprecated_dibels`
+ [ ] Manual QA in production

PR #1975, opened 2018-08-14
Update jest to 23.x.x

alexsoble:
# Who is this PR for?

Developers.

Our `jest` version is old and we were capped from being auto-upgraded above 21.x.x.

The newer version comes with better error handling in async contexts. I got stuck for a bit on #1973 because `jest` was giving me an "Timeout - Async callback was not invoked within timeout" error, when the real issue was much simpler than that. I tried again with a higher `jest` version and the error was much simpler to resolve.

See https://github.com/facebook/jest/pull/4669 for details.

# What does this PR do?

Bumps `jest` major version and commits changes to snapshots. 

These new snapshots look better because crufty `width={undefined}` type styling gets removed in favor of no width attribute at all.


alexsoble:
selfie 🚢 cc @kevinrobinson

kevinrobinson:
@alexsoble awesome! 😄

PR #1977, opened 2018-08-16
Minimally add Bedford to PerDistrict; deploy script

alexsoble:
# Who is this PR for?

Bedford Public Schools.

Our newest school district, Bedford, needs to be added to our per-district config files.

If we launch a Heroku instance for Bedford without these changes, I would expect it to crash server-side because this line keeps us from launching an instance without setting a valid district key: 

```
raise_not_handled! unless VALID_DISTRICT_KEYS.include?(@district_key)
```

# What does this PR do?

Add Bedford to PerDistrict files so that we can launch a minimal Heroku instance. 

Also adds the "bedford" remote to our deploy script so that code will automatically deploy there.


alexsoble:
🚢 selfie

PR #1978, opened 2018-08-16
Edit "New Instance" documentation along with Bedford set up

alexsoble:
# Who is this PR for?

Developers.

# What does this PR do?

+ Update `new_instance.md` document while I set up Bedford's instance, so that the steps in the doc better match the steps of setting up a new district.
+ Remove verbose docs that weren't adding much extra value, keep a consistent checklist style.


kevinrobinson:
@alexsoble This doc lives in another place now and we could close this PR, is that right?

alexsoble:
@kevinrobinson Yes!

alexsoble:
@kevinrobinson To be more specific, I think this is in a good place to merge. It deletes some content, moves other content around within the doc, and moves other content to a separate checklist outside of public GitHub.

kevinrobinson:
Ah I misunderstood - this is the overall doc I thought it was an older version of `new_sftp.md`.  Sweet, let's merge this sucker! 👍 🚢

alexsoble:
👍

PR #1980, opened 2018-08-17
Refactor session timeout warning into component; update timing logic and add client-side signout

kevinrobinson:
# Who is this PR for?
educators

When we decreased the session timeout, this revealed some problems with the way the browser-side session timeout works.  It wouldn't work correctly when the server session was forcibly timed out in two ways: 1) it would leave student data visible in the browser after the client knew the server session had timed out and 2) it would show the user a "please click this link" message that wouldn't do anything from the user's perspective, and would not work to renew the session that had already timed out.

# What does this PR do?
Adds a `<SessionRenewal />` component that tries to handle more of these cases, so that we can keep the timeout more aggressive (30m).  This will warn the user 2 minutes before it thinks the server timeout will be hit.  After that, the UI will try to sign out the user and navigate back to the sign-in page.  The intent here is to clear the browser screen and communicate to the user that their session has expired.  It's not precisely aligned with the server session timeout but for most pages will be roughly close.  If the server session has already timed out, and the users clicks the renew link, they are redirected on the UI side to the sign in page.

Since the tests use `fetchMock` and require async assertions, these run into test pollution when Jest runs them in parallel.  Splitting those test cases into separate files appears to resolve this.

Notably, this won't work for routes like the student profile that are still in `legacyRoutes` since this is done as part of `App.js`.  Ideally we'd finish that migration before the start of the school year so this works the same across all pages. 

screen shot 2018-08-14 at 2 34 17 pm
screen shot 2018-08-14 at 2 33 55 pm

# Checklists


kevinrobinson:
selfie FYI @alexsoble

alexsoble:
@kevinrobinson Great improvements to one of the oldest and funkiest parts of the codebase! Bye jQuery and bugs, hi React and solid tested functionality! 👍👍

PR #1982, opened 2018-08-17
Force vertical scrollbars on profile v3, even if not enough content

kevinrobinson:
# Who is this PR for?
educators

There's a slight jump in layout when navigating between tabs if there isn't enough vertical content (eg, no notes) to require a scrollbar.  With no scrollbar, the width changes and layout hops a bit.

# What does this PR do?
Always adds vertical scrollbar to the body on profile v3.


kevinrobinson:
selfie

PR #1984, opened 2018-08-17
Fix bug with mutating STAR scores for Highcharts

kevinrobinson:
# Who is this PR for?
educators

In profile v3, clicking around the UI would change the STAR scores shown.  This was happening because `<ElaDetails />` and `<MathDetails />` were calling `#reverse()` and mutating the arrays of scores.

# What does this PR do?
Updates the details components to not mutate their arguments, and updates the tab code in `<LightProfilePage />` to not expect these to be already sorted.


kevinrobinson:
selfie

PR #1994, opened 2018-08-20
Validate educators assigned to correct homerooms

alexsoble:

+ Educators at our new school district deployment assigned to incorrect homerooms.

# What does this PR do?

+ Changes the import process to select the correct homeroom for each educator based on `{name,school}`.

+ Adds a new index with `unique: true` to verify that all homerooms in the database are unique on `{name,school}`, so that there can only be one per combination. 

  + I verified that there are no duplicate homerooms on `{name,school}` in any of our databases, so this is a guard against data quality regression in the future. 

# Deployment

+ The migration should deploy without errors on all of our instances, since there are no homeroom duplicates on `{name,school}` in our instances. 

+ The import process should heal all incorrect homeroom assignments. 

```
# educators_importer.rb

# this line will now select the correct homeroom:
[92] homeroom = Homeroom.find_by(name: row[:homeroom], school: educator.school) 
...
# this line will heal the mis-assigned educator ids:
[97] homeroom.update!(educator: educator) 
```

### DIFF
--- a/app/models/educator.rb
+++ b/app/models/educator.rb
@@ -70,11 +70,6 @@ class Educator < ActiveRecord::Base
     educator_labels.map(&:label_key)
   end
 
-  def default_homeroom
-    return homeroom if homeroom.present?
-    raise Exceptions::NoAssignedHomeroom
-  end
-
   def default_section
     return sections[0] if sections.present?
     raise Exceptions::NoAssignedSections
--- a/spec/importers/file_importers/students_importer_spec.rb
+++ b/spec/importers/file_importers/students_importer_spec.rb
@@ -19,11 +19,6 @@ RSpec.describe StudentsImporter do
       let!(:brown) { School.create(local_id: 'BRN') }
 
       context 'no existing students in database' do
-
-        it 'creates Student records' do
-          expect { import }.to change { Student.count }.by(4)
-        end
-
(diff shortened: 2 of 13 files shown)


app/models/educator.rb
```diff
@@ -70,11 +70,6 @@ def labels
     educator_labels.map(&:label_key)
   end
 
-  def default_homeroom
```
alexsoble: This method raised an error in the test suite. When I searched for implementations of `:default_homeroom` in our app using Sublime, nothing turned up. Removing it didn't affect any other areas of the test suite. So I think this is unused and safe to remove.

spec/importers/file_importers/students_importer_spec.rb
```diff
@@ -19,11 +19,6 @@
       let!(:brown) { School.create(local_id: 'BRN') }
 
       context 'no existing students in database' do
-
-        it 'creates Student records' do
```
alexsoble: This is duplicative, same thing tested below.

alexsoble:
@kevinrobinson: This is the first PR improving how homerooms are handled in the import process. This fixes the immediate issue with homeroom mis-assignment to the wrong educators. I tested both the migration and the import code locally and saw that it healed the bug; see "Deployment" above. I'm going to 🚢 this one as a selfie.

PR #1995, opened 2018-08-21
HS import: Improve logging for StudentSectionAssignmentsImporter

kevinrobinson:
# Who is this PR for?
developers

`StudentSectionAssignmentsImporter` doesn't log about the changes it makes, so when there are problems reported about student sections, we have no visibility into what's happening each night.  The join tables don't have timestamps so we can't see when records have changed either.

# What does this PR do?
Updates `StudentSectionAssignmentsImporter` to add in more counting and logging around different failure cases.  Also updates it to use `RecordSyncer` for deleting records, and inlines `StudentSectionAssignmentRow` into the importer class.  Updates the specs to add an integration test testing the whole importer, and updates the existing unit tests to work as well.

Adds validations and foreign key constraints to `student_section_assignments` fields.  Adds timestamps to the `student_section_assignments` table, and the `data_migrations:timestamps_for_section_join_tables` rake task to fill these in with initiation values after running the migration.

# Checklists


kevinrobinson:
selfie

kevinrobinson:
On first run, this looks good.  The nil records likely are not new but also not a priority, I'll add them to the backlog.

```
StudentSectionAssignmentsImporter: RecordSyncer stats: {
  :passed_nil_record_count=>90,
  :invalid_rows_count=>0,
  :unchanged_rows_count=>18740,
  :updated_rows_count=>0,
  :created_rows_count=>1463,
  :marked_ids_count=>20203,
  :destroyed_records_count=>245
}
```

kevinrobinson:
Confirmed it's idempotent too.

Issue #1990, opened 2018-08-20
Profile v3: New page, emphasizing best light and improving layout


kevinrobinson:
This is an umbrella issue for revisions to the profile page for the start of the 2018 school year.

kevinrobinson:
done a long time ago!

PR #1997, opened 2018-08-21
Add feature switch for v3 profile page

kevinrobinson:

# Who is this PR for?
developers, educators (NB training on 8/23 specifically)

The profile v3 page isn't enabled yet, and there's no way to enable it.

# What does this PR do?
Adds an env variable for forcing this to be enabled, keeping `please` as a magic querystring param that allows the v2 profile page to still be accessed.


kevinrobinson:
selfie

PR #2000, opened 2018-08-23
Change default sort order for sections to order by term enrollment

kevinrobinson:
# Who is this PR for?
HS educators

The section enrollment panel doesn't visualize the schedule (which would be great and parity with what it looks like in Aspen directly), but it isn't even sorted by term.

# What does this PR do?
Sort enrollment by term.  Also adds a story and test for this component.

Separately, update the `TakeNotes` story for two districts since it had broken.

from the story:
screen shot 2018-08-23 at 12 27 33 pm

# Checklists


kevinrobinson:
selfie

PR #2002, opened 2018-08-23
Revise SessionRenewal to reload and re-use existing guards

kevinrobinson:
# Who is this PR for?
educators

The semantics in https://github.com/studentinsights/studentinsights/pull/1980 are a bit confusing, and by forcing a signout in each browser process, this can lead to confusing behavior with multiple open tabs.

This will also help clarify what's happening in testing the behavior of Devise::Timeoutable.

#  What does this PR do?
Changes the forced sign out to a reload, which accomplishes the same effect if the server session has expired (ie, clearing the screen of student data, clearly showing actions on the current would not be successful).  If the server session is still active, this reloads the page.

# Checklists


kevinrobinson:
selfie

PR #2006, opened 2018-08-24
Tweak student photo height in Homeroom Table

alexsoble:
# Who is this PR for?

K8 homeroom teachers!

# What does this PR do?

Setting `height` and `width` to the same value creates a squished square photo. Also 40px tall thumbnail is a bit too small to see the student well.


alexsoble:
selfie

PR #2011, opened 2018-08-24
Refactor to consolidate start of school year code

kevinrobinson:
# Who is this PR for?
developers

There's different implementations of the start of the school year, and I want one to be able to use for capping the query on the "high absences" insight box.

# What does this PR do?
Consolidates this into `school_year.rb`, mirroring `schoolYear.js`.  Also isolates some code related to school years that is only used for the PDF profile page and we don't want to be used elsewhere.

# Checklists


kevinrobinson:
selfie

PR #2013, opened 2018-08-24
Update home page to only show high absences in current school year

kevinrobinson:
# Who is this PR for?
educators

The home page suggests students with absences because they missed several days or withdrew from summer school.  This seems like it could be surprising on the first day of school for staff before students have started.

# What does this PR do?
Caps the query for high absences if looking back 45 days would cross the school year boundary.  Doesn't update the UI code to reflect this (so it still looks like it's querying back 45 days).

Taking advantage of https://github.com/studentinsights/studentinsights/pull/2011.


kevinrobinson:
selfie

PR #2015, opened 2018-08-26
Remove call to measure page load time in App

kevinrobinson:
# Who is this PR for?
developers

# What does this PR do?
Removes console noise; no one is actively using this.


kevinrobinson:
selfie

PR #2020, opened 2018-08-26
Update column widths for MySectionsPage for longer educator names

kevinrobinson:
# Who is this PR for?
HS educators

longer educator names get wrapped and then vertically truncated

# What does this PR do?
fixes it


kevinrobinson:
selfie

PR #2021, opened 2018-08-27
Profile v3: Minimal queue about insight box to start the year

kevinrobinson:
# Who is this PR for?
educators, primarily K8

The insight box doesn't work yet, and rather than making it work, it'd be better to avoid putting prominent "asks" in the product without an expectation or support for them used.

# What does this PR do?
This rolls back the "share an insight" box, which communicated intent about design previously, and for now only shows a "coming soon" kind of box if there are no insights for the student.

screen shot 2018-08-27 at 8 42 45 am

# Checklists
+ [x] Author checked latest in IE - Student profile v3


kevinrobinson:
selfie

PR #2022, opened 2018-08-27
Fix bug with looking up Homeroom when ambiguous param given

kevinrobinson:
# Who is this PR for?
educators

When linking to a homeroom by id (eg, from the profile page), the page doesn't load.  This is because the lookup process with `Homeroom.friendly.find(id_or_slug)` doesn't match as expected with the string version of an id.  As a guess, it looks first on the name field and so if any homeroom name matches the string version of an id, it will match that first rather than matching by id.  Alternately, perhaps it's doing a stricter lookup than `Homeroom.find` and failing to match on a string version of the id, the way that `#find` would.

# What does this PR do?
Fixes it by making explicit the lookup order, and removing the use of `friendly` to make queries.


kevinrobinson:
selfie

PR #2024, opened 2018-08-27
Fix typo

alexsoble:

"Somervile" > "Somerville" in profile Insights box.


alexsoble:
Already fixed.

PR #2026, opened 2018-08-27
Remove destroying empty homerooms

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2023.

# Who is this PR for?
developers, I think

I noticed this validating https://github.com/studentinsights/studentinsights/pull/2023.  I don't think there's any harm in empty Homerooms, and having this extra step complicate the semantics of "Insights says what the SIS says."  If we do want to add back in that we delete Homerooms like this, we should do it in the importer sync code, not elsewhere as a separate step of `ImportTask`.

# What does this PR do?
Removes this.

### DIFF
--- a/app/importers/import_task.rb
+++ b/app/importers/import_task.rb
@@ -173,9 +173,6 @@ class ImportTask
     begin
       log('Calling Student.update_recent_student_assessments...')
       Student.update_recent_student_assessments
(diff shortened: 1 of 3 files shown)


kevinrobinson:
selfie

PR #2028, opened 2018-08-28
Allow forcibly enabling class lists for particular educator_ids, for internal team

kevinrobinson:
# Who is this PR for?
developers, folks at WH potentially

Class lists are disabled, but folks want to recover some data.

# What does this PR do?
Let's individual users access class lists through an `EducatorLabel`.


kevinrobinson:
selfie

PR #2029, opened 2018-08-28
Add in text rendering for class lists, with EducatorLabel authorization

kevinrobinson:
# Who is this PR for?
WH educators

Class lists have different authorization rules that mean access doesn't stay static over time, particularly after the Aspen rollover and new homeroom and grade assignments for educators and students.  Some folks at WH want to access these lists, but we've turned off the feature.

# What does this PR do?
Uses the `EducatorLabel` in https://github.com/studentinsights/studentinsights/pull/2028 to grant access to an endpoint that renders class lists as text.  This is simpler than reworking the authorization rules, and reworking the UI code to work for historical class lists as well as for prospective class lists.  Those changes would be quite involved, and folks just want to see their class lists to do their work now, so this gives them the data in a more minimal format, without requiring moving student data through other systems.

This also provides a good way for developers to compare the result of the existing endpoints and UI if we want to extend those to work with historical lists as well, which I assume we will want to do in October or later this fall.

# Checklists

### DIFF
--- a/app/controllers/class_lists_controller.rb
+++ b/app/controllers/class_lists_controller.rb
@@ -44,6 +44,17 @@ class ClassListsController < ApplicationController
     render json: Herfindahl.new.with_dimensions_json(workspaces)
   end
 
+  # For getting access to class list data after the year has passed,
+  # until we rework the authorization to work at all times of the year.
+  def text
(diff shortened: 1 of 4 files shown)


kevinrobinson:
selfie

Issue #1939, opened 2018-07-24
Improve Homeroom and Section roster pages


kevinrobinson:
Revisions to K8 homeroom and HS roster for start of 2018-2019 school year.  Here's some starting points:
- add photos for K8 only
- keep IEP, 504, ELL remove other data by default
- supporting single "show data" option or "download with data"
- MCAS scores need dates
- add sorting carets to columns
- sorting doesn't work for all columns

kevinrobinson:
Some related notes:
- `HomeroomTable` and `SectionView` end up doing much of the same work, but with different code.  This is fine in general, but for differences across districts doubles the complexity.
- `SectionPage` ends up overflowing horizontal views and we should address this.
- Change "level of need" to just show presence of IEP or not.  Consider adding 504 and ELL status?
- Update visual style of "last note" to match levels page, counselor meeting page, and school dashboards (eg, blue badge like in https://github.com/studentinsights/studentinsights/issues/2643).

PR #2032, opened 2018-08-28
Add school, educator and student grade levels to Homeroom page

kevinrobinson:
# Who is this PR for?
K8 educators

Data quality issues around homeroom assignments and grade levels are a bit hidden.  Related to https://github.com/studentinsights/studentinsights/issues/1939.

# What does this PR do?
Adds in more info to the homeroom page, adjusts layout.

### before
screen shot 2018-08-28 at 12 04 52 pm

### after
screen shot 2018-08-28 at 11 58 40 am

# Checklists
+ [x] Author checked latest in IE - Homeroom page


kevinrobinson:
This PR also removes the link to the school page, and does the same for the profile v3 page.  Only some users can access these right now.  Later, we should update the school roster page to use the core authorizer code, and then we can allow anyone to use those pages (scoped to which students they have access to).  Or we could rework those pages to be more meaningful (sort of a day-to-day overview of the school using pieces of the home page, feed and my students that we have now).  But all out of scope for now, so removing the links.

kevinrobinson:
selfie

PR #2033, opened 2018-08-28
Switch class lists UI to text, allowing users to explore more

kevinrobinson:
# Who is this PR for?
WH educators

They're not finding what they expect for class lists; communicating about this might be hard without the visual.

# What does this PR do?
Changes links on the `/classlists` page to view the text version of lists rather than the full UI (which doesn't work for lists retrospectively to past years).


kevinrobinson:
selfie

Issue #2034, opened 2018-08-28
Investigate Educator records with local_id: nil


kevinrobinson:
In Somerville:
```
 Educator.where(local_id: nil).size
=> 293
```

kevinrobinson:
More broadly, `EducatorsImporter` treats `login_name` as the primary key, but other importers like `EducatorSectionAssignmentRow` use `local_id` for lookups.

`local_id` seems like it perhaps should be the actual primary key, and either way the `Educator` table needs constraints on it and `EducatorsImporter` needs to be paying attention to this.

I didn't look at New Bedford yet.

kevinrobinson:
Related, for Somerville HS educators only, from some email discussion:

> We can correctly import someone as an educator and they can sign in to Insights without a local_id.  But in other places like the section assignments, the export we receive uses local_ids rather than educator login ids, which is the specific gap here.  Here's an example row from educator_section_assignments.txt that is exported:

```
"local_id","course_number","school_local_id","section_number","term_local_id"
\N,"011","SHS","011-007","9"
```

> The problem here is we can't tell in Insights who the educator is since local_id is null.  So for that person they can sign into Insights, but can't see any of their sections.

> Looking forward, if we could can change the educator section export SQL to also include the loginID along with the local_id, then we could update Student Insights to work even when the educator doesn't have an local_id yet.

kevinrobinson:
With https://github.com/studentinsights/studentinsights/pull/2036, nil `local_id` values for educators are no longer meaningful.  We can keep importing these for debugging purposes when triangulating with the SIS, but this field has no impact on the product anymore.  https://github.com/studentinsights/studentinsights/pull/2039 tags these as deprecated in the SQL export.

PR #2035, opened 2018-08-28
Add login_name to EducatorSectionAssignment export SQL

kevinrobinson:
Addressing part of https://github.com/studentinsights/studentinsights/issues/2034.

# Who is this PR for?
HS teachers

This will help remove some friction for new staff, who get an Aspen record and `login_name` before a `local_id`, which is a slower process done through HR.

# What does this PR do?
Includes `login_name` as well, so we can cut over to that after confirming we get it and looking at how it changes things.


kevinrobinson:
Also update the years for each script.  Per John these should be 2019 (when running them on 8/28/2018).

kevinrobinson:
selfie

PR #2036, opened 2018-08-28
Update EducatorSectionAssignmentsImporter to use login_name instead of local_id

kevinrobinson:
Part of https://github.com/studentinsights/studentinsights/issues/2034.

# Who is this PR for?
HS educators

In Somerville, the organizational that assigns educators a `local_id` in the SIS is slower than the one that assigns them a `login_name`.  This can mean educators start working but don't yet have a `local_id` for a bit of time, which effectively blocks them from accessing students in their sections within Insights.

# What does this PR do?
Changes the `EducatorSectionAssignmentsImporter` to use `login_name` instead of `local_id`, which is now available thanks to John running https://github.com/studentinsights/studentinsights/pull/2035.  In making this update, it also adds in logging and counting stats, and uses `RecordSyncer` to do the syncing and logging.

Also adds non-null constraints on the join table and presence validations.  There are no violations in production, just upping the guards on it.

# Checklists


kevinrobinson:
Still need to update the tests here.

kevinrobinson:
selfie

PR #2004, opened 2018-08-24
Profile v3: Allow clicking to view restricted note content inline, allow creating restricted notes inline

kevinrobinson:
Part of https://github.com/studentinsights/studentinsights/issues/1998 and https://github.com/studentinsights/studentinsights/issues/1990.

# Who is this PR for?
educators with access to restricted notes

# What problem does this PR fix?
https://github.com/studentinsights/studentinsights/pull/1999 moved the presence of restricted notes inline, but didn't yet allow educators with access to view the contents of restricted notes.  It also didn't yet allow educators to create new restricted notes.

# What does this PR do?
### viewing
Updates `<RestrictedNotesPresence />` to allow clicking to see the content of the restricted note.  Adds a new endpoint to allow fetching this.

Relatedly, refactors endpoint URLs and removes `event_note_attachments_controller` to consolidate into `event_notes_controller`.

Also updates copy on `<RestrictedNotesPresence />` to soften it.

### writing
Updates `<TakeNotes />` to support `showRestrictedCheckbox` prop, and updates call sites to only use it in the new profile page.  Relatedly, updates the param from `is_restricted` to `isRestricted` since the others are all camel case (we should remove this indirection, but for now this just removes the inconsistency).

# Screenshot (if adding a client-side feature)
### without access: presence inline only
screen shot 2018-08-24 at 12 17 54 pm

### with access: presence inline, able to click and view 
screen shot 2018-08-24 at 12 16 57 pm
screen shot 2018-08-24 at 12 16 52 pm

### with access: my notes page
screen shot 2018-08-24 at 12 54 00 pm

### with access: profile v3, writing
screen shot 2018-08-24 at 1 38 28 pm
screen shot 2018-08-24 at 1 38 33 pm
screen shot 2018-08-24 at 1 38 36 pm

### without access: profile v3 (no checkbox)
screen shot 2018-08-24 at 1 47 01 pm

### with access: profile v2 (unchanged)
screen shot 2018-08-24 at 1 38 45 pm

### with access: restricted notes page v2 (unchanged)
screen shot 2018-08-24 at 1 38 56 pm

# Checklists
+ [x] Author checked latest in IE - Student Profile v2
+ [x] Author checked latest in IE - Restricted notes v2
+ [x] Author checked latest in IE - Student Profile v3
+ [x] Author checked latest in IE - My notes
+ [x] Author improved specs for code in need of better test coverage
+ [x] Author included specs for new code

PR #2037, opened 2018-08-28
Limit use of jQuery methods

kevinrobinson:
# Who is this PR for?
delelopers

Part of https://github.com/studentinsights/studentinsights/issues/1758, which I updated to see what would be involved to migrate major versions.  This does one of the simplest changes just to make some progress today.

# What does this PR do?
Limits use of jQuery APIs (eg, remove `$.ajax` and implicit `jquery_ujs` patch, `$.Deferred()`).

Separately uses strong params in an endpoint that I needed to read through to understand what JS parameters were necessary when migrating to `fetch`.

# Checklists
+ [x] Author checked latest in IE - Student Profile v2
+ [x] Author checked latest in IE - Student Report PDF
+ [x] Author checked latest in IE - Service Uploads


kevinrobinson:
This needs manual QA still.

kevinrobinson:
Did some more manual testing; found two bugs.

One was an existing JS bug in deleting attachments from updating the server route but not the JS path in https://github.com/studentinsights/studentinsights/pull/2004/files#diff-99522d9297d5278a7368499381535d9fR35.  That is fixed in this PR and has a test now too.

The other is with this change, where discontinuing services would use the new `apiDeleteJson` function, which didn't parse the response's JSON.  That's fixed in this PR and this PR adds tests to each of those shared `apiFetchJson` functions.

kevinrobinson:
https://github.com/studentinsights/studentinsights/pull/2037/commits/a28c61c7d462f885900096aafc9b776326c87203 fixes an unrelated existing issue in development; the includes wouldn't include a lib class needed for the PDF download.  This didn't impact production.

kevinrobinson:
selfie

PR #2039, opened 2018-08-28
Add deprecation comments for educator local_id in SQL exports

kevinrobinson:
A note to developers about https://github.com/studentinsights/studentinsights/issues/2034.


kevinrobinson:
selfie

PR #2040, opened 2018-08-29
Use inclusive date ranges when querying for students with absences and in dashboards

kevinrobinson:
# Who is this PR for?
educators

The date range for these queries is exclusive and not inclusive.  In the profile v2 UI this was the case as well, and so the problem never surfaced.  Profile v3 fixed this to be inclusive and made this more visible by using the same "last 45 days" time window, which wasn't visible in the profile page before.

# What does this PR do?
Updates `InsightStudentsWithHighAbsences` to query with an inclusive time range, and updates `DashboardQueries` to do the same (at the response level, not the database query level).

### DIFF
--- a/app/lib/dashboard_queries.rb
+++ b/app/lib/dashboard_queries.rb
@@ -1,7 +1,8 @@
 class DashboardQueries
-  def initialize(educator)
+  def initialize(educator, options = {})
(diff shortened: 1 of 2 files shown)


kevinrobinson:
selfie

PR #2042, opened 2018-08-29
Fix bug in getFirstDateIndex

kevinrobinson:
# Who is this PR for?
admin

# What problem does this PR fix?
The counts in the absence dashboard are still off after https://github.com/studentinsights/studentinsights/pull/2040.

# What does this PR do?
Fixes a JS bug that led to filtering out the first day of school.

# Checklists
+ [x] Author included specs for new code

PR #2043, opened 2018-08-29
Update values for JS integration tests for absence dashboard after fixing date range bug

kevinrobinson:
Updates the expected values for JS integration tests after https://github.com/studentinsights/studentinsights/pull/2042.


kevinrobinson:
selfie

PR #2044, opened 2018-08-29
Limit data for absences dashboard

kevinrobinson:
# Who is this PR for?
admin

Initially this was an attempt at reducing the date range for the query, but we also want to speed this up for larger schools.

# What does this PR do?
Reworks the querying so there are two database queries, and that these are minimal for the data needed (eg, no old data that will be filtered out in JS, remove older `event_notes`).

# Checklists

### DIFF
--- a/app/lib/dashboard_queries.rb
+++ b/app/lib/dashboard_queries.rb
@@ -6,20 +6,51 @@ class DashboardQueries
   end
 
   def absence_dashboard_data(school)
-    student_absence_data = authorized_students_for_dashboard(school) do |students_relation|
-      students_relation.includes([homeroom: :educator], :dashboard_absences, :event_notes)
+    # authorization
+    students = authorized_students_for_dashboard(school) do |students_relation|
+      students_relation.includes([homeroom: :educator])
     end
-    students_with_events = student_absence_data.map do |student|
-      individual_student_absence_data(student)
+
+    # query for absences
+    absences = Absence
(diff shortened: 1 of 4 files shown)


kevinrobinson:
This could use a `PerfTest` to measure the impact on the larger New Bedford schools or SHS before shipping.  cc @alexsoble if this is helpful in the short term, feel free to grab this and do the perf test or to just ship it and see, otherwise I can finish this later this week.

kevinrobinson:
Here's some data on this.  The results are pretty variable, so here's I've included the values that show the most conservative impact:

### 
```
Somerville, using SHS as a test case, and running the PerfTest across the first 10 districtwide users:
  master:        median: 12075ms   p95: 13628ms
  this PR:       median:  4397ms   p95:  5379ms

New Bedford, using Normandin Middle as a test case, and running the PerfTest across the first 5 districtwide users:
  master:        median: 21777ms   p95: 22605ms
  this PR:       median:  7857ms   p95: 11016ms
```

Other test runs were faster/slower, but either way this should help significantly.  It doesn't impact the bytes over the wire, and verified this by looking at the byte size before and after this diff on both districts.

kevinrobinson:
selfie

kevinrobinson:
Confirmed this loads in under 5 seconds for SHS and Normandin in production:

screen shot 2018-09-10 at 1 46 09 pm

screen shot 2018-09-10 at 1 46 21 pm

PR #2045, opened 2018-08-29
WIP: Make Student Profile quote default text change per district 

alexsoble:
# Who is this PR for?

New Bedford Student Profile users.

All student profiles mention Somerville High School, could be confusing to folks outside Somerville. 

# What does this PR do?

Changes the text to say simply, "This is being piloted at another school" for New Bedford, Bedford, and any other school district besides Somerville. 

# Checklist 

+ [ ] Student has real transition note (unaffected) 

+ [ ] Student is in Somerville (unaffected)

+ [ ] Student is in New Bedford (new text)

### DIFF
--- a/app/assets/javascripts/helpers/PerDistrict.js
+++ b/app/assets/javascripts/helpers/PerDistrict.js
@@ -186,3 +186,11 @@ export function studentTableEventNoteTypeIds(districtKey, schoolType) {
   if (isSomervilleOrDemo) return [300, 301];  // Includes elementary/middle, Capuano early childhood,
                                               // and SPED.
 }
+
+export function studentProfileQuoteUpsell(districtKey) {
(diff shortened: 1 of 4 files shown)


kevinrobinson:
Closed for https://github.com/studentinsights/studentinsights/pull/2049.

@alexsoble The underlying issue related to showing this differently across districts hasn't been resolved, but right now the copy is clearer and sort of tries to upsell this to folks in other districts, or get them to ask more (rather than making it generic).  We can update this and if you have ideas of what this should look like to provide more context, that'd be great!  For this week my priority is still getting this out at SHS, and tolerating if that isn't perfect yet for Somerville K8 or other districts, but I'd also love to figure out how to have little inline upsells like this for various bits that can shine light on pilot work that certain districts are doing to ones that aren't doing that yet.

PR #2047, opened 2018-08-30
Profile v3: Mute colors for services

kevinrobinson:

# Who is this PR for?
educators

Services had an old color scheme that I don't think ever meant much to anyone, and it was only partially implemented for some services.  With the profile v3 changes, the semantic meaning of the colors don't match the meaning elsewhere on profile v3, and these aren't in the same palette either.

# What does this PR do?
Mutes the colors for services to gray.

screen shot 2018-08-30 at 3 51 59 pm
screen shot 2018-08-30 at 3 51 42 pm

# Checklists


kevinrobinson:
selfie

PR #2051, opened 2018-09-04
Filter out long hyphens when parsing transition note

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2049.

# Who is this PR for?
HS students, educators

In https://github.com/studentinsights/studentinsights/pull/2049, I updated the regex parsing in reaction a Robocop warning but did this wrong; so long hyphens aren't filtered out in parsing anymore.

# What does this PR do?
Fixes the regex, adds a test case.

# Checklists

### DIFF
--- a/app/lib/transition_note_parser.rb
+++ b/app/lib/transition_note_parser.rb
@@ -35,6 +35,6 @@ class TransitionNoteParser
   end
 
   def clean(text)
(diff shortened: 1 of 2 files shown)


kevinrobinson:
selfie

PR #2052, opened 2018-09-04
Fix JS transition note parsing to trim after replacing

kevinrobinson:
Follow on to https://github.com/studentinsights/studentinsights/pull/2051

# Who is this PR for?
HS students, educators

newlines mixed in with separators.

# Checklists


kevinrobinson:
selfie

PR #2054, opened 2018-09-05
Import historical STAR Math and Reading data for New Bedford

alexsoble:
# Notes 

+ As we expand to more schools, we need to backfill historical STAR data for those schools
+ @kevinrobinson, see email to you and Uri about this and how we might rethink our STAR import strategy to be more robust against this

# Deployment 

+ Merge, run import process to backfill historical STAR data (either for all students/schools or for students at our target schools)
+ Revert after import has run successfully so that we go back to tracking the normal STAR folder and filenames


alexsoble:
Running a one-off import job for New Bedford STAR now, with school scope limited to our target schools.

PR #2056, opened 2018-09-05
Update LDAP strategy to grab timeout from env

kevinrobinson:
# Who is this PR for?
developers, so this is more easily tunable.  also bumps to 30 seconds to rule out this in connectivity issues.

I updated the env for Somerville and New Bedford (but not Bedford, since it's not using this setup yet cc @alexsoble).


kevinrobinson:
talked about this with @alexsoble, mergin'

PR #2059, opened 2018-09-06
Allow STAR Math/Reading importers to skip and log invalid rows

alexsoble:
# Who is this PR for?

New Bedford STAR historical data backfill.

One of the rows in the historical STAR Math file contains a percentile rank of zero, which we considered invalid. The STAR Math and Reading importers are currently configured to blow up if they run across any invalid data, since we've never encountered invalid STAR data before in either district. 

# What does this PR do?

Instead of blowing up the STAR importer if a row is invalid, log the errors and continue importing rows.

### DIFF
--- a/app/importers/file_importers/star_math_importer.rb
+++ b/app/importers/file_importers/star_math_importer.rb
@@ -3,6 +3,7 @@ class StarMathImporter
   def initialize(options:)
     @school_scope = options.fetch(:school_scope)
     @log = options.fetch(:log)
+    @invalid_rows_count = 0
   end
 
   def import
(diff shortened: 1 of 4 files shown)


alexsoble:
I ran a local import task and verified that this successfully skips STAR rows with "zero" for percentile rank and does not blow up. I didn't run it all the way to the end since it's more than an hour (big file). My plan is to 🚢 this now and kick off a new import task in the New Bedford production instance.

alexsoble:
Counts before shipping and kicking off import task, for reference:

```
# New Bedford production

> StarMathResult.count
=> 24409
> StarReadingResult.count
=> 52262
```

kevinrobinson:
FWIW the were I think 4 records that were 0 when grepping the file itself, so hopefully that will line up with this awesome importer logging!

alexsoble:
Counts after shipping and kicking off import task:

```
> StarMathResult.count
=> 49461
> StarReadingResult.count
=> 52263

PR #2060, opened 2018-09-06
Remove references to S3 region, this is set in ENV

kevinrobinson:
# Who is this PR for?
developers

These regions aren't accurate any more, and in other places this is controlled by an ENV variable.

# What does this PR do?
Removes these, let ENV control this.


alexsoble:
👍  good improvement

kevinrobinson:
ah this will fix https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/209/

PR #2061, opened 2018-09-06
Update photo importer to move zip path into config

kevinrobinson:
This was broken in making the new SFTP box and we didn't notice since this job doesn't run regularly.  In migrating the S3 storage, I ran it to verify it still works which revealed this.

Resolves https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/210/


kevinrobinson:
selfie

Issue #2005, opened 2018-08-24
Fix LoginActivity page


kevinrobinson:
This looks like it might have regressed with the lodash upgrade in https://github.com/studentinsights/studentinsights/pull/1964:

<img width="1404" alt="screen shot 2018-08-24 at 3 37 43 pm" src="https://user-images.githubusercontent.com/1056957/44604376-b73ccd80-a7b3-11e8-826b-577716a2b7b0.png">

kevinrobinson:
Fixed by https://github.com/studentinsights/studentinsights/pull/2063

PR #2063, opened 2018-09-07
Update LoginActivity to remove old lodash use

kevinrobinson:
# Who is this PR for?
project team

https://github.com/studentinsights/studentinsights/issues/2005

# What does this PR do?
Removes use of old lodash function that's no longer supported.

screen shot 2018-09-07 at 6 17 11 am

# Checklists
+ [x] Author checked latest in IE - Login Activity


app/assets/javascripts/login_activity/LoginActivityPage.test.js
```diff
@@ -1,18 +1,41 @@
...
 import fetchMock from 'fetch-mock/es5/client';
+import LoginActivityPage from './LoginActivityPage';
+import loginActivityJson from './loginActivityJson.fixture';
 
-it('renders without crashing', () => {
+function mockFetch(props = {}) {
   fetchMock.restore();
-  fetchMock.get(
-    '/api/login_activity?created_at_or_before=1532105716&created_after=1529513716',
-    '[]'
```
kevinrobinson: Without any data in the fixture here, most of the code in the component wasn't getting run in the test before, which is why the test didn't catch the bug.  So here I grabbed the response from my local development environment, and then have it use that which is a fast way to get a rough smoke test.

app/assets/javascripts/login_activity/LoginActivityPage.test.js
```diff
@@ -1,18 +1,41 @@
...
+  mockFetch(props);
   const el = document.createElement('div');
-  ReactDOM.render(
-    <LoginActivityPage
-      queryStartTimestamp={'1529513716'}
-      queryEndTimestamp={'1532105716'} />, el);
+  ReactDOM.render(<LoginActivityPage {...props} />, el);
+});
+
+it('renders after fetching', done => {
```
kevinrobinson: Since the component fetches data, that promise doesn't resolve synchronously and so by the time the first test finishes, none of the rendering code has run.  This second test here is an async test that uses `done`, where we do the same thing, but then wait for the promise resolve and the rendering to happen before making assertions and telling Jest the test is done.

The setTimeout value is`0` milliseconds, which tells the JS engine to let any queued up work on the event loop go first (ie, for the mock fetch that we setup to immediately resolve to the fixture, run any work that's waiting on that).  So by the time the function in setTimeout is run, the rendering has already happened and we can make assertions, etc. and then tell Jest the test is done.

kevinrobinson:
@alexsoble let's use `<LoginActivity />` to look at usage this afternoon!

kevinrobinson:
selfie

PR #2066, opened 2018-09-07
Update Somerville DIBELS rules with "Benchmark" => "CORE"

alexsoble:
# Who is this PR for?

Somerville educators.

Didn't realize earlier that DIBELS benchmark value of "Benchmark" means "CORE".


alexsoble:
selfie

PR #2068, opened 2018-09-10
WIP: Separate educator login from email, handle Bedford auth system

alexsoble:
# Who is this PR for?

Bedford Public Schools.

New Bedford and Somerville allow for querying LDAP by educator email. Bedford uses a different system which requires a separate login name or "short username". 

# What does this PR do?

Splits `email` from `login_name` in the educators table. 

+ Adds a migration that creates `login_name` as a field, auto-fills all New Bedford and Somerville educators to have login names equal to their current emails, then applies `null: false` migration.
  + This means that we can cut over to authenticating on `login_name` instead of `email` without changing anything for Somerville or New Bedford educators.
+ Updates import code to automatically set login name and email as the same for New Bedford and Somerville educators, while letting the two be different for Bedford using `PerDistrict`.

# Blockers

Still waiting on Bedford to send an example educators export CSV that will show me how to set educator emails for Bedford. 

# Deploy 

+ [ ] Merge and ship 
+ [ ] Watch the outcome of the `AddLoginToEducatorsTable` migration
+ [ ] Manually check that all New Bedford and Somerville educators have login names set

### DIFF
--- a/app/importers/rows/educator_row.rb
+++ b/app/importers/rows/educator_row.rb
@@ -8,9 +8,12 @@ class EducatorRow < Struct.new(:row, :school_ids_dictionary)
   def build
     return nil if row[:login_name].nil? || row[:login_name] == ''
 
+    # TODO: find_or_initialize on login_name once all active educators
+    # have a login_name set.
     educator = Educator.find_or_initialize_by(email: email)
 
     educator.assign_attributes(
+      login_name: row[:login_name],
       state_id: row[:state_id],
       full_name: row[:full_name],
       staff_type: row[:staff_type],
@@ -32,7 +35,7 @@ class EducatorRow < Struct.new(:row, :school_ids_dictionary)
   private
 
   def email
-    PerDistrict.new.from_import_login_name_to_email(row[:login_name])
+    PerDistrict.new.from_import_login_name_to_email(row[:login_name], row[:full_name])
   end
 
   def is_admin?
--- a/app/importers/rows/educator_row.rb
+++ b/app/importers/rows/educator_row.rb
@@ -8,12 +8,13 @@ class EducatorRow < Struct.new(:row, :school_ids_dictionary)
   def build
     return nil if row[:login_name].nil? || row[:login_name] == ''
 
-    # TODO: find_or_initialize on login_name once all active educators
-    # have a login_name set.
-    educator = Educator.find_or_initialize_by(email: email)
+    email = email_from_row
+    login_name = login_name_from_row(email)
+    educator = Educator.find_or_initialize_by(login_name: login_name)
 
     educator.assign_attributes(
-      login_name: row[:login_name],
+      login_name: login_name,
+      email: email,
       state_id: row[:state_id],
       full_name: row[:full_name],
       staff_type: row[:staff_type],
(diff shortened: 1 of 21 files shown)


kevinrobinson:
@alexsoble Here's some feedback like you asked over email.  And to clarify, here's what I understand of the status overall.  The core pieces are 1) importing the new fields for Bedford and storing them in the Insights database, 2) updating the LDAP strategy to do an additional mapping for Bedford.  And we're blocked on shipping part 1 right now because we need Bedford pals to tell us the format of this, and then we also would need them to export the data first.

#### if you think the overall strategy is sound
- This change introduces risk that could lock out users across all districts, so it's great to review and talk it out!  Are there ways we can reduce that risk and deploy this more incrementally?  One option there might be: limit the scope of changes to only the two core pieces (1 import new fields for Bedford, 2 update LDAP strategy for Bedford), and then deploy changes to them independently in separate PRs.
- The change to make `login_name` a new concept for all districts leaks the complexity for Bedford into Somerville and New Bedford as well.  Is there a way to isolate the change to just Bedford?  One option there might be to import the additional fields for Bedford only and leave other districts unchanged (part 1).  And then for part 2, have the LDAP code call `PerDistrict.new.find_educator_from_login_text(login_text)` and have each district use whatever way it needs to map that `login_text` to a user record.

#### if you see any major problems or concerns? 
- There some changes here that seem tough to verify without deploying them (eg, Devise config changes, code that is `PerDistrict`).  Could we approach this more incrementally, or test this in the console, or in local development mode with different `DISTRICT_KEY` values, etc?

@alexsoble Let me know if this is helpful, I'm around all afternoon if it's helpful to talk more, or to grab any pieces that I can help with, just like you so helpfully did a few weeks back with the changes before the start of the school year :)

config/initializers/devise.rb
```diff
@@ -1,6 +1,8 @@
 # Use this hook to configure devise mailer, warden hooks and so forth.
 # Many of these configuration options can be set straight in your model.
 Devise.setup do |config|
+  config.authentication_keys = [:login_name]
```
kevinrobinson: Is this here so it can be called for localized text in `devise.en.yml`, or does it control something else?

db/migrate/20180907185759_add_login_to_educators_table.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    if PerDistrict.new.login_educator_with_email? && Rails.env.production?
+      puts "Backfilling login_name column with emails for #{PerDistrict.new.district_key}!"
+
+      Educator.find_each do |educator, index|
+        educator.login_name = educator.email
+        educator.save!
+
+        puts("  Updated #{index} educators.") if index > 0 && index % 100 == 0
+      end
```
kevinrobinson: If this raises in the middle of a run, could it lock out users that haven't been processed yet from signing in?  Alternately, after the deploy but before this migration, would all users be prevented from signing in?

app/importers/rows/educator_row.rb
```diff
@@ -8,9 +8,13 @@ def self.build(row)
   def build
     return nil if row[:login_name].nil? || row[:login_name] == ''
 
-    educator = Educator.find_or_initialize_by(email: email)
+    email = email_from_row
+    login_name = login_name_from_row(email)
+    educator = Educator.find_or_initialize_by(login_name: login_name)
```
kevinrobinson: This allows a bunch of edge cases for Somerville and New Bedford (where these aren't distinct concepts).  What happens if these values were different?  These also seem like places where tests across different district keys or `PerDistrict` return values would help make sure this works for the different cases (eg, first-time import and create, subsequence import and update).

alexsoble:
Hey Kevin, this is really helpful! I agree that it would be better to ship
this in smaller pieces, and to not touch the critical path for Somerville
and New Bedford. Thanks for those comments and suggestions. I’ll get to
work on that tomorrow AM!

Just so you know, I haven’t heard from Bedford about any data being added
to the SFTP box. And Uri doesn’t want me taking any shortcuts or hacks to
work around that. So all of this is working ahead of the actual data.

alexsoble:
Closing in favor of a more incremental approach. 😸

PR #2071, opened 2018-09-11
WIP: Login strategy and educator import for Bedford

alexsoble:
# Who is this PR for?

Bedford Public Schools.

Bedford does not use email address for sign on.

# What does this PR do?

Creates a new login path for Bedford staff, expressed in `per_district.rb`. This path uses a login name instead of an email for educators. 

Updates the front-end to remove HTML email validation from login input when a district uses something other than email as a login, like Bedford.

Builds off #2068, but designed to change the login path for Somerville and New Bedford educators as little as possible. 

# Verification: local login 

## Somerville

Steps:
1. `rake db:reset && yarn s`
2. Verified correct handling of no password and incorrect password 
3. Log in with correct password and email address:

TK

## Bedford 

Steps:
1. `rake db:migrate:reset`
2. Rails console: `require './spec/support/bedford_pals'; BedfordTestPals.create!`
3. `DISTRICT_KEY=bedford yarn s`
4. Verified correct handling of no password and incorrect password 
5. Log in with login name:

TK

# Verification:  educator import

## Somerville 

Steps:
1. Local import process
2. Verify in Rails console:

```
2.5.1 :001 > Educator.where(email: nil).count
 => 0
2.5.1 :003 > Educator.where.not(login_name: nil).count
 => 0
```

## Bedford 
1. Local import process: `DISTRICT_KEY=bedford thor import:start --source educators`
2. Verify in Rails console:

```
# Both email and login_name are validated for Bedford educators.
# Only email is enforced at the database level.

2.5.1 :001 > Educator.where(login_name: nil).count
 => 0
2.5.1 :003 > Educator.where(email: nil).count 
 => 0
```


app/config_objects/per_district.rb
```diff
@@ -117,15 +129,15 @@ def enable_counselor_based_feed?
...
+  def from_educator_row_to_email(row)
     if @district_key == SOMERVILLE
-      login_name + '@k12.somerville.ma.us'
+      row[:login_name] + '@k12.somerville.ma.us'
     elsif @district_key == NEW_BEDFORD
-      login_name + '@newbedfordschools.org'
+      row[:login_name] + '@newbedfordschools.org'
     elsif @district_key == BEDFORD
-      login_name + '@bedfordps.org'
+      row[:email]
```
alexsoble: This assumes that Bedford is sending us an educators file with an "email" column to their SFTP box.

Will was able to create this file in a one-time export, but the header was named "Email" instead of "email", and he hasn't sent it to the SFTP site through the cron job yet. 

These are small details but they all need to be in place before the Bedford educator import can run successful.

config/initializers/devise.rb
```diff
@@ -12,6 +12,8 @@
   # available as additional gems.
   require 'devise/orm/active_record'
 
+  config.authentication_keys = [ PerDistrict.new.educator_login_field ]
```
alexsoble: This is the change most likely to impact other parts of the system. 

Testing locally, I verified that:
1. If `config.authentication_keys` is set to `[:email, :login_name]`, authentication breaks for all districts.
2. This strategy allows authentication to work for both types of login.

spec/support/bedford_pals.rb
```diff
@@ -0,0 +1,32 @@
+# This class defines a set of users, schools, homerooms and students
+# that can be used for testing Bedford-style authorization paths.
+class BedfordTestPals
```
alexsoble: Minimal for now, meant to facilitate local testing against Bedford-style logins.

app/config_objects/per_district.rb
alexsoble: Update: Will confirmed that the file will be updated in SFTP tomorrow, and that the header will be lowercased.

alexsoble: Another update: I realized this is a blocker for importing educator records, but not a blocker for creating a developer user to test with.

alexsoble:
Overall, I'm satisfied with local development environment testing here, but I'd like to add integration tests to the specs to guard against error introduction and express the Bedford login path in specs. 

Specifically, I tried to add a `sign_in_feature_spec.rb` test with a Bedford context block. I ran into some issues handling ENV. I looked at this Thoughtbot essay as a guide: https://robots.thoughtbot.com/testing-and-environment-variables.

The "Override environment variables on a per-test basis" strategy didn't work for me, I got test pollution even when I restored the ENV variable at the end of the spec. 

The "Stub ENV" strategy didn't work well either.

When I tried to stub ENV like this:

```
allow(ENV).to receive(:[]).with('DISTRICT_KEY').and_return('bedford')
```

 I got errors like these:

```
expected: ("DISTRICT_KEY") 
  got: ("GEM_HOME")
Please stub a default value first if message might be received with other args as well.
```

I was looking at `should_use_mock_ldap_spec.rb` as an example here, since it stubs ENV all over the place, but I realized that stubbing ENV is different in a unit test versus an integration test. An integration test is going to make many, many more calls to ENV, which will require an elaborate and probably impractical amount of stubbing. 

I'm doing to do a little more research here, specifically around how I might stub ENV in a feature spec, or otherwise handle this differently. 

@kevinrobinson if you have any good ideas for how we might integration test login for two different districts in one test suite, let me know!

alexsoble:
Oooh, I'm going to look at `stub_const`: https://relishapp.com/rspec/rspec-mocks/v/2-11/docs/stubbing-constants

alexsoble:
This worked for solving the ENV stubbing issues:

```
stub_const('ENV', ENV.to_hash.merge('DISTRICT_KEY' => 'bedford'))
```

But now I'm looking at a new issue: the Rails app that Capybara's is testing against has been loaded up with `DISTRICT_KEY=somerville` default settings, so the Bedford path code fails. Looking at ways to stop/restart Capybara's underlying Rails app to model the Bedford app instance.

app/views/layouts/application.html.erb
```diff
@@ -22,23 +22,17 @@
...
-              <div class="sign-in-item">
-                <%= f.submit "Log in", class: "btn btn-primary" %>
-              </div>
-            <% end %>
-          </div>
+          <%= render partial: 'shared/navbar_signed_out', locals: {
+            login_field_label: login_field_label,
+            login_field_type: login_field_type,
+            login_field: login_field,
+            authenticity_token: session[:_csrf_token],
```
alexsoble: Reading up on CSRF tokens in Rails 5 to make sure I'm doing this right.

alexsoble:
@kevinrobinson Update: I implemented the `form_tag` helper you suggested, which worked great! At this point the specs were passing and local functionality looked good.

Next, I started implementing a new attribute name in the login form like we discussed, since we're not tied to using model-backed `:email` or `:login_name` and have the flexibility to choose a different attribute name that could make the code clearer. I chose `:username`. (ab96020dcf7d253d430452ee04832459cb6396fb)

Unfortunately, while I got sign on to work using this new attribute name, I ran into a new issue. Successful sign in works well, but unsuccessful sign in attempts throw an error page: 

unknown attribute username

This is reflected in the feature spec failures on this branch — the success test cases in the feature spec are green, the failed log-on test cases are red.

@kevinrobinson I'll stop work here and let you take this over going forward. You could back out the `:username` changes and use a model-backed field like `:login_name`, or figure out how to debug the bug. Based on my initial investigation, it looks like this may be related to code that's called by Devise, related to `authorization_keys` in `sessions_controller.rb`:

```
def new
  self.resource = resource_class.new(sign_in_params)
  clean_up_passwords(resource)
  yield resource if block_given?
  respond_with(resource, serialize_options(resource))
end

...
 
def serialize_options(resource)
  methods = resource_class.authentication_keys.dup
  methods = methods.keys if methods.is_a?(Hash)
  methods << :password if resource.respond_to?(:password)
  { methods: methods, only: [:password] }
end
```

Cheers and thank you so much for all your help progressing this forward @kevinrobinson!

kevinrobinson:
@alexsoble Awesome, that sounds great, thanks for all the great info and for talking through this today too!  I'll check this out tomorrow morning 👍

alexsoble:
@kevinrobinson We talked about this today, but we realized that the Bedford username needs to come with the `@k12...` suffix!

kevinrobinson:
I took forward @alexsoble's awesome work here, and split it into one PR doing the migrations and import (not super risky: https://github.com/studentinsights/studentinsights/pull/2082) and then another updating the login paths (risky: https://github.com/studentinsights/studentinsights/pull/2085).  The second one adds some more layers of tests and handles the email note mentioned above.

PR #2073, opened 2018-09-11
Remove references to Scout

kevinrobinson:
# Who is this PR for?
developers

# What problem does this PR fix?
cleanup

# What does this PR do?
remove references to a service we previously removed

PR #2074, opened 2018-09-11
Fix flaky assertion in TransitionNote authorizer spec

kevinrobinson:
Fixes https://travis-ci.org/studentinsights/studentinsights/builds/427366527, which was revealed from merging unrelated changes in https://github.com/studentinsights/studentinsights/pull/2073.


kevinrobinson:
selfie

PR #2075, opened 2018-09-11
Remove selenium-driver, update rubyzip

kevinrobinson:
Our code depends on `rubyzip` but we only pull it in transitively from `selenium-driver`, which we're not actually using.


kevinrobinson:
selfie

PR #2070, opened 2018-09-11
Profile v3: Improve display of programs, placements, IEP, ELL and educator contacts

kevinrobinson:
Part of https://github.com/studentinsights/studentinsights/issues/1990.

# Who is this PR for?
educators

# What problem does this PR fix?
The profile page doesn't show information about the student's program or placement, and doesn't show some details about IEP data or educator contacts.  English learning info is displayed in ways that don't live up to "best light" (eg, "Limited" as a standalone badge).

# What does this PR do?
Updates this, include test cases across a range of different permutations.

The idea is that there are five lines to use here, and these are filled bottom-up.  The bottommost is for "educator contacts" mirroring "family contacts" to the left, and includes any educators we know are working with that student, or points of contact.  Right now this includes just counselor, sped liaison, and staff who are actively providing services (which only has limited data coverage).  Above that is data about learning English, alongside home language to the left.  Above that is IEP data, program or placement data, and 504 data.  Total, this can take up to five lines of text.

This contains disabled code for the IEP dialog to display the IEP PDF inline, since the download experience is clumsy in IE11.  But this doesn't work cross-browser yet or with the CSP settings, so this is disabled here.  The server-side IEP code is refactored to be functional in development mode with a sample PDF doc.

Separately, this refactors some functions (but not all) related to how special education information is displayed to users and moves it into `helpers/specialEducation`.

Finally, `<HelpBubble />` is refactored to disable some styling and wrapper elements for these use cases, but this doesn't tackle that overall cleanup yet.

# Screenshot (if adding a client-side feature)
### before, labeled as "Limited" and without showing SPED placement
screen shot 2018-09-11 at 11 41 44 am

### after, clarify to "Limited English" and showing SPED placement
screen shot 2018-09-11 at 11 42 06 am

### before, without showing separate program
screen shot 2018-09-11 at 11 41 52 am

### after, showing assignment to separate program
screen shot 2018-09-11 at 11 41 56 am

### popup for educator contacts
screen shot 2018-09-11 at 11 42 10 am

### popup for IEP info
screen shot 2018-09-11 at 11 42 52 am

### popup for ACCESS data (unchanged)
screen shot 2018-09-11 at 11 42 16 am

# Checklists
+ [x] Author checked latest in IE - Student Profile
+ [x] Author checked latest in IE - My students
+ [x] Author checked latest in IE - Tiering page
+ [x] Author included specs for new code

PR #2076, opened 2018-09-11
Fix bug in IEP document branching

kevinrobinson:
Fix bug from https://github.com/studentinsights/studentinsights/pull/2070.


kevinrobinson:
selfie

PR #2072, opened 2018-09-11
Profile v3: Show IEP document inline when possible

kevinrobinson:
Part of https://github.com/studentinsights/studentinsights/issues/1990, follow-on to https://github.com/studentinsights/studentinsights/pull/2070.

# Who is this PR for?
educators

# What problem does this PR fix?
Some educators have said it's helpful that they can see the IEP-at-a-glance here.  But for IE users, depending on their setup, they may have to complete several steps before they can actually see the document.

### 1. confirm
screen shot 2018-09-11 at 2 49 10 pm

### 2. open
screen shot 2018-09-11 at 2 49 23 pm

### 3. choose viewer
screen shot 2018-09-11 at 2 49 30 pm

### 4. confirm opening
screen shot 2018-09-11 at 2 49 36 pm

### 5. reader application
screen shot 2018-09-11 at 2 49 42 pm

# What does this PR do?
Uses `PdfObject` to detect inline PDF viewing support, and if it's available, renders the PDF inline with `<object />`.  If not, falls back to the same download link.

I experimented with https://github.com/mozilla/pdf.js a pdf, to target helping users with older unconfigured browsers, but stopped as it's not intended as a drop-in JS dependency.

# Screenshot (if adding a client-side feature)
### Firefox
screen shot 2018-09-11 at 2 29 08 pm

### Safari
screen shot 2018-09-11 at 2 30 00 pm

### Chrome
screen shot 2018-09-11 at 2 28 58 pm

### IE11, with Acrobat installled, on initial view
screen shot 2018-09-11 at 2 50 34 pm

### IE11, with Acrobat installled, on closing and re-opening dialog
This shows something different, with more Acrobat toolbars and menus, not sure why.
screen shot 2018-09-11 at 2 51 31 pm

### Fallback (IE, no PDF viewer)
screen shot 2018-09-11 at 2 53 05 pm

# Checklists
+ [x] Author checked latest in IE - Student Profile

PR #2077, opened 2018-09-11
Update PDF to optimistically show inline, rather than detecting

kevinrobinson:
Follow-up to https://github.com/studentinsights/studentinsights/pull/2072.

PDFObject's detection is broken for the Firefox pdf.js viewer specifically.  So this PR updates the strategy to optimistically show the PDF inline, and removes `PDFObject` rather than trying to detect.  I must have mixed this up in testing https://github.com/studentinsights/studentinsights/pull/2072.

Also updates CSP for `X-Frame-Options` to allow sameorigin, since Firefox blocks `<object />` requests without this (which seems to not follow the CSP rule for `object-src`).

### Firefox
screen shot 2018-09-11 at 6 17 33 pm

### Safari
screen shot 2018-09-11 at 6 22 24 pm

### Chrome
screen shot 2018-09-11 at 6 17 45 pm

### IE11 with Acrobat viewer
screen shot 2018-09-11 at 6 19 15 pm

### IE11 without, upsell
screen shot 2018-09-11 at 6 21 58 pm

### DIFF
--- a/app/assets/javascripts/student_profile/LightHeaderSupportBits.js
+++ b/app/assets/javascripts/student_profile/LightHeaderSupportBits.js
@@ -126,18 +126,17 @@ export default class LightHeaderSupportBits extends React.Component {
     if (!hasAnySpecialEducationData(student, iepDocument)) return null;
     
     const specialEducationText = prettyIepTextForSpecialEducationStudent(student);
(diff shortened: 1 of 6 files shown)


kevinrobinson:
selfie

PR #2067, opened 2018-09-10
Fix bug in LoginFilter patch

kevinrobinson:
# Who is this PR for?
developers

# What problem does this PR fix?
bug in https://github.com/studentinsights/studentinsights/pull/2063 (https://github.com/studentinsights/studentinsights/pull/2063/files#diff-04861b3b13d716503cbd7bb02bb53d27R142).

PR #2078, opened 2018-09-13
Update yarn in Travis

kevinrobinson:
I think there's something wrong with our test setup across envs (but it could be just my machine).  I think that https://github.com/studentinsights/studentinsights/pull/2067 should have started failing the build in https://travis-ci.org/studentinsights/studentinsights/builds/426721325 since the test wasn't updated.  The assertion should be finding `41` tooltip divs, not `47` as there and still on master.

But the build didn't break, that test explicitly passes:
screen shot 2018-09-13 at 9 12 55 am

And none of the other test runs subsequently have failed.  Yet on my machine, this fails when I check out to master and run the tests.  So something isn't right, and reading the code it looks to me like this should be failing in the same way it is locally now for me.

In looking at the Travis logs, I saw yarn was behind.  I doubt this has anything to do with anything, but updating it and fixing the version seems good.  This PR also adds a `console.log` to the particular test so on the branch & master Travis builds I can see more about what's happening.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ branches:
 
 before_install:
   - export TZ=America/New_York
+  - curl -o- -L https://yarnpkg.com/install.sh | bash # see https://github.com/travis-ci/travis-ci/issues/7471#issuecomment-288832948
 
 install:
   - . $HOME/.nvm/nvm.sh
--- a/app/assets/javascripts/login_activity/LoginActivityPage.test.js
+++ b/app/assets/javascripts/login_activity/LoginActivityPage.test.js
@@ -35,7 +35,7 @@ it('renders after fetching', done => {
 
   setTimeout(() => {
     expect(el.innerHTML).toContain('Login Attempts, Past 30 days');
-    expect(el.querySelectorAll('div.tooltip').length).toEqual(47);
+    expect(el.querySelectorAll('div.tooltip').length).toEqual(41);
     done();
   }, 0);
 });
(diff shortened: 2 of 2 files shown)


kevinrobinson:
huh super concerning that this happened... https://travis-ci.org/studentinsights/studentinsights/builds/428158358 failed now as it should, just by upgrading yarn and adding a console.log... going to remove the console.log and see if that fails, then maybe try one more run with reverting the yarn change to see what happens

kevinrobinson:
Also restarting https://travis-ci.org/studentinsights/studentinsights/builds/427430455 to see if that fails now (just guessing about some caching shared across Travis runs, since something is going on there that I don't understand yet).

kevinrobinson:
Yeah so this seems super concerning and I don't know how to understand what happened here.  For reasons I don't understand, I shipped a test bug in https://github.com/studentinsights/studentinsights/pull/2067 and that wasn't caught in Travis across multiple test runs for three days.  Worse, inspecting the logs show that the specific test that should have failed passed.  I don't understand how that could have happened, and it was only revealed by running the Jest tests locally.

I don't remember test failures locally on Tuesday, so it's possible that something about this particular test or the test setup isn't working or I'm not understanding.  That could make sense on one machine (eg, maybe stale caching in the JS build or in the Jest runner or somewhere else in here), but I don't understand how that could happy on multiple Travis runs.  I don't understand why upgrading yarn influenced this, but it may be that adding the `before_install` step or upgrading yarn caused something to be invalidated in Travis.  From looking at https://docs.travis-ci.com/user/caching/#how-does-caching-work and our config, it seems that somehow a stale yarn cache influenced this.  But I don't see how that could have impacted this particular test result.

So I don't understand the root cause, but this is fixed now and I don't have good ideas about how to debug further that would be reasonably time efficient, going to move on.

I'll ship this PR with the Travis yarn upgrade (not directly related, but seems good to be explicit and updated) and the fix for the test bug.

kevinrobinson:
selfie

PR #2080, opened 2018-09-14
Remove DelayedJob

kevinrobinson:
We don't use this anymore.

Also moves the import tasks in `jobs` (which aren't Rails jobs) into the `importers` folder.


kevinrobinson:
selfie

PR #2082, opened 2018-09-17
Add Educator login_name and update import process

kevinrobinson:
This is a smaller part of https://github.com/studentinsights/studentinsights/pull/2071, pulled off to deploy in two pieces - one that handles updating the import and a second than handles update the login.

# Who is this PR for?
Supporting Bedford's different login process.

We transform `login_name` on import now, this migrates Somerville and New Bedford to fill this in, enabling the next step to switch the import process to use `login_name` as the primary key across districts.  This is effectively how Somerville and New Bedford both work now (the email is just a suffix), but Bedford handles these differently, so this change will make it so that this is generalized across districts, and email becomes secondary.


kevinrobinson:
selfie

PR #2084, opened 2018-09-17
Fix time in LoginActivity test, have it use nowFn from context

kevinrobinson:
# Who is this PR for?
developers

This JS test is tied to the wall clock, so it keeps failing over time.  This is the root problem discussed in https://github.com/studentinsights/studentinsights/pull/2078.

# What does this PR do?
Fixes the time in the test.


kevinrobinson:
selfie

PR #2085, opened 2018-09-17
Update login path and Devise config to support Bedford login

kevinrobinson:
This is the second part of https://github.com/studentinsights/studentinsights/pull/2082, split out from https://github.com/studentinsights/studentinsights/pull/2071.

# Who is this PR for?
Bedford educators

Bedford educators have a `login_name` but also have two sets of email addresses, one that is used for email and the other that is used for authenticating with their LDAP server.  This makes several changes to the Devise config and login path to enable differences here across district, building on the first step in https://github.com/studentinsights/studentinsights/pull/2082 to import `login_name` for all districts.

# What does this PR do?
1. Update login form and Devise config.  Updates the form to be a plain textbox, decoupled from the Rails model, update the translation text to be agnostic.
2. Update LDAP model to express that `authentication_keys` should be `login_text` and then update our model to include a noop writer that the Devise controller uses (similar to `password=`), and adds tests for this to the Educator model.
3. Update the LDAP strategy to call into `PerDistrict` for finding a user from `login_text` and for finding the `ldap_login` for a given user.  Add tests at all the layers for this.
4. Update the `MockLDAP` class to call into `PerDistrict` to determine if the the `ldap_login` exists in the database.  Also move `ShouldUseMockLDAP` to a class method on `MockLDAP`.
5. Change the Devise config to make the routing and controller actions explicit (useful for debugging) and cleanup some config (eg, removing http_authenticatable which we never use and wouldn't work anyway with SSL redirects).

screen shot 2018-09-17 at 1 58 26 pm

# Checklists
+ [x] Author checked latest in IE - Login


kevinrobinson:
selfie

PR #2088, opened 2018-09-17
Remove LoadDistrictConfig, refactoring into PerDistrict

kevinrobinson:
# Who is this PR for?
developers

There's some older per-district config that's in `LoadDistrictConfig`, this consolidates into into `PerDistrict`.  This PR doesn't change how individual fields are read from the yaml - they're still not expressed in `PerDistrict` and so the code doesn't have information about which values are missing because the district doesn't support, or missing in a way that's a bug.


kevinrobinson:
selfie

PR #2089, opened 2018-09-17
Add presence for login_name

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2082, now that this is migrated.  There's already a non-null constraint on the database field.


kevinrobinson:
selfie

PR #2092, opened 2018-09-18
Hold out schoolwide_access and can_view_restricted_notes

kevinrobinson:
Revert a part of https://github.com/studentinsights/studentinsights/pull/2082; separate refactoring coming to remove `admin` altogether and these defaults.


kevinrobinson:
selfie

PR #2094, opened 2018-09-18
Simplify interface for filter feeds

kevinrobinson:
Straight refactor to simplify, from looking at this in investigating https://github.com/studentinsights/studentinsights/pull/2092.


kevinrobinson:
selfie

PR #2095, opened 2018-09-18
Remove School state_id, work towards tightening validations

kevinrobinson:
Splitting off pieces from https://github.com/studentinsights/studentinsights/pull/2088.

# Who is this PR for?
developers

There's an `state_id` field on the `School` model that we don't use and can't read in for any districts anymore, so this removes it.  The `school_type` field is similar in that we don't read it in anywhere (it was probably set manually in the past), but we do rely on this.

# What does this PR do?
In https://github.com/studentinsights/studentinsights/pull/2088 and subsequent work, the idea is to encode this manual data entry into the `yaml` files for each district.  This PR just removes the old field, and relaxes the validations on `school_type` so we can manually run a migration, then in the next PR add in more constraints and validations on `school_type`.  https://github.com/studentinsights/studentinsights/pull/2088 will ultimately move this into the `yaml` files so that we can enforce that the DB matches what's in source for these, and so that we can include this as a init task for new districts (eg, Bedford).

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -1,11 +1,22 @@
 class School < ActiveRecord::Base
   extend FriendlyId
   friendly_id :local_id, use: :slugged
-  validates :local_id, presence: true, uniqueness: true
-  validate :validate_school_type
+
+  # These are internal to Insights, not from the SIS
+  VALID_SCHOOL_TYPES = [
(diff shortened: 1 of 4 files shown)


kevinrobinson:
selfie

PR #2096, opened 2018-09-18
Add district name to HTML head tag

kevinrobinson:
So that the district name appears in the browser tab or window title.  Mostly for developers.


kevinrobinson:
selfie

PR #2097, opened 2018-09-18
Add rake task for updating School records

kevinrobinson:
# Who is this PR for?
developers

Make the workflow for loading new district data more explicit.  We can use this for Bedford.

# What does this PR do?
With https://github.com/studentinsights/studentinsights/pull/2088, Schools are specified in the district yaml file.  This syncs from that yaml file to production and is intended to be run manually.  The intention is that it's harder for this to drift, and for one-off import tasks to mistakenly import data we don't need.

### DIFF
--- a/lib/tasks/schools.rake
+++ b/lib/tasks/schools.rake
@@ -1,0 +1,32 @@
+desc 'For syncing district.yml files to School records in production'
+namespace :schools do
+  task update: :environment do
+    school_definitions = PerDistrict.new.school_definitions_for_import
+    puts "Found #{school_definitions} school_definitions..."
+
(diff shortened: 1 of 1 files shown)


kevinrobinson:
selfie

kevinrobinson:
Looks like this:
```
Found [{"name"=>"Lane Elementary", "local_id"=>"Lane", "school_type"=>"ES"}, {"name"=>"Davis Elementary", "local_id"=>"Davis", "school_type"=>"ES"}] school_definitions...
Checking Lane...
  Creating school Lane...
Checking Davis...
  Creating school Davis...
Done.
```

PR #2100, opened 2018-09-18
Raise when CsvDownloader file is zero bytes

kevinrobinson:
# Who is this PR for?
developers

not being alerted on 0 byte files

# What does this PR do?
raise, we'll get alerted, we can talk to district pals


kevinrobinson:
selfie

PR #2102, opened 2018-09-19
Fix bug in error message when raising on zero byte CSV

kevinrobinson:
Fix for inside the `if` for https://github.com/studentinsights/studentinsights/pull/2100.


kevinrobinson:
selfie

PR #2103, opened 2018-09-20
Update low grades endpoint to include any notes by default

kevinrobinson:
# Who is this PR for?
HS educators

In https://github.com/studentinsights/studentinsights/pull/2057, I think this was the intent but the param in the controller meant this still only applied to NGE and 10GE notes.  This is different than what the copy suggests and what I remember as the intention.

# What does this PR do?
Updates controller to use whatever the insight class indicates as default.


kevinrobinson:
selfie

Issue #1661, opened 2018-04-27
Absences and discipline: Add dropdown to filter by Grade, House and Counselor


kevinrobinson:
This is important for SHS housemasters and counselors.  For the UX, one suggestion is to add a `react-select` dropdown only the high school and only for Somerville (see `PerDistrict`).

- [x] add filter to absences page (https://github.com/studentinsights/studentinsights/pull/2105)
- [x] add filter to discipline page (https://github.com/studentinsights/studentinsights/pull/2186)
- [ ] add filter to tardies page

PR #2105, opened 2018-09-21
Absences: Add filters for house, counselor and grade to fork from current page

kevinrobinson:
# Who is this PR for?
SHS housemasters and counselors

Housemasters and counselors at SHS focus on particular groups of students, but the absences page shows them all students they have access to in the school.  Part of https://github.com/studentinsights/studentinsights/issues/1661.

# What does this PR do?
It creates a new page (eg, `/schools/shs/absences/v2`), split from the current absences page, so we can compare and verify no regressions before releasing.

The absences page previously did a bunch of work to precompute different values based on the various filters, which made it responsive to user interactions.  This expanded the amount of computation and code for each permutation of filters, requiring changes throughout to add a new filter.  So this PR builds on the existing page, but filters these on-demand and adds memoization to expensive computations so that it's still responsive.

It removes the monthly rate, since this doesn't make sense over the default 45 day time period.  For now this just replaces it with a single bar.  I also looked at using a single number (since that's essentially what this is showing) but that highlights it as if it were a top-line metric and we want the emphasis to be on the list of students, not the metric.

This also changes the click interaction for selecting a homeroom, so that clicking highlights the homeroom in orange, rather than showing it separately in the table.

Separately, this removes the CSS from all dashboard pages so that they render as expected within Storybook, and as a migration step moves them into `dashboardStyles` but doesn't make any other changes.

### current
screen shot 2018-09-21 at 10 53 40 am

### v2
screen shot 2018-09-21 at 10 49 48 am
screen shot 2018-09-21 at 10 49 59 am

# Checklists
+ [x] Author checked latest in IE - Absences v2
+ [x] Author checked latest in IE - Absences


kevinrobinson:
selfie

app/assets/javascripts/school_absences/SchoolAbsencesDashboard.js
```diff
@@ -0,0 +1,404 @@
...
+  componentWillReceiveProps() {
+    this.memoize = memoizer();
+  }
+
+  shouldIncludeExcusedAbsences() {
+    const {excusedAbsencesKey} = this.state;
+    return excusedAbsencesKey === ALL_ABSENCES;
+  }
+
+  filteredStudents(options = {}) {
```
kevinrobinson: @edavidsonsawyer hello! :)  This is the pull request I mentioned, where the goal is to be able to write functions like `filteredStudents` and `filteredAbsences` that just do simple ifs for whatever filters we want.

So here I wrote this naively, then looked in the Chrome profiler to see what was slowest, and then added the calls to `this.memoize` which memoizes different computations so they only have to be done once.  So even if the code I wrote calls `this.filteredStudents` lots of times it only has to actually compute it once.

edavidsonsawyer: @kevinrobinson this is cool! It's a much cleaner approach. I also like switching out the monthly rate since as you mentioned that doesn't make a lot of sense over the time ranges users are seeing (and its faster!). 

I do wonder whether it would be useful to have some way of looking at attendance over time when it comes to the 90 day and school year ranges, though I'm not sure whether anyone was getting value out of that. But then, the whole visualization of this should probably change somewhere in the future since, like you said, the student list is where the action is at.

kevinrobinson: @edavidsonsawyer Yeah good question about attendance over time!  I was wondering that too, and ended up deciding to not do that, and to minimize the attendance rate number here (keeping the single bar chart where you have to hover to see a number vs. seeing the number more prominently).  The intention there was to focus attention on individual students and immediate actions they can take, rather than understanding aggregate distributions or patterns or trends.  In other words, we're trying to help educators notice things happening with individual students in a way that's directly tied to actions they can take, rather than presenting them with an abstract metric that they'll be held responsible to drive up or down, as people might do in other kinds of settings.

PR #2107, opened 2018-09-21
Store historical grades during import

kevinrobinson:
# Who is this PR for?
enabling looking at this over time in the HS

nothing, it's enabling experimental work

# What does this PR do?
stores `HistoricalGrade` records during the section/grade importer.


kevinrobinson:
selfie

PR #2110, opened 2018-09-24
Update how ACCESS and WIDA data is shown in profile page

kevinrobinson:

# Who is this PR for?
educators, particularly HS NEST teachers and HS NGE/10GE teachers

The data on ACCESS, WIDA Model and WIDA Screener are merged together, and so subtests from different dates are shown as if they were the same test time.  Also, the numbers don't mean anything to classroom teachers if they aren't familiar with WIDA scoring rubrics.

# What does this PR do?
Organizes the scores by composite and subtest, shows the scores on a visual scale, adds in data about when the subtest score is from.  Also links to WIDA interpretation guides.

This PR does not update the import-side code to preserve data about the particular test that was taken, but that would be a good future improvement.

This also aims to not break the profile v2 page, but doesn't spend any time making things look good there, as that's deprecated and not visible to users anymore.

screen shot 2018-09-24 at 1 11 37 pm

# Checklists
+ [x] Author checked latest in IE - Student Profile v2


kevinrobinson:
selfie

PR #2112, opened 2018-09-24
Update absence imported to use PerDistrict parsing of boolean fields for Bedford

kevinrobinson:
# Who is this PR for?
Bedford pals

Their export uses string values for true/false in attendance data, not integer values like others.

# What does this PR do?
Updates this to be `PerDistrict.`


kevinrobinson:
selfie

PR #2101, opened 2018-09-18
Update PerDistrict handling for Bedford students

kevinrobinson:
# Who is this PR for?
Bedford educators

# What problem does this PR fix?
Bedford exports data in a slightly different format, and includes some data we weren't expecting.

# What does this PR do?
Update `PerDistrict` to import `house` `counselor` and `sped_liaison`.  Updates UI code for the student profile around:
```
program_assigned
limited_english_proficiency
sped_placement
sped_level_of_need
```

PR #2113, opened 2018-09-24
Support excused absences filter for Bedford

kevinrobinson:
Follow on to https://github.com/studentinsights/studentinsights/pull/2101 and https://github.com/studentinsights/studentinsights/pull/2112.


kevinrobinson:
selfie

PR #2117, opened 2018-09-25
Stricter date parsing in student importer

kevinrobinson:
# Who is this PR for?
Bedford educators

Bedford exports currently include invalid registration dates, which they're fixing, but we aren't noticing because of lax date parsing, and are setting these to incorrect dates.

# What does this PR do?
Switches to strict date parsing for dates in the student importer.


kevinrobinson:
selfie

PR #2118, opened 2018-09-25
Update valid grades for Bedford

kevinrobinson:
# Who is this PR for?
Bedford educators

They use different codes for younger grade levels.

# What does this PR do?
Update this to be maximal across districts (rather than per-district), and refactor to remove different Ruby definitions of grade levels to all use `GradeLevels`.


kevinrobinson:
selfie

PR #2119, opened 2018-09-25
Update StudentsImporter to set nil rather than empty string values

kevinrobinson:
# Who is this PR for?
developers

If a field is missing, `StudentsImporter` will set it as nil.  If it's there, but an empty string, it will set it as `""`.  For making things with most of these enum fields, these mean the same thing.

# What does this PR do?
Updates the importer to set `""` as nil.  It also improves some `Student` validations, moving them up to use `inclusion: in` and tightening `local_id`.

# Checklists

### DIFF
--- a/app/controllers/admin/educators_controller.rb
+++ b/app/controllers/admin/educators_controller.rb
@@ -10,7 +10,7 @@ module Admin
     end
 
     def edit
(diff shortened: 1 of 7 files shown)


kevinrobinson:
selfie

PR #2120, opened 2018-09-25
Tighten student validations to exclude empty strings

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2119, enforcing this in validations now.  Mostly requires updating test setup code.


kevinrobinson:
selfie

Issue #1880, opened 2018-07-10
Improve way "homerooms" are created and assigned to educators


kevinrobinson:
This particularly impacts New Bedford.  See @kevinrobinson for details.

alexsoble:
@kevinrobinson Looking into this further — I ran the following script against production New Bedford data to see if there are any homeroom teachers currently assigned to incorrect homerooms: 

```
Educator.all.each do |educator|
  if educator.homeroom.present? && educator.school.present?
    raise 'School mismatch!!' if educator.school != educator.homeroom.school
  end
end ; nil
```

The script didn't raise, and I did some manual inspection too to verify the same thing. So for the two schools currently imported, no New Bedford educator account is assigned to an incorrect homeroom.

kevinrobinson:
@alexsoble nice, awesome to check this!  I fixed all this up manually when I removed all the other students and educators and homerooms.

One tiny clarification - I think it might be risky to consider these homerooms correct.  They are still identified with names that won't let us discriminate between homerooms across schools.  And the way that homerooms get created or updated is in the student importer, where it's based on the student's school.  So I think to check this all the way around it involves looking at whether the students in that homeroom also match.  There's that circular dependency in the importers between students, educators and homerooms.   `EducatorsImporter` runs first, and that expects homerooms to exist.  Then `StudentsImporter` runs and that creates the student records and homeroom records.

kevinrobinson:
https://github.com/studentinsights/studentinsights/pull/2023 fixed some things, but there are a few other edge cases impacting different educators/students/homerooms:
- [ ] because of past configurations, there are stale Homeroom records (eg, created from a Student import row, that wouldn't be created now)
- [ ] in Somerville the SIS has semantically confusing data (eg, student in the Brown school but the CAP-141  homeroom).  This gets created at `CAP-141` within the Brown (from https://github.com/studentinsights/studentinsights/pull/1994) which is isolated and good, but misleading in Insights as it is in the SIS.  Probably nothing to do here, might be better to mirror bad SIS data and surface it than hide it.
- [ ] handle stale Student records (eg, NB#443 who moved to a school that is not imported, and so is never updated or removed - maybe mark them as archived?)
- [ ] handle stale Educator records (mark them as archived?)

PR #2123, opened 2018-09-25
Create assigned_homeroom_id key on Educator table, copy to bootstrap, and update importers

kevinrobinson:
# Who is this PR for?
educators, developers

Part of https://github.com/studentinsights/studentinsights/issues/1880#issuecomment-416420431.

Multiple educators can be assigned to the same homeroom, but Insights doesn't allow this now.  This results in the importer process thrashing each day over who is assigned to the homeroom, and effectively alternating who will be authorized to view those students.  This isn't user-impacting right now, and it came up from validating the Bedford imports.

The overall idea is to migrate the key to the Educator model, and do this in pieces:
- [ ] add new key, copy values into it
- [ ] run import
- [ ] update read side (feed, homerooms_controller)
- [ ] drop old key
- [ ] migrate column name and read code

Because these are related, once we start we should finish them all to avoid allowing drift during the migration.

# What does this PR do?
This PR aims to do the first two items.

# Checklists

### DIFF
--- a/app/importers/file_importers/educators_importer.rb
+++ b/app/importers/file_importers/educators_importer.rb
@@ -76,18 +76,18 @@ class EducatorsImporter
       return
     end
 
-    # Find the matching Educator record if possible and sync it
-    maybe_educator = EducatorRow.new(row, school_ids_dictionary).match_educator_record
-    @educator_syncer.validate_mark_and_sync!(maybe_educator)
-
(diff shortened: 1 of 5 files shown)


kevinrobinson:
https://github.com/studentinsights/studentinsights/pull/2162 pointed to this again, looking into `Homeroom` records failing the validation on `name`.

PR #2125, opened 2018-09-25
Update assessment date parsing to be per-district

kevinrobinson:
# Who is this PR for?
Bedford

Works around export format being different.


kevinrobinson:
CsvRowCleaner already does this, closing.

PR #2127, opened 2018-09-25
Update IncidentCard on feed to skip missing location

kevinrobinson:
# Who is this PR for?
Bedford educators

Most Bedford incident don't have location set, so this looks strange.

# What does this PR do?
Updates it to remove the line if no location.


kevinrobinson:
selfie

PR #2130, opened 2018-09-25
Disable Bedford assessment import

kevinrobinson:
Tabling this for now, and will pull in iteratively after they get up and running, so updating the config to reflect that.


kevinrobinson:
selfie

PR #2132, opened 2018-09-26
Top Level Discipline Incident Type filter

edavidsonsawyer:
# Who is this PR for?
Discipline Dashboard Users 

Discipline incidents require different responses depending on the type of incident, so it's unhelpful to just show total incidents. More useful is to break down the incidents by category so users can compare similar types of behevior. 

# What does this PR do?
Allows users to select the type of discipline incident they would like to see data for. 

It also simplifies the structure of the discipline dashboard, ~~shamelessly copying~~ building off of the work @kevinrobinson has done with the absence dashboard. 

discipline

# Checklists

## Javascript QA

## Automated Testing


edavidsonsawyer:
Hey @kevinrobinson this is ready for a look. For now I have the categories in a dropdown - seeing them all at once was overwhelming. Probably a lot of that has to do with the demo data which randomly assigns incident categories. I think the schools will have a few incident types that dominate, but I don't know for sure. 

I would like to chat with you about how to handle certain selection cases that a user can get into here which might look confusing but it's a little involved to explain without a screen share. Maybe at today's call or some other time this afternoon?

kevinrobinson:
@edavidsonsawyer Sure!  I'm free 2pm-4pm this afternoon, so how about we sync on things with Uri altogether and then you and dive into any more details after?  I'll also take a look at this PR now too so I'll be prepared to be helpful :)

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -41,12 +45,56 @@ export default class SchoolDisciplineDashboard extends React.Component {
...
-  filterIncidentDates(incidentsArray) {
-    const {nowFn} = this.context;
-    const {timeRangeKey} = this.state;
-    const range = momentRange(timeRangeKey, nowFn());
-    return incidentsArray.filter((incident) => {
-      return moment.utc(incident.occurred_at).isBetween(range[0], range[1]);
+  allDisciplineIncidents() {
+    return this.memoize(['allDisciplineIncidents'], () => {
+      const {dashboardStudents} = this.props;
+      return _.flattenDeep(_.compact(dashboardStudents.map(student => this.mergeDisciplineData(student.discipline_incidents, student))));
```
kevinrobinson: Can you say more about why the transformation in `allDisciplineIncidents` and `mergeDisciplineData` is necessary to do up front?

It looks to me like the flattening has to get undone later one with regrouping by student, and the merge fn looks to me like it's changing the names of the keys and doing little transformations like rendering a moment as a user-facing string.  These seem like they could be moved from being precomputed up front to done on-demand wherever they are needed in the `render` methods.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -41,12 +45,56 @@ export default class SchoolDisciplineDashboard extends React.Component {
...
+  }
+
+  filteredDisciplineIncidents(disciplineIncidents) {
+    return this.memoize(['filteredIncidents', this.state, arguments], () => {
+      const {nowFn} = this.context;
+      const {timeRangeKey, selectedIncidentCode} = this.state;
+      const range = momentRange(timeRangeKey, nowFn());
+      return disciplineIncidents.filter(incident => {
+        if (!moment.utc(incident.occurred_at).isBetween(range[0], range[1])) return false;
+        if (incident.incident_code !== selectedIncidentCode && selectedIncidentCode !== ALL) return false;
```
kevinrobinson: nice!

kevinrobinson: Does the filter for the selected time/date/etc belong here too?

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -96,13 +144,8 @@ export default class SchoolDisciplineDashboard extends React.Component {
     });
   }
 
-  //For grades and classrooms, the students table should only show the relevant students
```
kevinrobinson: nice!

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -117,23 +160,29 @@ export default class SchoolDisciplineDashboard extends React.Component {
     const {timeRangeKey} = this.state;
     const {school} = this.props;
     const chartOptions = [
-      {value: 'location', label: 'Location'},
-      {value: 'time', label: 'Time'},
+      {value: 'incident_location', label: 'Location'},
+      {value: 'exactTime', label: 'Time'},
```
kevinrobinson: let's keep the case for keys the same within the same data structure :)

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -186,22 +235,10 @@ export default class SchoolDisciplineDashboard extends React.Component {
...
-        id: student.id,
-        first_name: student.first_name,
-        last_name: student.last_name,
-        latest_note: student.latest_note,
-        events: studentDisciplineIncidentCounts[student.id] || 0,
-        grade: student.grade
-      });
-    });
-
+    const selectedIncidents = this.state.selectedCategory ? //if the user is looking at a subgroup of incidents
```
kevinrobinson: I think this branch and grouping is the kind of thing we want to remove, and isolate in methods like `filteredStudents` or `filteredDisciplineIncidents`.  The way it is now, if we want to add a filter for grade next week, we'd have to rework this and make the index more complex to support that.  If we can isolate the filtering, then updating filters just requires updating the filter function.

Could this method work with calling `filteredStudents` or `filteredIncidents`?

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -186,22 +235,10 @@ export default class SchoolDisciplineDashboard extends React.Component {
...
+                              groupedIncidents[this.state.selectedCategory] :
+                              allIncidents;
+    const rows = selectedIncidents ? this.studentDisciplineIncidentRows(selectedIncidents) : [];
```
kevinrobinson: I'm wondering about how we can remove the work that `studentDisciplineIncidentRows` is doing in order to create a list of `[{student, disciplineIncidentCount}]` that respects the various filters.

Here's what I see here, let me know if I'm following.  Overall, this component gets a list of students with discipline incidents for each, and it stores in state the info about the various filters.  .  `render` flattens all the incidents and filters them and calls this method with the flattened incidents.  This method then has to call `studentDisciplineIncidentRows` to group those back together by students so it can count them.

Check out how this works in the [absences page](https://github.com/studentinsights/studentinsights/blob/master/app/assets/javascripts/school_absences/SchoolAbsencesDashboard.js#L150), where it's essentially:
```
this.filteredStudents().map(student => {
  return {
    ...student,
    absenceCount: this.filteredAbsences(student).length
  }
})
```
Would a similar approach help simplify here?

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -216,17 +253,6 @@ SchoolDisciplineDashboard.contextTypes = {
...
-    student_id: PropTypes.number.isRequired, //ID of student involved in incident
-    location: PropTypes.string, //Place where incident occurred
-    time: PropTypes.string, //Time of day for incident - NULL if no specific time recorded
-    classroom: PropTypes.string, //Name of student's homeroom teacher
-    student_grade: PropTypes.string, //Grade of student
-    day: PropTypes.string, //Day of week on which incident occurred
-    offense: PropTypes.string, //Specific type of incident
-    student_race: PropTypes.string, //Race of student
-    occurred_at: PropTypes.string, //Date for incident, used in filtering specific date ranges
-  })).isRequired,
```
kevinrobinson: This seems like good checking for dev/test and documentation of intent.  Is there a way to keep it and say `dashboardStudents` has an array of objects that have the `discipline_events` field in this shape?

kevinrobinson:
re: UX, the idea in other places is that filters are in the upper left, and time range is on the far right.  The breakdown seems fine where it is for now, but check out how this is laid out in the absences page:

screen shot 2018-09-28 at 10 33 42 am

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: Sure thing! The problem I'm trying to solve is that the groups we want aren't attributes of a single object. Some are student attributes , some are discipline incident attributes, and some are computed from the discipline incident. I find it more readable and less confusing to have one function that "owns" all the attributes we care about for grouping and then just _groupBy to get our categories. That way whenever we want to add some other category, we have a place to put that logic and see it all together.

This isn't exactly precomputed since this method is only called in the main render method and the results are passed down to the other methods that need them. Unless I'm misunderstanding something it should still be happening on demand (that's why I removed the old SchoolWideDisciplineIncidents component that really was precomputing all this stuff).

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: I don't think so. I'm using this to trim the incidents that fall outside the date range or selected incident type. The time of day, day of week, etc are different. Those get sorted into categories, and when the user selects that part of the bar chart they filter the student list, but the incidents still show up in the bar chart.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: Oops! Yeah, I was experimenting and left one of these in.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: Let me think about that. All the rows method is doing is taking a slice of the displayed incidents (based on the selected category) and reducing it to the information that the students table needs to display them. It does check to see if it should be using a subgroup, but it isn't changing or regrouping the flattened incidents. 

The main difference here between the absence dashboard is that we don't show students without discipline incidents so it makes more sense to think of this as a map of the discipline incidents with student data attached rather than a map of each student that we would then have to filter for students with no incidents.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: I think we can do that. Not all students will have this information, but there's no reason we couldn't validate the structure for those that do.

edavidsonsawyer:
Thanks @kevinrobinson! I did respond here but I think it's more helpful if we talk later, I just wanted to remember my thoughts.

kevinrobinson:
@edavidsonsawyer that sounds great, I'll check through your comments here too beforehand, thanks!

edavidsonsawyer:
Hi @kevinrobinson, when you get the chance take a look at this. It was a little more involved than I expected, but it does simplify the render methods and condense the functional logic.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -41,38 +45,99 @@ export default class SchoolDisciplineDashboard extends React.Component {
...
-    //if a user selects a category and then moves to a time range with no incidents within that category, return an empty object
-    if (!incidents) return studentDisciplineIncidentCounts;
-    incidents.forEach((incident) => {
-      studentDisciplineIncidentCounts[incident.student_id] = studentDisciplineIncidentCounts[incident.student_id] || 0;
-      studentDisciplineIncidentCounts[incident.student_id]++;
+  filterStudents(students) {
+    const {selectedChart, selectedCategory} = this.state;
+    return students.filter(student => {
+      if (selectedCategory && selectedChart === 'grade' && student.grade !== selectedCategory) return false;
+      if (selectedCategory && selectedChart === 'race' && student.race !== selectedCategory) return false;
```
kevinrobinson: `race` looks like it's leftover from past iterations, let's clean this up and remove it if it's not in the product anymore.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -41,38 +45,99 @@ export default class SchoolDisciplineDashboard extends React.Component {
...
+    }
+    case 'grade': {
+      const groupedStudents = _.groupBy(students, group);
+      const categories = this.sortedGrades(Object.keys(groupedStudents));
+      const seriesData = categories.map(category => {
+        return this.getIncidentsFromStudents(groupedStudents[category]).length;
+      });
+      return {categories, seriesData};
+    }
+    case 'incident_code': case 'incident_location': {
```
kevinrobinson: Should `incident_code` be removed, since it's now a filter up top instead?

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -41,38 +45,99 @@ export default class SchoolDisciplineDashboard extends React.Component {
...
+      const categories = this.sortedByIncidents(groupedIncidents);
+      const seriesData = categories.map(category => groupedIncidents[category].length);
+      return {categories, seriesData};
+    }
+    case 'time': {
+      const incidents = this.getIncidentsFromStudents(students);
+      const groupedIncidents = _.groupBy(incidents, incident => {
+        return incident.has_exact_time ? moment.utc(incident.occurred_at).startOf('hour').format('h:mm a') : "Not Logged";
+      });
+      const categories = this.sortedTimes(Object.keys(groupedIncidents));
```
kevinrobinson: I think there's a UX bug here here, in that for breakdowns that refer to stable known categories, we should set these values to be fixed so this looks like the scale people will expect (eg, like with `sortedDays` here rather than inferred from the data in `sortedTimes`).

This is what I see locally, check out the x scale:
screen shot 2018-10-02 at 10 09 31 am

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -41,38 +45,99 @@ export default class SchoolDisciplineDashboard extends React.Component {
...
+      const seriesData = categories.map(category => groupedIncidents[category].length);
+      return {categories, seriesData};
+    }
+    case 'day': {
```
kevinrobinson: If I try to use this locally, it doesn't work and I see this stack trace:

screen shot 2018-10-02 at 10 04 49 am

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -96,13 +161,33 @@ export default class SchoolDisciplineDashboard extends React.Component {
...
+        last_name: student.last_name,
+        grade: student.grade,
+        latest_note: student.latest_note,
+        events: this.filterIncidents(student.discipline_incidents, selectedCategory).length
+      };
+    });
+  }
+
+  allIncidentTypes(students) {
+    return _.uniqBy(this.getIncidentsFromStudents(students), 'incident_code').map(incident => incident.incident_code);
```
kevinrobinson: This method is called `allIncidentTypes` but it looks like it's ultimately calling `filterIncidents`.  Is this a naming thing or is this a bug?

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -96,13 +161,33 @@ export default class SchoolDisciplineDashboard extends React.Component {
...
   }
 
-  //For grades and classrooms, the students table should only show the relevant students
-  groupStudents() {
-    if (this.state.selectedChart === 'grade' && this.state.selectedCategory) {
-      return this.props.dashboardStudents.filter(student => student.grade === this.state.selectedCategory);
-    } else if (this.state.selectedChart === 'classroom' && this.state.selectedCategory) {
-      return this.props.dashboardStudents.filter(student => student.homeroom_label === this.state.selectedCategory);
-    } else return this.props.dashboardStudents;
+  sortedByStudent(groupedStudents) {
```
kevinrobinson: This method looks like it sorts by incident count, not student count?

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: I meant to ask about that. I wasn't sure we necessarily wanted to remove the ability to compare different incident types, though I'm not sure there's much useful information to be gained from that.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: Oops, that's a bug - I added the filter to the student incident counts later, but we don't want this list filtered.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: I'd imagine this looks less awkward with non-random data, but what do you think an appropriate scale would be? It should be possible to display incidents outside of school hours, but a 24 hour scale feels unnecessary. Maybe I'll try it with full 24 hours and we can see if it looks okay in production.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: I thought you guys had mentioned race so I was leaving it open as an option. I'll take it out until we're sure we want it.

edavidsonsawyer:
Thanks @kevinrobinson this is helpful!

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
kevinrobinson: Yeah, if it's not being used let's remove it.  We've got source control so always easy to bring things back :)

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
kevinrobinson: Yeah I agree that there's probably not a lot happening in most schools at 1am :)  Maybe 7am - 3pm and bucket anything else into "other"?

edavidsonsawyer:
Okay @kevinrobinson this has been tightened up some. I'm planning on opening two more PRs separately for chart highlighting a la absences and a possible change to the reset filter to make it less jarring. They don't need to be tied to this feature, though.

edavidsonsawyer:
@kevinrobinson that's an interesting bug. I'll open a second PR for that one. 

I was poking around more tonight and I've also seen some odd behavior when there are a small number of incidents, so it seems fine to pause this until that's settled.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -78,7 +77,15 @@ export default class SchoolDisciplineDashboard extends React.Component {
...
-    return incident.has_exact_time ? moment.utc(incident.occurred_at).startOf('hour').format('h:mm a') : "Not Logged";
+    const hour = moment.utc(incident.occurred_at).startOf('hour').format('h:mm a');
+    const timeFormat = "HH:mm a";
+    const schoolStart = moment("7:00 am", timeFormat);
+    const schoolEnd = moment("3:00 pm", timeFormat);
+    return !incident.has_exact_time
+            ? "Not Logged"
+            : moment(hour, timeFormat).isBetween(schoolStart, schoolEnd)
+            ?  hour
+            : "Other";
```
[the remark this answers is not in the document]
edavidsonsawyer: In a moment of weakness, I thought parsing them as UTC might give us times outside the 7am-3pm range in the timestamp. But you're right, it's not a problem.

edavidsonsawyer:
@kevinrobinson before changing the bar chart component, is this the wonky behavior you saw? 
breakdownbug

If so, that's a (now fixed) mistake in this component. The student list wasn't being reset when the discipline types changed. So you might have been looking at the three students with incidents "off campus", and then changed the incident type to "cell phone" and the filters would show there were no students in those filters. Since the list and chart should be showing the same data, this component now resets the selected category when changing the top level incident type filter. 

I haven't found a problem I can trace back to the bar chart, but I'll keep looking. I'd prefer not to remove the shouldComponentUpdate check entirely since it's a UI fix, not a performance optimization. It's jarring to have the chart completely redraw every time a new homeroom (or discipline category) is selected.

edavidsonsawyer:
@kevinrobinson nevermind the above, there will still be a problem updating the chart when the event totals are the same - I thought the categories were passed in with the series data here. I'll open that other PR.

edavidsonsawyer:
@kevinrobinson I've reviewed the filtering with these last changes and it's consistent. The charts also no longer rely on the lifecycle method in dashboardbarchart to prevent animating, so that can also be removed (separate PR).

kevinrobinson:
@edavidsonsawyer great, let's 🚢!

PR #2133, opened 2018-09-27
Fix bug with only including NGE/10GE students in low grades box, fix level 1 tier bug

kevinrobinson:
# Who is this PR for?
HS teachers

Fixes a bug with https://github.com/studentinsights/studentinsights/pull/2057, which intended to release the "low grades" box to all HS teachers, but didn't update a filter that filtered students down to only 9th and 10th grade.

# What does this PR do?
Removes that filter and adds a test.

# Checklists

### DIFF
--- a/app/lib/experimental_somerville_high_tiers.rb
+++ b/app/lib/experimental_somerville_high_tiers.rb
@@ -118,11 +118,11 @@ class ExperimentalSomervilleHighTiers
     ].compact
     return Tier.new(2, tier_two_triggers, data) if tier_two_triggers.size > 0
 
-    # Level 1: 1 F and 2 Ds
+    # Level 1: 1 F or 2 Ds
     #   OR less than 95% attendance over last 45 days
     #   (no discipline involved)
     tier_one_triggers = [
-      (:academic if data[:course_failures] == 1 and data[:course_ds] >= 2),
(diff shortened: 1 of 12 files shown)


kevinrobinson:
Also add a new test pal for this test case, Kylo, and update other specs to consider him.

kevinrobinson:
This also fixes a bug with the level 1 calculation, where it would only match and adds a slew of test cases to that method.

app/lib/experimental_somerville_high_tiers.rb
```diff
@@ -118,11 +118,11 @@ def decide_tier(data, options = {})
     ].compact
     return Tier.new(2, tier_two_triggers, data) if tier_two_triggers.size > 0
 
-    # Level 1: 1 F and 2 Ds
+    # Level 1: 1 F or 2 Ds
     #   OR less than 95% attendance over last 45 days
     #   (no discipline involved)
     tier_one_triggers = [
-      (:academic if data[:course_failures] == 1 and data[:course_ds] >= 2),
+      (:academic if data[:course_failures] == 1 || data[:course_ds] >= 2),
```
kevinrobinson: This is the separate bug for levels.

kevinrobinson:
This also fixes a bug with the level 1 calculation, where it would only match and adds a slew of test cases to that method.

kevinrobinson:
selfie

PR #2134, opened 2018-09-27
Revert mistaken change to level 1 calculation

kevinrobinson:
Reverts part of https://github.com/studentinsights/studentinsights/pull/2133, I misunderstood this and the code before was correct, but the comment was wrong.


kevinrobinson:
selfie

PR #2136, opened 2018-09-28
IEP import: Add validations and add migrations for new fields

kevinrobinson:
# Who is this PR for?
educators

There's no validations that the IEP document filename and student references are consistent.  We don't store multiple documents now and so folks can't look back and we can't notify on changes, etc.

# What does this PR do?
Adds validations to the document to sanity-check the filename, and adds new fields for tracking more information about the file, allowing us to detect changes, only store new documents, and change the association to store historical documents.

After this will come:
- [ ] update association and read path
- [ ] run new import
- [ ] run backfill with all files
- [ ] add non-null constraints on new fields

Then we'll have data on how this changes over time, enabling folks to look back, us to highlights changes, etc.

### DIFF
--- a/app/models/iep_document.rb
+++ b/app/models/iep_document.rb
@@ -1,4 +1,30 @@
 class IepDocument < ActiveRecord::Base
   belongs_to :student
   validates :file_name, presence: true, uniqueness: true
+  validate :validate_file_name
+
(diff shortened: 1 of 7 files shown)


kevinrobinson:
selfie

PR #2138, opened 2018-10-01
Make explicit that order matters in event note spec

kevinrobinson:
This failed and I went to change `eq` to `contain_exactly` but order matters, so make that explicit and remove the `EventNote.all` indirection.  Not sure where flakiness could come from here, but can't reproduce easily so punting for now.


kevinrobinson:
selfie

PR #2139, opened 2018-10-01
Revert read path change that isn't migrated yet

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2137, part of the migration here isn't done yet, need to update the importer and ship that before cutting this over.


kevinrobinson:
selfie

PR #2140, opened 2018-10-01
Fix bug in IEP storer bucket config

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2139.


kevinrobinson:
selfie

PR #2141, opened 2018-10-01
IEP importer: Store all documents, digest files to track changes

kevinrobinson:
Part of https://github.com/studentinsights/studentinsights/pull/2136.

# Who is this PR for?
educators

# What does this PR do?
Updates the IEP importer to store all versions of the documents, digesting the file contents.  It also updates the key for storing this in S3 to remove information from the filename.

With https://github.com/studentinsights/studentinsights/pull/2137, the read path should support this but fall back for existing records.

After deploying this, we can run the import on existing exports, and then run a backfill on older ones.
# Screenshot (if adding a client-side feature)

# Checklists
+ [x] Author improved specs for code in need of better test coverage
+ [x]  Author included specs for new code

PR #2144, opened 2018-10-01
Fix bugs in downloading IEP PDF zips

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2141, the `mkdirp` calls for the contents of the zip files would collide with the zip filename itself.


kevinrobinson:
selfie

PR #2145, opened 2018-10-01
Add database constraints for new IepDocument fields, update demo data

kevinrobinson:
Part of https://github.com/studentinsights/studentinsights/pull/2136.

# Who is this PR for?
developers

There aren't constraints for new fields on `IepDocument`, since we needed to migrate first.

# What does this PR do?
The migration is done; adds in database constraints and updates demo data to be consistent with them.  Verified this is good to deploy to Somerville; the migration will fail on the demo site and we can rebuild the database.

Separately, limits serialization of this data in the profile page; no need to send anything other than "there is a document," so this just sends down an id if there is one and null if not.


kevinrobinson:
selfie

PR #2149, opened 2018-10-02
Add constraints and validations to Section table

kevinrobinson:
# Who is this PR for?
developers

Several of these fields must be present, and we want to make some new things that rely on this data.

# What does this PR do?
Enforce presence of required fields (based on current production data) with constraints and validations.


kevinrobinson:
selfie  this will require rebuilding the demo site

PR #2150, opened 2018-10-02
HS Levels: Filter out sections that aren't in the current term

kevinrobinson:
# Who is this PR for?
HS educators

The "match" process for matching actual sections to the four categories of subjects is a heuristic and not 100% accurate.  One problem is that the server returns all sections, regardless of whether they're in the current term.  This messes up the naive match heuristic, which can match on a section in a future term.  The UI code doesn't have data on the term right now, so it needs it or the server needs to do this.

# What does this PR do?
Removes sections not in the current term from the server response.  Also updates tests and tests setup which weren't careful about handling time before.

# Checklists
+ [x] Author checked latest in IE - Levels prototype


kevinrobinson:
selfie

PR #2151, opened 2018-10-02
Add whitelist feature to EducatorsImporter

kevinrobinson:
# Who is this PR for?
Bedford educators

Generally we want to limit access, and scope this by school.  Sometimes pilot folks are in central office positions or other kinds of positions, and we want to whitelist them for access.  Previously we've done that with patching the database manually.

# What does this PR do?
Brings this under config.

# Checklists


kevinrobinson:
selfie

PR #2152, opened 2018-10-02
Add notes about Educator school optional value"

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2151.  The current validations will prevent us from importing educators outside pilot schools (eg, central admin project leads).  Relax this to let us keep this all in code and config rather than just modifying the db.


kevinrobinson:
selfie

PR #2153, opened 2018-10-02
Cleanup on educators importer change

kevinrobinson:
Follow on to https://github.com/studentinsights/studentinsights/pull/2152, that stray `!` is a bug :\


kevinrobinson:
selfie

PR #2154, opened 2018-10-04
Fix bug with homeroom label sort order on Absence page

kevinrobinson:
# Who is this PR for?
admin educators

The sort order on the absences by homeroom doesn't match the series data.

# What does this PR do?
Fixes that.


kevinrobinson:
selfie

PR #2155, opened 2018-10-04
Fix bug with findings homeroom by slug or id

kevinrobinson:
# Who is this PR for?
K8 educators

`homerooms_controller` uses the `friendly_id` gem to patch different Rails methods and allow querying by user-friendly slug or by id.  https://github.com/studentinsights/studentinsights/pull/2022 tried to fix this but didn't fix it all the way since normal Rails type coercion is still in place and can lead to bugs like this:

```
irb> Homeroom.find_by_id('7-263')
=> #<Homeroom id: 7, name: "H-001", ... >
```

# What does this PR do?
This PR does explicit type checking and conversions to avoid this kind of bug.  Longer term, we should just remove this gem and only accept one type of param - there's no real value in the URL format here that's worth any complexity.

# Checklists

### DIFF
--- a/app/controllers/homerooms_controller.rb
+++ b/app/controllers/homerooms_controller.rb
@@ -55,13 +55,22 @@ class HomeroomsController < ApplicationController
 
   # Calling `Homeroom.friendly.find` with a string version of an id will first
   # not match on id, and will match on the name for another homeroom with that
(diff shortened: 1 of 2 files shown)


kevinrobinson:
selfie

PR #2156, opened 2018-10-04
StudentsImporter: Mark students missing from export as `missing_from_last_export`

kevinrobinson:
# Who is this PR for?
educators, originally came up for folks in New Bedford

For some schools in New Bedford, there have been student withdrawals this year.  The export code on their side expresses this by removing the students from the export list (Bedford does the same).  In Insights, `StudentsImporter` just doesn't process these rows.  This means students who have withdrawn are still shown on homeroom rosters and other places in Insights when they shouldn't be.

The root cause of this mismatch is because Somerville's export process is different - it includes all students and sets different `enrollment_status` values.

# What does this PR do?
Updates `RecordSyncer` to add a new public method, `process_unmarked_records!` which is a more generic version of the existing `delete_unmarked_records!` method.  Updates  `StudentsImporter` to use the more generic method, and to set `missing_from_last_export: true` for each Student record within scope of the import that is missing, instead of deleting the record.

On the read path, `Student.active` now also considers this bit in addition to `enrollment_status`.

# Checklists


kevinrobinson:
selfie, this has an env flag for controlling the read path, so I'll do 1) deploy / migrate 2) run import task and verify 3) update read path

PR #2158, opened 2018-10-05
Profile: Only allow revising your own notes, limit scope of change for revising

kevinrobinson:
# Who is this PR for?
educators

Folks are concerned about revising notes, and want to limit how much this is possible.  Also, currently if a user edits a note it updates the `recorded_at` and bumps it up in the feed and profile in a way that is unexpected.

# What does this PR do?
Prevents users from editing notes someone else wrote, in the UI and in the controller.

# Checklists
+ [x] Author checked latest in IE - Home page feed
+ [x] Author checked latest in IE - My notes


kevinrobinson:
selfie

PR #2161, opened 2018-10-09
Migrate to ApplicationRecord

kevinrobinson:
Part of the Rails 5 migration, this came up from wanting to run a one-off task that checks a validations across all models.


kevinrobinson:
selfie

PR #2163, opened 2018-10-09
Remove validation and constraint on Student state_id

kevinrobinson:
We don't use this anywhere, and student records can be missing this in Somerville when they're first added.  This came up with deploying https://github.com/studentinsights/studentinsights/pull/2160


kevinrobinson:
selfie

PR #2169, opened 2018-10-09
Log to Rollbar if precomputed query falls through

kevinrobinson:
Make this more visible.  These can drift if importer tasks change the students imported.


kevinrobinson:
selfie

PR #2172, opened 2018-10-11
HS Levels: Add EducatorLabel for section-based feed, enabling schoolwide access for data coordinators

kevinrobinson:
# Who is this PR for?
HS data coordinators, dept. heads potentially

Data coordinators work as teachers, but also have other responsibilities where they have access to other student data.  Right now there's no way to have both concepts.

# What does this PR do?
Adds a new `FeedFilter` based on sections.  This way HS teachers can use this for normal product features, while also having `schoolwide_access` for authorization and particular admin functions.

# Checklists

### DIFF
--- a/app/config_objects/per_district.rb
+++ b/app/config_objects/per_district.rb
@@ -131,7 +131,6 @@ class PerDistrict
     end
   end
 
-
   # In the import process, we typically only get usernames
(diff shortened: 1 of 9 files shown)


kevinrobinson:
selfie

PR #2171, opened 2018-10-11
Remove bar chart shouldcomponentupdate check

edavidsonsawyer:
# Who is this PR for?
Dashboard users

# What problem does this PR fix?
In order to prevent charts from rendering each time a user clicked on a homeroom, the dashboard bar chart component had a check to prevent updating unless the data changed. However, this isn't optimal for performance or a best practice for lifecycle methods.

# What does this PR do?
Removes the shouldComponentUpdate method from the dashboard bar chart, and instead turns off animation in the tardies dashboard (this was done already in the absence dashboard).

# Checklists

+ [x] Author checked latest in IE - Absence Dashboard
+ [x] Author checked latest in IE - Discipline Dashboard
+ [x] Author checked latest in IE - Tardies Dashboard
+ [ ] Reviewer checked latest in IE - Absence Dashboard
+ [ ] Reviewer checked latest in IE - Discipline Dashboard
+ [ ] Reviewer checked latest in IE - Tardies Dashboard

PR #2177, opened 2018-10-11
Simplify pull request template

kevinrobinson:
# Who is this PR for?
developers on the team

@edavidsonsawyer In looking at https://github.com/studentinsights/studentinsights/pull/2171, I was thinking about if it made sense to revise the pull request template, and a bit the way we work on these things.  There's some checklist items about author and reviewing doing manual QA and tests, and I thought it would make sense as a default to have the author have responsibility for manual QA and tests.  So that generally if there's a pull request, the author is confident it works correctly and is safe to ship.

Of course if there's something tricky and more hands would be important to make sure things work correctly then that's great and we can just do that :)

We could chat more on Friday, or feel free to comment on here or propose any other improvements too!

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,32 +7,23 @@
 # Screenshot (if adding a client-side feature)
 
 # Checklists
+*Which features or pages does this PR touch?*
++ [ ] Home page
++ [ ] Student Profile
++ [ ] Homeroom
++ [ ] Section
++ [ ] My Sections
++ [ ] My Notes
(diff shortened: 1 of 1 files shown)


edavidsonsawyer:
@kevinrobinson yeah, I think that makes sense! Most of these PRs are well-contained and pretty reasonable for the author to QA. On those occasions that we're pushing something really big, critical, or we can't be certain of the impact it's easy enough to ask for another look.

kevinrobinson:
@edavidsonsawyer Sounds good!  And when we need to do more manual QA or anything like that it's always just an @ away :)

I'll merge this in, thanks!

PR #2180, opened 2018-10-15
Add Rollbar notifications when catching unauthorized; update 422 page

kevinrobinson:
# Who is this PR for?
developers, educators

Add more visibility into unauthorized exceptions.  For Rails issue 21948 with rejected sign in because of stale browser tab, this update the 422 and this brands that in the case that does happen.


kevinrobinson:
selfie

PR #2182, opened 2018-10-15
Levels: Fix bug with matching on multiple courses in department, clarify matching, add link to source

kevinrobinson:
# Who is this PR for?
HS educators

When students are in multiple courses within a department, they might have a grade in one course that would meet an academic trigger to provide more support.  Yet when the Insights UI does this match, it just takes the first course that matches in the department, and so that grade doesn't appear in the table for that department.

# What does this PR do?
This PR updates the table UI to match on courses in the department, then courses with grades, and take the lowest (since the intent is to find places where students could benefit from more supports).

It also adds a link to the source code to the levels calculation, factored out in https://github.com/studentinsights/studentinsights/pull/2181 to make that clearer.

Showing link to source code:
screen shot 2018-10-15 at 12 21 12 pm

# Checklists
+ [x] Levels

### DIFF
--- a/.mention-bot
+++ b/.mention-bot
@@ -5,7 +5,7 @@
   "alwaysNotifyForPaths": [
     {
       "name": "kevinrobinson",
-      "files": ["config/**/*.rb", "Procfile", ".eslintrc"],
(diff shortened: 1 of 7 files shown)


kevinrobinson:
selfie

PR #2183, opened 2018-10-15
Levels: Try overflow hidden for Chromebook IE wiggle bug

kevinrobinson:
# Who is this PR for?
HS educators on Chromebooks

There's a bug where the layout constantly resizes and the page jumps continually.  This is from `react-virtualized` resizing, and I suspect is an interaction between the browser scrollbars and the `AutoSizer`.

# What does this PR do?
Adds in `overflow: hidden` to the container, which in Web Inspector on an impacting Chromebook resolves the issue (I haven't been able to reproduce on a VM).


kevinrobinson:
selfie

PR #2185, opened 2018-10-15
Fix firstMatch in levels page

kevinrobinson:
Fixes bug in https://github.com/studentinsights/studentinsights/pull/2184 and from https://github.com/studentinsights/studentinsights/pull/2182 originally.


kevinrobinson:
selfie

PR #2186, opened 2018-10-16
Add House, Grade, and Counselor Filters to Discipline Dashboard

edavidsonsawyer:
# Who is this PR for?

Discipline dashboard users

To be useful in high schools, educators need to be able to look at discipline incidents by house, grade, and counselor.

# What does this PR do?

Adds a top level house filter for schools that support it. cf #1661 

disciplinefilters

# Checklists

+ [x] Discipline


kevinrobinson:
@edavidsonsawyer The tests are failing here, and the checklists in the pull request description aren't finished, so I didn't review this, but let me know when I should!

re: design, in https://github.com/studentinsights/studentinsights/issues/1661#event-1622846657 the issue is to add filters for grade, house and counselor.  If you want to ship those incrementally, that's great!  If shipping incrementally, then the order matters so we don't reorder the filters on folks while shipping incrementally :)

You can use the absences dashboard as a model here, with the specific bits (eg, unexcused / incident type) coming first, and then the same other filters coming after in the same order:
screen shot 2018-10-16 at 8 48 48 am

edavidsonsawyer:
@kevinrobinson yeah, this and the other PR are still WIP, I'll ping you when they're all set. I hadn't thought about reordering the filters - I was trying to get this quickly useful for high school educators, but it shouldn't be a problem to add all filters to this PR.

app/lib/dashboard_queries.rb
```diff
@@ -96,7 +96,9 @@ def shared_student_fields(student)
       :first_name,
       :last_name,
       :grade,
-      :id
+      :id,
```
edavidsonsawyer: @kevinrobinson I wanted to call this out since it also means sending that info to the tardies dashboard. My thinking is eventually we'll want to add these filters there as well, and it doesn't do any harm, but if we want to keep this compartmentalized then this can be done specifically for the data sent in the discipline api.

edavidsonsawyer:
@kevinrobinson All right, I've tested out these filters with and without large incident counts. They all look like they work and appear when needed based on the school. When you have the chance, I'd appreciate a code review.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.test.js
```diff
@@ -1,30 +1,113 @@
...
+// work with wrapper components, that's why there's extra setup to translate
+// between
+function elWrappedInContext(el, context) {
+  const {districtKey, nowFn} = context;
+  return withNowMoment(nowFn(), (
+    <PerDistrictContainer districtKey={districtKey}>{el}</PerDistrictContainer>
   ));
+}
+
+it('displays house for SHS', () => {
```
kevinrobinson: nice tests!  👍

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.test.js
```diff
@@ -1,30 +1,113 @@
...
 import {
   createStudents,
-  testSchool
+  testSchool,
+  testHighSchool
 } from '../DashboardTestData';
 import SchoolDisciplineDashboard from './SchoolDisciplineDashboard';
-import {withDefaultNowContext} from '../../testing/NowContainer';
 
+jest.mock('react-virtualized'); // doesn't work in test without a real DOM
```
kevinrobinson: FWIW if you want to use snapshot tests on something that uses `react-virtualized` you can do that using a similar approach to the one in https://github.com/studentinsights/studentinsights/pull/2196.  It's fine to not do that here, just sharing about a new way to approach this that I was trying.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -55,12 +67,17 @@ export default class SchoolDisciplineDashboard extends React.Component {
...
+    const {selectedChart, selectedCategory, grade, house, counselor} = this.state;
     return students.filter(student => {
-      if (selectedCategory && selectedChart === 'grade' && student.grade !== selectedCategory) return false;
-      if (selectedCategory && selectedChart === 'homeroom_label' && student.homeroom_label !== selectedCategory) return false;
-      if (selectedCategory && !this.filterIncidents(student.discipline_incidents, true).length) return false;
+      if (grade !== ALL && student.grade !== grade) return false;
+      if (house !== ALL && student.house !== house) return false;
+      if (counselor !== ALL && student.counselor !== counselor) return false;
+      if (selectedCategory && shouldFilterSelectedCategory) { //used by the student list when a user selects a category within a chart
+        if (selectedChart === 'grade' && student.grade !== selectedCategory) return false;
```
kevinrobinson: This seems like there are two separate bits of state related to the grade filtering.  One is for if the user uses the select filter, and the other is if they click on a grade within a chart.

Could these be a single bit of state?  In other words, when the user clicks on a grade within a chart, would it simplify to call `this.setState({grade})` or to call `this.setsState({homeroom})`?

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -187,7 +204,7 @@ export default class SchoolDisciplineDashboard extends React.Component {
   }
 
   studentTableRows(students) {
-    const filteredStudents = this.filterStudents(students);
+    const filteredStudents = this.filterStudents(students, true);
```
kevinrobinson: If you revise the argument to be `options = {}` this can help make it more obvious what's happening at the call site.  Right now, reading this line of code, there's no indication of what `true` means.

Using the `options = {}` argument let's you write code like this at the call site:
```
this.filterStudents(students, { shouldFilterSelectedCategory: true })
```

and express explicitly what you want that function to be doing.  I'd recommend this as a good rule of thumb in general.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -201,18 +218,38 @@ export default class SchoolDisciplineDashboard extends React.Component {
...
-  allIncidentTypes(students) {
-    const incidents = students.map(student => {
-      return student.discipline_incidents ? student.discipline_incidents : [];
-    });
-    return _.uniqBy(_.flatten(incidents), 'incident_code')
-    .map(incident => incident.incident_code);
+  allIncidentTypes() {
+    const {dashboardStudents} = this.props;
+    const incidents = _.uniqBy(_.compact(_.flatten(dashboardStudents.map(student => student.discipline_incidents))), 'incident_code');
+    return incidents.map(incident => incident.incident_code).sort();
```
kevinrobinson: As a small style suggestions, if this is just trying to get all the codes, maybe flat map then uniq then sort (like you do in the other methods)?  It doesn't seem like the whole object of `incidents` is used.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.test.js
```diff
@@ -1,30 +1,113 @@
...
 
-  it('renders a date range selector', () => {
-    expect(dash.find('SelectTimeRange').length).toEqual(1);
-  });
+it('renders a date range selector', () => {
+  const context = testContext({districtKey: 'somerville'});
+  const props = {school: testSchool};
+  const el = testEl(props);
+  const dash = mount(elWrappedInContext(el, context));
+  expect(dash.find('SelectTimeRange').length > 0).toEqual(true);
```
kevinrobinson: Given that this code now works differently per district, it would be good to add at least smoke tests for that.  Here's how this works in the absences page: https://github.com/studentinsights/studentinsights/blob/master/app/assets/javascripts/school_absences/SchoolAbsencesDashboard.test.js#L211

app/assets/javascripts/school_administrator_dashboard/DashboardTestData.js
```diff
@@ -1,9 +1,16 @@
-export function testSchool() {
-  return {
-    local_id: 'HEA',
-    name: 'Arthur D. Healey'
-  };
-}
+export const testSchool = {
+  local_id: 'HEA',
+  name: 'Arthur D. Healey'
+};
```
kevinrobinson: Can you say more about changing these from functions to shared global objects?  One benefit of functions is that there's no possibility of a test polluting the value, since you're creating a new test object each time.  If it's a global object, it opens up a class of bugs where test code or real code could update the object and break completely unrelated tests.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: We could certainly do that. It might be awkward to treat selecting a grade within a chart different from selecting other categories within a chart. We would still have need state for selectedCategory, but have an exception when that category is a grade. Since there are two ways you might filter a grade and the application behavior is different depending on which is selected, it seemed to make sense to hold that as two different bits of state. 

I imagine moving forward that we might not need the grade/homeroom charts with the top level filters, in which case this problem should go away.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: Thanks for this! I was sure there was a better syntax I was missing.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
edavidsonsawyer: This was an attempt to make it more readable, since this one has to iterate through students and then incidents. Are you saying it would be preferable to have it all on one line?

app/assets/javascripts/school_administrator_dashboard/DashboardTestData.js
edavidsonsawyer: Ah, you're right. I was thinking of this as static fixture data, but there's no reason to switch while it's still a shared component.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.test.js
edavidsonsawyer: @kevinrobinson I've got most of your suggestions implemented, but the CI build is hassling me about these snapshots. Is there something I should be sending in this branch to be able to write snapshots?

kevinrobinson: @edavidsonsawyer Yeah!  So with snapshots, you need to update this locally and push up the changes.

Conceptually, snapshots are the same as any other test - your PR has to include how the expectations for the tests are different.  The only wrinkle is that Jest makes the snapshot expectations for you.  It should prompt you to do this when you run `yarn test` locally, something like this:

image

Here's a direct link to the Jest docs about this: https://jestjs.io/docs/en/snapshot-testing.html#updating-snapshots or if you our setup isn't working quite right or is confusing, let me know and I can take a look with you so we can make this work nicely and remove any friction.

edavidsonsawyer: @kevinrobinson got it now, sorry that took so long. I forgot to merge master into this branch and it was messing up the snapshots here, but not locally.

kevinrobinson: @edavidsonsawyer no worries!  Is this ready to change from WIP and to deploy now?  I can do that later today if this is ready to 🚢 👍

edavidsonsawyer: @kevinrobinson sure is! thanks again for the review!

kevinrobinson: no problemo!

PR #2191, opened 2018-10-16
Levels: Add snapshots task to capture changes over time

kevinrobinson:
# Who is this PR for?
HS educators

The levels calculation changes, and as section assignments change, we don't have enough information preserved to compute this over time, and see how this changes for a student, or how services and supports change over time in response.

# What does this PR do?
Adds a rake task to snapshot these into a model.

This also refactors `SomervilleHighTiers` to allow calling it an an unsafe way without an authorization check for tasks like this.

# Checklists
+ [x] Levels page


kevinrobinson:
selfie

PR #2192, opened 2018-10-16
Feed: Add ELL-based feed filter for teachers in NEST and ELL department

kevinrobinson:
# Who is this PR for?
teachers in NEST and ELL department at HS

Teachers with schoolwide access who work only with ELL students see students in their feed that they won't focus on, since they're outside their core caseload.

# What does this PR do?
Adds a new `FeedFilter` to filter feed students based on their ELL status.

# Checklists
+ [x] Home page


kevinrobinson:
selfie

PR #2194, opened 2018-10-16
WIP - Student searchbar: Migrate to react-autocomplete

kevinrobinson:
# Who is this PR for?
educators

Chips away at jQuery autocomplete dependency.  Right now the way the student searchbar matches it won't match last name, first name or with partial matches.

# What does this PR do?
Updates the student searchbar so that it matches any part of the student name in any order (eg, either last name or first name first is fine).  Replaces `autocomplete` with `react-autocomplete`, keeping the same `sessionStorage` caching.

Also updates the district page, since it had relied on the global searchbar code too.  This will be migrated to client-rendering in the future, so making this factoring perfect isn't a goal in this PR.

Functionally, revise the label text ahead of potentially adding a new searchbar for the feed notes, and add a placeholder text to the input.

# Checklists
+ [x] Core, navbar
+ [x] District page
+ [ ] Included specs for changes
+ [ ] Improved specs for existing code in need of better test coverage
+ [ ] Manual testing made more sense here


kevinrobinson:
Closing, moving this stale work forward in https://github.com/studentinsights/studentinsights/pull/2485.

PR #2195, opened 2018-10-18
Work around Google Forms bug with losing data within fields

kevinrobinson:
# Who is this PR for?
HS educators and students

Running the student meeting importer this morning, it had a few rows with invalid student local ids, and so I looked into them.  Turns out this appears to be coming from a bug in Google Forms :(

In the Forms UI, it appeared that a specific value was entered for the `Student ID Number` field that was invalid.  This showed up in the individual and aggregate responses views, and also when downloading the file as a CSV.  At first I figured this was a data entry mistake.

To communicate about this with teachers, I exported the Form data into Google Sheets, and went to highlight the rows that were invalid so they could choose to fix them or leave them.  Surprisingly, one of the invalid student ID numbers was not in the spreadsheet.  It appeared as if Forms had truncated the last few characters, while the exported Sheet had a full proper student ID value.

This unfortunately seems like it must be a bug in the Forms product and so I reported this as best I could in the product forums here: https://productforums.google.com/forum/#!topic/google-education/FDVz_uM7-I8;context-place=forum/google-education

Separately, when exporting from Google Forms to Google Sheets, Forms changes the format of the data and renames the username field, as well as changes the timestamp format.  This PR adds ways to deal with both and captures this in comments.

# Checklists
+ [x] Student Profile


kevinrobinson:
selfie

PR #2196, opened 2018-10-19
Levels: Fix bug with rendering grade strengths

kevinrobinson:
# Who is this PR for?
HS educators

On IE, background colors for "strengths" grades aren't being set, resulting in those grades not appearing on the page.  This doesn't impact other browsers, which interpret `rgb(x,y,z,a)` as if it were `rgba(x,y,z,a)`, which IE ignores the rule.

# What does this PR do?
Fixes it.  Adds an unrelated test to the ordering function in JS, since that's what I initially thought the problem was.

This PR also adds a new method for testing, `unvirtualize` which can be used in tests to swap in an naive unvirtualized implementation of `react-virtualized` `<Table />` that can be used for snapshot testing.  And this adds a snapshot test for `<StudentsLevelsTable />` that will catch regressions in individual styling of the cells going forward.

screen shot 2018-10-18 at 7 35 39 pm

# Checklists
+ [x] Levels page

### DIFF
--- a/app/assets/javascripts/levels/StudentLevelsTable.js
+++ b/app/assets/javascripts/levels/StudentLevelsTable.js
@@ -20,7 +20,6 @@ import {
   STUDY_SKILLS
 } from './Courses';
 
-
 // Render a virtualized table, with information and triggers on the
 // left columns and supports on the right.  Exports the description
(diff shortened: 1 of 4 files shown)


kevinrobinson:
selfie

PR #2197, opened 2018-10-19
Update Levels link to be from EducatorLabel, to enable one-off access too

kevinrobinson:
# Who is this PR for?
districtwide admin, project lead

The SHS Levels link is switched different than with labels, so there's no way to add one-off access.

# What does this PR do?
Moves this into `dynamic_labels` and stills also it to be set with a normal static `EducatorLabel`.

Note that this is just about the link, not authorization.

# Checklists
+ [x] Levels
+ [x] Navbar

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -31,13 +31,13 @@ Layout/EmptyLines:
 Layout/Tab:
   Enabled: true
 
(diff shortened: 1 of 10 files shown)


kevinrobinson:
selfie

PR #2200, opened 2018-10-21
Roster: Improve precomputing keying for school overview, simplify authorization

kevinrobinson:
# Who is this PR for?
schoolwide admin, mostly at SHS

There have been instances of the school roster page falling back to doing queries that should be precomputed (since they're pretty slow).  This is happening because the way this was initially written relied on some coordinated timing that may no longer be true.  This code also uses some older duplicate authorization code.

# What does this PR do?
Reworks the precomputing:

1. Adds a new keying strategy that removes the "date" from the key. Relaxes a uniqueness constraint on `key` to support this.
2. Updates the write path to use the new keying strategy, and updates the read path to query for the latest document at the key, adding in some other guardrails around freshness of the document.
3. Remove the duplicate authorization code
4. Add precompute job to indexing step at the end of the import task

# Checklists
+ [x] School Overview
+ [x] Import task

### DIFF
--- a/app/controllers/schools_controller.rb
+++ b/app/controllers/schools_controller.rb
@@ -100,25 +100,17 @@ class SchoolsController < ApplicationController
   # Results an array of student_hashes.
   def load_precomputed_student_hashes(authorized_student_ids)
     begin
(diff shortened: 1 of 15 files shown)


kevinrobinson:
selfie

PR #2201, opened 2018-10-22
Add notes review script

kevinrobinson:
# Who is this PR for?
educator teams who want to reflect on ways they're using notes in Student Insights

# What does this PR do?
Adds a command-line script for getting them this data, for sharing via district Google Drive.

# Checklists
+ [x] Add spec
+ [x] Manual testing made more sense here

PR #2203, opened 2018-10-22
Fix whitespace bugs from NotesReview

kevinrobinson:
Bug from https://github.com/studentinsights/studentinsights/pull/2201, fixing https://travis-ci.org/studentinsights/studentinsights.


kevinrobinson:
selfie

PR #2202, opened 2018-10-22
Profile: Improve display of ELL status and WIDA level

kevinrobinson:
# Who is this PR for?
educators

# What problem does this PR fix?
ELL teachers at SHS asked to revise the way ELL and FLEP data is represented to clarify for classroom teachers in particular.

# What does this PR do?
It revises the single link about language status and factors that out into `<LanguageStatusLink />`.  It revises the designations to be either "English Learner", "Former English Learner" or "Fluent English".  For English learners, it removes the text description of the labels in favor of the numeric level only.  For students in Somerville with their `program` set to "SEIP" or "Wavered SEIP" these are blacklisted and not shown in Insights, based on feedback from SHS ELL teachers noted in comments.

There are related changes made to the `<AccessPanel />` component, and several different kinds of tests and refactoring to improve how this data is accessed in `language.js`.

# Screenshot (if adding a client-side feature)
### Example from story
screen shot 2018-10-22 at 2 55 58 pm

### Story and test showing combinations across districts and values
screen shot 2018-10-22 at 2 00 40 pm

### Example modal
screen shot 2018-10-22 at 1 35 25 pm

# Checklists
*Which features or pages does this PR touch?*
+ [x] Student Profile
+ [x] Levels page
+ [x] Included specs for changes
+ [x] Improved specs for existing code in need of better test coverage

PR #2205, opened 2018-10-22
Student export: Add ELL entry and transition date for Somerville

kevinrobinson:
For adding another bit to follow-on to https://github.com/studentinsights/studentinsights/pull/2202.


kevinrobinson:
selfie

PR #2210, opened 2018-10-26
Add dependent:destroy to EventNote for revisions, update student meeting importer column

kevinrobinson:
# Who is this PR for?
HS educators and students

The importer wasn't working because of a column change so this updates that.  But it also wasn't working because when for old notes that have revisions, `event_note.destroy!` would raise as violating a foreign key constraint since it would orphan references on `EventNoteRevisions`.

This means an educator was revising a note imported from the survey (which we might not want, but don't currently restrict).  A future improvement might be to check for that in the importer and log about it, to close the loop back with the educator, but this PR doesn't do this and just destroys the revision with no logging or notice (which was the intent of the existing importer code).

# What does this PR do?
Adds `dependent: destroy` to the `EventNote` model for revisions.  And updates the column name for the student local id.

# Checklists
+ [x] Student meeting importer


kevinrobinson:
selfie

PR #2212, opened 2018-10-26
StudentsImporter: Update StudentRow to set missing_from_export: false depending on export config

kevinrobinson:
Related to https://github.com/studentinsights/studentinsights/pull/2211/files, but split out to see the diffs.  See that PR for description.

Updates `StudentRow` to set `missing_from_last_export: false` for imported rows, depending on the export setting for the district.


kevinrobinson:
selfie

PR #2215, opened 2018-10-29
School overview: Decouple query from homeroom page, refactor querying

kevinrobinson:
# Who is this PR for?
school admin

The school overview page is still falling through in some circumstances, and not reading the precomputed data.  A bunch of this code is doing loops with n+1 queries naively, and might be straightforward to optimize and remove the complexity of precomputing altogether, but this is coupled with the homeroom page and so involves updating other callers too.

# What does this PR do?
Decouples the school overview query methods from the homeroom page queries, inlining the relevant bits into the homeroom controller, and refactoring the mixin into a new `SchoolOverviewQueries` class.

This class does one optimization to unroll repeated queries to serialize students, and to query for most recent events, batching those into single queries.

Separately, this adds a new query param to the controller to allow forcing on-demand querying, to get a rough sense of what the delta is from the end-user's perspective on production datasets.

# Checklists
+ [x] Homeroom
+ [x] School Overview


kevinrobinson:
selfie

PR #2216, opened 2018-10-29
Allow querying on demand for school overview, and echo that back

kevinrobinson:
Follow on to https://github.com/studentinsights/studentinsights/pull/2215, which forgot to thread query param from HTML page to api request.


kevinrobinson:
selfie

PR #2217, opened 2018-10-29
StudentsImporter: Improve improting for house, sped liaison, counselor and update validations

kevinrobinson:
# Who is this PR for?
educators, primarily in Bedford

There are some special cases in Bedford around these values, and in the process of working around them, I found that there are many empty string values and no validations on these.

# What does this PR do?
Before this PR, I manually patched the `Student` records in each district to set empty strings or "N/A" for `sped_liaison` in Bedford to `nil`.

This PR adds validations enforcing that.  It also moves the parsing code for `sped_liaison` and `counselor` into `PerDistrict`.  And it disables importing `sped_liaison` for Bedford, since these values are all set to "N/A" right now.  Even though we are getting data in this field, it's not meaningful, so we effectively are not receiving this data at all and no value in importing it.

# Checklists
+ [x] Core 

### DIFF
--- a/app/config_objects/per_district.rb
+++ b/app/config_objects/per_district.rb
@@ -269,11 +269,11 @@ class PerDistrict
   # the full name).
   def parse_counselor_during_import(value)
     if @district_key == SOMERVILLE # eg, Robinson, Kevin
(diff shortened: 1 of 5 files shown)


kevinrobinson:
selfie

PR #2218, opened 2018-10-29
Profile: Hide sheltered immersion from program

kevinrobinson:
# Who is this PR for?
Bedford educators

The `program_assigned` field contains a mix of different information across districts.  The profile page now uses it to tell information about special education placement, and so for Somerville we hide values here about sheltered language immersion.  The values for this are different in Bedford.

Also, before shipping https://github.com/studentinsights/studentinsights/pull/2217, for students in Bedford where "N/A" was exported as the `sped_liaison`, this would incorrectly suggest there was an IEP with more info because it appeared as if there was a SPED liaison.

# What does this PR do?
Hides the value for Bedford.  Also doesn't consider `sped_liaison` when deciding whether to show IEP link.  If that's all there is, it's likely a data quality bug upstream.

# Checklists
+ [x] Student Profile

### DIFF
--- a/app/assets/javascripts/helpers/specialEducation.js
+++ b/app/assets/javascripts/helpers/specialEducation.js
@@ -9,7 +9,7 @@ export function hasAnySpecialEducationData(student, maybeIepDocument) {
   if (disability) return true;
   if (levelOfNeed) return true;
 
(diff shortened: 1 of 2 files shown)


kevinrobinson:
selfie

PR #2223, opened 2018-10-30
AssessmentImporter: Update MCAS for Bedford

kevinrobinson:
# Who is this PR for?
Bedford educators

Folks in Bedford want to import MCAS assessment data, but the format of their export is different than Somerville's, so we can't just turn this on.

# What does this PR do?
Makes the MCAS import process work differently using `PerDistrict`:
1. Update the `X2AssessmentImporter` to parse different things within the same file with `PerDistrict#choose_assessment_importer_row_class`.
2. Update the `McasRow` class to be more explicit about matching, factors out differences across districts into `PerDistrict#normalized_subject_from_mcas_export`.

Some other related cleanup:
3. Removes sanitization of `assessment_growth` currently applied across all assessments, since it conflicts with Bedford MCAS data.  For now this is just removed so the only guard is database type constraint (there are no validations).
4. Removes `:assessment_date` as a field handled by `CsvRowCleaner`, and push this out to individual importers (eg, `DibelsRow` and `AccessRow`).  This is aiming to remove that class altogether to make importers more explicit about any transformations.

# Checklists
+ [x] Profile
+ [x] Overview
+ [x] X2AssessmentImporter


kevinrobinson:
hmm, got what looks like a flaky test error: https://travis-ci.org/studentinsights/studentinsights/builds/448425967?utm_source=github_status&utm_medium=notification, going to restart it for now and ignore

kevinrobinson:
selfie

PR #2224, opened 2018-10-30
Improve RecordSyncer alert notification context

kevinrobinson:
# Who is this PR for?
developers

Not enough context to know what importer or model is triggering this.

# What does this PR do?
Adds in the call stack.

Not sure this is super useful at all, but this is a tiny change to try it out.


kevinrobinson:
selfie

PR #2225, opened 2018-10-30
Allow setting asset host in ENV, for adding CDN

kevinrobinson:
For incrementally deploying config to add a CDN for assets.


kevinrobinson:
selfie

PR #2227, opened 2018-11-01
Remove implicit type coercion for MCAS rows

kevinrobinson:
# Who is this PR for?
educators, came up in new Bedford import

There's implicit type coercion in the importer; this can let unexpected values be entered as `0` integer values when they probably should be `nil`.

# What does this PR do?
Explicitly converts types in the importer, adds a validation to the model for `0` values which don't make sense conceptually.

# Checklists
+ [x] Student Profile
+ [x] Overview


kevinrobinson:
selfie

PR #2229, opened 2018-11-02
Update school_local_id for Bedford assessment export

kevinrobinson:
Update this to fix bug that left this empty for all rows.


kevinrobinson:
selfie

PR #2237, opened 2018-11-05
Profile: Fix propTypes warning on team membership

kevinrobinson:
Fixing a noisy propType warning, adding a test.


kevinrobinson:
selfie

PR #2239, opened 2018-11-06
Fix bug with sending null valued SGP to UI

kevinrobinson:
# Who is this PR for?
Bedford educators

Bedford doesn't have SGP data, but these charts still appear on the profile page.  This is from the server sending down data points that have null, when it should be filtering these out.

# What does this PR do?
Filters these values out; same for scale score.

# Checklists
+ [x] Student Profile


kevinrobinson:
selfie

PR #2241, opened 2018-11-08
Remove Jekyll config added through Github UI; move to gh-pages instead

kevinrobinson:
# Who is this PR for?
developers

I tried setting this to the `docs` folder first, but think it's better to keep code and website on distinct branches.  This file was committed from the UI; this removes it since this is on the `gh-pages` branch now.


kevinrobinson:
selfie

PR #2243, opened 2018-11-16
Search notes: Add initial version, behind EducatorLabel feature switch

kevinrobinson:
# Who is this PR for?
all educators, came up first with HS counselors

There's no great way for a team of educators to look back and reflect on how they're talking about particular groups of students, or how they're talking about them within particular support structures (eg, note types).  There's also no great way for educators to look back at what topics or issues have come up across students (eg, who they've talked about financial aid with, who is getting in-classroom supports for oral reading fluency).  This is one way to explore helping out.

One mock of what this might look like is this (not all done in this PR):
14 home-search-financial-aid
15 home search financial aid notes

# What does this PR do?
Adds a new page at `/search/notes` for searching notes, and an endpoint for responding to queries.  It does not allow searching restricted notes (this would be something to add, just separately with a closer security review).

The server code also adds a new feature - searching across your own notes, all your students, or home feed students only.  The UI doesn't use this yet, but the intent is to support a gradual release for different settings; that's probably something to wire up next week.

I also ran a few ad-hoc tests on production data, and the query seems fast enough to start, even for ones that would return larger data sets.  This is likely something we'll need to profile with a `PerfTest` across roles since the authorization checks are different costs based on roles.  There's a bunch of space to optimize there if we need to later.

Separately, this PR factors out the serialization of event note cards from the feed, so we can re-use that format and the UI rendering code completely here.

screen shot 2018-11-16 at 12 29 19 pm

# Checklists
+ [x] Search notes page
+ [x] Home page


kevinrobinson:
selfie

PR #2244, opened 2018-11-17
Add Rack::Attack

kevinrobinson:
# Who is this PR for?
students, families

# What does this PR do?
Adds more layers of defense against attacks from attack bots, mostly throttling repeated login requests.

# Checklists
+ [x] Core 

This code will run locally and in tests, but production config is a bit different.  Will deploy this to demo first and then roll out after that.


kevinrobinson:
selfie

PR #2247, opened 2018-11-20
Block more php and asp scripts in Rack

kevinrobinson:
Blocking a common source of noise and attacks.

Also logging blocked requests.


kevinrobinson:
selfie

PR #2252, opened 2018-11-20
Search: Swap in fulltext, allow querying by time range, clamp start_time and limit

kevinrobinson:
# Who is this PR for?
educators

Searching notes didn't work with case mismatches, and didn't do any processing like stemming.  It also didn't allow setting the time range, and would let clients query for all data.

# What does this PR do?
First, it uses Postgres fulltext operators for matching text instead of LIKE.  This includes stemming and filtering stop words.  There's no ranking; all results are still time-ordered.  I looked at [websearch_to_tsquery](https://www.postgresql.org/docs/11/textsearch-controls.html) which would be great but this is new in Postgres 11.  It was just released as [beta on Heroku](https://devcenter.heroku.com/changelog-items/1511), but this requires a database migration so cutting it from this PR.

Second, it allows searching by time range, defaulting to just the school year.

Third, the server clamps the time range that the client sends and the limit that it sends as well.

screen shot 2018-11-20 at 5 44 54 pm

# Checklists
+ [x] Search notes

### DIFF
--- a/app/assets/javascripts/components/SelectTimeRange.js
+++ b/app/assets/javascripts/components/SelectTimeRange.js
@@ -1,19 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import SimpleFilterSelect from './SimpleFilterSelect';
-import {firstDayOfSchoolForMoment} from '../helpers/schoolYear';
+import {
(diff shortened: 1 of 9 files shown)


kevinrobinson:
selfie

PR #2253, opened 2018-11-21
Remove Mixpanel

kevinrobinson:
This removes Mixpanel, which we had used to look at aggregate product usage data (personally identifiable was not sent; no student data was sent).

Remove this is great because it a) reduces security risk, b) reduces privacy risk and c) we haven't really used the JS data anyway.  We can use within-app data for this and talk to people instead!


kevinrobinson:
selfie

PR #2254, opened 2018-11-21
Add `rel` for target blank

kevinrobinson:
This is security improvement (eg https://stackoverflow.com/questions/50709625/link-with-target-blank-and-rel-noopener-noreferrer-still-vulnerable#50709724)


kevinrobinson:
selfie

PR #2258, opened 2018-11-21
Cleanup app.json and old ENV variables

kevinrobinson:
And enable one-click deploys for curious people again!

screen shot 2018-11-21 at 1 18 46 pm


kevinrobinson:
selfie

PR #2259, opened 2018-11-21
Cleanup: Remove ActionMailer config from production

kevinrobinson:
Remove this and related config; we only use the simpler Mailgun HTTPS API right now.


kevinrobinson:
selfie

PR #2262, opened 2018-11-21
Re-enable New Bedford assessment import

kevinrobinson:
The upstream issues with their SIS are resolved now.


kevinrobinson:
selfie

PR #2263, opened 2018-11-21
Remove Google Font loader

kevinrobinson:
Removing dependency on external services and whitelisting in CSP.  We can distribute these to a CDN separately, and if they load a few more bits the first time they are cached that's a good tradeoff to be able to control this ourselves.


kevinrobinson:
selfie

PR #2264, opened 2018-11-21
Fix weekly report bug for notes with no school

kevinrobinson:
follow on to https://github.com/studentinsights/studentinsights/pull/2253


kevinrobinson:
selfie

PR #2265, opened 2018-11-22
Try something else for flaky timing spec

kevinrobinson:
Aiming at helping https://travis-ci.org/studentinsights/studentinsights/builds/458182204#L2665.


kevinrobinson:
selfie

PR #2271, opened 2018-11-27
Remove shoulda-helpers and launchy

kevinrobinson:
# Who is this PR for?
developers

cleanup, removing dependencies

# What does this PR do?
Removes use of `should-matchers`, which was only in these files.  Also removes `launchy` which was added prior to this repo and isn't being used.

# Checklists


kevinrobinson:
selfie

PR #2274, opened 2018-11-27
Enable multifactor authorization flows using Authenticator app and SMS

kevinrobinson:
This is a replacement for https://github.com/studentinsights/studentinsights/pull/2248, since it batches in much more automated and manual testing.

# Who is this PR for?
students, families, districtwide admin

We don't have multifactor authentication enabled for any users yet, and it's hard to update the login flow since it's embedded in the core layout, and because changes here involve interop with Devise.

# What does this PR do?
First, factors out the sign in form to its own page, outside of the navbar. This enables better isolation here, and some styling is improved and cleaned in the process.

Second, this updates the form and authentication setup in Devise and our strategy to allow an additional `login_code` parameter. For the standard login, this is set to a placeholder value on the form (since Devise requires all parameters are always passed).

Third, this adds JS to the sign in page that allows a multifactor login flow (see below). This PR does not enable this flow for any users.

Fourth, this adds server models and endpoints for two kinds of multifactor authentication flows: an using an authenticator app or SMS.  The authenticator app is more secure but more friction, with the idea that it can be for developer accounts to start, while SMS could be enabled for project leads and districtwide admin.  Migrating users will take a while and have to be done in phases; this PR assumes this is not enabled for anyone to start.  These endpoints are protected by `Rack::Attack` and by `ConsistentTiming`.

Fifth, this adds more layers of tests for all files touched, including more edge cases to the authentication specs, specifically ones that test whether it guards against invalid upstream and downstream behavior. It also factors out the actual call to the LDAP server into `LdapAuthenticator` and adds another layers tests there as well.

### Sign in styling, with "Use multifactor login" link
This form works the same as before, just also sending a placeholder `login_code` to keep the Devise API the same across both flows.
screen shot 2018-11-27 at 7 13 24 pm

### Multifactor sign in, step 1
No server interaction.
screen shot 2018-11-27 at 7 13 30 pm

### Multifactor sign in, step 2
screen shot 2018-11-27 at 7 13 50 pm

### Updated invalid login message, same for all reasons, including multifactor required
screen shot 2018-11-27 at 7 14 09 pm

### Required fields warnings across browsers
screen shot 2018-11-20 at 12 26 04 pm
screen shot 2018-11-20 at 12 25 44 pm
screen shot 2018-11-20 at 12 25 35 pm

# Checklists
+ [x] Sign in
+ [x] Navbar
+ [x] Core 

Deploying requires some config changes as well.  To deploy, I'll deploy to demo first and do more manual testing there as well before deploying this to any production sites.  The first production deployments will also omit configuration that prevent SMS authentication altogether.


kevinrobinson:
🚢

PR #2275, opened 2018-11-27
Check in Gemfile.lock and fix bundler version

kevinrobinson:
Fix for https://github.com/studentinsights/studentinsights/pull/2271.


kevinrobinson:
selfie

PR #2280, opened 2018-12-01
Fix CSS bug with status message

kevinrobinson:
This styling regressed in https://github.com/studentinsights/studentinsights/pull/2274


kevinrobinson:
selfie

PR #2281, opened 2018-12-03
Update instructions to crawlers; remove referrer altogether

kevinrobinson:
# Who is this PR for?
students, families

# What does this PR do?
Tell the nice search engines not to crawl anything beyond the root page; make referrer header policy stricter and remove it altogether.

# Checklists
+ [x] Manual testing made more sense here

PR #2283, opened 2018-12-03
Rails wants origin headers

kevinrobinson:
Rails wants this: https://rollbar.com/somerville-teacher-tool/somerville-teacher-tool/items/307/ so let's revert part of https://github.com/studentinsights/studentinsights/pull/2281


kevinrobinson:
selfie

PR #2284, opened 2018-12-03
Cleanup: Remove server-side profile v2 code

kevinrobinson:
# Who is this PR for?
developers

Removes surface area of endpoints and of product code.

# What does this PR do?
Simplify and remove old code, cleaning up from https://github.com/studentinsights/studentinsights/issues/1990.  This also removes the `Authorizer#is_authorized_for_deprecated_transition_note_ui` method which was marked deprecated and was now no longer used after this cleanup.

This makes the minimal change to `App.js` for updated routing; JS code can be cleaned up separately.

# Checklists
+ [x] Student Profile


kevinrobinson:
selfie

PR #2285, opened 2018-12-04
Levels: Fix sort bug for counselor note in table

kevinrobinson:
# Who is this PR for?
HS educators

# What does this PR do?
Fix bug with sorting by last counselor note in HS levels page.

# Checklists
+ [x] Levels


kevinrobinson:
selfie

PR #2289, opened 2018-12-06
Update reading and math chart labels so they're clearer for HS studen…

kevinrobinson:
…ts in Testing tab

# Who is this PR for?
HS educators

The labels aren't explicit about what subject test the chart is for, which matters for HS when these are in the same testing tab.

# What does this PR do?
Clarifies chart titles.

# Checklists
+ [x] Student Profile


kevinrobinson:
selfie

PR #2295, opened 2018-12-08
Update logging to log hashed session and educator identifiers, remove view logging

kevinrobinson:
# Who is this PR for?
developers

# What problem does this PR fix?
We can't easily see requests across a session when debugging, and there's noise in logs about view rendering that we don't use.

# What does this PR do?
Adds two log tags for session and educator identifiers.  Both are hashed from the real value using a secret to reduce risk of exposure if logs were leaked, and shortened for legibility.  These values are write-only and never read or used for anything.

# Screenshot (if adding a client-side feature)
Example of log output, showing request, session and session identifiers.
```
[r:bcf387ed] [s:ca347062] [e:7930d8d2] Started GET "/home" for 127.0.0.1 at 2018-12-08 12:13:02 -0500
[r:bcf387ed] [s:ca347062] [e:7930d8d2] Processing by UiController#ui as HTML
[r:bcf387ed] [s:ca347062] [e:7930d8d2] Completed 200 OK in 72ms (Views: 63.5ms | ActiveRecord: 3.2ms)
...
[r:a9b20f92] [s:ca347062] [e:7930d8d2] Started GET "/api/home/feed_json?educator_id=[FILTERED]&limit=[FILTERED]&time_now=[FILTERED]" for 127.0.0.1 at 2018-12-08 12:13:04 -0500
[r:a9b20f92] [s:ca347062] [e:7930d8d2] Processing by HomeController#feed_json as JSON
...
[r:42046f1d] [s:ca347062] [e:7930d8d2] Started GET "/api/home/students_with_high_absences_json?educator_id=[FILTERED]&limit=[FILTERED]" for 127.0.0.1 at 2018-12-08 12:13:04 -0500
```

Updated this so it still logs "no value" if these can't be read:
```
[r:11b556ce] [s:] [e:]
```

# Checklists
*Which features or pages does this PR touch?*
+ [x] Core 

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Manual testing made more sense here

PR #2296, opened 2018-12-08
Fix config bug for logger change

kevinrobinson:
Fix bug in https://github.com/studentinsights/studentinsights/pull/2295 after refactoring.


kevinrobinson:
selfie

PR #2294, opened 2018-12-08
504: Add initial importer for Somerville, minimal UI for data quality checks

kevinrobinson:
# Who is this PR for?
students, families, educators - HS counselors to start

# What problem does this PR fix?
504 data isn't visible in Student Insights.

# What does this PR do?
Builds on https://github.com/studentinsights/studentinsights/pull/2287 and adds an initial importer for the "ed plans" from Somerville but not yet for accommodations.  Adds a minimal UI to show this when enabled by the `enable_viewing_504_data_in_profile` label, so we can start looking at data quality with folks, which has been a concern previously.

# Screenshot (if adding a client-side feature)
screen shot 2018-12-08 at 10 53 17 am

# Checklists
*Which features or pages does this PR touch?*
+ [x] Student Profile

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Improved specs for existing code in need of better test coverage
+ [x] Manual testing made more sense here

PR #2298, opened 2018-12-08
Fix bug in EdPlan importer client call

kevinrobinson:
Fix for https://github.com/studentinsights/studentinsights/pull/2294


kevinrobinson:
selfie

PR #2297, opened 2018-12-08
Fix bug with educator_id logging

kevinrobinson:
Another fix for https://github.com/studentinsights/studentinsights/pull/2296.

PR #2299, opened 2018-12-08
Fix educator id logging

kevinrobinson:
One more for https://github.com/studentinsights/studentinsights/pull/2297, tried adding a spec but didn't see a clear way to get at the log output here.  The approach for testing log scrubbing or Rollbar log scrubbing didn't work the same as I hoped they might.


kevinrobinson:
selfie

PR #2300, opened 2018-12-08
Fix reading downloaded ed plan file

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2298; other importers use `CsvDownloader`, but this keeps this more direct.


kevinrobinson:
selfie

PR #2304, opened 2018-12-10
Search: Allow marking notes as restricted

kevinrobinson:
# Who is this PR for?
students, families.  first users SHS counselors

If notes with sensitive information are entered but not marked as restricted, there's no way for educators to resolve this without emailing for help.  This works well and they can do this quickly, but we should support educators doing this directly.

Adding search magnifies this problem since it's simpler to search for particular topics that might be sensitive, and so we can use search to enable educators to resolve any privacy concerns for students before releasing it more widely.

# What does this PR do?
Adds a "Mark restricted" link to notes on the search page only.

This involves changing `FeedView` to support threading per-note props at render time, and then adds a new `MutableFeedView` that adds in this UI element and handles making server requests and optimistically updating the local UI as well.  This PR then makes use of this in `SearchNotesPage` only.  The metadata for search results doesn't update in response to this, and that's something to improve in a separate PR; this is aimed at enabling the most critical use case only.

This also fixes an existing bug in `SelectTimeRange` impacting the search page - the `momentRange` fn was mutating a moment that it was passed, which led to excluding notes from the current day in search.

screen shot 2018-12-10 at 8 47 59 am
screen shot 2018-12-10 at 8 48 04 am

# Checklists
+ [x] Home page
+ [x] Search notes


kevinrobinson:
selfie

PR #2307, opened 2018-12-10
504: Remove unused data from importer, tighten some validations

kevinrobinson:
Building on https://github.com/studentinsights/studentinsights/pull/2287.

# Who is this PR for?
developers

# What does this PR do?
Removes data that either a) isn't used or b) always has the same value from the importer and model, and tightens some validations on fields that are always present.  Also adds `dependent: destroy` for accommodations.

These validations require removing 3 records from the production database; this is already done.

# Checklists
+ [x] Student Profile


kevinrobinson:
selfie

PR #2310, opened 2018-12-10
Disable 504 link if no actual document (edge case)

kevinrobinson:
This comes up in some development data, but not in real production data.


kevinrobinson:
selfie

PR #2313, opened 2018-12-11
District: Rework district page to React, swap big search for embedded work board

kevinrobinson:
This will conflict with https://github.com/studentinsights/studentinsights/pull/2194, the intention is to ship this first, then revisit https://github.com/studentinsights/studentinsights/pull/2194 which should be a bit simpler.

# Who is this PR for?
project leads, districtwide users

# What does this PR do?
Reworks the districtwide page, which was server rendered and relied on code in `index.js`, so that it's rendered in React and the dependency on `index.js` can be removed.

Also removes the redundant search bar, and instead embeds the project work board from the public website.

screen shot 2018-12-11 at 1 23 26 pm

# Checklists
+ [x] District


kevinrobinson:
selfie

PR #2315, opened 2018-12-11
504: Update plan_504 method on student record to read EdPlan for Somerville

kevinrobinson:
Related to https://github.com/studentinsights/studentinsights/pull/2287.

# Who is this PR for?
Somerville educators

The `plan_504` value coming from the SIS export isn't accurate; it overcounts students who have an ed plan set to "previous" status.

# What does this PR do?
To work around this, this PR adds a method to the `Student` model to patch this for Somerville.  We could update the student export to solve this upstream, but this is a faster change and self-contained.  Aside from the complexity downside, this also means doing another query for each read of `plan_504`, which could be expensive for calls like `Student.active.as_json` with an eager include.

This also updates the JS function that decides about showing "504 plan" so that it is stricter, which isn't directly related to the server-side change (since the server-side change for Somerville keeps the same API with the UI).

# Checklists
+ [x] Student Profile
+ [x] School Overview

### DIFF
--- a/app/assets/javascripts/helpers/PerDistrict.js
+++ b/app/assets/javascripts/helpers/PerDistrict.js
@@ -48,18 +48,22 @@ export function orderedDisabilityValues(districtKey) {
   return ORDERED_DISABILITY_VALUES_MAP[districtKey] || [];
 }
 
-// Includes if they "exited" the 504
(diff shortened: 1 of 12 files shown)


kevinrobinson:
selfie

PR #2317, opened 2018-12-11
Remove iframe strategy, just put work board inline

kevinrobinson:
Using an iframe was never a good idea; CSP blocks this as it should and there's no upside to relaxing the CSP for this.  I can paste it inline each week.


kevinrobinson:
selfie

PR #2321, opened 2018-12-12
Update cookie name to be different per district

kevinrobinson:
These are already isolated by domain, this just adds another layer of isolation.


kevinrobinson:
selfie

PR #2323, opened 2018-12-12
Fix usage of session key

kevinrobinson:
Fix https://github.com/studentinsights/studentinsights/pull/2321.


kevinrobinson:
selfie

PR #2324, opened 2018-12-13
504: Include 504 importer in x2 import group

kevinrobinson:
This wasn't included, which meant `thor import:start` didn't include these.


kevinrobinson:
selfie

PR #2329, opened 2018-12-17
Mulitfactor: Update TestPals to use stable TOTP secret in development and demo

kevinrobinson:
# Who is this PR for?
test users, developers

The TOTP secret for test users are generated anew each time, but this is a pain for development and demo, where we want this to be stable across builds so that users can test it.

# What does this PR do?
Fixes the secret for the `TestPals` admin user.  This only impacts dev and demo.  The 444s are intended as a weak signal that it's an artificial value (`fake` isn't allowed in the format).


kevinrobinson:
selfie

PR #2330, opened 2018-12-17
Fix test bug on profile that did not fix time during test

kevinrobinson:
This fixes a bug that is impacting test failures like https://travis-ci.org/studentinsights/studentinsights/builds/469166107?utm_source=github_status&utm_medium=notification because of today's date.  This code should all grab the time from context, so that tests work consistently.


kevinrobinson:
selfie

PR #2334, opened 2018-12-17
Profile: Update educator contacts to respect PerDistrict setting for counselor field

kevinrobinson:
# Who is this PR for?
Bedford educators

Bedford exports this value, but the data is only administrative and not meaningful

# What does this PR do?
Updates the profile to respect the `PerDistrict.js` setting for this, and updates the server method with a note about this.


kevinrobinson:
selfie

PR #2335, opened 2018-12-19
Demo: Tune down MORE_DEMO_STUDENTS for demo site

kevinrobinson:
Tuning this down to avoid hitting thresholds and spending more on the demo site.


kevinrobinson:
selfie

PR #2341, opened 2018-12-21
Relax constraint on F&P that was too strict

kevinrobinson:
This should allow multiple data points for each student.


kevinrobinson:
selfie

PR #2344, opened 2019-01-08
Fix spec setup bug with time parsing

kevinrobinson:
Fix spec setup that had an implicit "this year", addressing https://travis-ci.org/studentinsights/studentinsights/builds/476949452


kevinrobinson:
selfie

PR #2345, opened 2019-01-08
Teams: Migration to store season

kevinrobinson:
# Who is this PR for?
HS students, educators, coaches

Allows tracking current season for sports teams, not just school year.

# What does this PR do?
Adds a `season_key` field.


kevinrobinson:
selfie

PR #2349, opened 2019-01-10
Student Profile Heatmap

edavidsonsawyer:
# Who is this PR for?

Educators

Users of the discipline dashboard can look at a heatmap of aggregated discipline incidents. It may be useful to see a heatmap for an individual student to find patterns in discipline incidents.

# What does this PR do?

Adds a heatmap of a student's individual discipline incidents to the student profile in the Incident History section

heatmap_student

# Checklists
+ [x] Student Profile
+ [x] Discipline

### DIFF
--- a/app/assets/javascripts/components/DisciplineScatterPlot.js
+++ b/app/assets/javascripts/components/DisciplineScatterPlot.js
@@ -58,8 +58,8 @@ export default class DisciplineScatterPlot extends React.Component{
           }}
           title={{text: this.props.titleText}}
           yAxis={[{
-            min: this.props.yAxisMin,
-            max: this.props.yAxisMax,
+            min: 420, //7 AM
+            max: 960, //3 PM plus a gutter category
             reversed: true,
             showLastLabel: false,
             tickInterval: 60,


edavidsonsawyer:
@kevinrobinson here's the heatmap adapted slightly for the student profile. One improvement that we'll need to make is to set the school hours in one location and import into this and the dashboard component so the times don't get out of sync. I want to check in with you first, though, to find out whether there are other components that might use a feature like that so it can be done all at once.

app/assets/javascripts/student_profile/__snapshots__/IncidentHeatmap.test.js.snap
```diff
@@ -0,0 +1,25 @@
...
+  style={
+    Object {
+      "display": "flex",
+      "flex": 1,
+      "padding": 10,
+      "width": "100%",
+    }
+  }
+>
+  <div
```
kevinrobinson: This snapshot isn't testing much since `HighchartsWrapper` doesn't render much itself directly.  If you add in `jest.mock('../components/HighchartsWrapper', () => 'highcharts-wrapper');` before the `it` calls in the test, that will mock the component, and then snapshot test will verify that `<HighchartsWrapper />` is called with the same arguments over time.

If you add that in and then run `yarn jest app/assets/javascripts/student_profile/IncidentHeatmap.test.js` you can see the diff and how the snapshot becomes more useful in checking that the way the component is called stays the same over time.

app/assets/javascripts/school_administrator_dashboard/DashboardScatterPlot.js
```diff
@@ -20,7 +20,7 @@ export default class DashboardScatterPlot extends React.Component{
   }
```
kevinrobinson: This seems great!  In this PS, can you also move this file into `components` since we'll re-use it across the `school_administrator_dashboard` and `student_profile` folders?

app/assets/javascripts/student_profile/IncidentHeatmap.js
```diff
@@ -0,0 +1,47 @@
...
+import PropTypes from 'prop-types';
+import moment from 'moment';
+import DashboardScatterPlot from '../school_administrator_dashboard/DashboardScatterPlot';
+
+export default class IncidentHeatmap extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  getincidentTimeAsMinutes(incident) {
```
kevinrobinson: This works but `getincidentTimeAsMinutes` looks like it's copy-pasted from `SchoolDisciplineDashboard`.  Could you move it (and `areMinutesWithinSchoolHours`) to a plain function that `DashboardScatterPlot` exports?  That way both callers can use that same function, and maybe that helps with what you were asking about re: "setting the school hours in one location?"

app/assets/javascripts/student_profile/LightBehaviorDetails.js
```diff
@@ -122,14 +123,19 @@ export default class LightBehaviorDetails extends React.Component {
...
     return (
-      <div style={{width: '50%'}}>
-        {filteredDisciplineIncidents.map(incident => (
-          <IncidentCard
-            key={incident.id}
-            style={{background: '#eee', marginBottom: 20}}
-            incidentCard={incident}
-          />
-        ))}
+      <div className="IncidentHistory" style={{display: 'flex'}}>
```
kevinrobinson: This class name doesn't look like it's used for anything, can we remove it?  In general, let's use names to label the root element inside a component (with the same name as the component), or for things that need proper CSS, for testing or other JS interop.

app/assets/javascripts/student_profile/__snapshots__/IncidentHeatmap.test.js.snap
edavidsonsawyer: Ah, thanks for catching that!

edavidsonsawyer:
@kevinrobinson thanks for the comments! They all sound good to me and have all been implemented

PR #2350, opened 2019-01-11
Set profile heatmap width

edavidsonsawyer:
# Who is this PR for?

Educators

The student profile discipline incident heatmap isn't set at 50% width, causing it to compress the incident cards.

# What does this PR do?

Sets the chart width so it takes half the space.

+ [ ] Home page

+ [ ] Manual testing made more sense here


kevinrobinson:
@edavidsonsawyer it looks great! :)

PR #2352, opened 2019-01-16
Reading: Allow teams to enter benchmark reading data

kevinrobinson:
# Who is this PR for?
K5 reading teams: classroom teachers, reading specialists, special education teachers, ELL teachers, literacy coaches

Data is entered in spreadsheets now, which requires reading specialists to set them up and maintain them as enrollment, IEP status, etc. shift over time (or they become stale).  Across the district, this involves making ~100 separate spreadsheets each year.

The spreadsheet format works great for entry, and also enables educators to make other views of the data on their own.  But this takes data manipulation for looking across time over grades (eg, at DORF WPM from 2nd > 3rd grade), and is hard to use directly for critical instruction decisions like planning reading groups.

# What does this PR do?
Adds a prompt on the home page for entering reading benchmark data, linking to a page for entering data in the same kind of grid format everyone's already used to.  This is intended for the first pilot, so doesn't handle the general case (which we'll still need to discover across grade levels and schools).  Next work is for the pilot of using this data within an A&R planning meeting for grouping, and then we'll iterate on that or grow to the next team, and migrate and remove past prototypes around reading data.

Deploying has no impact - releasing requires changes to some ENV config and labels to make this visible and authorized for specific educators.

### home page prompt
screen shot 2019-01-16 at 9 36 59 am

### entry
screen shot 2019-01-16 at 9 36 41 am

# Checklists
+ [x] Home page
+ [x] Reading benchmark data entry page


kevinrobinson:
selfie

PR #2353, opened 2019-01-16
Profile Heatmap Additions

edavidsonsawyer:
# Who is this PR for?

Educators

# What does this PR do?

Makes the incident history and profile heatmap more prominent in the student profile page and makes heatmap reflect the clean slate filter. Also improves the heatmap tooltip and restores a missing spec for the heatmap component.

heatmap_student

# Checklists
+ [x] Student Profile
+ [x] Discipline


app/assets/javascripts/student_profile/LightBehaviorDetails.js
```diff
@@ -134,7 +134,7 @@ export default class LightBehaviorDetails extends React.Component {
           ))}
         </div>
         <div style={{width: '50%'}}>
-          <IncidentHeatmap incidents={this.props.disciplineIncidents}/>
+          <IncidentHeatmap incidents={filteredDisciplineIncidents}/>
```
kevinrobinson: 👍 😄

app/assets/javascripts/components/DisciplineScatterPlot.js
```diff
@@ -110,6 +114,7 @@ DisciplineScatterPlot.propTypes = {
   seriesData: PropTypes.array.isRequired, // array of JSON event objects.
   titleText: PropTypes.string, //discipline dashboard makes its own title
   measureText: PropTypes.string.isRequired,
+  studentChart: PropTypes.bool.isRequired, //flag to indicate the chart is displaying individual student data
```
kevinrobinson: Can you accomplish this with the current API, by passing different `tooltip`s?

This change would modify the generic scatterplot component to now have code that's specific to the profile page and dashboard page.  It looks like you could have them pass in the `tooltip` they need, and then this component stays generic.

edavidsonsawyer: @kevinrobinson you're fast on the draw! I was actually looking at this and I agree that it's fine to pass a function here instead of a flag. At the time I thought there might be other reasons to distinguish behavior based on whether it was for a student or school.

kevinrobinson: 👍 also the test failure https://travis-ci.org/studentinsights/studentinsights/builds/480544138?utm_source=github_status&utm_medium=notification seems wacky, you can try just restarting it and if it still fails ping me and I can help debug:

image

edavidsonsawyer: @kevinrobinson the old turn-it-off-turn-it-on-again trick does it again

kevinrobinson: cool, yeah there's got to be a race condition in that spec related to testing scrubbing data sent to Rollbar, but not sure where.  So if it's happening .05% of the time then not so bad but yeah flaky tests are no bueno...

kevinrobinson:
@edavidsonsawyer awesome, 🚢 ing!

kevinrobinson:
@edavidsonsawyer I think there's three good reasons to update the chart to show the current school year.  First, I think the strong default assumption for everything is "this school year" unless there are places where we really explicitly call that out (eg, past 45 days).  I don't think anyone who's trying to dig into discipline incidents is going to read the "clean slate" copy and notice what it says :)  Second, students' schedules often change significantly from year to year; for students transitioning schools (say 8th grade to HS) it will be completely different.  This means showing two years will hide any patterns that there might be related to the schedule or school day in the current year.  Third, patterns from the current school year are the only thing that's potentially actionable.

So my vote would be to make the chart reflect the school year and label that explicitly like 2018-2019 school year.

Separately, it also seems like a good idea to clarify the clean slate message so it's clearer that it's the current and previous school year.

PR #2354, opened 2019-01-17
Reading: disable sorting of columns with inputs

kevinrobinson:
follow-on to https://github.com/studentinsights/studentinsights/pull/2352

# Who is this PR for?
Reading teams

The input elements are controlled, and the table is sorted.  This leads to the table re-sorting on keystroke, and the cursor focus not working as expected when the table is sorted by the column you are entering data in.

# What does this PR do?
I tried just making this uncontrolled but it led to the sorting not working at all (which I can't understand), so for now just disabling sort on the input columns.

# Checklists
+ [x] Reading data entry


kevinrobinson:
selfie

PR #2356, opened 2019-01-17
Reading: Remove download, send back empty JSON for client, improve notice on network failure

kevinrobinson:
From https://github.com/studentinsights/studentinsights/pull/2352

The client expects valid JSON, so an empty response body raises a clientside error.  This also makes the message on network failure more visible, so that if it does happen folks don't type in all the data without realizing.

Also, the approach to downloading CSV doesn't work with the state in the `doc` object, so this removes that for now.


kevinrobinson:
selfie

PR #2357, opened 2019-01-17
Profile: Update UI check for IEP to use student record field as authoritative if PDF conflicts

kevinrobinson:
# Who is this PR for?
educators, students, families

There are multiple bits of data about special education and IEPs - previously `hasAnySpecialEducationData` was used to decide whether to consider the student as having an active IEP, but this was overly inclusive.

# What does this PR do?
Changes `hasAnySpecialEducationData` to `hasActiveIep`, which uses the student record fields as authoritative, regardless of the IEP PDF document.

# Checklists
+ [x] Student Profile
+ [x] Benchmark reading data


kevinrobinson:
selfie

PR #2358, opened 2019-01-18
Prevent scatterplot from displaying with too many data points

edavidsonsawyer:
# Who is this PR for?
Educators

When there are hundreds of discipline incidents, the scatterplot will have too many data points to show useful information.

# What does this PR do?

Prevents the scatterplot from displaying when there are more than 500 incidents

limit_heatmap

# Checklists
+ [x] Discipline

### DIFF
--- a/app/assets/javascripts/components/ScatterPlotOverloadMessage.js
+++ b/app/assets/javascripts/components/ScatterPlotOverloadMessage.js
@@ -1,0 +1,15 @@
+import React from 'react';
+
+export default function ScatterPlotOverloadMessage() {
+  return (
+    <div style={styles.message}>
+      <span>There are too many students to display this data. You can select another chart type or try filtering the students.</span>
+    </div>
+  );
+}
+
+const styles = {
+  message: {
+    padding: 20
+  }
+};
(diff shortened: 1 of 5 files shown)


edavidsonsawyer:
@kevinrobinson this is a little different than we talked about - it doesn't completely disable the heatmap for the high schools, but prevents it from showing over a threshold of 500 incidents. We could change it to simply not be available, but if someone wants to look by grade, house, or code this would keep it available. Thoughts?

kevinrobinson:
@edavidsonsawyer I like the idea here of preserving some part of this feature, but I think we should avoid messages like this.  I think for users, while these kinds of message are good for describing what is happening, the problem isn't really that "there are too many students" it's that "the team knows this won't work for what you need in your school, but we chose to leave you this message instead of fixing it" :)  In general, if a feature is important, we should make it work instead of adding messages.  Or if this case, where we decided with educators that it wasn't the most important next thing to work on, I think it's better to just remove the option.

This especially matters for people who haven't looked at this page yet.  When Uri shows it to them, we don't want their first experience on page load to be a message about how something doesn't entirely work and how to work around it - better to let them see what we think is most useful and dig into the things that do fully work.

For the tuning - 500 seems like it would be when there's something on the order of 100 data points in each (day, hour) timeslot.  Is that how you're thinking about it?

edavidsonsawyer:
@kevinrobinson that's a reasonable way to look at it. I'll close this PR in favor of a version that just hides the scatterplot option for large sets.

kevinrobinson:
@edavidsonsawyer Sounds good!  It'd be awesome to make this work for larger data sets too, but I think based on the conversation and input that'd be a good improvement, but not the most important thing for right now.  I'm excited about finishing off the last few bits here too so we can work more on some things related to reading and MTSS!  :)

PR #2362, opened 2019-01-21
Hide scatterplot for large schools

edavidsonsawyer:
# Who is this PR for?
Educators 

For large schools, the scatterplot in the discipline dashboard has too many datapoints to show useful information.

# What does this PR do?
Removes the scatterplot feature when there are more than 500 discipline incidents to display when first loading

# Checklists
+ [x] Discipline

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
+++ b/app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
@@ -49,6 +49,13 @@ export default class SchoolDisciplineDashboard extends React.Component {
     this.memoize = memoizer();
   }
 
+  //Remove scatter plot option when there are too many incidents to show patterns
+  componentDidMount() {
+    if (this.getIncidentsFromStudents(this.props.dashboardStudents).length > 500) {
+      this.setState({selectedChart: 'incident_location', scatterPlotAvailable: false});
+    }
+  }
+
   filterIncidents(disciplineIncidents, options = {}) {
     return this.memoize(['filteredIncidents', this.state, arguments], () => {
       if (!disciplineIncidents) return [];
@@ -315,10 +322,10 @@ export default class SchoolDisciplineDashboard extends React.Component {
 
   render() {
     const {districtKey} = this.context;
-    const {timeRangeKey, selectedChart, grade, house, counselor} = this.state;
+    const {scatterPlotAvailable, timeRangeKey, selectedChart, grade, house, counselor} = this.state;
     const {school, dashboardStudents} = this.props;
     const chartOptions = [
-      {value: 'scatter', label: 'Day & Time'},
+      ... scatterPlotAvailable ? [{value: 'scatter', label: 'Day & Time'}] : [],
       {value: 'incident_location', label: 'Location'},
       {value: 'time', label: 'Time'},
       {value: 'homeroom_label', label: 'Classroom'},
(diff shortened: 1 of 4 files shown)


edavidsonsawyer:
@kevinrobinson here's the simpler version of disabling the scatter plot for large schools. I'm not sure what's up with the build error so I'll try to look into that later on today.

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.test.js
```diff
@@ -48,80 +47,115 @@ function renderShallow(props = {}) {
...
 it('renders a scatter plot', () => {
   const context = testContext({districtKey: 'somerville'});
-  const props = {school: testSchool()};
+  const props = {school: testSchool(), dashboardStudents: createStudents(setDate())};
   const el = testEl(props);
   const dash = mount(elWrappedInContext(el, context));
   expect(dash.find('DisciplineScatterPlot').length > 0).toEqual(true);
 });
 
+it('does not render a scatter plot when there are more than 500 incidents', () => {
```
kevinrobinson: nice!

app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/SchoolDisciplineDashboard.js
```diff
@@ -49,6 +49,13 @@ export default class SchoolDisciplineDashboard extends React.Component {
     this.memoize = memoizer();
   }
 
+  //Remove scatter plot option when there are too many incidents to show patterns
+  componentDidMount() {
+    if (this.getIncidentsFromStudents(this.props.dashboardStudents).length > 500) {
+      this.setState({selectedChart: 'incident_location', scatterPlotAvailable: false});
+    }
+  }
+
```
kevinrobinson: Can you say more about why this is here, rather than computed in `render`?  I don't follow why it's using the lifecycle hook.

edavidsonsawyer: It's here because the state has to change (from default scatter to some alternative), which can't be done in render. I could probably set the initial state conditionally in the constructor, but it seemed cleaner to just use getIncidentsFromStudents in the mounted component. It shouldn't be much of a performance issue since we cache the result of that function with memoizer.

kevinrobinson: ah, thanks!  Got it, it's because this is set in `initialState` in the constructor, that's super helpful.  Yeah, the better way to do this is describe this explicitly in the constructor, and then let `initialState` decide what `selectedChart` should be based on those props.  Here that would involve factoring out the incident and filtering functions to be plain functions (rather than relying on `this.state` and `this.props`) and that seems a larger change than this is worth.  Using the lifecycle hook implies that it's because of something related to the DOM or outside of React, and then leads to double-rendering on load.

If you tested this and it works without a visible jump, let me know and let's 🚢 as is.  And thanks for explaining and helping me learn! 👍

edavidsonsawyer: Great! I have tested this locally with schools above and below the 500 student threshold, and in the former case the scatter plot is cleanly removed. I would like to double check in production just to be sure the threshold is appropriate, but that can probably come from feedback from users.

kevinrobinson: @edavidsonsawyer awesome, thanks! 🚢 ing now and I'll verify as well when it's out.

kevinrobinson:
@edavidsonsawyer looks good! 👍  It looks like there's a separate issue with SHS, which I'd best is from the call to `latest_note` (since a larger MS school in New Bedford with less notes loads just fine).  I'll add that to the list of things to get this functional enough for Somerville general release.

PR #2365, opened 2019-01-22
Add discipline code back to filter

edavidsonsawyer:
# Who is this PR for?
Educators

Schools have many different codes for discipline incidents, but there is no easy way to see how many of each type they have across the school.

# What does this PR do?
Adds an option to the discipline dashboard to view discipline incidents broken down by code.

incident_codes

# Checklists
+ [x] Discipline

### DIFF
--- a/app/assets/javascripts/components/SelectDisciplineIncidentType.js
+++ b/app/assets/javascripts/components/SelectDisciplineIncidentType.js
@@ -12,7 +12,7 @@ export default function SelectDisciplineIncidentType({type, onChange, types, sty
   return (
     <SimpleFilterSelect
       style={style}
-      placeholder="Incident Type..."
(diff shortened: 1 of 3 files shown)


edavidsonsawyer:
@kevinrobinson I forgot about this! It's a quick change that we can put in whenever it's wanted.

kevinrobinson:
@edavidsonsawyer Awesome!  ⚡️  

When you get a chance, can you also rename the filter to "Incident code" instead of "Incident type" too?  I think that's my mistake from before, the label you added here made me think of it and folks will see "code" within Aspen too.  Thanks!

edavidsonsawyer:
@kevinrobinson my bad! I meant to change that be in sync with the other language anyway. Should be all set now!

PR #2368, opened 2019-01-23
Reading: store grouping state locally, snapshot grouping state and post to server

kevinrobinson:
Builds on https://github.com/studentinsights/studentinsights/pull/2367

# Who is this PR for?
K5 reading teams

# What does this PR do?
This adds the "additional groups" grouping UI, and lifts state of groupings up to maintain it across navigating between groups.  It adds an `Autosaver` components that tracks changes and autosaves posting to the server to store that work in the database.  The schema is minimal and not intended as final; this is just enabling saving during the first pilot (and not yet even reading these into the UI yet).

(no visual feedback)

# Checklists
+ [X] Reading grouping page

### DIFF
--- a/app/assets/javascripts/components/Lifecycle.js
+++ b/app/assets/javascripts/components/Lifecycle.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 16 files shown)


kevinrobinson:
selfie

PR #2369, opened 2019-01-23
Update README, removing product bits and points to www instead

kevinrobinson:
An update here, removing old dated bits and pointing to the website for product instead.


kevinrobinson:
selfie

PR #2370, opened 2019-01-24
Session renewal: Add instrumentation and disable force reloading

kevinrobinson:
# Who is this PR for?
educators, primarily in K8 MTSS

Follow-on to https://github.com/studentinsights/studentinsights/pull/2366, which doesn't seem to have resolved the problem.

# What does this PR do?
Adds more instrumentation, removing the reload step and messaging about sessions expiring.  This doesn't change the server-side timeout at all.  It only impacts the client-side UX when someone leaves a tab open for a long time - it leaves the current tab visible and removes the indication that the session has expired.

Going to try this for a meeting structure this morning, then revisit.

# Checklists
+ [x] Core 


kevinrobinson:
selfie

PR #2371, opened 2019-01-25
Restricted notes: Clarify that admin can also mark notes restricted

kevinrobinson:
# Who is this PR for?
educators

The UI suggests that the original author of the note marked it restricted.  With an initial limited release of https://github.com/studentinsights/studentinsights/pull/2243 and https://github.com/studentinsights/studentinsights/pull/2304, admin can do this as well, so the text is misleading.

# What does this PR do?
Just adds copy clarifying.  Before a general release later, it'd be better to make explicit who marked the note as restricted (so folks know who to follow up with), but this is a good first step for now.

# Checklists
+ [x] Student Profile
+ [x] My notes


kevinrobinson:
selfie

PR #2374, opened 2019-01-28
Discipline Dashboard release

edavidsonsawyer:
# Who is this PR for?
Educators

The discipline dashboard is available in production, but there are no live links to the feature. 

# What does this PR do?
Adds links to the discipline dashboard alongside the tardies/absences dashboards

admin_disc_links
teacher_disc_links

# Checklists
+ [x] Home page

### DIFF
--- a/app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/__snapshots__/SchoolDisciplineDashboard.test.js.snap
+++ b/app/assets/javascripts/school_administrator_dashboard/discipline_dashboard/__snapshots__/SchoolDisciplineDashboard.test.js.snap
@@ -36,40 +36,6 @@ exports[`snapshots works for bedford 1`] = `
       <div
         style={
           Object {
-            "background": "red",
-            "color": "white",
-            "fontWeight": "bold",
-            "padding": 20,
-          }
-        }
-      >
-        <span>
-          This is an experimental prototype!
-        </span>
-        <span
-          style={
-            Object {
-              "marginLeft": 10,
-            }
-          }
-        >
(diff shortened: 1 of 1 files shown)


edavidsonsawyer:
@kevinrobinson I have this ready to go but we probably want to hold off on this request until we have a solution for the SHS loading time. I'm working on that today.

kevinrobinson:
@edavidsonsawyer Awesome!  This looks good, and can we remove the experimental banner too when this is ready to go out?  The other wrinkle here is that some users have additional links on the navbar - you can sign in as the `jodi` or `harry` user to see this.  I think it will fit with a 1024px wide screen, but can you confirm?

re: SHS loading, I think the issue is the n+1 where it queries for all students, then for each one queries for discipline incidents.  The absences page batches these queries up and might be a model to check out: https://github.com/studentinsights/studentinsights/blob/master/app/lib/dashboard_queries.rb#L8.  If there are ways to make smaller changes that'd be great to try, but just using the same approach as the absences page is fine too, which might be easier without profiling.

edavidsonsawyer:
@kevinrobinson thanks for flagging the users with extra links. I tested with both of them at a bunch of resolutions and everything seems to fit. One note is that I don't think the Jodi user is supposed to have access to the dashboards (she doesn't now). Is that not correct?

kevinrobinson:
hi eric, awesome!  yeah sorry I think I mixed up the names then

edavidsonsawyer:
@kevinrobinson okay, the banner is out, so this should be ready when we want to merge it.

PR #2359, opened 2019-01-18
Reading: Refactor profile to use IepDialog component

kevinrobinson:
# Who is this PR for?
developers

# What does this PR do?
Refactor student profile to use split-out IepDialog component from reading work.

# Checklists
*Which features or pages does this PR touch?*
+ [x] Student Profile
+ [x] Reading benchmark data

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [X] Manual testing made more sense here

PR #2376, opened 2019-01-28
Profile: Fix bug with rendering IepDialog in Bedford

kevinrobinson:
# Who is this PR for?
Bedford educators

https://github.com/studentinsights/studentinsights/pull/2359 introduced a regression to the profile page for Bedford only.  This led to the profile page not loading, and Rollbar errors were ambiguous (which would be improved with React 16 error boundaries), so I mistakenly thought this was related to trying to interact after session expiration.

# What does this PR do?
Fixes the bug; wraps the return value from `<IepDialog />` in an element rather than a plain string (https://reactjs.org/docs/error-decoder.html/?invariant=109&args[]=t) and renames it to `<IepDialogLink />` to clarify and add tests and stories.

# Checklists
+ [x] Student Profile
+ [x] Reading entry
+ [x] Reading grouping


kevinrobinson:
selfie

PR #2377, opened 2019-01-28
SessionRenewal: Rework to probe server to warn about session expiration instead of using heuristic

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2370 and https://github.com/studentinsights/studentinsights/pull/2366.

# Who is this PR for?
educators

There were some user reports of being signed out unexpectedly, and in investigating and in looking at Rollbar errors related to `.slice` in parsing dates, this looks like it's coming from users interacting with the server (eg, to save a note) after the session has expired.  In that case, it means the educators loses what they just typed in.

# What does this PR do?
Reworks `<SessionRenewal />`.  The core loop is probing the server to see if the user is still signed in.  This is with a new `/educators/probe` endpoint that skips the normal Devise Timeoutable behavior of keeping the session alive.  This endpoint returns how much time is left in the session, and then `<SessionRenewal />` shows an expiration notice if it's within the warning window.  Renewing the session makes the same request to the server as before.

This works better than the previous approach because it avoids UI-side heuristics, and just checks with the server.  If the session has already expired, the UI redirects to close the page, since any data from interactions would be lost anyway.  The expiration message should make it clear to users this would happen beforehand, and this PR changes the background color to a warning orange instead of blue.

This revises the Rollbar instrumentation and keeps it in place, so we can verify this is working as expected.

# Checklists
+ [X] Core 

+ [X] Included specs for changes
+ [X] Manual testing made more sense here

### DIFF
--- a/app/assets/javascripts/components/SessionRenewal-One.test.js
+++ b/app/assets/javascripts/components/SessionRenewal-One.test.js
@@ -1,60 +1,52 @@
-import React from 'react';
-import ReactDOM from 'react-dom';
-import {withDefaultNowContext} from '../testing/NowContainer';
 import fetchMock from 'fetch-mock/es5/client';
-import SessionRenewal from './SessionRenewal';
-
-
-export function testProps(props = {}) {
-  return {
-    sessionTimeoutInSeconds: 2,
-    warningTimeoutInSeconds: 1,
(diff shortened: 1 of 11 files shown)


kevinrobinson:
selfie

PR #2373, opened 2019-01-28
Restrict student discipline scatter plot to school year

edavidsonsawyer:
# Who is this PR for?
Educators

# What problem does this PR fix?
The student profile has a discipline incident heatmap that includes incidents for the previous school year, which is unlikely to be useful in finding patterns in the student's behavior.

# What does this PR do?
Restricts the heatmap in the profile to the current school year only. 

# Checklists
*Which features or pages does this PR touch?*
+ [x] Student Profile

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Improved specs for existing code in need of better test coverage
+ [x] Manual testing made more sense here

PR #2378, opened 2019-01-29
Profile: Restrict student discipline scatter plot to school year (v2) #2373 

kevinrobinson:
This is the same as https://github.com/studentinsights/studentinsights/pull/2373, but adds a commit adding the snapshot test back in.  cc @edavidsonsawyer


kevinrobinson:
🚢

PR #2388, opened 2019-01-30
Sessions: Include querystring when forcibly reloaded

kevinrobinson:
This is to gather more information to debug issues where educators report they are being signed out and their session is expiring more quickly than they expect.


kevinrobinson:
selfie

PR #2389, opened 2019-01-31
Session renewal: Fix bug from IE11 fetch polyfill

kevinrobinson:
The `whatwg-fetch` polyfill kicks in in IE11, and it doesn't support non-transparent redirection, or the `redirected` flag, since `XmlHttpRequest` follows redirects without any indication to the caller.  So this changes to request JSON so the server responds with a 401 instead and avoids the issue.

This also updates the instrumentation, and makes a noop refactor to `apiFetchJson` to expose the plain method.


kevinrobinson:
selfie

PR #2387, opened 2019-01-30
Student voice: Fix timezone bug on imports; release profile insights with feature switch for Q2 self-reflection

kevinrobinson:
Building on https://github.com/studentinsights/studentinsights/pull/2382, https://github.com/studentinsights/studentinsights/pull/2381, https://github.com/studentinsights/studentinsights/pull/2383, https://github.com/studentinsights/studentinsights/pull/2384 and https://github.com/studentinsights/studentinsights/pull/2379.

# Who is this PR for?
SHS educators, students

# What problem does this PR fix?
There was a bug with importing timestamps from Sheets, so these timestamps were in the wrong timezone and then wouldn't be visible in the feed like they should be.

# What does this PR do?
This reworks timestamp parsing for all survey importers, moving it into `ImportMatcher` and adding a test.

This also removes the `?winterinsights` switch for profile insights, enabling it for all students and educators.  It adds instead a `PROFILE_INCLUDE_Q2_SELF_REFLECTION_INSIGHTS` ENV switch for controlling whether the Q2 self-reflection is included, or whether it is just from "What I want my teachers to know about me."

# Checklists
*Which features or pages does this PR touch?*
+ [x] Home page
+ [x] Student Profile

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Improved specs for existing code in need of better test coverage

PR #2392, opened 2019-02-04
Student voice: Decouple profile insights and grades reflection switches

kevinrobinson:
Fix bug between https://github.com/studentinsights/studentinsights/pull/2384 and https://github.com/studentinsights/studentinsights/pull/2387, where the feature switch controlling whether Q2 self-reflection appears in the "insights box" also would remove it from the grades tab.  This decouples the two.


kevinrobinson:
selfie

PR #2393, opened 2019-02-04
Student voice: Fix timezone parsing in importing form timestamps

kevinrobinson:
Makes sure Rails is always running in UTC, and fix check for EDT when parsing Sheets timestamps that don't say the timezone explicitly.


kevinrobinson:
selfie

PR #2394, opened 2019-02-04
Fix bug in self-reflection for grade refactoring

kevinrobinson:
Fixing bug from https://github.com/studentinsights/studentinsights/pull/2392.


kevinrobinson:
selfie

PR #2397, opened 2019-02-05
Maintenance: Upgrade to React 16.7, add RollbarErrorBoundary

kevinrobinson:
# Who is this PR for?
developers, educators

We don't have good visibility into certain kinds of errors within React components.

# What does this PR do?
Upgrades React-related dependencies from 15.4 to 16.7.  Adds `RollbarErrorBoundary` within `<App />`.

Also upgrades a bunch of libraries in the process (eg, babel versions).

# Screenshot
error-boundary

# Checklists
+ [x] Core 


kevinrobinson:
This came up from https://github.com/studentinsights/studentinsights/pull/2398 and related.

kevinrobinson:
Okay, so this does a few other things now.  Updates to Babel 7, to resolve problems with the Jest config that didn't surface locally but did in Travis.  In the process of trying to understand why the Travis config was different, I made three changes to the build config.

First, I updated the build to split the JS and Ruby tests, and parallelize them across VMs.  Second, I updated the way the build cache works for yarn.  These speed things up a minute or two, but not much.  I tried unsuccessfully to get Travis to cache Ruby gems, since those take 2.5min to install, but can't get this to work with `bundler: true` or with manually specifying the gems folder to cache.  Punting that for now.

This also surfaced that the Travis env set a timezone different than in production, and so I updated tests to all run in UTC to match the production deployment.  This updates specs that were written to run in `America/New_York` to be correct for UTC now.

kevinrobinson:
selfie

PR #2398, opened 2019-02-05
Revert "Revert "Student voice: Move fall student voice survey into profile notes list""

kevinrobinson:
Reverts studentinsights/studentinsights#2396

double-revert for https://github.com/studentinsights/studentinsights/pull/2395

PR #2400, opened 2019-02-05
Student voice: Update differences between what shows in profile notes / profile insights

kevinrobinson:
Moving over from https://github.com/studentinsights/studentinsights/pull/2398.


kevinrobinson:
This branch is broken now; this will need some refactoring to split out: profile notes from profile insights.

kevinrobinson:
selfie

PR #2407, opened 2019-02-13
Upgrade jQuery UI

kevinrobinson:
bumping versions

EDIT: also bundling in jQuery UI since these are coupled.  In the process, this involved updating jQuery UI override styles, and so this has a bunch of unrelated CSS cleanup too.

EDIT: this also reduces the after-gzip bundle sizes a little bit too, CSS by 14% and JS by 7%.

### before
screen shot 2019-02-15 at 12 25 45 pm

### after
screen shot 2019-02-15 at 12 45 50 pm


kevinrobinson:
bundling in https://github.com/studentinsights/studentinsights/pull/2412 since these are coupled

kevinrobinson:
selfie

PR #2410, opened 2019-02-13
Use CDN domain in more CSP rules

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2408.


kevinrobinson:
selfie

PR #2411, opened 2019-02-13
Upgrade to jQUery 3.3

kevinrobinson:
Upgrading, along with https://github.com/studentinsights/studentinsights/pull/2407 and https://github.com/studentinsights/studentinsights/pull/2409.


kevinrobinson:
selfie

PR #2413, opened 2019-02-14
Remove use of new Date() causing test bugs

kevinrobinson:
This is in a few places, and this updates them to use `context.nowFn` now, since these tests started failing today.


kevinrobinson:
selfie

PR #2417, opened 2019-02-15
Session renewal: Update debug logging

kevinrobinson:
This updates debug logging, bumping versions, and downgrading these to `info` since things seem good.  If this checks out in another week or two, we could remove altogether.


kevinrobinson:
selfie

PR #2418, opened 2019-02-15
Guard sessionStorage usage more

kevinrobinson:
Removing noise from either folks visiting the demo site without cookies or scrapers.


kevinrobinson:
selfie

PR #2419, opened 2019-02-15
Update font files

kevinrobinson:
Some files specified in CSS were missing, which came up as monitoring noise.


kevinrobinson:
selfie

PR #2421, opened 2019-02-16
Update for upcoming changes to Heroku build

kevinrobinson:
For https://help.heroku.com/P5IMU3MP/heroku-node-js-build-script-change-faq


kevinrobinson:
simplified, although seems like there's still too many yarn installs, but one less than before and this removes a monkey path, so 🚢 ing for now

PR #2422, opened 2019-03-05
Add audits for vulnerabilities in Ruby dependencies

kevinrobinson:
For auditing Ruby dependencies for CVEs, using https://github.com/rubysec/bundler-audit and https://github.com/civisanalytics/ruby_audit


kevinrobinson:
selfie

PR #2423, opened 2019-03-05
Sign in:  Revise session expired copy 

kevinrobinson:
### before
screen shot 2019-03-05 at 11 33 08 am

### after
screen shot 2019-03-05 at 11 33 27 am


kevinrobinson:
selfie

PR #2428, opened 2019-03-14
Update Brakeman

kevinrobinson:
Bump version, ensure latest and run all checks, with documented safelist.


kevinrobinson:
selfie

PR #2430, opened 2019-03-19
Authorization: Update authorizer to expose reason why educator is authorized

kevinrobinson:
# Who is this PR for?
educators, students, families

Educator's can't see right now who else has access to a student's record.  Making this visible could help educators and district project leads catch any mistakes in how they've set up roles and permissions.

# What does this PR do?
This PR just refactors the `Authorizer` class to expose that internally; there's no product changes here yet.

# Checklists
+ [x] Core 


kevinrobinson:
selfie

PR #2432, opened 2019-03-21
Home page feed: Allow flag for showing photos

kevinrobinson:
# Who is this PR for?
educators, students

# What problem does this PR fix?
Most students have photos, but these aren't visible on the home page feed.

# What does this PR do?
Adds photos in, with a querystring flag for now to support auditing.  If there is no photo, falls back to showing nothing there (rather than placeholder).

# Screenshot (if adding a client-side feature)
Screen Shot 2019-03-21 at 12 38 04 PM

# Checklists
*Which features or pages does this PR touch?*
+ [x] Home page
+ [x] Student Profile

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Improved specs for existing code in need of better test coverage
+ [x] Manual testing made more sense here

PR #2433, opened 2019-03-21
Home page feed: Design tweaks and add snapshot test for StudentPhotoCropped

kevinrobinson:
Styling follow-on to https://github.com/studentinsights/studentinsights/pull/2432, adds a snapshot test too.


kevinrobinson:
selfie

PR #2440, opened 2019-03-29
Layout: Update min-width on body, navbar sizing

kevinrobinson:
# Who is this PR for?
educators

Two layout issues: horizontal scrollbars showing unnecessarily on 1024px wide IE11 browsers, and navbar width leading to text wrapping for particular users when many links.

# What does this PR do?
Updates body min-width to match workaround JS function and removes that fn.  Fixes the width of the searchbar to prevent jumping when font is initially loaded (there was a bug in the CSS before), and use placeholder instead of label to reduce its overall width while increasing size that's interactable.

Screen Shot 2019-03-29 at 12 02 55 PM

# Checklists
+ [x] Layout and navbar


kevinrobinson:
selfie

PR #2441, opened 2019-03-29
Profile: Remove resizable textarea from DraftNote, add snapshots and stories

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2440, adding `resize: none` CSS and tests/stories to catch.


kevinrobinson:
selfie

PR #2444, opened 2019-05-01
Class lists: Enable naming lists for other purposes (eg, science classes)

kevinrobinson:
# Who is this PR for?
middle school teams

Middle school teams also sometimes make different sets of class lists (eg, subject-specific classes like science, different sets of class lists for a dual-language program  When folks did this last year, they'd make multiple class lists and describe them in the comments, but this wasn't visible in the class lists.

# What does this PR do?
Moves this up front, and asks teams to name each list, with a default suggestion of "homerooms."

Screen Shot 2019-05-01 at 9 40 34 AM
Screen Shot 2019-05-01 at 9 39 36 AM

# Checklists
+ [x] Class list view
+ [x] Class list creator


kevinrobinson:
selfie

PR #2445, opened 2019-05-01
Allow demo to use Somerville school year calculations to demo levels page

kevinrobinson:
This is for the levels page, which is enabled in demo, but raises because the school year calculations aren't supported in the demo env.  For the demo page to work, it's fine to use the Somerville definitions.


kevinrobinson:
selfie

PR #2447, opened 2019-05-01
Upgrade react-beautiful-dnd

kevinrobinson:
One minor change suggested by the upgrade docs, no issues.  Now's a good time to upgrade, before doing some other manual testing here.

PR #2448, opened 2019-05-01
Class lists: Add student photo to class list creator

kevinrobinson:
# Who is this PR for?
K8 teaching teams

Now with more human :)

# What does this PR do?
Adds in student photos, and an icon if there's a recent note (within the last 45 days).

This also revealed a test bug with the way students were "consistently" hashed into buckets for stories and tests.  This is changed to only look at the student id now, and this led to a bunch of trashing in tests and stories, and some misconfiguration between the react-test-renderer for snapshots and react-beautiful-dnd that I couldn't figure out, but probably came from https://github.com/studentinsights/studentinsights/pull/2447 and weren't until trying to add more kinds of snapshots to catch regressions that before were only visible in stories.  There's some more tests here, but the root issue isn't resolve and is mostly worked around, with some comments added (eg, `Invariant failed: provided.innerRef has not been provided with a HTMLElement.`).

Screen Shot 2019-05-01 at 11 26 33 AM

# Checklists
+ [x] Class lists


kevinrobinson:
selfie

PR #2450, opened 2019-05-03
Add quilts experiment visualization

kevinrobinson:
# Who is this PR for?
internal team

This experiments with one way to visualize more complex conceptions of identity in aggregate datatsets, aiming for a way that challenges monolithic notions of identity and could be connected to more individualized information.  This is just an internal experiment, at `/schools/:id/explore/quilts`


kevinrobinson:
selfie

PR #2451, opened 2019-05-03
Class list: Add equity check for diversity groups

kevinrobinson:
# Who is this PR for?
K8 teaching teams

# What does this PR do?
Adds a check for racial and ethnic diversity, using a framing for diversity chosen from talking with folks in Somerville about what they thought would be most important for teaching teams to consider as they're creating class lists.

image

# Checklists
+ [x] Class lists


kevinrobinson:
selfie

PR #2454, opened 2019-05-07
Class list: Fix bug with branching on STAR/DIBELS in profile

kevinrobinson:
# Who is this PR for?
K8 teaching teams

STAR data should only be shown for students going into grade 3 or later, but the branching in the inline profile wasn't correct since there was no `grade` field sent down on the student record.

# What does this PR do?
Uses the `gradeLevelNextYear` to fix the branching.

# Checklists
+ [x] Class lists


kevinrobinson:
selfie

PR #2462, opened 2019-05-08
Class lists: Update list to be narrower with short school names and shorter dates

kevinrobinson:
# Who is this PR for?
K8 admin

# What problem does this PR fix?
With names and staff names, these can be a bit long and overflow 1024px wide.  No one likes horizontal scrollbars :)

# What does this PR do?
Add in short school names for Somerville, and cut date rendering from "Wednesday" to "Wed".

# Screenshot (if adding a client-side feature)
Screen Shot 2019-05-08 at 4 19 17 PM

# Checklists
*Which features or pages does this PR touch?*
+ [x] Class lists view

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Manual testing made more sense here

PR #2463, opened 2019-05-08
Class lists: Update class lists view to sort with short names

kevinrobinson:
Fix for https://github.com/studentinsights/studentinsights/pull/2462.


kevinrobinson:
selfie

PR #2464, opened 2019-05-09
Equity experiments: Add more durable links to experiments

kevinrobinson:
# Who is this PR for?
internal team

There are various experiments around equity questions that we've looked at internally, but these aren't in a place where they could even really be demoed internally.

# What does this PR do?
Adds links into the district page (controlled by a label), and refactors the URL paths and controller endpoints.


kevinrobinson:
selfie

PR #2468, opened 2019-05-10
Reading: Refactor to clarify importers, placeholder reading importer for Somerville

kevinrobinson:
# Who is this PR for?
internal team, developers, working towards K8 teaching teams

# What does this PR do?
"importer" files have different interfaces, so rename files to "processor" if they just take a file directly (eg, on the command line).  This will help clarify as we automate more of this and use other transport methods (eg, Google API).

Also adds a placeholder `SomervilleMegaReadingImporter` file with comments for collaboration.


kevinrobinson:
selfie

app/importers/reading/somerville_mega_reading_importer.rb
```diff
@@ -0,0 +1,67 @@
+class SomervilleMegaReadingImporter
```
kevinrobinson: @edavidsonsawyer I thought this might be helpful for talking more concretely about what it might look like to write a reading importer.  This is placeholder for talking about that, with the idea that this might be able to combine parts of `MegaReadingImporter` with some kind of "Fetcher" class from https://github.com/studentinsights/studentinsights/pull/2442.

edavidsonsawyer: Nice! I've been changing around the import script to work more flexibly as a class with fewer opinions about what it's fetching. It should integrate well with these changes. If there's time on the call today we can talk that over quickly.

kevinrobinson: awesome! 👍

PR #2467, opened 2019-05-10
Reading: Add debug view for seeing progress and global distributions

kevinrobinson:
# Who is this PR for?
Internal team, K8 literacy coaches

# What problem does this PR fix?
There's no way to see what data is imported, as we start working on this across more data points and groups, this will become increasingly complicated to understand :)

# What does this PR do?
Creates a "debug" reading page that shows the overall status for all students.

Also factors out `<FountasAndPinnellBreakdown />` from `<ClassroomStats />` for re-use here.

# Screenshot (if adding a client-side feature)
Screen Shot 2019-05-10 at 10 47 48 AM

# Checklists
*Which features or pages does this PR touch?*
+ [x] Reading debug
+ [x] Class lists

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Manual testing made more sense here

PR #2469, opened 2019-05-10
Fix handlings of missing F&P

kevinrobinson:
Fixes bug in handling of `includeMissing=false` from https://github.com/studentinsights/studentinsights/pull/2467.


kevinrobinson:
selfie

PR #2471, opened 2019-05-16
Update brakeman, force SSL in two places

kevinrobinson:
We already set `force_ssl` in `application_controller` but Brakeman recommends doing it in `production.rb` so saying it twice seems fine :)


kevinrobinson:
selfie

PR #2475, opened 2019-05-17
Reading: Enable exporting reading debug data

kevinrobinson:
# Who is this PR for?
internal team, reading specialists or literacy coaches

# What problem does this PR fix?
There's no way to get all the reading data out for other analysis.  As we migrate more data in, this keeps other uses open.

Also factors out `<DownloadIcon />` svg.

# Screenshot (if adding a client-side feature)
Screen Shot 2019-05-17 at 10 26 06 AM

# Checklists
*Which features or pages does this PR touch?*
+ [x] Reading debug

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Manual testing made more sense here

PR #2476, opened 2019-05-17
Reading: Include benchmark_assessment_key in debug export

kevinrobinson:
Missing bit from https://github.com/studentinsights/studentinsights/pull/2475.


kevinrobinson:
selfie

PR #2477, opened 2019-05-24
Counselor Meetings: Add initial version for feedback and testing

kevinrobinson:
# Who is this PR for?
SHS counseling team

# What problem does this PR fix?
A page for counselors to students they haven't connected with in a while, especially in situations with large caseloads.

# What does this PR do?
Adds a new page at `/counselors/meetings`.

This is based on the awesome design and UX work by @ciscomeneses and @amyo423 in https://docs.google.com/presentation/d/1_uhHiq2j_gr0_Uu0tf1meSU27Z0lnjTjgMV3bgzwKkY/edit#slide=id.g55fc9767bc_0_11, and their awesome engineering work building most of this as well 👍 🔥 

There's no link visible yet, and this is behind a feature switch.

# Checklists
*Which features or pages does this PR touch?*
+ [x] Counselor meetings

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Manual testing made more sense here

PR #2478, opened 2019-05-24
Counselor meetings: Limit scope to SHS and 9-12, regardless of authorization

kevinrobinson:
Follow on to https://github.com/studentinsights/studentinsights/pull/2477, explicitly limiting to SHS students.


kevinrobinson:
selfie

PR #2479, opened 2019-05-24
Counselor Meetings: Style tweaks, inclusive time filter

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2477, update styling for narrow screens, and make time filter inclusive (eg, include today).


kevinrobinson:
selfie

PR #2483, opened 2019-05-28
Transition notes: Show new transition inline in profile list after saving

kevinrobinson:
# Who is this PR for?
8th grade counselors

New transition notes don't appear inline, so it's not clear that they are saved.

# What does this PR do?
Fixes that, updating the UI on confirmed save.  This punts on editing inline still.

# Checklists
+ [x] Student Profile


kevinrobinson:
selfie

PR #2484, opened 2019-05-28
Searchbar: Mitigations for searching through many students

kevinrobinson:
# Who is this PR for?
districtwide admin

Autocomplete freezes for a bit with ~10k students.

# What does this PR do?
Mitigates it by adding a min-length, delaying the responsiveness, and adding a CSS scroll overflow.  The real problem is rendering so many items in the list, but jQuery UI hooks don't let us simply change that.  https://github.com/studentinsights/studentinsights/pull/2194 is the better path beyond short-term mitigations.

# Checklists
+ [x] Core 


kevinrobinson:
selfie

PR #2485, opened 2019-05-29
Student searchbar: Migrate to React component and optimize for larger lists 

kevinrobinson:
This builds on https://github.com/studentinsights/studentinsights/pull/2194 and replaces it.

# Who is this PR for?
districtwide admin, school admin in large buildings

This searchbar can be quite slow for districtwide admin or schoolwide admin in large buildings.

The existing jQuery autocomplete UI doesn't have good hooks for truncating the results returned, and uses a library we'd like to remove as a dependency to limit pieces of UI we depend on.

Right now the way the student searchbar matches it won't match last name, first name or with partial matches.

# What does this PR do?
Updates the student searchbar so that it matches any part of the student name in any order (eg, either last name or first name first is fine). Replaces autocomplete with react-autocomplete, keeping the same sessionStorage caching.

Revise the label text ahead of potentially adding a new searchbar for the feed notes, and add a placeholder text to the input.

Add some optimizations for truncating large lists (eg, >500 students), and in that case offer to link to the My students page instead.

This shows fake names from (fake_names.rb):
image

image

# Checklists
+ [x] Core 


kevinrobinson:
selfie

PR #2486, opened 2019-05-29
Revert "Student searchbar: Migrate to React component and optimize for larger lists "

kevinrobinson:
Bug on IE11, reverting to recover and fix.

Reverts studentinsights/studentinsights#2485


kevinrobinson:
selfie

PR #2487, opened 2019-05-29
Double revert for Student searchbar: Migrate to React component and optimize for larger lists

kevinrobinson:
Double revert with a patch.  Original PR was https://github.com/studentinsights/studentinsights/pull/2485, revert was https://github.com/studentinsights/studentinsights/pull/2486.

This patches the issue in IE11, which led to the contents of the page being blank, since the error was in `index.js`.  The problem was that `document.querySelectorAll(selector).forEach` isn't supported in IE11.  This adds a polyfill, and I verified manually that this fixes the issue locally.

Patch is commit on top, https://github.com/studentinsights/studentinsights/pull/2487/commits/0623c26312851e31527601b444b95bbfd7e0c132.


kevinrobinson:
selfie

PR #2491, opened 2019-06-01
Transitions: Update filters for transitions page

kevinrobinson:
# Who is this PR for?
8th grade counselors, 9th grade counselors and assistant principals

# What problem does this PR fix?
Folks can sort and search, but not obvious filter for starred / has a note / school.

# What does this PR do?
Adds them.  Also updates default sort to "Starred for discussion?" and adjust the column widths.

# Checklists
*Which features or pages does this PR touch?*
+ [x] Transitions

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Improved specs for existing code in need of better test coverage
+ [x] Manual testing made more sense here

PR #2492, opened 2019-06-01
Transitions: Adjust widths, fix bug with duplicate schools

kevinrobinson:
Fix to https://github.com/studentinsights/studentinsights/pull/2491/files, duplicate schools in the select.


kevinrobinson:
selfie

PR #2493, opened 2019-06-06
Reader profile: Initial prototype for design work

kevinrobinson:
# Who is this PR for?
Folks working with K1 students

Pulling together different data sources about early reading, and putting them in a frame that can help discussion protocols (eg, in MTSS).

# What does this PR do?
Adds a reader profile inline to the "Reading" tab of the profile for students.  It's behind a label, and contains its own error boundary to isolate it and let us iterate in ways that allow educators to critique this with production data for real students.

One new element is the `pdf-reader` gem for parsing IEP PDF documents as text, and some prototype-quality regular expressions cleaning that.  Another is adding `lunr` in JS to do searches within note and IEP documents, and this is also only prototype-quality at this point.

This PR factors out `<CleanSlateFeedView />` from the `<InlineStudentProfile />` for class lists, to re-use it here, and refactors the counselor meeting page to use this as well.

There are many design elements that are still prototype-quality or experimental, and the assumption is that this will change significantly over the summer before the fall.

Minimal:
Screen Shot 2019-06-06 at 11 48 02 AM

# Checklists
+ [x] Student Profile


kevinrobinson:
selfie

PR #2495, opened 2019-06-07
Transition notes: Support limited transition from 5th > 6th in Somerville

kevinrobinson:
# Who is this PR for?
Students and families in 5th grade at once school, transitioning schools

Help support counselors and admin in transitions across years and buildings.

# What does this PR do?
Extends the 8th > 9th grade transition to be used in a similar way for counselors and admin in one school for students transitioning schools at 6th grade.  This maps mostly directly to the 8th > 9th grade transition, in particular with how restricted information is handled.

Separately, other schools with program shifts where students switch buildings will do this also, but in different ways (eg, Somerville Argenziano program, Bedford elementary schools), and those import processes can be tackled separately in the summer so they're all ready for the fall.  They may re-use parts of the model or UI here but for now this PR punts on that to the minimal change needed during the time before the end of the school year.

As a caveat, for folks working K5 with districtwide access, the "Jump to next student" isn't smart enough to know they are only doing this for a smaller subset of students, but they can workaround using the "School transitions page."

# Checklists
+ [x] School transitions
+ [x] Student profile

### DIFF
--- a/app/assets/javascripts/student_profile/LightNotesDetails.js
+++ b/app/assets/javascripts/student_profile/LightNotesDetails.js
@@ -117,9 +117,7 @@ export default class LightNotesDetails extends React.Component {
   renderTakeNotesButton() {
     const {currentEducator, student} = this.props;
 
-    // Only K8 counselors with access can write transition notes,
-    // and only for 8th graders.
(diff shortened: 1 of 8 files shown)


kevinrobinson:
selfie

PR #2497, opened 2019-06-07
Transitions: Fix default sort order for starred

kevinrobinson:
https://github.com/studentinsights/studentinsights/pull/2491 tried to do this, but the sort direction was backwards.


kevinrobinson:
selfie

PR #2498, opened 2019-06-07
Transitions: Fix bug in which students are in scope for elementary transitions

kevinrobinson:
Fixes a bug in https://github.com/studentinsights/studentinsights/pull/2495 with the scoping for 5th > 6th grade students changing schools.


kevinrobinson:
selfie

PR #2499, opened 2019-06-07
My Students: Fix sort order for program

kevinrobinson:
Support sorting by program properly, since the "pretty" text isn't the same as the raw field value.


kevinrobinson:
selfie

PR #2501, opened 2019-06-07
Reader Profile: Moving parsing and segmenting IEP text to server

kevinrobinson:
# Who is this PR for?
K3 educators

This is still just experimental, but moving parsing of IEP text to the server to run better data quality audits on it.

# What does this PR do?
This is focused on just the dates first, but has experimental bits for the service grid and segmenting by sentences too.

# Checklists
+ [x] Reader profile


kevinrobinson:
selfie

PR #2502, opened 2019-06-11
Transitions: Add label to allow editing; fix bug with starred in edit dialog

kevinrobinson:
For 8th grade counseling team, allowing whitelisting for editing when needed.


kevinrobinson:
selfie

PR #2505, opened 2019-06-11
Transitions: Fix IE11 layout bug on dialog

kevinrobinson:
Fix layout styling bug for IE11 only, and write background colors as rgba.


kevinrobinson:
selfie

PR #2507, opened 2019-06-27
District admin: Export 'wide' students spreadsheet

kevinrobinson:
# Who is this PR for?
district admin

district admin doing other kinds of rostering for summer school that want to be able to work from data within Student Insights, but permission and authorization is different and idiosyncratic for different summer school programs

# What does this PR do?
Adds a districtwide export link for exporting a "wide" CSV with columns of data about students.  This isn't all data, but a good baseline for one-off rostering cases like this that aren't common or important enough to support directly.

Screen Shot 2019-06-27 at 9 59 45 AM

# Checklists
+ [x] School Overview
+ [x] District admin


kevinrobinson:
This bundles commits for fixing Ruby whitespace issues that were sitting on master.

In response to looking at a security vulnerability report in eslint, it turns out that the vulnerability is very low risk for the project.  Since we were a bit behind here, I decided to update eslint and related packages anyway, including updating code and tests to conform to new whitespace and special character rules.

PR #2509, opened 2019-06-27
Security: Add SECURITY.md with note about responsibile disclosure

kevinrobinson:
This adds a written security policy, linking to the security section in the privacy policy and adding a note for how to do do responsible disclosure.


kevinrobinson:
selfie

PR #2510, opened 2019-06-27
Update robots.txt to show icon, logo and description for search engines

kevinrobinson:
Adding more signals to users.


kevinrobinson:
selfie

PR #2511, opened 2019-06-28
Import: Add DataFlow descriptions to importer classes, with tests, for making visible in UI

kevinrobinson:
# Who is this PR for?
project leads

The log of import tasks is somewhat visible, but it's more intended for developers.  There's no way for district project leads to see importers directly, particularly for other kinds of imports that are semi-automated, or from different sources other than regular SIS syncs.

# What does this PR do?
Adds some minimal description in code of what these are, intended to be put in a UI for users, alongside other kinds of data flows not yet visible (eg, student voice, reading data, transitions).

I'm mot sure doing this in code over docs adds real value, but trying it out for now.

# Checklists
+ [ ] Imports


kevinrobinson:
selfie

PR #2513, opened 2019-07-01
Reduce default options for wide students export

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2507, reducing default options for export so it's faster from the web.


kevinrobinson:
selfie

PR #2517, opened 2019-07-03
Services: Show searchable, filterable list of all services for authorized students

kevinrobinson:
# Who is this PR for?
product team, project leads

The roster overview page gives visibility into services in a school, but not across the district and not for all time.  As we're importing services from transition forms, or services from reading programs, this makes that more visible.

# What does this PR do?
Adds a "services" page, with a table listing all these.  Adds it to the district page for now.  It would be the starting point for where to add a global export as well (eg, for other analysis).

Screen Shot 2019-07-03 at 5 06 44 PM

# Checklists
+ [ ] Services page


kevinrobinson:
selfie

PR #2523, opened 2019-07-18
Reading Sheets Import and Process prototype

edavidsonsawyer:
# Who is this PR for?
Educators

Reading data from schools is kept in spreadsheets that must be manually moved for the data to reach Insights 

# What does this PR do?
Introduces Importer and Processor classes that will fetch sheets from a google drive folder and process the data so it can be added to Insights. 

This assumes the sheets within a grade follow a consistent template, but the specific template can be flexible between grades. 

# Checklists
+ [X] Core 

+ [X] Included specs for changes
+ [X] Manual testing made more sense here


edavidsonsawyer:
@kevinrobinson I just realized I wasn't rerunning the tests when I reopened this, so I don't think these failures are from this branch. 

That aside, when you have a chance, would you take a look and see if this looks suitable for the reading sheets import?

kevinrobinson:
@edavidsonsawyer This is awesome progress!  👍   Apologies on the delay in not noticing this earlier in the week.

I'm assuming this is built towards the format in `example 3 (somerville 2nd grade reading)` is that right?  If so, I think there's three good next steps to do here, and then let's merge this is and hopefully the timing will align where we have some production data ready then too.  Let me know if you want to chat more over email or video too!

First, let's update the format to use a new sheet called `dev SPS Reading Benchmarks (7/19/19)` that is added in drive.  This is the proposed new format for the coming year for benchmarks, and I marked the older reading examples as deprecated now.  You'll notice there's a hidden row at the top with explicit computer-friendly column keys, and then there are two rows for the humans using the sheets :)  So after parsing the CSV, the processor code can just skip the first two rows when processing.  So let's update this to use that sheet format, and you could update the processor to be able to handle all the tabs at once, or just starting with one tab/grade is great too.

Second, let's remove any computed values (eg, how many levels of growth).  The new format should discourage this more explicitly, but either way if there is anything that's computed we don't need to import it as we can just re-compute ourselves.  There are sometimes subtle differences here, and removing that will just simplify a bit.

Third, there are validations on the `ReadingBenchmarkDataPoint` model that only allow adding `assessment_key` values that are whitelisted.  That's a small step to help us stay sane when adding a lot of different values.  So the test here is awesome for testing the processor code, but the new processor code introduces new `assessment_key` values that aren't whitelisted, so if we tried `rows.each {|row| ReadingBenchmarkData.create!(row) }` many would fail (eg, for new values like `growth_f_and_p`).  So if the processor has code to add new data points, let's add them into the whitelist in `ReadingBenchmarkDataPoint` too.  And the tests can check this by doing something like above and then making assertions on them.

Also, I think it's great that this PR punts the "import" step for now, which will have to handle the "sync" process where data points change over time.  We can do that separately, so focusing on the processor class seems like great scoping. 👍

kevinrobinson:
@edavidsonsawyer re test failures, this is what I see here: https://travis-ci.org/studentinsights/studentinsights/jobs/561886950#L3144

Screen Shot 2019-07-26 at 12 44 25 PM

So these are just linter failures, you can fix most of the minor things like whitespace by running `rubocop -a` and if it can't fix them automatically it will show you where the issue is.

db/schema.rb
```diff
@@ -13,6 +13,7 @@
 ActiveRecord::Schema.define(version: 2019_06_06_165748) do
 
   # These are extensions that must be enabled in order to support this database
+  enable_extension "fuzzystrmatch"
```
kevinrobinson: ah thanks for catching and adding this! 👍

app/importers/reading/mega_reading_processor.rb
```diff
@@ -19,12 +19,14 @@ def initialize(educator_id, options = {})
...
 
-  def import(file_text, options = {})
+  def process(file_text, options = {})
     reset_counters!
 
     # parse
     rows = []
-    StreamingCsvTransformer.from_text(@log, file_text).each_with_index do |row, index|
+    #last row before data is the header
+    string = (@header_rows_count > 1) ? file_text.lines[@header_rows_count-1..-1].join : file_text
```
kevinrobinson: Can you say more about moving this before the CSV is parsed, here rather than afterward in `flat_map_rows`?

edavidsonsawyer: Sure thing. `StreamingCSVTransformer` needs a string that can be parsed as a csv. This is here to trim header rows that we aren't going to use later to reference data points, assuming here the last line before the data begins is the "real" header. I haven't been able to find a cleaner way to specify which row in the csv to treat as the header. `flat_map_rows` just takes a csv row as an argument. It needs to reference the header data. 

An alternative approach would be to define the headers explicitly in code rather than trying to get them from the csv.

kevinrobinson: Ah got it, thanks!  👍 Yeah so the problem is that in that template, the actual "rows" in the second row since the first row has other stuff helpful for people entering data, got it.

The new reading template format reverses that - the first row is hidden from people, so it makes for simple parsing, and then there some formats have a a few rows without any real data, with explanations for people about how to use the sheet, etc.

So this makes total sense as a workaround, and then we should be able to remove it with the new format.

edavidsonsawyer: That's great! Since this is likely to be sensitive to formatting changes, there are probably a few pieces here that might need to change with new formats. That shouldn't be a problem as long as the final template works well for everyone using it.

kevinrobinson: Yep, let me know if you want to chat more.  It's possible there will be some drift in the format from now until September, but I'm guessing it'll just be things like names to columns, or maybe skipping a different number of header rows or something like that.

edavidsonsawyer:
@kevinrobinson This is now set up to use the new template, which simplifies the process quite a bit. It also removes the calculated fields and whitelists the new data points. 

I'm imagining the importer can call the sheets it wants, and feed each individual csv (the tab, in this case) to the processor before importing the results. If you think it's important to have the processor work on entire spreadsheets at once that's doable as well, we'd just configure the processor to handle an array of csvs.

app/importers/reading/mega_reading_processor.rb
```diff
@@ -107,7 +111,9 @@ def match_student(row)
     return [student.id, nil] if student.present?
 
     # fuzzy
-    last_first_name = row['Name']
+    last_first_name = row['student_last_names_first_names']
```
kevinrobinson: The new template has the name, but it also has `student_local_id` and for now let's peg it to just that.  So that means we can use the key `student_local_id` rather than the other ones above, and that we can remove the fuzzy matching.

When I used the fuzzy matching previously trying this on the console, I would typically try processing, see all the errors, then go change the sheet to fix ambiguous names or typos, and re-run until it was correct.  I think that kind of workflow will be harder for educators to do themselves, or for us to make tools for.  So if we just use the lasid, then the code is simpler and the import rule is simpler: *to import data, you have to add the lasid to the sheet, here's how to look it up if you're not sure or here's a master sheet with all student names and lasids you can grab from* etc.

app/importers/reading/mega_reading_processor.rb
```diff
@@ -19,11 +19,13 @@ def initialize(educator_id, options = {})
...
   end
 
-  def import(file_text, options = {})
+  def process(file_text, options = {})
     reset_counters!
 
     # parse
     rows = []
+    #last row before data is the header
+    # string = (@header_rows_count > 1) ? file_text.lines[@header_rows_count-1..-1].join : file_text
```
kevinrobinson: Let's remove the comment if we don't need it anymore.

app/importers/reading/somerville_mega_reading_importer.rb
```diff
@@ -47,10 +48,11 @@ def reset_counters!
...
   end
 
-  def read_or_fetch_csv
+  def read_or_fetch_sheet
     raise 'unfinished'
-    # TODO read from google or mock and return `streaming_csv` that responds
-    # to #each_with_index
+    # fetcher = GoogleSheetsFetcher.new
+    # fetcher.get_spreadsheet(SHEET_URL)
+    # TODO add sheet url to ENV
```
kevinrobinson: I'm figuring you left this so I can add in the env bits when shipping, thanks!  If there's something else to know about here just let me know.  I think this will be folder-based, like this: https://github.com/studentinsights/studentinsights/blob/master/app/importers/transitions/bedford_davis_transition_notes_importer.rb#L58

spec/models/reading_benchmark_data_point_spec.rb
```diff
@@ -36,6 +36,46 @@
         benchmark_assessment_key: :dibels_dorf_acc,
         json: { value: 96 }
       })
+      ReadingBenchmarkDataPoint.create!({
```
kevinrobinson: nice! 👍

spec/importers/reading/reading_processor_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+  describe 'integration test' do
+    it 'works on happy path' do
+      pals = TestPals.create!
+      students = create_test_students(pals)
+      processor = MegaReadingProcessor.new(pals.uri.id, header_rows_count: 3)
+      reading_data = processor.process(fixture_file_text)
+      puts reading_data[0].as_json(except: [:student_id, :imported_by_educator_id])
+
+      expect(reading_data.size).to eq 2
+      expect(reading_data[0].as_json(except: [:student_id, :imported_by_educator_id])).to eq([
```
kevinrobinson: These tests are great!  Can you update to also verify the student is correct?  This would be a major problem if it were not :)  Since the student record will have a different id reach run, you can store the record in a variable, and then reference its id in the test assertion (eg, `donald = Student.create!(..)` and then later check that `student_id: donald.id`)

app/importers/reading/somerville_mega_reading_importer.rb
kevinrobinson: Can you also either comment on what you're thinking about for the other methods here that still `raise 'unfinished'?  I'm figuring that some of them will still need to be done in the next pass (eg, doing the "syncing" with existing records in the database) but that others are now stale and can be removed.

app/importers/reading/mega_reading_processor.rb
kevinrobinson: Should `header_rows_count` be updated to two?  In the template it looks like there's the header row and then two additional rows for educators explaining things and we'd want to skip both those additional rows.  We could rename this too if it's ambiguous what "header" here means too.

kevinrobinson: addressed in https://github.com/studentinsights/studentinsights/pull/2544/files#diff-f01c15c5800dc5ef4b4a64b7ad07baadR23

app/importers/reading/mega_reading_processor.rb
kevinrobinson: addressed in https://github.com/studentinsights/studentinsights/pull/2544/files#diff-f01c15c5800dc5ef4b4a64b7ad07baadL100

spec/importers/reading/reading_processor_spec.rb
kevinrobinson: addressed in https://github.com/studentinsights/studentinsights/pull/2544/files#diff-f9626d0b749afc86a13a97049c70cceaR36

kevinrobinson:
Here's our notes talking through the cases for update.  I think we can do the "importer" work without doing this first, and just add the data points we find.

# step 1
```
class MegaReadingImporter
  def import
    # get all the tabs
    # get all the rows in the tab, where row is like `(eric, ORF WPM, 7, fall)`

    # just add them all to the database
    rows.each {|row| ReadingBenchmarkDataPoint.create!(row) }
    end
```

That won't handle if we run it twice - we'll get extra data points, but that's okay as a good first step.

### step 2
```
# Importer is in charge of everything!
class MegaReadingImporter
  def import
    # get all the tabs
    # get all the records in the tab

    # it just gives the syncer all the rows as new active record models
    # but it doesn't delete or handle anything else, it still is just creating all the records twice
    # but does that through the syncer, tracks stats and stuff
    end
```

### step 3
After that I might be:
```
# Importer is in charge of everything!
class MegaReadingImporter
  def import
    # get all the tabs
    # get all the records in the tab

    # it knows how to call the syncer in the right ways
    # use the syncer to do smarter syncer for different cases
    end
```

Those cases might be as simple as: "delete everything that ever came from this form."  But there might be important edge cases we want to handle, like the teacher deleting a row because a student changed homeroom.  And we'll have to translate those to simple guidance for educators too.

Alternately, we might just want to say "if the student is in the sheet, update their records for that time period in the sheet."

We don't have to handle all those edge cases, simple is better!  We could say "if you do X, Y will happen" and "here are two edge cases that would break, but we're not handling now."

kevinrobinson:
@edavidsonsawyer FYI I pushed some other commits on https://github.com/studentinsights/studentinsights/pull/2544 related to the processor code.  You should be able to pull them all in, but if there's conflicts you're not sure how to resolve just ping me and I am happy to help with merging!

edavidsonsawyer:
Hi @kevinrobinson this should be ready for a look.

app/importers/reading/somerville_mega_reading_importer.rb
```diff
@@ -1,67 +1,90 @@
...
-
-    # TODO add dependencies as instance variables
-    # @student_ids_map = ::StudentIdsMap.new
+  def initialize(school_year, options = {})
+    @school_year = school_year
+    @log = options.fetch(:log, Rails.env.test? ? LogHelper::Redirect.instance.file : STDOUT)
     # @matcher = options.fetch(:matcher, ImportMatcher.new)
-    @syncer = ::RecordSyncer.new(log: @log)
-    reset_counters!
+    # @syncer = ::RecordSyncer.new(log: @log)
```
kevinrobinson: There's some leftover commented out code here and in other places.  Let's remove all that before merging.

app/importers/reading/somerville_mega_reading_importer.rb
```diff
@@ -1,67 +1,90 @@
+# For importing reading data from Google Sheets or a local folder of csvs.
```
kevinrobinson: We removed the local folder option to setup mocks in tests, right?

app/importers/reading/somerville_mega_reading_importer.rb
```diff
@@ -1,67 +1,90 @@
...
+    @valid_student_names_count = 0
+    @invalid_student_name_count = 0
+    @valid_data_points_count = 0
+
+    processor = MegaReadingProcessor.new(read_uploaded_by_educator_from_env, @school_year)
+    streaming_csvs.each_with_index do |tab, index|
+      rows, meta = processor.process(tab.tab_csv)
+      log("processed #{index} sheets.")
+      rows.each_with_index {|row, index| import_row(row, index)}
+      valid_student_names_count, valid_data_points_count, invalid_student_name_count = meta.values_at(
```
kevinrobinson: I don't think you need to pull these metrics out here, the processor class already logs these stats: https://github.com/studentinsights/studentinsights/blob/master/app/importers/reading/mega_reading_processor.rb#L43 - but we'll need to pass the `log` instance along when creating the processor instance.

app/importers/reading/somerville_mega_reading_importer.rb
```diff
@@ -1,67 +1,90 @@
...
     log('Starting loop...')
-    reset_counters!
-    streaming_csv.each_with_index do |row, index|
-      import_row(row)
-      log("processed #{index} rows.") if index % 1000 == 0
+    @valid_student_names_count = 0
+    @invalid_student_name_count = 0
+    @valid_data_points_count = 0
+
+    processor = MegaReadingProcessor.new(read_uploaded_by_educator_from_env, @school_year)
```
kevinrobinson: Ah cool, this is different than what I was thinking, two 😄 😄 are better than one 😄 .  This approach re-uses the process instance for each tab, which would be good for when things are stable since we can see aggregate stats across all sheets and tabs.

When deploying some other importers, I found it useful to create a new processor class for each tab, since otherwise the `stats` metrics are aggregated across all sheets and tabs, and when you see something like `invalid_student_name_count: 3` you don't know which sheet or tab to investigate or debug.  In  https://github.com/studentinsights/studentinsights/blob/master/app/importers/transitions/bedford_davis_services_importer.rb#L76 you can see how this works for example.

Let's go with this approach here and then we can see what comes up when deploying.

app/importers/reading/somerville_mega_reading_importer.rb
```diff
@@ -1,67 +1,90 @@
...
-    # TODO the main loop!  has to respect options passed to the
-    # importer (eg, school_ids_ids filter)
+    benchmark_data_point = matching_student_insights_record_for_row(row)
+    benchmark_data_point.save!
+  end
+
+  def matching_student_insights_record_for_row(row)
+    benchmark_data_point = ReadingBenchmarkDataPoint.find_or_initialize_by(
+      student_id: row[:student_id],
+      educator: Educator.find(row[:educator_id]), #because sheets don't explicitly link an educator to a record
```
kevinrobinson: Can we can remove the `#find` call, and just set `educator_id` directly?

app/importers/reading/somerville_mega_reading_importer.rb
```diff
@@ -1,67 +1,90 @@
...
+    streaming_csvs = @fetcher.get_tabs_from_sheet(sheet_id)
   end
 
-  def read_or_fetch_csv
-    raise 'unfinished'
-    # TODO read from google or mock and return `streaming_csv` that responds
-    # to #each_with_index
+  def read_uploaded_by_educator_from_env
+    educator_login_name = ENV.fetch('READING_IMPORTER_UPLOADED_BY_EDUCATOR_LOGIN_NAME', '')
+    uploaded_by_educator = Educator.find_by_login_name(educator_login_name).id
```
kevinrobinson: If env isn't set, the find call will return nil, and then the call to `#id` will raise.

Check out how `StudentVoiceSurveyImporter` uses `.try(:id)` to guard, so that if the env variable is not set, the more informative error below will raise.

app/importers/reading/somerville_mega_reading_importer.rb
```diff
@@ -1,67 +1,90 @@
...
-    @syncer = ::RecordSyncer.new(log: @log)
-    reset_counters!
+    # @syncer = ::RecordSyncer.new(log: @log)
+    @fetcher = options.fetch(:fetcher, GoogleSheetsFetcher.new)
   end
 
   def import
-    streaming_csv = read_or_fetch_csv
-    if streaming_csv.nil?
+    streaming_csvs = read_or_fetch_sheet
```
kevinrobinson: This importer class should work on a folder, not just a single sheet.  Check out https://github.com/studentinsights/studentinsights/blob/master/app/importers/transitions/bedford_davis_transition_notes_importer.rb#L43 for a model of how this is set in the env and how to change the call to the fetcher.

edavidsonsawyer: Gotcha. I had assumed the template sheet reflected the way the data would be entered - all in one sheet. If instead it will be many sheets in a folder, we'll do that.

kevinrobinson: Sounds good!

app/importers/reading/somerville_mega_reading_importer.rb
edavidsonsawyer: Okay, I'll leave this as-is for now, but it wouldn't be much trouble to use a new instance per tab if we need to chase down problems with the sheets.

edavidsonsawyer:
@kevinrobinson I've made the changes recommended above. I also gave this a shot using a test account with the fetcher and the template sheet, and it seemed to work well with a handful of students across the different grades.

kevinrobinson:
@edavidsonsawyer Great!  Merging and going to try kicking the tires here. 👍 🎉

PR #2524, opened 2019-07-18
Monitoring: Rollbar functions can't be called without explicitly setting 'this'

kevinrobinson:
Rollbar reporting in `RollbarErrorBoundary` wasn't working because our code assigned the reporting function to a variable without binding 'this'.  This led to the Rollbar library code throwing and dropping the errors. 👣 🔫 

# What does this PR do?
Removes use of variable.


kevinrobinson:
selfie

PR #2521, opened 2019-07-16
Search: Allow web search syntax when searching notes

kevinrobinson:
# Who is this PR for?
educators

# What does this PR do?
Supports typical search syntax like +/- and quoted phrases.

# Checklists
*Which features or pages does this PR touch?*
+ [x] Search notes

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Manual testing made more sense here

PR #2525, opened 2019-07-18
Defer freezing Object.prototype until after vendor code

kevinrobinson:
Fixing "malformed field ref string" in https://rollbar.com/somerville-teacher-tool/studentinsights/items/408 coming from lunr.

This was introduced inhttps://github.com/studentinsights/studentinsights/pull/2521/files#diff-9fcaa1e447bc24f5885b46562d22090dR1, trying to add another layer of defense for JS.  While that work surfaced the Object.prototype pollution from moment.js, this bug came from a related issue, in that the common pattern `Foo.prototype.toString = function() { ...}` no longer works if `Object.prototype` is frozen.  This led to errors in lunr at read time, because building the index didn't work as expected because `toString` methods couldn't be defined using the pattern above.

There's probably not a lot of value here, so just moving this freeze until after our code loads.  If there are more problems, not sure it's worth keeping at all.

### DIFF
--- a/ui/freeze.js
+++ b/ui/freeze.js
@@ -1,13 +1,16 @@
-// Prevent further Object.prototype pollution from happening
+// Prevent further Object.prototype pollution from happening after
+// our code initially loads.
 // (example: https://snyk.io/vuln/SNYK-JS-SETVALUE-450213)
-
+//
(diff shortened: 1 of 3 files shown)


kevinrobinson:
selfie

PR #2526, opened 2019-07-19
Somerville: Disable STAR importer

kevinrobinson:
The vendor is doing some kind of migration, and started this for Somerville in June 2019.  This changed the format of the export, and there are other aspects to this that are moving slowly.  This PR turns off importing from the vendor altogether in Somerville, and district folks have the ball for collaborating with the vendor on these changes and then moving forward again in the fall.


kevinrobinson:
selfie

PR #2527, opened 2019-07-23
Services: Update sort for timeline column

kevinrobinson:
Update to https://github.com/studentinsights/studentinsights/pull/2517, fixing how the "timeline" sorting column works to be more intuitive.


kevinrobinson:
selfie

PR #2528, opened 2019-07-25
Notes: Processors for self-serve note imports

kevinrobinson:
# Who is this PR for?
educators, students

# What does this PR do?
Adds processor classes for importing questionairre-style forms from educators, and restricted notes.  These are for use during the transition to next school year, but also with the assumption that these formats will be shared more widely for self-serve tools.

# Checklists
+ [x] Manual import processes


kevinrobinson:
selfie

PR #2529, opened 2019-07-26
Experiment: Add Sorbet first pass

kevinrobinson:
This is an experiment trying out a first step of adoption for Sorbet, a gradual type checker.  This basically does the defaults, and has a few commits of me understanding what the next steps would be and reverting them.  It also adds `sorbet-rails`.

The two issues that come up to increase type checking are use of `class Foo > Struct.new(:bar)`, and use of Rails finder methods like `Educator.find_by_login_name`.  Both could be reworked with straightforward syntax changes.  It did find one real bug on the first file that I added `typed: true` to, which is fixed here and I'll fix separately.

The runtime checking works for all tests, so we could move forward with deploying this to demo, shake the tires and then deploy.  The incremental adoption experience seems far better than other gradual typing tools, with different grades of checks per files.  And unrolling this altogether would just involve removing (or ignoring) the comments and the gems.

Going to let this sit here for a bit.


kevinrobinson:
This is stale, and no plans to do this now, so closing.  The relevant info is here if we want to revisit in the future.

PR #2530, opened 2019-07-26
Fix bug in feed filter for ELL educators

kevinrobinson:
thanks sorbet! https://github.com/studentinsights/studentinsights/pull/2529


kevinrobinson:
selfie

PR #2531, opened 2019-07-26
Fix Rollbar binding

kevinrobinson:
😞 similar to https://github.com/studentinsights/studentinsights/pull/2524


kevinrobinson:
selfie

PR #2535, opened 2019-07-29
Service: Add Bedford service types, part1

kevinrobinson:
For importing from transition surveys, and then updating UI.  Also adds columns for more metadata about services, from docs.


kevinrobinson:
selfie

PR #2536, opened 2019-07-29
Service: Import social emotional services from Bedford transition forms

kevinrobinson:
# Who is this PR for?
Bedford students, families, educators

# What does this PR do?
Imports the Bedford transition forms, processing them to update the services for the year.  On top of https://github.com/studentinsights/studentinsights/pull/2535.

# Checklists
*Which features or pages does this PR touch?*
+ [x] Student Profile
+ [x] School Overview

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [X] Included specs for changes
+ [X] Manual testing made more sense here

PR #2537, opened 2019-07-30
Service: Add Bedford service types, part2

kevinrobinson:
Migration step 2 for https://github.com/studentinsights/studentinsights/pull/2535, split out to isolate test changes.

Next is https://github.com/studentinsights/studentinsights/pull/2536.


kevinrobinson:
hmm, can't make sense of this: https://travis-ci.org/studentinsights/studentinsights/jobs/565628102#L3137

Screen Shot 2019-07-30 at 2 57 22 PM

It seems the migration is silently failing, and just isn't doing what it says it's doing ❓ 😕

kevinrobinson:
k, `.column_names` is [cached](https://apidock.com/rails/ActiveRecord/Base/column_names/class).

kevinrobinson:
huh today I learned table structure is cached across migration files, never would have expected that.  references:
- https://stackoverflow.com/a/11355761/5711971
- https://api.rubyonrails.org/classes/ActiveRecord/Migration.html
- https://github.com/rails/rails/issues/6939#issuecomment-6738014

And splitting it into separately PRs or commits doesn't resolve the issue either, since all migrations run at once in a clean Travis environment anyway.  Still not sure why this didn't reproduce locally though; I thought I had dropped and re-run all migrations.  Either way, shipping this and moving on.

kevinrobinson:
selfie

PR #2538, opened 2019-07-30
Service: Add Bedford service types, part3

kevinrobinson:
After https://github.com/studentinsights/studentinsights/pull/2537, since the migration can't happen in the same code as adding this to the database seeds.  Note that tests on this PR will still fail until we merge https://github.com/studentinsights/studentinsights/pull/2537.


kevinrobinson:
closing, figured this out in https://github.com/studentinsights/studentinsights/pull/2537#issuecomment-516581271.

PR #2539, opened 2019-07-30
Fix spec bug with assuming same order

kevinrobinson:
Well down the rabbit hole of being frustrated with the build today :\

Fixes https://travis-ci.org/studentinsights/studentinsights/jobs/565549159#L3317, which looks like it's from a test bug mistakenly assuming the same order for records.


kevinrobinson:
selfie

PR #2540, opened 2019-07-30
Force sslmode=require

kevinrobinson:
This is already set in production.  More context is in internal docs, ask @kevinrobinson.


kevinrobinson:
selfie

PR #2541, opened 2019-08-16
Fix tests with assertions or setup mistakenly using wall clock

kevinrobinson:
Came up in https://travis-ci.org/studentinsights/studentinsights/jobs/572807904 from the new school year.


kevinrobinson:
selfie

PR #2542, opened 2019-08-20
Services: UI for new elements describing services, update available choices and non-academic phaselines

kevinrobinson:
This updates the services UI in a few different places to show more information about service types, and support more ways that this differs across districts.  It's driven by changes for folks in Bedford K5 schools.

This PR will change the services available in Bedford on deploy.  There is an env switch for setting `SERVICES_INFO_DOC_URL` across districts, and this just won't be shown if not set.  And there's a label that controls that per-educator as well (`show_services_info`).  Deploying consists of verifying changes for Bedford, incrementally deploying info across districts for test users.  After that we can remove the label check in a subsequent PR.

### Student profile, services
Allow adding different service types, and include information about those service types if available.
Screen Shot 2019-10-21 at 5 31 42 PM

### Student profile, services list
Previously there was text describing different services, but this didn't apply to all districts.  This PR replaces that with a link to the chart the district has made describing the meaning of different services.  The intention is that if there's discussion about changes to be had, it's better to push those to docs and educator discussions rather than through `help@studentinsights.org`:

image

Screen Shot 2019-10-22 at 12 07 16 PM

### Student profile, attendance and behavior
Show different service types as phaselines on these graphs, depending on per-district config.
Screen Shot 2019-10-21 at 4 30 14 PM

Screen Shot 2019-10-21 at 4 30 06 PM


kevinrobinson:
selfie

PR #2546, opened 2019-08-23
Educators import: Mark `missing_from_last_export` and show project leads in permissions tools

kevinrobinson:
# Who is this PR for?
Project leads across districts

Typically project leads or district IT staff remove educators from the authentication system (eg, LDAP), which means they are not allowed to use Student Insights at all, regardless of past permissions.

By design, the district LDAP system controls who has access regardless of any data within Student Insights.  But Student Insights stores records for staff and students after they have been removed from the daily export, so that it can maintain past information (eg, who wrote a note about a student).  This distinction isn't tracked or shown.

# What does this PR do?
Updates the importer and educator model to track this with `missing_from_last_export`, similar to the `StudentsImporter`.   It also backports a fix for an edge case that hasn't come up to the `StudentsImporter`.

Uses this info in permissions tools for project leads so this process is visible for them.  Removes some old cruft from the administrate config.

This PR doesn't update the authentication process, but a separate one will add a defensive layer beyond the district's own authentication check that ensures the educator record was still present in the latest SIS export.

### adjust permissions list
Screen Shot 2019-08-23 at 12 04 17 PM

### view
Screen Shot 2019-08-23 at 12 09 18 PM

### edit
Screen Shot 2019-08-23 at 12 04 30 PM

# Checklists
+ [x] Educators import
+ [x] Project lead permissions tools

### DIFF
--- a/app/dashboards/educator_dashboard.rb
+++ b/app/dashboards/educator_dashboard.rb
@@ -1,5 +1,6 @@
 require "administrate/base_dashboard"
 
+# See https://administrate-prototype.herokuapp.com/getting_started
 class EducatorDashboard < Administrate::BaseDashboard
   # ATTRIBUTE_TYPES
   # a hash that describes the type of each of the model's fields.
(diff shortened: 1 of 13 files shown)


app/dashboards/educator_dashboard.rb
```diff
@@ -8,35 +9,26 @@ class EducatorDashboard < Administrate::BaseDashboard
   # which determines how the attribute is displayed
   # on pages throughout the dashboard.
   ATTRIBUTE_TYPES = {
-    school_id: SchoolNameField.with_options(searchable: false),
-    homeroom: HomeroomNameField.with_options(searchable: false),
-    students: Field::HasMany.with_options(searchable: false),
-    interventions: Field::HasMany.with_options(searchable: false),
-    id: Field::Number.with_options(searchable: false),
-    email: Field::String.with_options(searchable: false),
-    encrypted_password: Field::String.with_options(searchable: false),
```
kevinrobinson: These seem scary but most haven't existed on the model in a long time.  Without digging into Administrate all the way, these appear to be only limiting what options are available for other config, based on how this behaved editing them here.

That being said, another audit pass on administrate would be a good security step.

kevinrobinson:
selfie

kevinrobinson:
Done this migration, verified across districts.

PR #2547, opened 2019-08-23
Fix administrate use of active instead of active?

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2546; the administrate setup is apparently is cool with you describing fields that don't exist on the model and doing its best to try to display them anyway.

Also remove a line that looks like it used to be a permissions guard, but doesn't make sense anymore given there are more restrictive guards elsewhere and the check is in a place unrelated to what the boolean class is about.


kevinrobinson:
selfie

PR #2568, opened 2019-08-29
Home page feed: Show homerooms for all schools but SHS

kevinrobinson:
# Who is this PR for?
FC students, educators

For SHS, homerooms are administrative and not super meaningful, and this previously was removed from the home page feed.  But it was done by checking for high schools, when it should have been more tightly scoped.  This led to homeroom and teacher info not being shown in the feed for FC students.

# What does this PR do?
Updates the check to only hide for SHS in Somerville in particular.

# Checklists
+ [x] Home page
+ [x] Student Profile
+ [x] My Notes

### DIFF
--- a/app/assets/javascripts/feed/FeedCardFrame.js
+++ b/app/assets/javascripts/feed/FeedCardFrame.js
@@ -99,36 +99,42 @@ const styles = {
 };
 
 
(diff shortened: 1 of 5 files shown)


kevinrobinson:
selfie

PR #2569, opened 2019-08-29
Monitoring: Send error objects to Rollbar separately so it can serialize

kevinrobinson:
JS properties in `Error` objects are non-enumerable by default, so `{error, moreInfo}` means Rollbar serializes the error as `{}`.  Update the way we call to keep these separate params.


kevinrobinson:
selfie

PR #2572, opened 2019-08-29
Sign in: Add another layer of defense on our side, whether educator is active

kevinrobinson:
District IT systems are responsible for authenticating users, but Student Insights has an additional layer of defense that checked the user matches a known user from the SIS export.  Building on https://github.com/studentinsights/studentinsights/pull/2546, this adds another layer of defense to the sign in process on our side, before communicating with districts, and essentially checks that the user was also included in the most recent export.  This further protects district authentication systems.


kevinrobinson:
selfie

PR #2573, opened 2019-08-29
Precompute: Limit to active educators

kevinrobinson:
Building on https://github.com/studentinsights/studentinsights/pull/2546, and similar to https://github.com/studentinsights/studentinsights/pull/2562, limit precomputing school overview data to active educators only.

Separately, move the specs into folders that match the code (leftover from an old refactor probably).


kevinrobinson:
selfie

PR #2575, opened 2019-08-29
Profile insights: Update SHS prioritization

kevinrobinson:
# Who is this PR for?
SHS students, families, educators

The fall student voice surveys at SHS aren't showing in the insights box, since previously we favored mid-year reflections if they were present (eg, in the middle of the year).  This doesn't make sense at the start of the year, but hadn't been updated yet.

# What does this PR do?
Update profile insights box to include fall surveys, and previous year surveys as well.  We can make this smarter later, probably should ask for all insights (with timestamps) then sort, or have them express some prioritization.

# Checklists
+ [x] Student Profile

### DIFF
--- a/app/lib/profile_insights.rb
+++ b/app/lib/profile_insights.rb
@@ -4,10 +4,11 @@ class ProfileInsights
     @time_now = options.fetch(:time_now, Time.now)
   end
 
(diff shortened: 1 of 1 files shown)


kevinrobinson:
selfie

PR #2576, opened 2019-08-29
My students: Remove school link for all

kevinrobinson:
This link is misleading; not all folks have access so they click in and get denied.  A few teachers have run into this in the last few days, so removing the link for now and can add a check back later.  Folks with districtwide access don't lose anything other than convenience clicking in.


kevinrobinson:
selfie

PR #2583, opened 2019-09-04
Maintenance: Remove older homeroom and student factories

kevinrobinson:
Maintenance that came up working on homeroom authorization.

This also documents in specs some unexpected behavior related to `Homeroom#grade`, which we should migrate away from separately (towards querying students on-demand)


kevinrobinson:
selfie

PR #2584, opened 2019-09-04
Authorizer: Update homeroom method to be built from student methods

kevinrobinson:
# Who is this PR for?
students, families, educators

This fixes an authorization gap between the methods answering "what homerooms can an educator access" and "what students can an educator access"?  For users with access to the homeroom page, this can result in two problems.  This has come up at the start of this school year for this first time.

First, the homeroom dropdown shows homerooms where they wouldn't have access to view full profiles for any students.

Screen Shot 2019-09-04 at 12 28 44 PM

Second, if using the dropdown to visit other homeroom pages, those pages would show name and other information about those students, but the student authorization check would prevent access to the profile or to the student photo.  This does result in authenticated educators being able to view enrollment information, and other information including disability status.

Screen Shot 2019-09-04 at 12 28 36 PM

Student-level authorization rules blocked photos from loading, or from full profiles from being accessed.  Monitoring on this is how we came to learn about the problem (there have been no reports from educators).

The core difference in semantics between the previous homeroom-level authorization method and   student-level authorization is that the previous homeroom method would include homerooms at the same grade level in the same school as the educator's homeroom, if it was assigned.  This would also result in older homerooms (with no students listed) being incorrectly included in the list.

# What does this PR do?
- [x] Move previous homeroom authorization method from `Educator` to `Authorizer`, rename it as deprecated and require callers to pass `acknowledge_deprecated:true`.
- [x] Audit that there are no calls to previous authorization method outside of tests.
- [x] Add new method, built off of student-level authorization.
- [x] Add new tests to verify expected behavior of new method, including that it no longer includes homerooms at the same grade level and school, and that it excludes homerooms with no active students.  Add parity tests showing the behavior is the same as the previous method, and describe all differences in test cases.  Add option to describe differences with "no students" homerooms in test cases.
- [x] Fix old spec setup bugs where `grade: 'K'` was set on homeroom directly, instead of `KF` on student model.  Add explicit validation of grade on `Homeroom` model as well.
- [x] Add `PerfTest` to test for performances regressions, verify no issues across districts.
- [x] Run manual data quality checks in production.
- [x] Add more test cases verifying current behavior to `HomeroomsController`

This does not cut over the behavior in `HomeroomsController`, so nothing should change by shipping this, and I'll do a final check on this before cutting over in a separate PR.

# Checklists
This won't change anything as-is, but when cut over, it will influence:
+ [x] Homeroom
+ [x] Core, authorization


kevinrobinson:
selfie

PR #2585, opened 2019-09-04
Maintenance: Remove authorizer instance variable on ApplicationController

kevinrobinson:
This is for after https://github.com/studentinsights/studentinsights/pull/2584 lands.

In https://github.com/studentinsights/studentinsights/pull/2584, it came up that instance variables on `ApplicationController` leak across calls to controller actions in rspec tests.  This is not how I thought it worked, but turns out all the other uses of tests like this never came across this particular combo.  Those kinds of tests are useful for exhaustively testing authorization rules for a given setup (eg across combinations of things like `[educator, student]`).

This removes the instance variable in `ApplicationController` related to authorization; it's not doing much there.  This has no impact on authorization rules.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -109,6 +109,6 @@ class ApplicationController < ActionController::Base
 
   private
   def authorizer
-    @authorizer ||= Authorizer.new(current_educator)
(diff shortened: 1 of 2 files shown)


kevinrobinson:
selfie

PR #2586, opened 2019-09-04
Authorization: Enable homerooms controller cutover 

kevinrobinson:
ENV feature switch for cutting over https://github.com/studentinsights/studentinsights/pull/2584.


kevinrobinson:
selfie

PR #2588, opened 2019-09-04
Homeroom paths: Match by id only, not slug

kevinrobinson:
# Who is this PR for?
K8 educators with homeroom access

This has bitten us twice before: https://github.com/studentinsights/studentinsights/pull/2022 and https://github.com/studentinsights/studentinsights/pull/2155.  Another case came up again verifying https://github.com/studentinsights/studentinsights/pull/2584.

In this case, there was a collision between `id` and `slug` spaces for homerooms, which caused the navbar link to be generated with a slug, but be interpreted as if it were an id in the controller lookup code.

# What does this PR do?
Removes `/homerooms/:slug` and only generates and supports `/homerooms/:id`.  There's still a footgun here because the `url_helpers.homeroom_path(homeroom)` will use slug, but that's for another day.  Verifying https://github.com/studentinsights/studentinsights/pull/2584 is more important, and this bug makes that more complicated. 

# Checklists
+ [x] Navbar links
+ [x] Homeroom

### DIFF
--- a/app/assets/javascripts/homeroom/HomeroomPage.js
+++ b/app/assets/javascripts/homeroom/HomeroomPage.js
@@ -21,8 +21,8 @@ export default class HomeroomPage extends React.Component {
   }
 
   fetchJson() {
-    const {homeroomIdOrSlug} = this.props;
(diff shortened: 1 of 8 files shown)


kevinrobinson:
selfie

PR #2589, opened 2019-09-05
Homeroom navigation: Update navigator to use id, not slug

kevinrobinson:
Following on https://github.com/studentinsights/studentinsights/pull/2588, this should have been part of that PR.

# Who is this PR for?
K8 educators with homeroom access

The dropdown on the homeroom page was still using slugs, after https://github.com/studentinsights/studentinsights/pull/2588 switched to id-only.

# What does this PR do?
Fixes bug in making homeroom URL so it uses `id`.

Also adds `Homeroom#grades` so that there is a replacement for the deprecated `#grade` method, and sends it down to the UI.  Updates the UI to show this as a suffix, which is helpful anyway and immediately useful for verifying.

Screen Shot 2019-09-05 at 10 42 13 AM

# Checklists
+ [x] Homeroom


kevinrobinson:
selfie

PR #2590, opened 2019-09-05
Homerooms: Default grades list to active students only

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2589.  This came up since the homeroom page shows conflicting info in the dropdown (computed from `Homeroom#grades`) and the description of the homeroom the user is viewing (computed from the list of students, which the server limits to only active students).


kevinrobinson:
selfie

PR #2591, opened 2019-09-05
Homerooms: Remove link if no active students

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2589, which used `Authorizer#homerooms` and filtered out homerooms with no active students.  This leads to an edge case where an educator is assigned a homeroom with no active students having a link that doesn't work for them in the navbar.  This removes the link in that case - it wouldn't have done anything anyway so this essentially hides stale data coming from the SIS.


kevinrobinson:
selfie

PR #2592, opened 2019-09-05
Homeroom: Remove students_count, update link to require active students

kevinrobinson:
Following-on from https://github.com/studentinsights/studentinsights/pull/2584.

The semantics of `students_count` are unlikely to be what a developer wants; usually these should be active students, so let's remove this.  Separately fixes a bug in an edge case in production data.

Also optimizes `Educator.active` to one SQL query rather than two joined in memory.


kevinrobinson:
selfie

PR #2593, opened 2019-09-05
Home feed: Update PerfTest case

kevinrobinson:
Tried out an optimization here for a particular HS case that is slower than others, but it didn't improve things.  Updates the `PerfTest` test case for `Feed`, which reveals the slowness is from their authorization path.


kevinrobinson:
selfie

PR #2594, opened 2019-09-05
Maintenance: Add indexes for Student.active and student voice feed query

kevinrobinson:
Looking at `Student.active.explain` and the query in `FeedForStudentVoice.student_voice_surveys_for_cards`.

PR #2595, opened 2019-09-05
Update some specs to be more precise and strict

kevinrobinson:
Revealed by https://github.com/studentinsights/studentinsights/pull/2594.


kevinrobinson:
selfie

PR #2598, opened 2019-09-09
My students: Add homeroom column and filter

kevinrobinson:
# Who is this PR for?
K8 educators with schoolwide access, came up for folks in Bedford

# What problem does this PR fix?
No obvious way to filter by homeroom on the "My students" page

# What does this PR do?
Adds a column and supports sorting, filtering, and searching.

Factors `FilterStudentsBar` out from `my_students` into `components`, since it was already being re-used by the counselor meeting page.  Updates tests for new cases, and for regressions on older cases.

# Screenshot (if adding a client-side feature)
image

# Checklists
*Which features or pages does this PR touch?*
+ [x] My students
+ [x] Counselor meetings

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Improved specs for existing code in need of better test coverage
+ [x] Manual testing made more sense here

PR #2599, opened 2019-09-09
Add sort by default for homeroom select

kevinrobinson:
Related https://github.com/studentinsights/studentinsights/pull/2598.


kevinrobinson:
selfie

PR #2600, opened 2019-09-10
Reading: Template migration scripts

kevinrobinson:
# Who is this PR for?
SHS reading teachers

The template for reading data had a few small mistakes (formatting for one column was percents like [0.0-1.0] instead of [0-100], one header was missing).

# What does this PR do?
Adds Google Apps scripts for running migrations across sheets or folders of sheets.  This runs through a district Google account as a Google Apps project, with permissions and other things done there rather than in Student Insights, or via its integration, which is still read-only.


kevinrobinson:
selfie

PR #2604, opened 2019-09-17
Notes audit: Allow limited to not restricted notes

kevinrobinson:
For internal team collaborating on train-the-trainer style workshops.

Also fixes unrelated eslint bug.


kevinrobinson:
selfie

PR #2608, opened 2019-09-19
New Bedford: Update school scoping for trainings

kevinrobinson:
From talking with project leads there, for updating workshops.


kevinrobinson:
selfie

PR #2609, opened 2019-09-19
New Bedford: Add migration tools for changing school scope

kevinrobinson:
# Who is this PR for?
internal team, helping with limiting scope for work related to permissions in New Bedford

Folks in New Bedford are focusing training and workshop efforts in particular schools, differently than the previous scope.  https://github.com/studentinsights/studentinsights/pull/2608 updated to match that scope, but this leaves some drift.

We haven't reduced scope like this before, and the `school_scope` concept in the importers is for supporting incremental deployments, debugging import jobs, or other kinds of partial imports where we only want to update a slice of the db state.  So there aren't tools now for migrating previously imported data.

# What does this PR do?
Adds `SchoolScopeMigrator` class, which has methods for understanding the `Student` and `Educator` records, and which would currently be in scope and which would currently not be, based on the school definition files.  This info can be useful for understanding what's happening, and there are specific methods here for migrating `Student` and `Educator` records in response to reducing the scope of the imports.  It uses the `missing_from_last_export` and `#active` interface on those models.

Separately, these could be purged as part of a more rigorous enforcement of retention policies, but that isn't done here.  This is the 'mark' but there is no 'sweeping' here.

Example output:
```
Finding schools...
There are <N> School records within scope.


Starting Educators...
  Found <N> active and <N> total Educator records.
  Found <N> Educator records within scope...
  Ensuring missing_from_last_export:true for <N> stale Educator records...
  Actually setting missing_from_last_export:true for <N> currently active stale Educator records...
  Done updating Educators.


Starting Students...
  Found <N> active and <N> total Student records.
  Found <N> Student records within scope...
  Ensuring missing_from_last_export:true for <N> stale Student records...
  Actually setting missing_from_last_export:true for <N> currently active stale Student records...
  Done updating Students.


All done.
```
# Checklists
+ [x] Core 

### DIFF
--- a/app/lib/school_scope_migrator.rb
+++ b/app/lib/school_scope_migrator.rb
@@ -1,6 +1,6 @@
-# For manual or semi-automated migrations related to changing school scope 
+# For manual or semi-automated migrations related to changing school scope
 # in definition files or in import jobs.
-# 
+#
 # For example, if removing a school, you might update the definition file,
 # then use this class to run a migration that marks all Student and Educator
 # records that used to be in scope as inactive.
--- a/app/lib/school_scope_migrator.rb
+++ b/app/lib/school_scope_migrator.rb
@@ -1,0 +1,98 @@
+# For manual or semi-automated migrations related to changing school scope 
+# in definition files or in import jobs.
+# 
+# For example, if removing a school, you might update the definition file,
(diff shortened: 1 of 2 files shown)


kevinrobinson:
selfie

PR #2610, opened 2019-09-19
Import: Include SchoolScopeMigrator as index task

kevinrobinson:
Build on https://github.com/studentinsights/studentinsights/pull/2609, and after reviewing this manually, let's automate this to reduce the potential for drift over time.  If there are issues in this process, we'd rather find them sooner than later.


kevinrobinson:
selfie

PR #2612, opened 2019-09-19
Sections: Refactor to allow sections in New Bedford middle schools

kevinrobinson:
# Who is this PR for?
New Bedford students, families, educators

New Bedford exports only homeroom assignment data from the SIS, but this doesn't work well for most middle school educators.  District folks are updating that separately.

When that is fixed, there are assumptions and configuration and product differences that don't make sense for their MS use case.  One example is that in Somerville where sections are used, there is always grade data as well.  Another issue is that there's no way to deploy in stages, where we import and validate, then enable for authorization.

# What does this PR do?
1. Authorization
Authorization code now respects `PerDistrict#enabled_sections?` which can be overridden by an env variable as well.

2. Endpoints
The section-related endpoints respect the same method.

3. Section links
This adds another layer, `PerDistrict#allow_sections_link?`, allowing incremental deployment where we enable the feature but don't make it visible to everyone yet.

4. Student profile UI
This updates the profile UI to use a different path for New Bedford, where the sections are shown on the profile but without grades like in SHS.  This involves changes to `LightProfilePage` and also pushing code out into `PerDistrict.js`.  **EDIT: last commit backs out this actual change for now, to test first**

Screen Shot 2019-09-19 at 1 20 37 PM

# Checklists
+ [x] Navbar
+ [x] Student Profile
+ [x] Section
+ [x] My Sections
+ [x] Core 


kevinrobinson:
selfie

PR #2613, opened 2019-09-19
New Bedford: Temporary sections importer

kevinrobinson:
Building on https://github.com/studentinsights/studentinsights/pull/2612.

# Who is this PR for?
New Bedford students, families and educators

In coordinating with district IT folks, we're waiting a bit for some upstream work.  This is trying to advance work like https://github.com/studentinsights/studentinsights/pull/2612 and enable doing some data quality checks on our end, and enable related product changes before the upstream work is done.  This PR is mostly risk mitigation against timing on the upstream work.

# What does this PR do?
Adds a temporary importer class, using an interim export file format from the district that they'll migrate over to the existing spec formats in the next week or two.  So this composes and patches existing importers.  If for some reason we need to go forward with the mitigation plan in the next few weeks, we'll need to refactor these to remove things like overriding private methods, but there's no forcing function on that yet.

The work here has been manually verified in different ways, and there's no intention for this to be automated.

# Checklists
+ [x]  Section
+ [x] My Sections
+ [x] My Students
+ [x] Core 

### DIFF
--- a/app/importers/file_importers/courses_sections_importer.rb
+++ b/app/importers/file_importers/courses_sections_importer.rb
@@ -23,20 +23,23 @@ class CoursesSectionsImporter
   end
 
   def import
-    return unless remote_file_name
-
-    log('Starting loop...')
     reset_counters!
 
-    streaming_csv = CsvDownloader.new(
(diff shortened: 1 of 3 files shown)


kevinrobinson:
selfie

PR #2615, opened 2019-09-19
Section: Improve header and navigation to other sections

kevinrobinson:
# Who is this PR for?
SHS educators, New Bedford educators

# What problem does this PR fix?
The section page has an older select box for navigating, and the format of the header doesn't match the way people talk about sections.  This was less important when only used by folks at SHS, but more important now that MS folks in New Bedford will use this as well.

# What does this PR do?
Updates the header, and replaces the select dropdown with one that allows typing to filter, and is more responsive for larger lists.

Related to https://github.com/studentinsights/studentinsights/pull/2612, but branching independently to be merged separately.

# Screenshot (if adding a client-side feature)
### revised header
Screen Shot 2019-09-19 at 5 29 53 PM

### navigating to another section
Screen Shot 2019-09-19 at 4 14 22 PM

### grades in Somerville
Screen Shot 2019-09-19 at 4 14 53 PM

### no grades in New Bedford
Screen Shot 2019-09-19 at 4 14 49 PM

# Checklists
*Which features or pages does this PR touch?*
+ [x] Section

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Improved specs for existing code in need of better test coverage
+ [x] Manual testing made more sense here

PR #2617, opened 2019-09-20
Sections: Limit to sections in district_school_year

kevinrobinson:
More visible since https://github.com/studentinsights/studentinsights/pull/2615, this list should really only be for the current year (older ones have no enrolled active students anyway, so are just noise here).


kevinrobinson:
selfie

PR #2619, opened 2019-09-24
Reading: Fix bug in heuristic about moving in processor

kevinrobinson:
This conditional had a bug, so it was running even when not intended.  Also remove some duplicate stats logging.


kevinrobinson:
selfie

PR #2620, opened 2019-09-24
Reading: Update migration script with attempt at lasid warning

kevinrobinson:
Just checking in updates from attempting a migration that didn't work as simply as expected.


kevinrobinson:
selfie

PR #2621, opened 2019-09-24
Reading: Add validations for different assessments, not yet applied

kevinrobinson:
Adds a validation function for reading data points, not yet applied to the model.  This is part of data quality checks on fall assessment data.

It also adds a transformation function in the processor, specifically to handle how importing data produced by formulas works.


kevinrobinson:
selfie

PR #2623, opened 2019-09-25
Reading: Update GoogleSheetsFetcher to recur into folders and batch calls

kevinrobinson:
# Who is this PR for?
Somerville K5 students, reading specialists, literacy coaches

Previous PRs like #2619 #2620 #2621 worked out the kinks running the importer task over individual schools, but running it over the whole district ran into a few issues related to fetching data.  First, `#get_tabs_from_folder` assumed that all files within a folder were sheets, but turns out this isn't always the case.  Second, it seems that the `in parent` query in Google Drive searches only within the immediate "children" of a folder and not recursively which I think I was assuming.  Third, when updating to search recursively, we hit API request quotas.

# What does this PR do?
1. Updates the methods making API calls to explicitly limit results to files that are sheets or folders.
2. Adds a `recursive: true` option to `#get_tabs_from_folder`.  
3. Uses a batch method to get values from all sheets in a spreadsheet in one request.
4. Handle the special case of the API returning `nil` when calling `#values` on a newly added sheet.
5. Add optional logging that was useful in debugging.

### DIFF
--- a/app/importers/reading/google_sheets_fetcher.rb
+++ b/app/importers/reading/google_sheets_fetcher.rb
@@ -28,17 +28,36 @@ require 'csv'
 
 class GoogleSheetsFetcher
   # returns [Tab]
-  def get_tabs_from_folder(folder_id)
-    sheet_ids = get_sheet_ids(folder_id)
+  # optionally recur into folders with recursive: true
+  # Note that this may hit API quotas, and this class
+  # doesn't batch http requests.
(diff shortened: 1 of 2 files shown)


kevinrobinson:
@edavidsonsawyer This fetcher class is getting some good mileage! :)  I'm going to merge this now, but it'd be awesome if you could review and share any comments or feedback whenever you have a chance.

PR #2622, opened 2019-09-24
Reading: Minimal page for debugging import process by homeroom

kevinrobinson:
# Who is this PR for?
internal team

# What problem does this PR fix?
Stats for auditing import process by homeroom.

# What does this PR do?
Adds a new internal page for reviewing imported reading data, grouped by homeroom for a particular assessment period.  This is for internal debugging, not for district admin to use or accountability.

# Screenshot (if adding a client-side feature)
Screen Shot 2019-09-24 at 4 11 59 PM

# Checklists
*Which features or pages does this PR touch?*
+ [x] Reading by Homeroom page

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Manual testing made more sense here

PR #2624, opened 2019-09-25
Fix link bug in reading homerooms debug tool

kevinrobinson:
Bug fix from https://github.com/studentinsights/studentinsights/pull/2622.


kevinrobinson:
selfie

PR #2625, opened 2019-09-25
Reading: Clarify project lead links and page titles

kevinrobinson:
Clarifying link names and page titles.


kevinrobinson:
selfie

PR #2626, opened 2019-09-25
Reading: Set importer to fetch recursively, updating logging

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2623, running importer in standalone fashion, no more `dry_run: true`.


kevinrobinson:
selfie

PR #2628, opened 2019-09-25
Reading: Enforce validations, except for F&P

kevinrobinson:
Hold off on F&P, as there's a bunch of entry and interpretation complexity there.  For others, this has been audited and looks good!


kevinrobinson:
selfie

PR #2629, opened 2019-09-25
Reading: Update spec for recursive fetching

kevinrobinson:
Fix for not updating spec in https://github.com/studentinsights/studentinsights/pull/2626.


kevinrobinson:
selfie

PR #2630, opened 2019-09-25
Reading: Update specs to respect validations

kevinrobinson:
From more strictness in https://github.com/studentinsights/studentinsights/pull/2628.


kevinrobinson:
selfie

PR #2634, opened 2019-09-25
Cleanup: Add notes about validations across models

kevinrobinson:
Capturing some scripts and notes looking at validations.  Separately migrated some older `Section` records that were missing `term_local_id`.


kevinrobinson:
selfie

PR #2640, opened 2019-09-27
Reading: Update importer job  for daily automation

kevinrobinson:
# Who is this PR for?
K5 reading specialists, literacy coaches, teachers

# What problem does this PR fix?
The import of reading data from template sheets is still semi-manual.

# What does this PR do?
Adds the importer to the normal set of importer jobs, and sets it up to run for Somerville.  Also wraps the importer in a transaction on the table, to cautiously rollback the job in the case of any errors.

# Checklists
*Which features or pages does this PR touch?*
+ [x] Student Profile
+ [x] Import task

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Manual testing made more sense here

PR #2641, opened 2019-09-27
Reading: Add safeguards to importer for incremental use

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2640.  This removes logging of tab names, which may contain educators names.  It also adds a warning about calling the importer with an explicit `folder_id`, and doesn't call `syncer#delete_unmarked_records` if an explicit folder is passed.

This prevents against the mistake I just made :) of doing a test run on a subfolder, which then destroyed other records because the scope of the sync is for the whole time period.

It remains to be seen whether recording the URL or doing some other more complex method of setting the sync scope is worthwhile, so for now just adding these guardrails seems best.


kevinrobinson:
selfie

PR #2642, opened 2019-09-27
Reading: Fix bug with categories in debug page, update meaning of grey bars, tighting F&P classification

kevinrobinson:
# Who is this PR for?
internal team

# What does this PR do?
Fixes bug with intensive/strategic categories in DIBELS, fixes sort in popup tables, and updates how "missing" is counted for DIBELS and F&P.  For "missing" it's the count of students where we can't make an interpretation, either because:

1. there aren't thresholds set
2. we don't know how to interpret the value entered
3. there is no assessment data

The relative value of these different kinds of info will probably change over time.

# Checklists
+ [x] Reading debug


kevinrobinson:
selfie

PR #2636, opened 2019-09-25
Sections: Update controller guards, incremental deploy with importer

kevinrobinson:
# Who is this PR for?
New Bedford educators

# What does this PR do?
Updates section controller guards to remove special case for Somerville, so it can work in incremental deployment for New Bedford.  Also update temporary importer to allow injecting `file_text` directly for manual test runes.  And mark some `Educator` methods as deprecated that came up in looking at this.

# Checklists
*Which features or pages does this PR touch?*
+ [x] Section

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [ ] Included specs for changes
+ [ ] Manual testing made more sense here

PR #2648, opened 2019-09-30
Sections: Update temporary importer for incremental deploys

kevinrobinson:
# Who is this PR for?
New Bedford educators

# What does this PR do?
Split out from https://github.com/studentinsights/studentinsights/pull/2636, this makes updates to support incremental imports and manual testing.


kevinrobinson:
And an unrelated bump for RubyZip.  The zip bomb attack vector isn't particularly relevant for us, since these are coming from trusted channels with other layers of defense and not end users.

kevinrobinson:
selfie

PR #2650, opened 2019-10-01
Home: Add index to optimize imported form queries

kevinrobinson:
For the home page.


kevinrobinson:
selfie

kevinrobinson:
cuts query cost to 20% of what it was

PR #2651, opened 2019-10-01
Home feed: Optimize query for event notes

kevinrobinson:
For the home page, from `Feed#event_note_cards`


kevinrobinson:
selfie

kevinrobinson:
using indexes now in explain

PR #2652, opened 2019-10-01
Home page: Optimize query for authorized students

kevinrobinson:
# Who is this PR for?
all educators

# What does this PR do?
This optimizes `authorized { Student.active }` for educators with section assignments, impacting the home page feed, my students page, and many other parts of the product.

It does this by removing calls to `Educator#labels` that end up doing authorization checks to see if the educator has any sections, in the process of computing dynamic labels.  Since the only labels needs in the authorization method itself are static, optimize the call there to only look at static labels.

I also initially reordered the condition checks related to housemaster authorization; these would run even if the feature was disabled through an env flag.  This actually led to consistently slower p95 values on the `PerfTest`, so I noted it and backed that out.

Data from Somerville, for sample of 5% of educators:
```
> PerfTest.authorized(0.05);nil
before: all_for_educator          median: 911ms   p95: 7290ms
after:  all_for_educator          median: 423ms   p95: 993ms


> PerfTest.feed(0.05);nil
# before:    p50: 739    p95: 3939
# after:     p50: 713    p95: 1486
```

New Bedford will soon be impacted by this change, but for now the speedups only impact Somerville folks.

Separately, after this work and measurement, I noticed the `PerfTest` class defaults to all educators, and so updated that to use `Educator.active` going forward.

# Checklists
+ [x] Home page
+ [x] My Students
+ [x] Core 

### DIFF
--- a/app/lib/authorizer.rb
+++ b/app/lib/authorizer.rb
@@ -104,7 +104,10 @@ class Authorizer
     begin
       should_consider_sections = options.fetch(:should_consider_sections, PerDistrict.new.enabled_sections?)
       return :districtwide if @educator.districtwide_access?
-      return :housemaster if @educator.labels.include?('high_school_house_master') && student.grade == '8' && EnvironmentVariable.is_true('HOUSEMASTERS_AUTHORIZED_FOR_GRADE_8')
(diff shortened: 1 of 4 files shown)


kevinrobinson:
selfie

PR #2654, opened 2019-10-01
Authorization: Centralize section authorization in Authorizer

kevinrobinson:
# Who is this PR for?
K8 educators in New Bedford

# What does this PR do?
As we're prepping to release section-based authorization for New Bedford educators, this centralizes all authorization related to Sections into the `Authorizer` class.

```
# Example usage:
authorized { Section.all } # authorized
authorized { educator.sections } # assigned
```
This revises and adds more test cases to verify this all works as expected.  Also related to older issue writeups (eg, https://github.com/studentinsights/studentinsights/issues/1080#issuecomment-350572396).

# Checklists
*Which features or pages does this PR touch?*
+ [x] Home page
+ [x] Section
+ [x] My Sections
+ [x] My Students
+ [x] Core 

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Improved specs for existing code in need of better test coverage
+ [x] Manual testing made more sense here

PR #2656, opened 2019-10-02
Section authorization: Optimize method to make it faster

kevinrobinson:
# Who is this PR for?
Educators with section assignments

https://github.com/studentinsights/studentinsights/pull/2654 centralizes section authorization, and makes more use of these methods that previously.  This slowed down the section page for queries like `authorized { Section.all }`.

# What does this PR do?
Adds two optimizations, adapted from `#why_authorized_for_student?`, comparing foreign key values rather than model values to remove that query, and adding a `#to_a` call to the end of the association, which causes Rails to cache the result of that query across calls.  I'm not sure why this influences Rails' caching, but it does and `PerfTests` here show the same thing as with `#why_authorized_for_student?`.

Also adds minimal smoke tests to `PerfTest` methods, just verifying that they still run and don't break.  It does not make assertions about performance.

```
puts PerfTest.section_authorization_pattern(0.01);nil

# before
1%,  median: 1917ms    p95: 2490ms

# after
 1%, median:  417ms    p95:  613ms
 5%, median:    2ms    p95:  511ms
10%, median:    2ms    p95:  518ms
(caveat: distribution is bimodal since some educators don't have sections)
```

# Checklists
+ [x] Section
+ [x] Authorization

### DIFF
--- a/app/lib/authorizer.rb
+++ b/app/lib/authorizer.rb
@@ -153,10 +153,14 @@ class Authorizer
     return false unless PerDistrict.new.enabled_sections?
     return true if @educator.districtwide_access?
 
-    return false if @educator.school.present? && @educator.school != section.course.school
+    return false if @educator.school_id.present? && @educator.school_id != section.course.school_id
 
     return true if @educator.schoolwide_access? || @educator.admin?
-    return true if section.in?(@educator.sections)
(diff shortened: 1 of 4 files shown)


kevinrobinson:
selfie

Issue #1925, opened 2018-07-21
Import Tardy records for CSV rows with both absence and tardy bits set


kevinrobinson:
Fan out in cases where both flags are set (ie, Aspen represents what's happening on that day, not a list of events).

In other words, `matching_insights_record_for_row ` (https://github.com/studentinsights/studentinsights/blob/master/app/importers/file_importers/attendance_importer.rb#L146) can fan out to two records, not just one.  `attendance_event_class` needs to be updated as well.

kevinrobinson:
Done in https://github.com/studentinsights/studentinsights/pull/2658!

PR #2658, opened 2019-10-02
Allow import of tardy and absence events in single record

edavidsonsawyer:
This addresses #1925 by creating both an absence and a tardy if both exist in the record.

### DIFF
--- a/app/importers/file_importers/attendance_importer.rb
+++ b/app/importers/file_importers/attendance_importer.rb
@@ -141,8 +141,8 @@ class AttendanceImporter
     end
 
     # Skip if not absence or tardy
-    record_class = attendance_event_class(row)
-    if record_class.nil?
+    record_classes = valid_attendance_event_classes(row)
+    if record_classes.empty?
       @skipped_other_rows_count += 1
       return
     end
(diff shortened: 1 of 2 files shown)


app/importers/file_importers/attendance_importer.rb
```diff
@@ -189,7 +199,7 @@ def attendance_event_class(row)
     is_absence = per_district.is_attendance_import_value_truthy?(row[:absence])
     is_tardy = per_district.is_attendance_import_value_truthy?(row[:tardy])
 
-    if is_absence && is_tardy then nil
+    if is_absence && is_tardy then "both"
```
kevinrobinson: This seems like a good point to make the change, but using a string here means that `attendance_event_class` sometimes returns a class and sometimes returns a special string value.  How about renaming the method to reflect that change and use a symbol to indicate it's a special value?  Or alternately, you could have it return a list of classes (eg,`[Absence, Tardy]`).

edavidsonsawyer: Makes sense. How about we rename `attendance_event_class` and it just returns an array of valid classes for the event. That way we're always dealing with the same kind of thing. 

One question I've got. It's fair to assume the detailed fields should be the same if the row contains both an absence and a tardy?

kevinrobinson: Sure, that sounds good 👍 

re: the detailed fields, the shape of the export file is the same (eg, these tests don't look like they use a fixture CSV, but the test setup reflects this too), and the shape of the Insights models are the same (eg, check out `schema.rb`).  Whether the meaning of these fields beyond their types are the same (eg, what do the letter codes within these fields mean), or whether data entry and quality is the same (eg, does "excused" mean the same thing across places within the district) is more complicated :)

But I think for this PR we can ship it and it'll make tardy counts more accurate, and then we can look at the deeper level within Insights records if those kinds of questions come up down the line.

kevinrobinson:
🚢 !

PR #2660, opened 2019-10-02
Courses: Add uniqueness index

kevinrobinson:
This uniqueness is enforced in a validation that is slow and triggers multiple queries across models; add a uniqueness constraint to to this in the DB, maybe we'll remove the validation later.


kevinrobinson:
selfie

PR #2662, opened 2019-10-02
Maintenance: Remove get_process_mem

kevinrobinson:
We used this previously to look at memory profiling, and based on https://github.com/studentinsights/studentinsights/pull/1427#issuecomment-361080049 maybe it wasn't even that helpful.

This started showing warnings when running specs with the latest Ruby 2.6.x upgrade, so this just removes it.


kevinrobinson:
selfie

PR #2665, opened 2019-10-03
Maintenance: Tune Rollbar throttling to be stricter

kevinrobinson:
Throttling this to stay well under the next billing threshold.  Talk to @kevinrobinson for more about the underlying issue, which is unrelated to this codebase.


kevinrobinson:
selfie

PR #2666, opened 2019-10-04
Assessment importer: Update spec avoid relying on record order

kevinrobinson:
Trying to address the flaky spec failure in https://travis-ci.org/studentinsights/studentinsights/jobs/593200540#L2961.  It failed for the first time on an unrelated code change, and I can't reproduce it on a partial run locally even with the same seed value.


kevinrobinson:
selfie; still can't reproduce after 5 Travis runs so going to ship this since it's an improvement and move on

PR #2667, opened 2019-10-04
Attendance importer: Track student not matched separately, add logging and specs

kevinrobinson:
# Who is this PR for?
developers

There had been some false-positive warnings from importer jobs, and for the attendance job I looked into this more in verifying that https://github.com/studentinsights/studentinsights/pull/2658 had the intended effect.  This is coming from differences across districts in the scope of different exports.

For example, in Bedford the export XML in StudentInsightstudentdef.txt filters by only active students, but the same filter isn't applied to the attendance export.  This means we receive a large number of attendance events for past students.  This was tracked and `RecordSyncer` properly alerted on this being more than we might expect.  But since this is known, this PR updates the attendance importer to track this condition separately.  This PR also adds an integration test for this importer.

We should fix this upstream, which would also reduce the data being exported from districts, but that's work for another time.

# Checklists
+ [x] Attendance importer

### DIFF
--- a/app/importers/file_importers/attendance_importer.rb
+++ b/app/importers/file_importers/attendance_importer.rb
@@ -28,9 +28,9 @@ class AttendanceImporter
     @inclusive_date_range = create_inclusive_date_range(options)
 
     @student_ids_map = ::StudentIdsMap.new
-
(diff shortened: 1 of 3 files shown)


kevinrobinson:
selfie

PR #2673, opened 2019-10-17
School roster: Add minimal download CSV button

kevinrobinson:
# Who is this PR for?
New Bedford MS folks

We added this back in 2016 in a particular format as the first way to export to PowerBI, and then removed it in December 2016.  We noticed in January 2018 that it wasn't here anymore, but didn't take action since no one was missing it :)  I think we didn't immediately re-add it since the old way we used in Healey originally wasn't fast enough for a larger school like Normandin.  This past June (#2507) I optimized how this works for a districtwide list, but that was only for project leads so includes things like student address and race that we'd want to scrub from an export on the overview page.

# What does this PR do?
This adds CSV download, deferred to user action.  We'll deploy and see if it works well enough with large schools, or whether we need to swap to the server side.  This doesn't download the exact same data as is shown in the UI; this PR is just aimed at seeing if this works well enough with larger schools.

Also separately updates the server-side export (only exposed to project leads) to use a safelist rather than grabbing all fields on the model.

Screen Shot 2019-10-17 at 4 34 03 PM
Screen Shot 2019-10-17 at 4 34 09 PM

# Checklists
+ [x] School Overview

### DIFF
--- a/app/assets/javascripts/components/DownloadIcon.js
+++ b/app/assets/javascripts/components/DownloadIcon.js
@@ -1,5 +1,10 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
-export default function DownloadIcon() {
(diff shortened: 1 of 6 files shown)


kevinrobinson:
selfie

PR #2675, opened 2019-10-22
Maintenance: Upgrade to node 12

kevinrobinson:
It's now in long-term support.  The current build setup pins the dev and production node version in `package.json`, but floats to the latest LTS in Travis.  This means that when a new LTS is released, the build fails like this: https://travis-ci.org/studentinsights/studentinsights/jobs/600978116  This seems fine, and a healthy kind of prompt, so keeping it rather than fixing the version in Travis as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "studentinsights",
   "version": "0.1.0",
   "engines": {
-    "node": "^10.13.0"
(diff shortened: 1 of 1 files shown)


kevinrobinson:
selfie

kevinrobinson:
selfie

PR #2676, opened 2019-10-22
Maintenance: Remove transitional flags to complete upgrade to Rails 5.2

kevinrobinson:
This removes flags that were added as part of the upgrade process to Rails 5.2, in preparation for upgrading to 6.0 next.

It's possible that `per_form_csrf_tokens` interacts with some JS in a way that wasn't revealed in integration tests, so I'll test this manually on the demo site first.


kevinrobinson:
FWIW, I don't think this per-form CSRF flag actually provides additional protection as far as I can tell.  The UJS adapter also seems to blow away any per-form tokens and set them all to the global token (eg, https://github.com/rails/rails/blob/b2eb1d1c55a59fee1e6c4cba7030d8ceb524267c/actionview/app/assets/javascripts/rails-ujs/utils/csrf.coffee#L25).  The global token seems like it is always also valid based on their test cases (eg, `test_accepts_global_csrf_token` in https://github.com/rails/rails/blob/99d12b950756c7d2a4afff538bcf04b40eda82bf/actionpack/test/controller/request_forgery_protection_test.rb#L910).  Other issues confirm this (see github/rails/rails/32930, closed without a fix).

kevinrobinson:
selfie

PR #2680, opened 2019-10-25
Service info: Update to set per-district as well as with label

kevinrobinson:
Rolling out https://github.com/studentinsights/studentinsights/pull/2542 more widely.


kevinrobinson:
selfie

PR #2681, opened 2019-10-29
STAR: Update after breaking changes to vendor export

kevinrobinson:
# Who is this PR for?
Somerville educators

The vendor changed export formats (eg, https://github.com/studentinsights/studentinsights/pull/2526).

# What does this PR do?
This makes the minimal update to the new format, and factors out the importer code for reading and math to a single file since it's now the same.  This only really makes sense since the current integration is minimal and not expected to change.

Separately, remove `instructional_reading_level` from `StudentAssessment` and `StarReadingResult` since there's no educational or product use for these right now.

Updates the tests to be integration tests, since the refactoring here didn't work with the lower-level testing strategy.  Also move star importer bits into their own folder.

# Checklists
+ [x] STAR importers


kevinrobinson:
selfie

PR #2683, opened 2019-10-30
Reflection: Patterns within notes

kevinrobinson:
# Who is this PR for?
SHS educators in a workshop, to start

# What does this PR do?
Adds a reflection tool for looking at patterns in how students are written about within Student Insights.  This is still early, but the idea is to center how individual students are talked about (rather than how individual educators write), and to focus on students that are written about the most within Student Insights.  The intention is to help focus ask questions about reflecting whole-child and strengths-based perspectives.  Notes are scoped to student-level authorization, and to the current school year only (anything restricted is excluded, even if the educator would have access in other contexts).

This page is behind a label, and not exposed in the UI right now.  It'll be prototyped in a workshop first before any wider training or release.

# Screenshot (if adding a client-side feature)
(word cloud omitted)

Screen Shot 2019-10-30 at 11 13 27 AM

# Checklists
*Which features or pages does this PR touch?*
+ [x] Reflection on notes

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Manual testing made more sense here

PR #2686, opened 2019-11-01
Reflection: Update photo to remove cropped as default; add className to StudentPhoto

kevinrobinson:
Fix some spec bugs in https://github.com/studentinsights/studentinsights/pull/2684 and https://github.com/studentinsights/studentinsights/pull/2683.


kevinrobinson:
selfie

PR #2682, opened 2019-10-29
STAR: Support backwards compatibility too, fix bug with school filter 

kevinrobinson:
# Who is this PR for?
Builds on https://github.com/studentinsights/studentinsights/pull/2681.

# What problem does this PR fix?
Adds back support for backwards compatibility since New Bedford's format didn't update in the way I had thought.  Also fixes bug https://github.com/studentinsights/studentinsights/pull/2681 introduced with the school filter and adds a test case for it.

# Checklists
*Which features or pages does this PR touch?*
+ [x] STAR importers

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Improved specs for existing code in need of better test coverage

PR #2688, opened 2019-11-01
STAR: Update to support different time formats

kevinrobinson:
One more difference from https://github.com/studentinsights/studentinsights/pull/2682, the time format strings are different and I missed that earlier.


kevinrobinson:
selfie

PR #2691, opened 2019-11-04
Security: Update Rack:Attack to alert separately for targeted URLs 

kevinrobinson:
# Who is this PR for?
developers

Splitting alerting precedence for different kinds of blocked requests.  Also move blocklist/noise rule above datacenter rule to avoid higher-priority alerting on those noise URLs.

# Checklists
+ [x] Rack attack config


kevinrobinson:
selfie

PR #2693, opened 2019-11-05
Maintenance: Update Rollbar Ruby gem and tighten configuration

kevinrobinson:
More explicitly describe some privacy-related configuration.  Make sure default Rails person tracking is disabled.  Within Rollbar, only store anonymized identifiers and partially anonymized IP addresses, for the purpose of investigating issues.  Retention policy there is 30 days.


kevinrobinson:
selfie

PR #2694, opened 2019-11-05
Maintenance: Update secure_headers gem

kevinrobinson:
Related to https://github.com/studentinsights/studentinsights/pull/2693, which in the past had conflicts with different versions of this gem.  So updating `secure_headers` first and then will do Rollbar.  No issues expected.


kevinrobinson:
selfie

PR #2695, opened 2019-11-05
Maintenance: Upgrade Rollbar.js version

kevinrobinson:
Related to https://github.com/studentinsights/studentinsights/pull/2693, which does this on the Ruby side.  The JS side has a bit more risk as the distribution package has changed (now combined across Node and the browser), and recent versions include several changes in config related to Telemetry and Person tracking which we want to disable here.

Reviewed changes for all versions, and each config option.  To deploy, verifying on demo first.


kevinrobinson:
selfie

PR #2696, opened 2019-11-05
Rollbar: More defensive whitelist for field, remove config from being sent

kevinrobinson:
Building on https://github.com/studentinsights/studentinsights/pull/2695, manual review looks good, but adding more defensiveness to scrubbing in case, and removing sending Rollbar config each time.


kevinrobinson:
selfie

PR #2697, opened 2019-11-05
Security: Updated Rack::Attack alerting for specific routes only

kevinrobinson:
Alert separately only if this matches a valid route, which is the primary concern.  Requests for `/garbage` are lower priority.


kevinrobinson:
selfie

PR #2698, opened 2019-11-05
Monitoring: Remove config from Rollbar events

kevinrobinson:
One more for https://github.com/studentinsights/studentinsights/pull/2696, not sure why `sendConfig:false` isn't doing what I expect, but moving on.


kevinrobinson:
selfie

kevinrobinson:
no impact; have to look at the source but not really worth it, moving on for now.

PR #2699, opened 2019-11-05
Import: Close SFTP session more aggressively

kevinrobinson:
This came up looking at some server logs, and it's been updated on the server side already.  This aims to limit the API of `SftpClient` to a single method for opening a connection, downloading a file, and closing the connection.  In the current implementation, no callers are ever explicitly closing the connection.


kevinrobinson:
selfie

PR #2700, opened 2019-11-06
Maintenance: Update gitignore to be more defensive about files

kevinrobinson:
# Who is this PR for?
developers

Adding more defenses against developers accidentally checking data they're working with locally into source control.

# What does this PR do?
Updates gitignore to aggressively ignore files with extensions like `*.csv` or `*.json`, especially if they are not in the expected test folders.

Also clean up some other old files that aren't used.  And rename csv fixture files that were `.txt` to be `.csv`, and XML files to be `.xml`.


kevinrobinson:
selfie

PR #2701, opened 2019-11-06
Maintenance: Create DistrictConfigLog to enable more self-serve district maintenance

kevinrobinson:
For now, this is moving some district configuration that district IT folks provide out of config code and into the database, where it could be stored, viewed and edited in a more self-serve way.

Concretely, that's done by moving where `PerDistrict` reads from into `DistrictConfigLog`.  It typically expresses differences directly in code, reads from ENV variables, or reads config from files like `district_somerville.yml`.

The idea is that `PerDistrict` reads secrets and info about services from ENV that's controlled by developers, but that other information (eg, URLs for sheets or drive folders being synced, safelists for particular features, etc) should be read from `DistrictConfigLog` where it's a step closer to being exposed to project leads and district IT staff.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,6 +67,7 @@ _site/
 !/app.json
 !/jest.json
 !/package.json
+!/config/sftp_filenames_development_fixture.json
 !/spec/importers/helpers/data_flows_for_other_importers.json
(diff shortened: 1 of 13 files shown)


kevinrobinson:
selfie

PR #2702, opened 2019-11-06
Auditing: auditjs added to Travis build

kevinrobinson:
This is a different index than npm/yarn.  Let's see if it catches anything useful, or if it's just noise.


kevinrobinson:
selfie

PR #2703, opened 2019-11-06
Auditing: Add rails_best_practices checks to Travis build, do cleanup

kevinrobinson:
This adds a subset of https://github.com/flyerhzm/rails_best_practices, does cleanup on the items it found, and adds the check to the Travis build.

These include:
- removing unused methods
- calling ActiveRecord operations and ignoring the return value (change to bang methods)
- removed some old rake tasks related to schools (synced differently now)

Some other config was just noise, so those are disabled here in `rails_best_practices.yml`.

EDIT: I also looked at `UseBeforeFilterCheck` which seems potential useful but is too noisy.

This also updates administrate to be more explicit, and limit open routes as helpfully pointed out by `RestrictAutoGeneratedRoutesCheck`.

```
# removed
admin/educators#create
admin/educators#new
admin/educators#destroy

# kept, explicitly now
admin/educators#index
admin/educators#show
admin/educators#edit
admin/educators#update
```


kevinrobinson:
example of how failures look in the build:
Screen Shot 2019-11-06 at 3 15 33 PM

kevinrobinson:
selfie

PR #2704, opened 2019-11-06
Security: Remove bcc from LoginChecker

kevinrobinson:
No need for bcc in email, it's tracked other places.


kevinrobinson:
selfie

PR #2705, opened 2019-11-07
STAR: Abort importers if filenames aren't set in config

kevinrobinson:
The check for `remote_file_name` being set is too strict for districts that don't use this; this refactors it to work as expected and adds tests for that.


kevinrobinson:
selfie

PR #2707, opened 2019-11-07
Fix assertion to remove spec flakiness

kevinrobinson:
This came up from the failure in https://travis-ci.org/studentinsights/studentinsights/jobs/608894197, which is a strange one since a) I can't reproduce locally on a few partial test runs with the same test seed and b) this code has been like this for a long time and never failed this way.  It's possible that with enough full runs and the same seed this would come up again, but not sure and that would take too long to test.

Regardless, the failure isn't meaningful anyway - these assertions should be checking set membership not the order of a list so this PR changes that.


kevinrobinson:
selfie

PR #2708, opened 2019-11-07
Services: Move filtering of provider suggestions to client, narrow scope of list

kevinrobinson:
# Who is this PR for?
educators

The service UI has an old autocomplete, and this worked by querying the server on each keystroke, and filtering a list on the server.  That meant unnecessary network chatter, sending user input in the query string (logged in some cases, exposed in some monitoring).  There were bugs in the filtering code on the server, and it didn't scope the query to the student's school like it intended (which we wouldn't want anyway, given the complexity of roles).

# What does this PR do?
Removes the server filtering, but adds scoping to providers for authorized students.  Request the list up front when the UI component is mounted and removes other UI config.

# Checklists
+ [x] Student Profile

### DIFF
--- a/app/assets/javascripts/student_profile/PageContainer.test.js
+++ b/app/assets/javascripts/student_profile/PageContainer.test.js
@@ -119,6 +119,13 @@ const helpers = {
 };
 
 describe('integration tests', () => {
+  beforeEach(() => {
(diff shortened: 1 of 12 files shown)


kevinrobinson:
selfie

PR #2709, opened 2019-11-08
Services: Limit autocomplete suggestions

kevinrobinson:
Small UX fix to https://github.com/studentinsights/studentinsights/pull/2708, avoiding suggesting a longer list on click.


kevinrobinson:
selfie

PR #2711, opened 2019-11-08
Rollbar: Allow district_key param from Ruby

kevinrobinson:
Behavior change related to https://github.com/studentinsights/studentinsights/pull/2693; this used to be allowed but scrubbing is more comprehensive, so `district_key` is getting scrubbed when we do still want it.


kevinrobinson:
selfie

PR #2712, opened 2019-11-13
STAR: Update config to read from export SFTP box instead of vendor

kevinrobinson:
# Who is this PR for?
New Bedford MS educators

IT pals in New Bedford are having issues with the STAR setup, either upstream or in working with other vendors.  They wanted to work around, so replicated the STAR export themselves.

# What does this PR do?
Parameterizes the `StarImporter` classes with methods for getting different clients, and reading config from different places.  The New Bedford setup uses this to connect to the SFTP export box rather than STAR directly, and reads the config (which is more sensitive) from `DistrictConfigLog` instead of yaml in source.

# Checklists
+ [x] STAR importers

### DIFF
--- a/app/config_objects/per_district.rb
+++ b/app/config_objects/per_district.rb
@@ -52,6 +52,32 @@ class PerDistrict
     yaml.fetch('star_filenames', {}).fetch(key, fallback)
   end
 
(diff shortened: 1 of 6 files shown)


kevinrobinson:
selfie

PR #2713, opened 2019-11-20
MCAS: Update score range, add more tests

kevinrobinson:
Also remove duplication in plotlines, and call into `mcasScores`.


kevinrobinson:
selfie

PR #2714, opened 2019-11-21
MCAS: Fix bucketing when on upper boundary, and lock graph scale to range definitions

kevinrobinson:
Fix on https://github.com/studentinsights/studentinsights/pull/2713


kevinrobinson:
selfie

PR #2715, opened 2019-11-21
MCAS: Fix method to call for score range, rename to clarify

kevinrobinson:
These function names were confusing, so this fixes the bug and the naming, on top of https://github.com/studentinsights/studentinsights/pull/2714.


kevinrobinson:
selfie

PR #2716, opened 2019-11-21
MCAS: Fix Highcharts params to min/max are respected

kevinrobinson:
The struggle from https://github.com/studentinsights/studentinsights/pull/2715 et al. is real 🙁  

Highcharts doesn't respect `min` or `max` because other params related to ticks influence it.  Try to be explicit everywhere to get it to do what we say.

With dev data (out of real distribution, fixed in this PR as well):
Screen Shot 2019-11-21 at 1 27 34 PM


kevinrobinson:
selfie

PR #2717, opened 2019-12-02
Object.values polyfill for IE11

kevinrobinson:
Resolves issue with internal /reading/debug page on IE11.


kevinrobinson:
selfie

PR #2718, opened 2019-12-02
Support language dropdown for Levels page in demo

kevinrobinson:
The parsing for language-related fields is the same for Somerville and Demo, except for this one method.  This fixes it, which raises on the levels page in demo.


kevinrobinson:
selfie

PR #2719, opened 2019-12-02
Notes: Fix to resizing text area when longer than screen height

kevinrobinson:
# Who is this PR for?
educators, driven from one style of MTSS meeting

When editing a note that is taller than the screen height, the 'resizing' bit of `ResizingTextArea` can lead to the page layout jumping, which can end up scrolling to the top of the textarea so that the cursor and focus isn't visible on the screen anymore.

# What does this PR do?
Updates the resizing code to use a different method.  Previously it set the element height to `auto`, waited for it to size, and then set the height to that value.  Now it uses a separate hidden element to compute the sizing, and then sizes the `textarea` to match.

This also adds notes on test cases, and leaves some styles for visual debugging.  Actual tests cases would be tough since jest/jsdom don't do actual text layout.

# Checklists
+ [x] Student Profile

### DIFF
--- a/app/assets/javascripts/components/ResizingTextArea.js
+++ b/app/assets/javascripts/components/ResizingTextArea.js
@@ -42,12 +42,14 @@ export default class ResizingTextArea extends React.Component {
     // purposes of sizing.  This limits the impact to when there is a trailing
     // newline, and the last character is about to wrap.
     this.span.textContent = this.el.textContent + '#';
-    // console.log('>> resize', this.el.style.height, '.', this.el.offsetHeight, this.el.scrollHeight, '|', this.span.style.height, '.', this.span.offsetHeight, this.span.scrollHeight);
+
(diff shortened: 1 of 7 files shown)


kevinrobinson:
:\ this works, but I can't reliably reproduce the original issue on master anymore :(

kevinrobinson:
Final edits; the compromise here is to essentially force a newline at the end of each `<textarea/>` so that the behavior is consistent across browsers.  I couldn't find another way to resolve a) the span having different height with a trailing newline and b) the textarea from scrolling on its own before the `change` event is fired, causing a visible flicker in IE11 in the time it takes to set scrollTop back to 0 in the handler.  The flicker would come up in the normal use case of hitting enter at the end of the text, so the workaround removing it seems better.

kevinrobinson:
selfie

PR #2720, opened 2019-12-03
Notes: Update CSS for hover and ResizingTextArea, add more comments

kevinrobinson:
Style regression on https://github.com/studentinsights/studentinsights/pull/2719; the hover and active CSS styles broke when changing the CSS class names and structure with the hidden element.  This fixes that and also adds in some comments on manual testing.


kevinrobinson:
selfie

PR #2721, opened 2019-12-04
Login: Enable email verification on login, for internal use only

kevinrobinson:
# Who is this PR for?
educators

More defenses for particular login security threats (see internal docs).

# What does this PR do?
Adds another field to `EducatorMultifactorConfig` for sending login codes via email, and add support for that in `MultifactorAuthenticator`, with a little refactoring to add an indirection layer for Mailgun calls.  Test cases to match all paths.

Deploying would have no impact, but would allow internal dogfooding.

# Checklists
+ [x] Login


kevinrobinson:
selfie

PR #2723, opened 2019-12-05
Import records: Fix bug blocking districtwide users not set as admin in SIS

kevinrobinson:
# Who is this PR for?
one project lead

Some project leads, or other folks with districtwide access, were incorrectly being blocked from accessing some the "SIS import history" tool.  This could happen if they didn't have the `admin` field also set upstream in the SIS at the time of the first account import.

# What does this PR do?
Updates to gate this to project leads only.  UI code already does that.

# Checklists
+ [x] District page


kevinrobinson:
selfie

PR #2724, opened 2019-12-05
Monitoring: Review and safelist some fields for alerting

kevinrobinson:
Relax a few more fields being sent to Rollbar, from scrubbing all in https://github.com/studentinsights/studentinsights/pull/2693.  These all use key names that are hopefully narrow as well as triggers for developers about sensitivity, so that we can keep scrubbing-by-default.


kevinrobinson:
selfie

PR #2725, opened 2019-12-09
Google API: Retry by default

kevinrobinson:
We've seen some transient network errors; retrying is disabled by default so this enables it.


kevinrobinson:
selfie

PR #2726, opened 2019-12-09
Testing: Update CoverageChecker to notice when file names change

kevinrobinson:
I noticed that `CoverageChecker` wasn't failing, even though some of the file names it referenced were renamed a little while back (eg, in the Rails 6 change https://github.com/studentinsights/studentinsights/pull/2679 with updating the locations of the Devise plugin files).

This updates it so that it fails if it can't find all files specified, adds two other files, and does some cleanup to remove dead spec code that I noticed poking around after a full coverage run.

EDIT: backed out one of the validation cleanups, needs more investigation.


kevinrobinson:
selfie

Issue #2685, opened 2019-11-01
Fix bug with handling unparseable dates in PDF export


kevinrobinson:
The error is: `TypeError: Unable to get property 'format' of undefined or null reference`, coming from the `ProfilePdfDialog.js` file.

This looks like most code in that component expects that `this.state.filterFromDate` will always be a valid Moment object, but the event handlers like `onFilterFromDateChanged` and others can set it to null if the text value isn't parseable as a valid date.

edavidsonsawyer:
@kevinrobinson it looks like this would be easy enough to fix by falling back to the existing date if the `filterFromDate` or `filterToDate` were null. However playing with the component it seems like the real problem might be that trying to change the date (eg. hitting delete) doesn't actually allow you to input a manual date and instead changes the date in unexpected ways, eventually leading to this error. The options I see are

1. Ensure keyboard import doesn't change the datepicker range, and rely on the modal calendars to set the dates,

2. Fix the input and make the date ranges editable while preventing invalid input

Option 2 is consistent with the service recorder so that seems like the way to go unless there's a reason I'm missing to stop users from manually setting pdf dates. Does that sound right to you?

kevinrobinson:
@edavidsonsawyer Cool, awesome investigation! 👍 

I agree that it's a UX issue primarily if folks can't just type a date in the text box.  I also agree that option 2 sounds like the right direction - users need to be able to:
- see default date sets when the open the dialog
- be able to use the datepicker to pick different dates
- be able to type dates in the textbox
- either get feedback on invalid dates or be prevented from clicking "Generate" when invalid (rare occurrence)

Given that, the code or internals for that component could change however they need to to meet those goals.

edavidsonsawyer:
@kevinrobinson great! I've got a branch that solves those problems so I'll add some tests and poke around with it to make sure it's working as expected and submit a PR

PR #2727, opened 2019-12-12
Fix student pdf dates

edavidsonsawyer:
# Who is this PR for?

Educators

cf #2685 in the student profile pdf report attempting to type in the date picker doesn't change the date and can produce an error.

# What does this PR do?

Allows input in the date picker boxes, disables the report and shows an error message when dates are invalid. 

pdf_generation

# Checklists
+ [x] Student Profile

### DIFF
--- a/app/assets/javascripts/student_profile/ProfilePdfDialog.js
+++ b/app/assets/javascripts/student_profile/ProfilePdfDialog.js
@@ -31,20 +31,20 @@ export default class ProfilePdfDialog extends React.Component {
 
   filterFromDateForQuery() {
     const {filterFromDateText} = this.state;
-    return this.formatDateTextForRails(filterFromDateText);
+    return this.formatDateText(filterFromDateText);
   }
 
   filterToDateForQuery() {
     const {filterToDateText} = this.state;
-    return this.formatDateTextForRails(filterToDateText);
+    return this.formatDateText(filterToDateText);
   }
 
-  // Normalize input date text into format Rails expects, tolerating empty string as null.
+  // Normalize input date text into format Rails expects
   // If the date is not valid, an error will be raised.
-  formatDateTextForRails(dateText) {
+  formatDateText(dateText) {
     const moment = toMoment(dateText);
     if (!moment.isValid()) throw new Error('invalid date: ' + dateText);
-    return moment.format('YYYY-MM-DD');
+    return moment.format('MM/DD/YYYY');
   }
 
   // Both dates need to be valid
--- a/app/assets/javascripts/student_profile/ProfilePdfDialog.js
+++ b/app/assets/javascripts/student_profile/ProfilePdfDialog.js
@@ -7,6 +7,7 @@ import {
   toSchoolYear,
   firstDayOfSchool
 } from '../helpers/schoolYear';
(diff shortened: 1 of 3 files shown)


edavidsonsawyer:
@kevinrobinson this CI failure looks like it's related to an audit of the npm packages, but I don't think any of these are related to this PR specifically. Have you seen this before?

app/assets/javascripts/student_profile/ProfilePdfDialog.js
```diff
@@ -134,18 +145,22 @@ export default class ProfilePdfDialog extends React.Component {
...
+              value={this.state.filterToDateText}
               onChange={this.onFilterToDateChanged}
               datepickerOptions={{
                 showOn: 'both',
                 dateFormat: 'mm/dd/yy',
                 minDate: undefined
               }} />
           </div>
+          <div style={{height: '2em'}}>
+            {!this.areDatesValid() && <div className="PdfDialogue-warning" style={styles.invalidDate}>Choose a valid date</div>}
```
kevinrobinson: It looks like you defined a `areDatesValid` variable up above, so you could use that here too.

Also, I don't see where `styles.invalidDate`is defined, should that be added with something like orange or red text, or maybe removed?

app/assets/javascripts/student_profile/ProfilePdfDialog.js
```diff
@@ -134,18 +145,22 @@ export default class ProfilePdfDialog extends React.Component {
...
+          </div>
         </div>
         <div style={{display: 'flex', justifyContent: 'flex-end'}}>
           <button
-            style={styles.studentReportButton}
+            style={{background: (areDatesValid) ? undefined : '#ccc'}}
```
kevinrobinson: This removes the `studentReportButton` styles, and the snapshot change shows that too, can you share more about why?

app/assets/javascripts/student_profile/ProfilePdfDialog.js
```diff
@@ -17,8 +17,8 @@ export default class ProfilePdfDialog extends React.Component {
     const nowMoment = context.nowFn();
     this.state = {
       includeRestrictedNotes: false,
-      filterFromDate: firstDayOfSchool(toSchoolYear(nowMoment)-1),
-      filterToDate: nowMoment
+      filterFromDateText: firstDayOfSchool(toSchoolYear(nowMoment)-1).format('MM/DD/YYYY'),
+      filterToDateText: nowMoment.format('MM/DD/YYYY')
```
kevinrobinson: Looking at this now, the end date should probably have been "now+1 day" also.  Otherwise it would cut off notes etc. from the current day.   If that seems right to you, want to add it in while we're in here?

app/assets/javascripts/student_profile/ProfilePdfDialog.js
edavidsonsawyer: This was just an oversight. I took it out to think about how to merge the styles and never added it back.

edavidsonsawyer:
@kevinrobinson Sorry for the delay on this, but all the above should be addressed.

kevinrobinson:
@edavidsonsawyer 👍 no worries! 🚢 🚢

PR #2733, opened 2020-01-03
Verify lockfile matches package.json

kevinrobinson:
Add https://github.com/lirantal/lockfile-lint as a low-cost defensive check for something we'd never catch in a PR manually.


kevinrobinson:
Bundled into https://github.com/studentinsights/studentinsights/pull/2735

PR #2734, opened 2020-01-03
Update JS selective resolutions

kevinrobinson:
Cleaning up some selective resolutions.


kevinrobinson:
Closed for simpler https://github.com/studentinsights/studentinsights/pull/2735

PR #2735, opened 2020-01-03
Update JS selective resolutions (second version)

kevinrobinson:
Less churn-y alternative to https://github.com/studentinsights/studentinsights/pull/2734/files


kevinrobinson:
selfie

PR #2736, opened 2020-01-06
Levels: Also export reversed name in download

kevinrobinson:
# Who is this PR for?
SHS admin

Folks want to sort by last name, but there's no simple way to recover this with formulas if names are exported as shown.

# What does this PR do?
Adds another column to the export only.

Screen Shot 2020-01-06 at 10 07 04 AM

# Checklists
+ [x] Levels


kevinrobinson:
selfie

PR #2737, opened 2020-01-08
Export: Swap in CSV package for escaping

kevinrobinson:
Fix for https://github.com/studentinsights/studentinsights/pull/2736; Google Sheets interprets TSVs properly by default but the version of Excel some folks have installed does not.

This swaps in `csv-stringify` to do this with proper escaping, which is better anyway.


kevinrobinson:
selfie

PR #2739, opened 2020-01-09
Revert 2738 revert 2737 patch/csv escaping

kevinrobinson:
Fixes https://github.com/studentinsights/studentinsights/pull/2737, which was reverted because the lib included non-ES5 features that didn't get compiled away.  This updates to pull the ES5-compat lib and avoid that, tested by local production build.


kevinrobinson:
selfie, verified in production demo site

PR #2740, opened 2020-01-09
Maintenance: Update apiFetchJson to reject on non-20x status codes

kevinrobinson:
# Who is this PR for?
developers

We sometimes get false-positive monitoring alerts about JS errors when folks click into stale pages after authorization has expired.  This can happen when a user wakes their computer after say an hour of inactivity, and they click before the probe has a chance to notice they've been signed out.

In that case, if they take some action on the page that makes an API call, the server will return say a 403.  But `apiFetchJson` passes through the JSON ignoring the status code.

# What does this PR do?
In the simplest common case, we want a naive `apiFetchJson` to reject if the response is not 200, so this adds that in.  Callers can always fetch in different ways for other use cases, and they still get access to the response JSON either way (which typically indicates more error info).

# Checklists
+ [x] Core 
+ [x] eg, Home page

### DIFF
--- a/app/assets/javascripts/helpers/apiFetchJson.js
+++ b/app/assets/javascripts/helpers/apiFetchJson.js
@@ -12,8 +12,25 @@ export function apiFetch(url, options = {}) {
 }
 
 
+// Return JSON either way, but fail the Promise on
+// an unexpected status code.  This is optimized for
+// the happy path.
+//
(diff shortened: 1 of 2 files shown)


edavidsonsawyer:
This is from a while back, but it's a good feature and we should have this in!

PR #2741, opened 2020-01-13
Reading: Update importer to better tolerate unmatched student ids

kevinrobinson:
# Who is this PR for?
K5 reading teachers, specialists, literacy coaches, folks in MTSS

There was a bug in a path of the reading importer where it was handling unexpected student id values, and this hadn't come up before since that path hadn't been exercised in production yet.

# What does this PR do?
This fixes the bug so it can tick the failure and continue on, and adds a test case for it.  This doesn't add an automated loop to flag these or circle back, which we may want to add later as we automate and smooth more of the loops and flows here.

It also removes the use of `ImporterMatcher` in this class, since the class uses some other logic for matching students, and so the matcher class it wasn't being used directly anymore except for ticking valid rows.

# Checklists
+ [x] Reading importer

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,8 +72,8 @@ _site/
 !/spec/importers/file_importers/data_flows_fixture.json
 
 *.csv
-!spec/fixtures/*.csv
-!spec/importers/*.csv
+!spec/fixtures/**/*.csv
+!spec/importers/**/*.csv
 
 *.sql
 !/x2_export/sql/*.sql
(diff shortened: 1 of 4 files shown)


kevinrobinson:
Also update `.gitignore` to allow CSV files in any folder within `spec/fixtures` or `spec/importers` without an explicit entry in gitignore. See https://github.com/studentinsights/studentinsights/pull/2700 for why that setup exists.

kevinrobinson:
selfie

PR #2742, opened 2020-01-13
STAR: Update config for importers to allow disabling importers

kevinrobinson:
# Who is this PR for?
developers

When there are upstream data quality issues, if we want to pause the STAR importers to avoid false-positive monitoring alerts, we have to remove the config values.  This isn't ideal since then we need to remember or record them somewhere.

# What does this PR do?
Adds `PerDistrict#is_star_import_enabled` so that config can remain in place, and we can just flip this switch if we need to globally disable while resolving issues upstream that might take a few days.

# Checklists
+ [x] STAR importers

### DIFF
--- a/app/config_objects/per_district.rb
+++ b/app/config_objects/per_district.rb
@@ -48,6 +48,22 @@ class PerDistrict
     yaml.fetch('school_definitions_for_import')
   end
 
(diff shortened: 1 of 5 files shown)


kevinrobinson:
selfie

PR #2744, opened 2020-01-13
Monitoring: Update Rollbar safelist to allow sync alerts to send more info

kevinrobinson:
# Who is this PR for?
developers

The safelist for Rollbar data wasn't allowing details about which importer, and which metric led to an alert being sent in `RecordSyncer`.  This made it harder to triage these alerts immediately so see if they were meaningful and needed action right away, or whether they were false-positive and the alerting needed to tuned.

# What does this PR do?
Adds keys to the Rollbar safelist (eg, `rollbar_safelist_alert_count`).

# Checklists
+ [x] Importing and syncing
+ [x] Monitoring


kevinrobinson:
selfie

PR #2745, opened 2020-01-14
Educators importer: Track missing email or login_name separately and skip

kevinrobinson:
# Who is this PR for?
developers

Some district SIS scripts include educator records without `login_name` or `email`.  This can happen for enough records that it leads to `RecordSyncer` alerting on it with its simple check for percentage of records that are missing or invalid or changed.  Those alerts are false positives.

# What does this PR do?
Track missing `email` or `login_name` separately instead of passing it to `RecordSyncer`.  This logs the counts, and avoids including those records in `RecordSyncer`'s stats or alerts.

# Checklists
+ [x] Educators importer


kevinrobinson:
selfie

PR #2746, opened 2020-01-14
Educators importer: Skip trying to match homerooms when none set 

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2745, prevent attempting to match a homeroom record if the value is empty.  This essentially just cleans up the metrics printed in logs, since matching on those empty homerooms would always fail as expected.


kevinrobinson:
selfie

PR #2747, opened 2020-01-22
Import: Check freshness when downloading file

kevinrobinson:
# Who is this PR for?
developers, district IT staff

If there are issues on the district side that prevents export jobs from running, we don't detect automatically when files are stale.

# What does this PR do?
Checks the modified timestamp when processing a file fetched through `SftpClient`.

# Checklists
+ [x] Importers


kevinrobinson:
selfie

PR #2748, opened 2020-01-22
Reading: January reader profile prototype

kevinrobinson:
# Who is this PR for?
MTSS leaders, reading specialists, instructional coaches

# What problem does this PR fix?
This is a prototype of the iteration called the "January reader profile."  More design info available in the "12/13 reading sketches" internal doc.

# What does this PR do?
This adds a new UI section on the reading tab.  It re-purposes existing endpoints, so other than the new `enable_reader_profile_january` label, this is a JS-only change.

For local dev and testing, this uses a grid of different test cases, with students of different grades at different time periods.  The data generation isn't super realistic yet beyond the shape; that'll be something to improve iteratively after this frame is out and validated.

This also doesn't yet do any real visualization of the data, since a key aspect of the design was the hypothesis of how de-emphasizing this might shape problem solving conversations.  That'll be another pass.

# Screenshot (if adding a client-side feature)
### story: example
Screen Shot 2020-01-22 at 4 06 03 PM

### story: click on an element
(images aren't loaded in the story)
Screen Shot 2020-01-22 at 4 06 19 PM

### story: examples across grades
Screen Shot 2020-01-22 at 4 05 49 PM

### story: full grid of (grades, time)
Screen Shot 2020-01-22 at 4 05 58 PM

### product: placement within page
Screen Shot 2020-01-22 at 4 21 22 PM

# Checklists
*Which features or pages does this PR touch?*
+ [x] Student Profile

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Included specs for changes
+ [x] Manual testing made more sense here

PR #2750, opened 2020-01-23
Reading: Fix letter naming tab, show more metadata when no information

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2748.

Fixes a bug with the wrong assessment key for letter naming fluency, and adds test case covering a scenario that would have revealed that bug.  Also makes some updates for how no information is handled.

# Checklists
+ [x] Student Profile


kevinrobinson:
selfie

PR #2760, opened 2020-02-07
Maintenance: Update eslintrc to restore no-console

kevinrobinson:
Removing `no-console` was a breaking change in `eslint:recommended` for eslint 6, and so introduced in https://github.com/studentinsights/studentinsights/pull/2759.  This adds it to our config, along with `no-alert` for good measure.  It separately removes `react/prefer-es6-class: off` since we migrated that previously and no longer need to relax it.

EDIT: Also mark current uses for confirm and alert (for some edge case UX situations not worth improving at the moment).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -23,13 +23,14 @@
     "indent": ["error", 2],
     "semi": ["error", "always"],
     "comma-dangle": ["error", "only-multiline"],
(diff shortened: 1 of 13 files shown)


kevinrobinson:
selfie

PR #2761, opened 2020-02-11
Reading: Add cohort chart and raw scores

kevinrobinson:
# Who is this PR for?
APs at MTSS, reading specialists, literacy coaches

When problem solving for a specific student (eg, in MTSS), it's sometimes challenging to make meaning out of benchmark screeners, especially for folks like APs who are less intimately familiar with benchmarks for each screener at each grade level.

Setting benchmarks supports this kind of triage, but after that folks still seek another level of understanding.  Raw scores provide this, as do chart visualizations that allow making inferences about growth.  But these leave a lot of space for different interpretations and predictions about future growth that aren't necessarily reasonable to make from just these data points, and those interpretations may not directly support problem solving about instructional changes or interventions.

As another way to approach this, charts comparing the student's score to other students in their grade cohort at school can be a more intuitive and actionable way for folks to triage.  It's trying to support folks in the more abstract "are things okay" or "should we prioritize attention or staffing" in a way that's scoped to what they have power over in the context of a problem-solving meeting.

# What does this PR do?
Adds cohort charts, and raw scores also in the process.  Abstracts the `BoxChart` to do this.

It de-colors the charts for percentiles and raw scores, to avoid implying that some kind of coloring for benchmarks is also being applied to the percentiles or elsewhere.  We'll have to test interpretation of all these, but the assumption is that this cohort chart is a secondary action here, and one that will be supported in PD training or coaching beforehand.

It also ports the JS F&P interpretation code to Ruby for use in computing percentiles, and factors the related JS code out to mirror the Ruby code, with comments on both sides.

Screen Shot 2020-02-12 at 11 47 50 AM

# Checklists
+ [x] Student Profile


Gemfile
```diff
@@ -77,7 +77,6 @@ group :development, :test do
   gem 'bourbon', '~> 4.3.2'
   gem 'capybara'
   gem 'database_cleaner'
-  gem 'descriptive-statistics'
```
kevinrobinson: separate, but unused since probably https://github.com/studentinsights/studentinsights/commit/d13e03af8b820ca95e5626afdc5f092d7b4d561c

kevinrobinson:
selfie

PR #2762, opened 2020-02-12
Fix Gemfile.lock for descriptive-statistics

kevinrobinson:
Fix for https://github.com/studentinsights/studentinsights/pull/2761; it updated the Gemfile but never ran `bundle install` so the deploy failed.


kevinrobinson:
selfie

PR #2763, opened 2020-02-12
EXPERIMENT: Update Travis to bundle install --deployment, verify mismatches are caught

kevinrobinson:
**This isn't intended to be merged**

In https://github.com/studentinsights/studentinsights/pull/2762 the build passed but deploy failed because the Gemfile didn't match the Gemfile.lock.  We should catch this in CI not at deploy time, so this makes an attempt at that, and mistakenly removes a gem to see if now the Travis build will fail.

**This isn't intended to be merged**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ before_install:
   - if [[ $TRAVIS_TEST_SUITE == "ruby" ]]; then gem install bundler; fi
 
 install:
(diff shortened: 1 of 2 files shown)


kevinrobinson:
This is what we're going for, from https://travis-ci.org/studentinsights/studentinsights/jobs/649572552
Screen Shot 2020-02-12 at 12 50 57 PM

PR #2764, opened 2020-02-12
Maintenance: Fail build when mismatch between Gemfile and Gemfile.lock

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2763/files, changing Travis script to fail build on Gemfile / Gemfile.lock mismatches, since these will (correctly) fail the deploy and we should catch this sooner.


kevinrobinson:
selfie

PR #2766, opened 2020-02-12
Maintenance: Fix call to bundle audit in CI

kevinrobinson:
Guessing this is related to https://github.com/studentinsights/studentinsights/pull/2764; seems like this line should have always been scoped by `bundle exec` (eg, https://github.com/rubysec/bundler-audit/issues/191#issuecomment-396675002)


kevinrobinson:
selfie

PR #2767, opened 2020-02-12
Reading: Fix images of Phoneme Segmentation Fluency materials

kevinrobinson:
This didn't get migrated when the carousel changed to support multiple images.


kevinrobinson:
selfie

PR #2768, opened 2020-02-26
Student photos: Update freshness check to yearly

kevinrobinson:
This is a good guardrail, but districts typically only update this yearly so express that here.


kevinrobinson:
selfie

PR #2769, opened 2020-02-26
STAR: Update freshness check by district

kevinrobinson:
Doing some clean up on this monitoring, related to https://github.com/studentinsights/studentinsights/pull/2768.  The mechanics of these syncs are different across districts, and so as a result so are some of the update cycles.


kevinrobinson:
selfie

PR #2771, opened 2020-02-26
Add comment about font setup

kevinrobinson:
Ran across this, which would have been a helpful explanation previously, so added a comment referring to it in the font CSS file.


kevinrobinson:
selfie

PR #2772, opened 2020-02-27
Maintenance: Simplify SFTP client params after bug

kevinrobinson:
In https://github.com/studentinsights/studentinsights/pull/2768 and https://github.com/studentinsights/studentinsights/pull/2769, there were bugs since the params for this call are a atypical, and end-to-end tests mocked that configuration.  We could improve that, but this simplifies the root problem of having to pass a nil `override_env` param before other config, especially since `override_env` is never used anywhere :)


kevinrobinson:
selfie

PR #2773, opened 2020-02-27
Patch/bedford special education edge cases

kevinrobinson:
# Who is this PR for?
MTSS facilitators, reading specialists, literacy coaches, K5 classroom teachers

# What does this PR do?
This is a first cut at the F&P materials.  It adds in some materials for levels A&B only, and adds a different way to choose materials.  For other screeners, these are standardized based on the student's grade and time of year, but the F&P protocol is based on moving up from their previous level, so students in the same cohort might be reading different materials.  This means these materials reflect what the student's instructional level is (while others show "what was assessed").

No work here to optimize images yet, there's a bunch more so punting that until we want to do this, then we can compress or shrink them all down in one shot.

Screen Shot 2020-02-27 at 4 12 09 PM

# Checklists
+ [x] Student Profile

### DIFF
--- a/app/assets/javascripts/helpers/specialEducation.test.js
+++ b/app/assets/javascripts/helpers/specialEducation.test.js
@@ -181,3 +181,30 @@ describe('#prettyIepTextForSpecialEducationStudent', () => {
     })).toEqual('IEP');
   });
 });
+
+
+describe('Bedford test cases', () => {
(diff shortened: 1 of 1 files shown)


kevinrobinson:
oops, nice write-up, but wrong branch :)

PR #2775, opened 2020-02-27
Reading: Update F&P thresholds

kevinrobinson:
This is on top of https://github.com/studentinsights/studentinsights/pull/2774, to avoid churning the snapshot tests (since that PR has some styling changes with wider impact, and this one does with adjusting thresholds).

# Who is this PR for?
MTSS facilitators, reading specialists, literacy coaches, K5 classroom teachers

This came up with literacy coaches, and so there are some slight revisions to the benchmarks here, but the main intention was to pick these for each of the screening time periods, and extend them out to fifth grade as well.  That way if there's screening, there's a way to support highlighting these as well.

# What does this PR do?
Updates the thresholds, and then some styling for `<ReadingScheduleGrid />`, which is used for visualizing the thresholds, and the full raw set of data points over time.  Both are intended more for validating data, but the substantive change is that the highlighting in the reader profile will change to match these thresholds.  We'll validate this with folks looking at the overall distributions for different cohorts, particularly after this next interval.

Screen Shot 2020-02-27 at 4 56 04 PM

# Checklists
+ [x] Student Profile
+ [x] Reading thresholds
+ [x] Reading benchmarks by cohort


kevinrobinson:
selfie

PR #2776, opened 2020-02-28
Special education: Add tests cases for current Bedford edge cases

kevinrobinson:
# Who is this PR for?
Bedford educator

There's some variation in manual processes for updating data about special education placement, previous identifications of disability, etc. in some upstream systems in Bedford.  When Student Insights finds potentially conflicting information like this, it surfaces that in the UI so that folks notice and can bridge any communication gaps.

# What does this PR do?
We're not yet sure if we want to keep this, make the "show conflicting information if its ambiguous how to interpret it" strategy more explicit and visible to users within Student Insights, or just address the underlying issues upstream.  For now, this just adds two edge cases as test cases, verifying that these are working the way we have come to understand things in some other communication.  If we decide on changes, we'll update these to reflect that.

# Checklists
+ [x] Student Profile
+ [x] School Overview


kevinrobinson:
selfie

PR #2782, opened 2020-03-17
Maintenance: Update rbnacl to resolve ffi deprecation warning

kevinrobinson:
Follow-on to https://github.com/studentinsights/studentinsights/pull/2781, which was an automated security patch.  It seems the automated patch updated unrelated gems, and while I reviewed the changelog for administrate, I didn't didn't notice the diff updated a bunch of unrelated packages.

One of the changes was to update ffi, which led to this deprecation warning when running tests or deploying:

```[DEPRECATION] Struct layout is already defined for class RbNaCl::HMAC::State. Redefinition as in /Users/me/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/rbnacl-7.1.0/lib/rbnacl/hmac/sha512.rb:105:in `<class:State>' will be disallowed in ffi-2.0.
```

So this bumps `rbnacl` to resolve that ([changelog](https://github.com/RubyCrypto/rbnacl/blob/master/CHANGES.md)).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -45,7 +45,7 @@ gem 'twilio-ruby'
 ### standalone libraries
 # security
 gem 'ipcat'
-gem 'rbnacl'
+gem 'rbnacl', '~> 7.1.1'
 gem 'rotp'
 gem 'rqrcode'
 gem 'zxcvbn-js', require: 'zxcvbn'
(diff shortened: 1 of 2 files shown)


kevinrobinson:
selfie

PR #2783, opened 2020-03-17
Maintenance: Ruby gem updates

kevinrobinson:
After https://github.com/studentinsights/studentinsights/pull/2781, I another `bundle update` and reviewed the changelog for each gem.


kevinrobinson:
selfie

PR #2784, opened 2020-03-19
Maintenance: Remove sass and bourbon

kevinrobinson:
https://github.com/studentinsights/studentinsights/pull/2783 ended up reaching a threshold for slug size in Heroku (see https://devcenter.heroku.com/articles/slug-compiler#slug-size).

To start I tried cleaning the remote repo (per https://thoughtbot.com/blog/how-to-reduce-a-large-heroku-compiled-slug-size) but that didn't change the size.  It did purge the build cache though, which revealed a deprecation notice about `sass-ruby`.  That led me to looking at the Gemfile for `sass-rails` and `bourbon`.  Neither are needed, since `sprockets` pulls in `sassc` are we haven't been using bourbon for a while either.  So this removes those gems, which should also reduce the slug size.

There are plenty of other ways to reduce the slug size too, but cleaning up these stale gems is an improvement anyway.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,6 @@ gem 'bootsnap', require: false
 gem 'jquery-rails'
 gem 'jquery-ui-rails', '~> 6.0.1'
 gem 'rails', '6.0.1'
-gem 'sass-rails', '~> 5.0'
 gem 'sprockets'
 gem 'thor'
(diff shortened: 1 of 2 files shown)


kevinrobinson:
selfie

PR #2785, opened 2020-03-19
Maintenance: Update AWS SDK to v3, limit to only S3 client 

kevinrobinson:
On top of https://github.com/studentinsights/studentinsights/pull/2784.

This is from a deprecation warning via email, and from looking at https://github.com/aws/aws-sdk-ruby/blob/master/V3_UPGRADING_GUIDE.md  No impact expected, other than that we can now pull in only the S3 client and not the rest of the AWS SDK.  So incidentally this should also reduce the slug size.

EDIT: turns out that there's a change in how credentials are read across versions, and the S3 client checks that config is present in the constructor.  Some tests use the real client, and then mock calls to it, but those now fail since the constructor raises that there is no config.

In development, this SDK behavior is a bit dangerous and not what we want.  If if there was local config or credentials in `~/.aws` the SDK would read that from disk, which is why these tests didn't fail when run locally.  It's dangerous because it's possible that test bugs could lead to exercising stale credentials a developer left on disk from doing unrelated work on the AWS CLI (even for a different project).  We never want this to be possible, so this adds `AWS_SDK_CONFIG_OPT_OUT`.

It also adds fake values to the ENV for keys and secrets in dev and test, so that the constructor doesn't raise and we can mock and test.  It looks like the SDK may now have more first-class ways to stub it that are baked in (eg https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/ClientStubs.html) but this PR didn't look more deeply into that.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,7 +28,7 @@ gem 'rack-attack'
 gem 'secure_headers'
 
 # services
-gem 'aws-sdk', '~> 2'
+gem 'aws-sdk-s3', '~> 1'
 gem 'dalli' # memcached for rack::attack
 gem 'google-api-client', "~> 0.28.7"
 gem 'net-ldap'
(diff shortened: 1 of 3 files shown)


kevinrobinson:
selfie

PR #2788, opened 2020-03-19
Maintenance: Update Rails, Sprockets

kevinrobinson:
The Rails update is from an XSS advisory that doesn't impact us.  But `bundle update rails` also pulled in Sprockets 4 (the `manifest.js` change.), which seems good while we are in maintenance-all-the-things mode.

EDIT: Also added monkey patch to work around gap bug between i18n and zeitwork in Rails 6.0.2.2, see https://travis-ci.org/github/studentinsights/studentinsights/jobs/664580684#L2793 for failure, and code comments for details.


kevinrobinson:
selfie

PR #2790, opened 2020-03-24
Maintenance: Guard check for sessionStorage even more

kevinrobinson:
This can raise in some edge cases where storage is disabled; in those cases on Safari even attempting to read `window.sessionStorage` raises.  This adds a more conservative guard around that so the caching is disabled instead of raising all the way up.


kevinrobinson:
selfie

PR #2792, opened 2020-03-25
Reading: More handling of nil or uninterpretable values for F&P ordering and percentiles

kevinrobinson:
# Who is this PR for?
K5 folks looking at reading profiles

In manual review, it came up that for some students the cohort percentiles for F&P would raise.  This happened on the server from trying to sort values that were nil, which could happen when unable to interpret the F&P level.

# What does this PR do?
This improves nil handling in a few places, and pushing that down all the way into `FAndPInterpret`.  `ComparableDataPoint` still implements a default ordering for null data points, since it has to tolerate underlying data points that may be nullable.  Separately, the real fix here is to filter all those uninterpretable data points out in the higher levels of the reader profile computation (if not, they'd be included in percentile computations).  Tests all around.

# Checklists
+ [x] Reader profile


kevinrobinson:
hrm, blocked on Travis issues at the moment

kevinrobinson:
selfie

PR #2794, opened 2020-03-27
Add GitHub Actions for CI

kevinrobinson:
We're having issues with Travis and still haven't been able to resolve them with their support team, so let's see what it takes to run tests with GitHub actions.  This aims to just run the JS tests.

EDIT: Iterated on this to get everything working.  Overall, the upside here is to remove Travis as a provider, which is a new benefit since after last week, when CI went down for a few days and we support couldn't help with even agreeing on whether it was up or not.  Based on the assumption that GitHub will have better stability (and investment and growth) going forward from here, this seems like an improvement.  We can remove the related permission grants to Travis as well, which were fairly permissive.

This also speeds up build time in both JS and in Ruby, and in particular so far the "time to spin up the build worker" has been close to zero.  With Travis, there have been several times where the time to spin up the worker has been significant, so this migration leaves CI faster overall.  I separately looked very briefly at parallelizing rspec but seems this still hasn't landed as first class yet (there are several ways to work around, but I didn't spend any time on it as this is already a large change and we already get improved test times).  Here's what this PR sees:
Screen Shot 2020-04-01 at 3 20 21 PM
Screen Shot 2020-04-01 at 3 23 37 PM

That's 3:45 for JS (compared to 4:24 in Travis) and 10:30 for Ruby (compared to 13:34 in Travis).  👍 😄 🕐 

I think the only thing we lose is the nice coloring in the build log, and there is some increased latency to see the output of a build step (eg, to see logs in rspec tests while they are running, for debugging drift between local tests/CI).  These seem minor compared to the speedups, but this is what is lost:
Screen Shot 2020-04-01 at 3 23 47 PM

Finally, there is some additional deployment risk since this changes what looked like a spacing typo in the `database.yml` config for production.  I fixed that spacing and deleted all the commented out bits that are irrelevant to make that kind of problem more visible.  This code bug has been present since https://github.com/studentinsights/studentinsights/pull/2540 but I also verified that it hasn't mattered and `sslmode` is set to require in all production environments because of the `PGSSLMODE` env variable.

So this is ready to 🚢 and then we can remove Travis and related permissions.


kevinrobinson:
success with the JS, and it's faster than Travis to start.  going to trying incrementally adding Ruby, which is more complicated because of Postgres, and then add caching and hopefully 🚢 

Screen Shot 2020-03-27 at 11 50 01 AM
Screen Shot 2020-03-27 at 11 
<img width=
Screen Shot 2020-03-27 at 11 41 15 AM

49 44 AM" src="https://user-images.githubusercontent.com/1056957/77775690-47dc2f00-7023-11ea-8b71-c8e47e7de602.png">
Screen Shot 2020-03-27 at 11 49 35 AM

kevinrobinson:
parallel jobs works as expected, and first set of Ruby steps look great 👍

kevinrobinson:
For posterity, what took forever here is that Rails wasn't reading `DATABASE_URL` at all, and printing out the ActiveRecord config finally revealed that but I still couldn't figure out why it wasn't reading it.  What ended up fixing it was setting `RAILS_ENV=test`, which I don't fully understand, since it's not like it was setting the config for dev or prod either.  🤷‍♂

kevinrobinson:
selfie

PR #2798, opened 2020-04-03
Maintenance: Remove Travis, add notes about GitHub CI

kevinrobinson:
Finishing off https://github.com/studentinsights/studentinsights/pull/2794.  I separately removed permission in GitHub related to Travis.


kevinrobinson:
selfie

PR #2799, opened 2020-04-03
Maintenance: Dedupe two yarn install runs in CI

kevinrobinson:
This runs twice (with caching) to do the initial install and then again to do it with `--check-files`.  No need to do this twice, which adds an extra ~20 seconds to the build.

This is on top of https://github.com/studentinsights/studentinsights/pull/2798.


kevinrobinson:
selfie

PR #2801, opened 2020-04-03
Maintenance: Sharding and parallelizing RSpec tests in CI

kevinrobinson:
This started as experimental, but going to ship it now 🚢 

This PR cuts total build time to ~4min!  The only downside was that the way the `CoverageChecker` worked each shard would fail, because none of the partial test runs reach 100% coverage on the expected files.  So I updated this to upload the coverage artifacts, and then added another step to merge them and run the coverage check.

This ran into a few issues:
- The download action didn't support multiple files, but the v2-preview does.  That has a bug with not properly unzipping artifacts though, so we work around that in here, with comments to the upstream issue which should be resolved soon.
- Unrelated: there were some severe latency spikes with cached folders for Ruby gems.  I commented upstream and it hasn't come up again this afternoon.
- Unrelated: there's an incident now impacting node download servers, so this works around and adds comments on that.

EDIT2: Verifying code coverage check works, [link](https://github.com/studentinsights/studentinsights/runs/565487008):
Screen Shot 2020-04-06 at 3 55 31 PM

EDIT3: example:
Screen Shot 2020-04-06 at 4 02 05 PM


kevinrobinson:
selfie

PR #2804, opened 2020-04-08
Maintenance: Fork wkhtmltopdf to remove unneeded binaries and reduce slug size

kevinrobinson:
Heroku has a slug size limit, and we ran into it a few weeks ago.  I figured it was from checking in large image assets related to reading, but turns out it was also driven by updating the `wkhtmltopdf-binary` gem.

That gem bundles up binaries for `wkhtmltopdf` on different platforms, and in that release it increased its coverage for more OS versions, which meant the size of the gem increased ~5x to almost 300MB, which is already at the soft limit for slug size just on its own.  On a dyno:

```
$ du -sh vendor/bundle/ruby/2.6.0/gems/* | sort -hr
288M	vendor/bundle/ruby/2.6.0/gems/wkhtmltopdf-binary-0.12.5.4
```

This PR adds comments to that issue upstream in the Gemfile, and swaps to a fork of the gem that limits the binaries to only Ubuntu 18 and OSX, which is all we need for this project.  That lets us keep the same developer setup and not add any other config complexity, while reducing the slug size by 100s of MB so we can deploy again.

For posterity, I also tried dropping in `rposborne/wkhtmltopdf-heroku` but it didn't just work, so instead of investigating that I switched to forking.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,8 +50,9 @@ gem 'rqrcode'
 gem 'zxcvbn-js', require: 'zxcvbn'
 
 # generating pdfs
+# See https://github.com/zakird/wkhtmltopdf_binary_gem/issues/55#issuecomment-552930066
+gem 'wkhtmltopdf-binary', :git => 'https://github.com/studentinsights/wkhtmltopdf_binary_gem.git'
 gem 'wicked_pdf'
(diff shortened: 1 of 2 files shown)


kevinrobinson:
selfie

PR #2805, opened 2020-04-08
Reading: Fix bug with supporting comparison to benchmark for F&P

kevinrobinson:
# Who is this PR for?
K5 reading folks

When calling `shouldHighlight`, the comparison of the current level with the benchmark level didn't properly support F&P.  In code, this was ambiguous whether that function would work for all `ReadingBenchmarkDataPoint` types or just DIBELS.

# What does this PR do?
Removes functions specific to DIBELS, and moves them into `readingData` to support F&P as well.  Updates the comparison for F&P to use interpretation and ordering specific to those levels.  Updates and adds some test cases.

# Checklists
+ [x] Reader profile


kevinrobinson:
selfie

PR #2806, opened 2020-04-08
Reading: Add in 4th grade winter threshold for F&P

kevinrobinson:
This was missing previously, so this PR updates to match what we had in internal docs.

# Checklists
+ [x] Reader profile


kevinrobinson:
selfie

PR #2807, opened 2020-04-10
Maintenance: Remove gzip workaround for artifact downloads in CI

kevinrobinson:
This is supposed to be fixed now, let's find out.


kevinrobinson:
Confirmed this is fixed upstream now: https://github.com/actions/download-artifact/issues/23#issuecomment-611704384

PR #2809, opened 2020-04-21
Maintenance: Update Rollbar handling in local dev, for manual testing

kevinrobinson:
# Who is this PR for?
This is for developers, trying to validate some edge cases locally.

Rollbar just wasn't provided locally, and there was a place in `<SessionRenewal />` that didn't guard for this and would raise.  This would only happen in local development.

# What does this PR do?
Updates the code setting up Rollbar to provide mock reporting functions if it doesn't load.  These just log to the console.  This way if other call sites don't guard properly local development still works.


kevinrobinson:
selfie

PR #2810, opened 2020-04-22
Maintenance: Update favicon and document title when session is close to expiring

kevinrobinson:
# Who is this PR for?
Educators within remote meetings

# What problem does this PR fix?
If someone is running a meeting remotely (eg, MTSS meeting), they start writing a note but then switch to other tabs or browser windows, and not be able to find the original tab they were working in.  In that circumstance, there's currently no indication to them that the tab is about to expire (session timeout is currently 30 minutes).

# What does this PR do?
This updates the favicon and document title so that there's a visual indication if this is a background tab.

# Screenshot (if adding a client-side feature)
This is on IE11, the UX is similar on other browsers as well.

### 1. in background tab
Screen Shot 2020-04-22 at 12 15 53 PM

### 2. favicon and document title change
Screen Shot 2020-04-22 at 12 12 45 PM

### 3. clicking in the tab, the timeout message is visible
Screen Shot 2020-04-22 at 12 13 02 PM

### 4. clicking revives the session and reset the favicon and title 
Screen Shot 2020-04-22 at 12 13 19 PM

# Checklists
*Which features or pages does this PR touch?*
+ [x] Core

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Manual testing made more sense here

PR #2812, opened 2020-04-30
Notes: Add autosaving drafts and extending session (no UX changes yet)

kevinrobinson:
# Who is this PR for?
all educators, particularly during remote meetings with multiple apps or windows

If someone is running a meeting remotely (eg, MTSS meeting), they start writing a note but then switch to other tabs or browser windows (eg, for videochat).  If they don't remember to circle back, they may lose what they had timed when the session times out.  https://github.com/studentinsights/studentinsights/pull/2810 aims to make that scenario more visible.

Similarly, if they prep for the remote meeting and load up the first student profile beforehand, but then that page sits for ~25 minutes before the meeting start, they might start typing the note, switch away to videochat, but have the session expire because there's no server interaction for so long.

# What does this PR do?
Debounces changes to draft notes, and sends them up to the server.  For now, the server writes these to `EventNoteDrafts` but doesn't otherwise use them.  This PR will keep the session alive in this scenario, and the next step will be the UI for discovering unsaved drafts and 'publishing' them as notes.  There's no UX changes in this PR.

The 'draft'  concept is a compromise between making it so folks don't have to think about saving at all (a la docs), and concerns that folks have about the potential for harm with immediately publishing as they type (eg, when typing through how to phrase privacy-sensitive issues).

Separately, this PR factors out a `RestrictedTextRedacter` that different models use to redact sensitive text in `#as_json` by default, and uses it in new `EventNoteDraft` model as well, with tests all around.

# Checklists
+ [x] Student profile, taking notes


kevinrobinson:
selfie

PR #2722, opened 2019-12-05
Admin: Fix breaking changes in administrate; remove some code on top

kevinrobinson:
# Who is this PR for?
project leads

# What problem does this PR fix?
In the Rails 6 PR, we bumped administrate too.  It broke but we missed it.  Related to https://github.com/thoughtbot/administrate/issues/1448.

# What does this PR do?
Rebuilds the custom views, removes one altogether and keeps the default except for our navbar up top.

# Screenshot (if adding a client-side feature)
Screen Shot 2019-12-05 at 10 15 06 AM

# Checklists
*Which features or pages does this PR touch?*
+ [x] Adjust permissions 

*Does this PR use tests to help verify we can deploy these changes quickly and confidently?*
+ [x] Manual testing made more sense here

PR #2813, opened 2020-05-07
Maintenance: Fix administrate pages from config bug

kevinrobinson:
# Who is this PR for?
project leads

The administrate admin pages for adjusting educator permissions were broken, probably in https://github.com/studentinsights/studentinsights/pull/2781.  I'm not sure how this got through so long, especially when there was a fair bit of manual testing to validate the gem update and other related gem updates around that time.

There was an authorization smoke test for these routes, but those didn't notice they were getting back Rails error pages.  This is the second consecutive time bumping administrate has broken these routes and we haven't caught it until a fair bit afterward (eg, https://github.com/studentinsights/studentinsights/pull/2722).

# What does this PR do?
Updates the authorization smoke test to explicitly check for Rails error pages, and fixes the root config problem, with comments referencing the upstream issue in administrate about the underlying change in Sprockets 4 that led to the config change.

# Checklists
+ [x] Project lead admin permissions pages

z+ [x] Improved specs for existing code in need of better test coverage

### DIFF
--- a/app/assets/config/manifest.js
+++ b/app/assets/config/manifest.js
@@ -1,9 +1,14 @@
 // This is the entrypoint for the Rails asset pipeline, and is new
 // as part of Sprockets 4.0.  See https://github.com/rails/sprockets/blob/master/UPGRADING.md#manifestjs
-//
+
+
 // Most JS is built separately, and most styles are defined in JS, but this
 // is the entry point for some older code still processed by the asset pipeline.
-//
 //= link_tree ../images
 //= link application.js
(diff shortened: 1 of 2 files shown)


kevinrobinson:
ugh github actions cache is down again :(

kevinrobinson:
The cache in GitHub actions is unreliable (https://github.com/studentinsights/studentinsights/pull/2815) so ran test locally and disabled it so we don't continually have these issues.  This PR is 🚢

PR #2814, opened 2020-05-07
Reading: Soft-release reader profile, with framing copy in front

kevinrobinson:
# Who is this PR for?
primarily for expanding the scope of validation and testing by projects leads (eg, with literacy coaches, reading specialists, classroom teachers)

The reader profile could be enabled individually with a feature switch, but we want to open this up a bit more.

# What does this PR do?
Because of coronavirus, the release and rollout this will likely happen outside of the PD structures that we'd initially been working towards.  So the work on framing the intention and how this connects to problem solving (versus framing centered on "data" or "graphs")  will have to get started in a lighter and more organic way.

This adds framing copy up front, with a "click to view" so that folks discovering this on their own get something minimal that hopefully helps with setting the frame and expectations.

Screen Shot 2020-05-07 at 10 39 07 AM

# Checklists
+ [x] Student Profile


kevinrobinson:
selfie

PR #2816, opened 2020-05-18
Class lists: Allow enabling authorization but hiding link for all

kevinrobinson:
Supporting different way of piloting for this year; enables for all but without links, which is done individually out of band.


kevinrobinson:
selfie, build failures already known and to be addressed separately

PR #2817, opened 2020-05-18
Maintenance: Remove workaround for zeitwerk eager loading bug

kevinrobinson:
This came from updating Rails, but ended up bundling a bunch in one.
- Rails update
- Zeitwerk monkey patch
- Error setting up the database: `PG::ObjectInUse: ERROR:  database "student_insights_test" is being accessed by other users. DETAIL:  There is 1 other session using the database.`  Not a GH issue since it still happens with only one strategy/matrix.
- Splitting up `bundle exec` tasks into their own commands makes this go away, but the setup fails on the immigrant check.  It doesn't fail on master.  There's a migration on master that wasn't checked in to `schema.rb` (could update CI to check for this).  This migration would have failed the immigrant check, since it didn't add foreign key constraints.  This PR adds a migration with those constraints.
- Brakeman update


kevinrobinson:
selfie

PR #2827, opened 2020-06-19
Maintenance: Update Rails and Rack patch releases

kevinrobinson:
Maintenance releases from notifications.


kevinrobinson:
selfie

PR #2831, opened 2020-07-15
rolling back to previous highcharts version

edavidsonsawyer:
# Who is this PR for?
All users

Student profile pdfs fail when the student has discipline incidents

# What does this PR do?
Rolls back to a previous version of Highcharts from the one set in #2826 . It's not immediately clear why upgrading Highcharts caused this problem, and we'll need to investigate and resolve that issue. Since the Highcharts upgrade wasn't necessary for any features, we can roll this back to make the pdf usable again. 

# Checklists
+ [x] Student Profile
+ [x] Student Report PDF
+ [x] Absences
+ [x] Tardies
+ [x] Discipline

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "chroma-js": "^2.1.0",
     "csv-stringify": "^5.3.6",
     "d3": "3.5.12",
-    "highcharts": "8.1.1",
+    "highcharts": "6.0.2",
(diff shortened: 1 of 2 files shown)


edavidsonsawyer:
This is a temporary fix, as we'll need to be able to keep Highcharts up to date going forward, but unblocking users for now while this is investigated.

PR #2834, opened 2020-07-27
Render pdf chart labels in html

edavidsonsawyer:
# Who is this PR for?
All users

Since highcharts 8.1.0, wkhtmltopdf produces a segfault when trying to render our charts with data labels. See, eg., https://github.com/wkhtmltopdf/wkhtmltopdf/issues/4233 and https://github.com/wkhtmltopdf/wkhtmltopdf/issues/4233

# What does this PR do?
Sets the datalabels to render as html. The practical upshot is that the label numbers are just plain black now. 

Before:
Screen Shot 2020-07-27 at 12 06 17 PM

After:
Screen Shot 2020-07-27 at 12 04 37 PM

# Checklists
+ [x] Student Report PDF


edavidsonsawyer:
Tested a variety of pdf options. This should be low impact at allows us to keep highcharts up to date, so shipping.

PR #2838, opened 2020-08-12
fixing rollbar error when only destroyed records present

edavidsonsawyer:
We have been getting an error from rollbar when the record_syncer runs complaining of an infinite value in the JSON. Turns out that we had a number of deleted records with no other record changes, making the percentage in the rollbar log infinite. Since deleted records aren't related to other records processed by the syncer, percentage isn't meaningful in this case.


edavidsonsawyer:
Premature! Closing until this error gets investigated a little further.

PR #2843, opened 2020-09-14
Disable reading sheets somerville

edavidsonsawyer:
For Fall 2020, assessments are going to be delayed for a while, so there's no reason to keep this import task active. We will re-enable once we have meaningful assessments to add.


edavidsonsawyer:
Shipping this, the failure on the audit check is a false positive that keeps coming back on us.

PR #2845, opened 2020-09-28
Revert "Temp section logging"

edavidsonsawyer:
Reverts studentinsights/studentinsights#2844

We got what we needed from these.


edavidsonsawyer:
We'll update brakeman later

PR #2853, opened 2020-11-29
only check sections assigned to educator unless districtwide

edavidsonsawyer:
# Who is this PR for?

Educators

Certain users have been experience long load times when accessing the student profile. Perf testing indicates that the section authorization check is the major factor. Authorizing all sections takes 9-13 seconds on average when the user has no schoolwide access. 

# What does this PR do?

Introduces a check to limit the sections passed to the authorizer only to those sections belonging to the user.

# Checklists
+ [x] Student Profile

The existing specs cover this change, as they already check for users with various permissions and this shouldn't functionally change the behavior.

### DIFF
--- a/app/controllers/profile_controller.rb
+++ b/app/controllers/profile_controller.rb
@@ -5,7 +5,8 @@ class ProfileController < ApplicationController
 
   def json
     student = authorized { Student.find(params[:id]) }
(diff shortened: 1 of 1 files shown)


edavidsonsawyer:
Selfie

PR #2877, opened 2021-05-12
Webpack 5 config changes

edavidsonsawyer:
On the upgrade to webpack 5, these configuration files were left behind. I'm not sure how this wasn't caught until now, but this should bring them in line with what's expected in the new version of webpack.

This is a further pass as these are causing the production deployment to fail.


edavidsonsawyer:
Closing this in favor of a different branch

PR #2886, opened 2021-06-17
Updates for merge-deep and glob-parent

edavidsonsawyer:
These are transitive dependencies handled here via resolutions. css-what is listed as vulnerable in the audits, but cf (https://github.com/cheeriojs/cheerio/issues/1924#issuecomment-856883288) this probably does not need to be fixed.


edavidsonsawyer:
Merging this since the failure is the above mentioned false positive.

PR #2890, opened 2021-07-20
Upgrade ruby to 2.7

edavidsonsawyer:
Ruby 2.6 is EOL next spring. In order to get a ahead of that change as well as take advantage of some optimization in ruby 2.7 we can upgrade this now. 

This did break the csv transformer. Digging into what changed in the new csv library, it no longer supports an encoding option that indicates the csv should be transcoded (ie "binary:utf-8"). However, it turns out csv never supported transcoding from an IO stream on initialization and would only read the input encoding. From 2.6.6. to 2.7 this is made explicit because it no longer parses the encoding option to separate input and output encodings. The upshot is that this option was never transcoding anyway so removing it should not change the behavior we see. 

The upgrade also resulted in several deprecation warnings that are resolved here related to the way ruby 2.7 handles keyword arguments (see https://www.ruby-lang.org/en/news/2019/12/25/ruby-2-7-0-released/).

Rollbar is updated here to accomodate another deprecation warning. This required a minor spec change as rollbar no longer sends the accesss_token in the payload (see https://github.com/rollbar/rollbar-gem/pull/966)


edavidsonsawyer:
Code checks and manual testing are both looking good here, so let's ship

PR #2893, opened 2021-08-06
Upgrade ruby to 2.7.4

edavidsonsawyer:
A minor bump for ruby for security fixes. This also includes new resolutions for tar across several libraries as well as an upgrade to auditjs just to keep up. None of these require code changes to support.


edavidsonsawyer:
Merging this. The audit failure is an xmldom vulnerability that shouldn't affect us and in any case the new version isn't published on npm yet.

PR #2900, opened 2021-09-09
Re Enabling Reading Benchmarks Import for 20201

edavidsonsawyer:
This re enables reading benchmark imports for Somerville now that we're getting this data again. There are some minor changes to the sheet incorporated as well, but nothing is changing on the ui side for the present.


edavidsonsawyer:
Audit flag is again from  xmldom which is not a concern for us in production.

